Issue #5, opened 2022-04-20
Reflection-based GenericRoundtripTest for ExpressionCreator methods (and failures)


curino:
(disclaimer this is me poking at the project to learn about it)

I wrote a simple reflection based test, that lists all methods in `ExpressionCreator` and for those using simple types (primitives and things like `Dates` or `BigDecimal` etc). I generate a synthetic set of params and invoke the method doing a roundtrip conversion (lifted from `LiteralRoundtripTest` which was only covering decimal so far). Two questions:

1. Do you agree it makes sense to have something like this in the codebase, to automatically cover (at least a few) cases as they get added to the system? Substrait strikes me as a project that will grow pretty broad and automating high-coverage tests seems worthy. (happy to post code if you agree it makes sense).

2.  I saw several of the methods failing on the roundtrip because the "nullability" seems not retained correctly. Is this by design and we should ignore it in the comparison or is it a bug? (example below with `timestamp` method).
 

```
Invoking timestamp ([boolean, long]) with values ([true, -5750464908882337892])

expected: <TimestampLiteral{nullable=true, value=-5750464908882337892}> but was: <TimestampLiteral{nullable=false, value=-5750464908882337892}>
Expected :TimestampLiteral{nullable=true, value=-5750464908882337892}
Actual   :TimestampLiteral{nullable=false, value=-5750464908882337892}
```

JamesRTaylor:
I think this would be a nice enhancement to improve the code coverage of our tests, @curino. Please feel free to submit a PR. Thanks!

curino:
Hi James, I did the PR and it was merged. The issue should be closed.

PR #6, opened 2022-04-20
Generic roundtrip test

curino:
Relates to: #5

### DIFF
--- a/core/src/main/java/io/substrait/expression/Expression.java
+++ b/core/src/main/java/io/substrait/expression/Expression.java
@@ -296,6 +296,7 @@ public interface Expression {
       ByteBuffer bb = ByteBuffer.allocate(16);
       bb.putLong(value().getMostSignificantBits());
       bb.putLong(value().getLeastSignificantBits());
+      bb.rewind();
       return ByteString.copyFrom(bb);
     }
   }
--- a/core/src/main/java/io/substrait/expression/Expression.java
+++ b/core/src/main/java/io/substrait/expression/Expression.java
@@ -296,7 +296,7 @@ public interface Expression {
       ByteBuffer bb = ByteBuffer.allocate(16);
       bb.putLong(value().getMostSignificantBits());
       bb.putLong(value().getLeastSignificantBits());
-      bb.rewind();
+      bb.flip();
       return ByteString.copyFrom(bb);
     }
   }
--- a/core/src/test/java/io/substrait/type/proto/GenericRoundtripTest.java
+++ b/core/src/test/java/io/substrait/type/proto/GenericRoundtripTest.java
@@ -5,7 +5,10 @@ import io.substrait.expression.Expression;
 import io.substrait.expression.ExpressionCreator;
 import io.substrait.expression.proto.ExpressionProtoConverter;
 import io.substrait.expression.proto.ProtoExpressionConverter;
-import org.junit.jupiter.api.Test;
+import org.junit.Assume;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.Arguments;
+import org.junit.jupiter.params.provider.MethodSource;
 
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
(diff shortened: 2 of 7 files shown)


core/src/test/java/io/substrait/type/proto/GenericRoundtripTest.java
```diff
@@ -0,0 +1,129 @@
...
+        } else if (type.equals(UUID.class)) {
+            return UUID.randomUUID();
+        } else if (type.equals(LocalDateTime.class)) {
+            return LocalDateTime.now();
+        } else if (type.equals(ByteString.class)) {
+            return ByteString.copyFromUtf8(UUID.randomUUID().toString());
+        } else if (type.equals(BigDecimal.class)) {
+            return BigDecimal.valueOf(rand.nextLong());
+        } else if (type.equals(Instant.class)){
+            return Instant.now();
```
jacques-n, line 124: same as comment above

core/src/test/java/io/substrait/type/proto/GenericRoundtripTest.java
```diff
@@ -0,0 +1,129 @@
...
+import java.time.Instant;
+import java.time.LocalDateTime;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Random;
+import java.util.UUID;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+public class GenericRoundtripTest {
```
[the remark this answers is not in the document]
curino: 100% this was just a quick draft for me to poke at things and demonstrate the idea.

core/src/test/java/io/substrait/type/proto/GenericRoundtripTest.java
```diff
@@ -0,0 +1,129 @@
...
+                continue;
+            }
+            selectedMethods.add(m);
+        }
+        return selectedMethods;
+    }
+
+
+    private Object valGenerator(Class<?> type) {
+        Random rand = new Random();
```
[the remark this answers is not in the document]
curino: I will look into it.

core/src/test/java/io/substrait/type/proto/GenericRoundtripTest.java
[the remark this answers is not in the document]
curino: As soon as I find some time I will address the comments and post a new version. Any thoughts on how to extend this past base types? I.e., do you have any easy to use infra to create example expression etc? (maybe from some parsed SQL or else)?

curino:
@jacques-n I fixed a good chunk of what you asked. 

Specifically the parametrization, and using junit reporting for uncovered cases. 

I also found and fixed the bug we had on round-tripping nullable fields (we missed to propagate the nullability in `ExpressionProtoConverter` for all/most types, it is now correct for the fields we cover and others more complex expressions don't seem to have nullability). 

I still need to address the comments around randomization not being tracked hence making repeatability of failure a pain (if you have pointers to good example let me know, otherwise I will look around).

curino:
I think I found and fixed another bug in Expression.toBytes() where we were missing to reset the `ByteBuffer` position and we generated an empty buffer as a copy (making the roundtrip for UUIDs not working). 

@jacques-n  let me know if you are ok to accept the 2 bugfixes and test collateral changes in one PR or prefer me to split them.

core/src/main/java/io/substrait/expression/Expression.java
```diff
@@ -296,6 +296,7 @@ public ByteString toBytes() {
       ByteBuffer bb = ByteBuffer.allocate(16);
       bb.putLong(value().getMostSignificantBits());
       bb.putLong(value().getLeastSignificantBits());
+      bb.rewind();
```
jacques-n, line 299: Should this be `bb.flip()`?

curino: Good point, I think in this case doesn't matter as we have a 16bytes and put in 2 longs, but generally I think `flip()` is likely the right answer. Updated.

curino: BTW I still need to look at how to make the rand() more trackable. I found something in junit to randomize the order of method execution, but nothing super nice/obvious to me to track this rand generation. @jacques-n  do you have a suggestion for that?

jacques-n: For now, let's just use a defined seed for all the random inputs so we avoid randomness. We can try to find something else later.

core/src/test/java/io/substrait/type/proto/GenericRoundtripTest.java
```diff
@@ -0,0 +1,145 @@
...
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Random;
+import java.util.UUID;
+
+public class GenericRoundtripTest {
+    static final org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(GenericRoundtripTest.class);
+
+    static Random rand = new Random(123);
```
jacques-n, line 31: Let's do `private final` instead of static. Random is not thread safe so best to avoid accidentally sharing.

jacques-n: Also best to add a comment that the value is not meaningful, just done to make sure test is repeatable.

jacques-n (approved):
Thanks for the contribution @curino !

core/src/test/java/io/substrait/type/proto/GenericRoundtripTest.java
curino: I guess you already merged. Should I send in an upate? (Or we can do it in future iterations, as I assume we should expand this to also cover complex types).

curino:
Thanks @jacques-n  for the prompt and useful feedback! Looking forward to slowly contribute more and more to the project!

PR #7, opened 2022-04-22
Use spotless to format Gradle Kotlin files

nastra:
This formats the Gradle Kotlin build files (.kts) via Spotless and applying the Google Format.

This also adds the following Gradle tasks:
```
spotlessApply - Applies code formatting steps to sourcecode in-place.
spotlessCheck - Checks that sourcecode satisfies formatting steps.
spotlessDiagnose
```

Additional info about spotless can be found at https://github.com/diffplug/spotless

It might also make sense to format all Java files using Spotless. If there's interest, then I could provide a PR that does that


rymurr (approved):
lgtm. I am all for not thinking about style here and in substrait-io/substrait-java

PR #8, opened 2022-04-25
Build: Add gradle task to init substrait submodule when building project 

nastra:
When cloning the repo and running `./gradlew clean jar` the build will currently fail, because it requires a manual step to init the git submodule via `git submodule update --init --recursive`.
    
This commit adds a Gradle task that does that and also includes that task when compiling the project.

This currently depends on https://github.com/substrait-io/substrait-java/pull/7 but will be rebased once it's merged


nastra:
@jacques-n or @rymurr can you review please?

PR #11, opened 2022-04-30
Support of parsing multiple create table statments. Add tpc-h smoke t…

jinfengni:
…est.

    The current status of 22 tpc-h queries:
            q1      OK
            q2      RexSubQuery not supported
            q3      OK
            q4      RexSubQuery not supported
            q5      OK
            q6      OK
            q7      Unable to convert call EXTRACT
            q8      ExceptionInInitializerError
            q9      Unable to convert call EXTRACT
            q10     OK
            q11     RexSubQuery not supported
            q12     ExceptionInInitializerError
            q13     Unable to find binding for call COUNT($1)
            q14     OK
            q15     Missing required parameter: '<sql>'
            q16     RexSubQuery not supported
            q17     RexSubQuery not supported
            q18     RexSubQuery not supported
            q19     OK
            q20     RexSubQuery not supported
            q21     RexSubQuery not supported
            q22     Unable to convert call SUBSTRING

### DIFF
--- a/isthmus/src/main/java/io/substrait/isthmus/SqlToSubstrait.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/SqlToSubstrait.java
@@ -29,6 +29,7 @@ import org.apache.calcite.rel.type.RelDataTypeFactory;
 import org.apache.calcite.rex.RexBuilder;
 import org.apache.calcite.schema.impl.AbstractTable;
 import org.apache.calcite.sql.SqlNode;
+import org.apache.calcite.sql.SqlNodeList;
 import org.apache.calcite.sql.SqlOperatorTable;
 import org.apache.calcite.sql.ddl.SqlColumnDeclaration;
 import org.apache.calcite.sql.ddl.SqlCreateTable;
@@ -66,8 +67,10 @@ public class SqlToSubstrait {
 
     if(tables != null) {
       for (String tableDef : tables) {
-        DefinedTable t = parseCreateTable(factory, validator, tableDef);
-        rootSchema.add(t.getName(), t);
+        List<DefinedTable> tList = parseCreateTable(factory, validator, tableDef);
+        for (DefinedTable t : tList) {
+          rootSchema.add(t.getName(), t);
+        }
       }
     }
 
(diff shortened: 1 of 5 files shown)


isthmus/src/main/java/io/substrait/isthmus/SqlToSubstrait.java
```diff
@@ -66,8 +67,10 @@ public Plan execute(String sql, List<String> tables) throws SqlParseException {
 
     if(tables != null) {
       for (String tableDef : tables) {
-        DefinedTable t = parseCreateTable(factory, validator, tableDef);
-        rootSchema.add(t.getName(), t);
+        List<DefinedTable> tList = parseCreateTable(factory, validator, tableDef);
```
jacques-n, line 70: This was designed to support a user passing multiple create commands by passing same arg multiple times.

`cmd ... -- create "t1..." -- create "t2..." -- create "t1..."`

Does that not work? If not, let's fix it. It looks like here you're having user pass multiple CREATEs as a single string.

isthmus/src/test/script/tpch_smoke.sh
```diff
@@ -0,0 +1,28 @@
...
+CMD=../../../build/graal/isthmus
+
+TPCH="../resources/tpch/"
+
+DDL=`cat ${TPCH}/schema.sql`
+QUERY_FOLDER="${TPCH}/queries"
+
+# echo $DDL
+
+for i in {1..5}; do echo $i; done
```
jacques-n, line 13: what is the purpose of this line?

isthmus/src/main/java/io/substrait/isthmus/SqlToSubstrait.java
jinfengni: The above syntax works.  Yes, the change proposed  is to allow user to pass a multiple statements block. For example, in tpch_smoke.sh, we could pass the whole contents in schema.sql, which contains 8 create table statements. Without the change, we have to separate schema.sql into 8 DDL strings, and use  
```
--create "create nation" --create "create orders" --create "create lineitem" .... 
```

The proposed change does not break existing usage. Use rnow could pass either single or multiple DDL statement in one --create string.

isthmus/src/test/script/tpch_smoke.sh
jinfengni: oops. I should have deleted this line.

isthmus/src/main/java/io/substrait/isthmus/SqlToSubstrait.java
jacques-n: It seems like an anti-pattern for a cli. Is there a reason you want to support all of the syntax in a single string? 

If you need this change, can you add a junit test case and smoke test for it as well?

jinfengni: I feel it convenient to support multiple table DDL in one --create string.  In case that DDL comes from an existing file, it could read from the file directly without manually copy / separate the DDLs before user could use this utility. 

But I do not have strong opinion. I could manually do the DDL editing, and use the existing syntax of this utility.

jacques-n: I don't have one really either. Seems like adding support for multiple creates in a single statement just makes things easier in certain cases. So maybe just update the help text for the cli to clarify this behavior and the test cases I mentioned?

jinfengni: - updated the help text for the cli of PlanEntryPoint. 
- added a new Junit test case in SimplePlansTest
- added tpch_smoke.sh as smoke test.

jacques-n: To clarify here, you're not executing the tpch smoke test as a precommit time test because it doesn't pass yet, right?

jinfengni: tpch_smoke.sh only runs the 7 queries that are supported. It would pass if I run manually. 

I added this tpch_smoke.sh to pr.ymal under 'Smoke test', follow 'smoke.sh' script.  I'm not clear how to run this smoke test, except for running it manually.

jacques-n (approved):
Got it, missed that change. Thanks!

PR #12, opened 2022-05-05
Converter for proto to pojo rel representation

JamesRTaylor:
Processing on Substrait plans in Java is much more clearly accomplished by working with the POJO representation rather than the serialized proto representation. The ProtoRelConverter class implements this with the help of the existing ProtoExpressionConverter class.

### DIFF
--- a/core/src/main/java/io/substrait/relation/ProtoRelConverter.java
+++ b/core/src/main/java/io/substrait/relation/ProtoRelConverter.java
@@ -1,0 +1,250 @@
+package io.substrait.relation;
+
+import static io.substrait.expression.proto.ProtoExpressionConverter.EMPTY_TYPE;
+
+import io.substrait.expression.AggregateFunctionInvocation;
+import io.substrait.expression.Expression;
+import io.substrait.expression.FunctionLookup;
+import io.substrait.expression.ImmutableExpression;
+import io.substrait.expression.proto.ProtoExpressionConverter;
+import io.substrait.function.SimpleExtension;
+import io.substrait.proto.AggregateRel;
+import io.substrait.proto.FetchRel;
+import io.substrait.proto.FilterRel;
+import io.substrait.proto.JoinRel;
+import io.substrait.proto.ProjectRel;
+import io.substrait.proto.ReadRel;
+import io.substrait.proto.SortRel;
+import io.substrait.type.ImmutableNamedStruct;
+import io.substrait.type.NamedStruct;
+import io.substrait.type.Type;
+import io.substrait.type.proto.FromProto;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Optional;
+
+/** Converts from proto to pojo rel representation TODO: AdvancedExtension, CrossJoin, Set */
+public class ProtoRelConverter {
+  static final org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(ProtoRelConverter.class);
+
+  private final FunctionLookup lookup;
+  private final SimpleExtension.ExtensionCollection extensions;
+
+  public ProtoRelConverter(FunctionLookup lookup, SimpleExtension.ExtensionCollection extensions) {
+    this.lookup = lookup;
+    this.extensions = extensions;
+  }
+
+  public Rel from(io.substrait.proto.Rel rel) {
+    var relType = rel.getRelTypeCase();
+    switch (relType) {
+      case READ -> {
+        return newRead(rel.getRead());
+      }
+      case FILTER -> {
+        return newFilter(rel.getFilter());
+      }
+      case FETCH -> {
+        return newFetch(rel.getFetch());
+      }
+      case AGGREGATE -> {
+        return newAggregate(rel.getAggregate());
+      }
+      case SORT -> {
+        return newSort(rel.getSort());
+      }
+      case JOIN -> {
+        return newJoin(rel.getJoin());
+      }
+      case PROJECT -> {
+        return newProject(rel.getProject());
+      }
+      default -> {
+        // TODO: add support for SET, EXTENSION_SINGLE, EXTENSION_MULTI, EXTENSION_LEAF, CROSS
+        throw new UnsupportedOperationException("Unsupported RelTypeCase of " + relType);
+      }
+    }
+  }
+
+  private Rel newRead(ReadRel rel) {
+    if (rel.hasVirtualTable()) {
+      return newVirtualTable(rel);
+    } else if (rel.hasNamedTable()) {
+      return newNamedScan(rel);
+    } else {
+      return newEmptyScan(rel);
+    }
+  }
+
+  private Filter newFilter(FilterRel rel) {
+    var input = from(rel.getInput());
+    // TODO: For case of subquery, we'll need to pass in ProtoRelConverter to
+    //       ProtoExpressionConverter with nested structure of Rel/Expression.
+    return Filter.builder()
+        .input(input)
+        .condition(
+            new ProtoExpressionConverter(lookup, extensions, input.getRecordType())
+                .from(rel.getCondition()))
+        .remap(optionalRelmap(rel.getCommon()))
+        .build();
+  }
+
+  private NamedStruct newNamedStruct(ReadRel rel) {
+    var namedStruct = rel.getBaseSchema();
+    var struct = namedStruct.getStruct();
+    return ImmutableNamedStruct.builder()
+        .names(namedStruct.getNamesList())
+        .struct(
+            Type.Struct.builder()
+                .fields(struct.getTypesList().stream().map(FromProto::from).toList())
+                .nullable(FromProto.isNullable(struct.getNullability()))
+                .build())
+        .build();
+  }
+
+  private EmptyScan newEmptyScan(ReadRel rel) {
+    var namedStruct = newNamedStruct(rel);
+    return EmptyScan.builder()
+        .initialSchema(namedStruct)
+        .remap(optionalRelmap(rel.getCommon()))
+        .filter(
+            Optional.ofNullable(
+                rel.hasFilter()
+                    ? new ProtoExpressionConverter(lookup, extensions, namedStruct.struct())
+                        .from(rel.getFilter())
+                    : null))
+        .build();
+  }
+
+  private NamedScan newNamedScan(ReadRel rel) {
+    var namedStruct = newNamedStruct(rel);
+    return NamedScan.builder()
+        .initialSchema(namedStruct)
+        .names(rel.getNamedTable().getNamesList())
+        .remap(optionalRelmap(rel.getCommon()))
+        .filter(
+            Optional.ofNullable(
+                rel.hasFilter()
+                    ? new ProtoExpressionConverter(lookup, extensions, namedStruct.struct())
+                        .from(rel.getFilter())
+                    : null))
+        .build();
+  }
+
+  private VirtualTableScan newVirtualTable(ReadRel rel) {
+    var virtualTable = rel.getVirtualTable();
+    List<Expression.StructLiteral> structLiterals = new ArrayList<>(virtualTable.getValuesCount());
+    for (var struct : virtualTable.getValuesList()) {
+      structLiterals.add(
+          ImmutableExpression.StructLiteral.builder()
+              .fields(struct.getFieldsList().stream().map(ProtoExpressionConverter::from).toList())
+              .build());
+    }
+    var converter = new ProtoExpressionConverter(lookup, extensions, EMPTY_TYPE);
+    return VirtualTableScan.builder()
+        .filter(converter.from(rel.getFilter()))
+        .remap(optionalRelmap(rel.getCommon()))
+        .rows(structLiterals)
+        .build();
+  }
+
+  private Fetch newFetch(FetchRel rel) {
+    var input = from(rel.getInput());
+    return Fetch.builder()
+        .input(input)
+        .remap(optionalRelmap(rel.getCommon()))
+        .count(rel.getCount())
+        .offset(rel.getOffset())
+        .build();
+  }
+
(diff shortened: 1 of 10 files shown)


nastra:
just fyi, once rebased you need to run [Gradle with JDK 17](https://github.com/substrait-io/substrait-java/blob/main/CONTRIBUTING.md#gradle--jdk-17) and then just do `./gradlw spotlessApply` to get the code properly formatted

JamesRTaylor:
Thanks, @jacques-n  and @nastra. All fixed up now.

core/src/main/java/io/substrait/relation/ProtoRelConverter.java
```diff
@@ -0,0 +1,286 @@
...
+import java.util.ArrayList;
+import java.util.List;
+
+/** Converts from proto to pojo rel representation TODO: AdvancedExtension, CrossJoin, Set */
+public class ProtoRelConverter {
+  static final org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(ProtoRelConverter.class);
+
+  private final FunctionLookup lookup;
+  private final SimpleExtension.ExtensionCollection extensions;
+  private final Type rootType;
```
jinfengni, line 31: what's rootType?  The root Rel's type? If that's the case, why would we pass rootType to `ProtoExpressionConverter` for the case of `NamedScan`, `VirtualTable`, `emptyScan`?

core/src/main/java/io/substrait/relation/ProtoRelConverter.java
```diff
@@ -0,0 +1,286 @@
...
+        throw new UnsupportedOperationException();
+      }
+    }
+  }
+
+  private Filter newFilter(FilterRel filterRel) {
+    Rel input = from(filterRel.getInput());
+    ProtoExpressionConverter expressionConverter =
+        new ProtoExpressionConverter(lookup, extensions, input.getRecordType());
+    Expression expression = expressionConverter.from(filterRel.getCondition());
```
jinfengni, line 87: For now, `ProtoExpressionConverter` would not handle expression containing subquery.  

Later on, if we extend the support of subquery in `ProtoExpressionConverter`, we need pass in a `ProtoRelConverter` to `ProtoExpressionConverter` because of the nested structure of Rel / Expression.
 
Maybe add a `TODO` mark here.

isthmus/src/test/java/io/substrait/isthmus/ProtoRelConverterTest.java
```diff
@@ -0,0 +1,87 @@
...
+import io.substrait.relation.ProtoRelConverter;
+import io.substrait.relation.Rel;
+import io.substrait.relation.RelConverter;
+import java.io.IOException;
+import java.util.Arrays;
+import org.apache.calcite.sql.parser.SqlParseException;
+import org.junit.jupiter.api.Test;
+
+public class ProtoRelConverterTest extends PlanTestBase {
+  private void assertProtoRelRoundrip(Plan p) throws IOException {
```
jinfengni, line 20: Good idea to use round trip to verify the conversion.

core/src/main/java/io/substrait/relation/ProtoRelConverter.java
JamesRTaylor: I've removed rootType as a member variable. Instead, I set it based on the context of where I'm calling it from. The two places I don't set it are in the case of a join and in the case of a virtual table. I'd like to pass in whatever ProtoExpressionConverter in all cases, but not sure what it needs.

core/src/main/java/io/substrait/relation/ProtoRelConverter.java
JamesRTaylor: Added TODO

JamesRTaylor:
Thanks for the review, @jinfengni. I've updated the PR.

core/src/main/java/io/substrait/relation/ProtoRelConverter.java
JamesRTaylor: @jacques-n - what should be passed through for `rootType`? It's only used once in ProtoExpressionConverter here:
```
   case ROOT_REFERENCE -> FieldReference.ofRoot(rootType, segments);
```

You end up with an NPE here with a null rootType.

jacques-n:
@JamesRTaylor , is kind of hard to decompose this change. It looks like there are 3 relatively discrete changes in this PR.

* Refactor functionlookup into separate interfaces and impls.
* Update to the latest Substrait upstream version
* Implement the proto => pojo conversion for rel nodes. 

Would you be willing to break these changes apart? 

Interestingly, it's very hard to review the submodule version update in Github. From what I can tell, the hash you're referencing isn't available in the substrait github repo on the main branch.

JamesRTaylor:
Changes are already in separate commits, @jacques-n. The submodule version update in Github isn’t necessary. I can revert that inadvertent change from the last commit.

isthmus/src/test/java/io/substrait/isthmus/ProtoRelConverterTest.java
```diff
@@ -0,0 +1,82 @@
...
+import io.substrait.proto.PlanRel;
+import io.substrait.relation.ProtoRelConverter;
+import io.substrait.relation.Rel;
+import org.apache.calcite.sql.parser.SqlParseException;
+import org.junit.jupiter.api.Test;
+
+import java.io.IOException;
+import java.util.Arrays;
+
+public class ProtoRelConverterTest extends PlanTestBase {
```
jacques-n, line 19: It looks like this tests don't verify any of the versions, just that we don't throw. Is that correct? Maybe add a comment to that effect throughout with a todo?

core/src/main/java/io/substrait/relation/ProtoRelConverter.java
```diff
@@ -0,0 +1,251 @@
...
+        for (AggregateRel.Grouping protoGrouping : aggregateRel.getGroupingsList()) {
+            List<Expression> expressions = new ArrayList<>(protoGrouping.getGroupingExpressionsCount());
+            for (io.substrait.proto.Expression protoExpression : protoGrouping.getGroupingExpressionsList()) {
+                Expression expression = expressionConverter.from(protoExpression);
+                expressions.add(expression);
+            }
+            Aggregate.Grouping grouping = Aggregate.Grouping.builder().expressions(expressions).build();
+            groupings.add(grouping);
+        }
+        List<Aggregate.Measure> measures = new ArrayList<>(aggregateRel.getMeasuresCount());
```
jacques-n: For most of these conversions, I'd be inclined to use streams instead of loops. I think it would be easier to read (and probably less code). Any reason you didn't?

For example
```
            List<Expression> expressions = new ArrayList<>(protoGrouping.getGroupingExpressionsCount());
            for (io.substrait.proto.Expression protoExpression : protoGrouping.getGroupingExpressionsList()) {
                Expression expression = expressionConverter.from(protoExpression);
                expressions.add(expression);
            }
```

becomes (roughly)

```
var expressions = protoGrouping.getGroupingExpressionsList().stream().map(expressionConverter::from).toList()
```

core/src/main/java/io/substrait/relation/ProtoRelConverter.java
```diff
@@ -0,0 +1,251 @@
...
+        this.extensions = extensions;
+        this.rootType = rootType;
+    }
+
+    public Rel from(io.substrait.proto.Rel rel) {
+        io.substrait.proto.Rel.RelTypeCase relType = rel.getRelTypeCase();
+        switch (relType) {
+            case READ -> {
+                ReadRel readRel = rel.getRead();
+                if (readRel.hasVirtualTable()) {
```
jacques-n: If you're avoiding nesting for all the other rels, it feels weird to do the nesting for readrel inline.

jacques-n (commented):
I generally prefer separate commits for things that are entirely disjunct. It makes it easier to later understand changes that are merged and also enable reverts more easily.

isthmus/src/test/java/io/substrait/isthmus/ProtoRelConverterTest.java
JamesRTaylor: The tests verify that a proto is convertable to a pojo and that when the pojo is converted back to a proto it is equivalent to the original proto.

jacques-n: Can you point out the assertions? I don't see them.

core/src/main/java/io/substrait/relation/ProtoRelConverter.java
jacques-n: If you want to reference a field from an incoming relation in a field reference, the reference not only needs the index, it also needs the data type for that reference. That's the purpose of rootType. It seems unecessary at the global level of rel converter since it would be consumed from one or more inputs of the rel we're currently working on.

isthmus/src/test/java/io/substrait/isthmus/ProtoRelConverterTest.java
```diff
@@ -0,0 +1,86 @@
...
+  private void assertProtoRelRoundrip(Plan p) throws IOException {
+    SimpleExtension.ExtensionCollection extensionCollection = SimpleExtension.loadDefaults();
+    FunctionLookup functionLookup = ImmutableFunctionLookup.builder().from(p).build();
+    ProtoRelConverter relConverter = new ProtoRelConverter(functionLookup, extensionCollection);
+    for (PlanRel planRel : p.getRelationsList()) {
+      io.substrait.proto.Rel protoRel1 = planRel.getRoot().getInput();
+      Rel rel = relConverter.from(protoRel1);
+      io.substrait.proto.Rel protoRel2 = new RelConverter(new FunctionCollector()).toProto(rel);
+      assertEquals(protoRel1, protoRel2);
```
JamesRTaylor, line 28: The assert is here, @jacques-n

jacques-n:
Can you rebase. Unfortunately it is showing this with all the changes.

JamesRTaylor:
Rebased. Will address your other comments in a new commit.

core/src/main/java/io/substrait/relation/ProtoRelConverter.java
JamesRTaylor: I've used streams for this method in my last commit. I find it harder to read and debug (partly because I'm not as used to writing the code that way), and in the end same to have the same number of lines. Take a look and let me know what you think. I'm fine either way - I can get used to those APIs too.

JamesRTaylor:
Please take a look when you have a chance, @jacques-n. I can convert other methods to use the stream APIs instead of for loops if you think it's an improvement. I'll squash all the commits after you approve the changes.

core/src/main/java/io/substrait/relation/ProtoRelConverter.java
JamesRTaylor: Do you think it'd make sense to have a two arg constructor for `ProtoExpressionConverter` and we can default rootType to some non null value that doesn't trigger that exception? Or the code could perhaps be tolerant of `rootType` being null?

core/src/main/java/io/substrait/relation/ProtoRelConverter.java
JamesRTaylor: If I only use streams for the inner loop, it still reads well. I'll push out a version of that.

JamesRTaylor:
PTAL, @jacques-n. I've updated it based on your comments. Thanks for the reviews.

core/src/main/java/io/substrait/relation/ProtoRelConverter.java
```diff
@@ -0,0 +1,249 @@
...
+                .toList())
+        .build();
+  }
+
+  private Join newJoin(JoinRel rel) {
+    Rel left = from(rel.getLeft());
+    Rel right = from(rel.getRight());
+    Type.Struct leftStruct = left.getRecordType();
+    Type.Struct rightStruct = right.getRecordType();
+    Type.Struct unionedStruct = Type.Struct.builder().from(leftStruct).from(rightStruct).build();
```
JamesRTaylor, line 231: Need to add more tests for joins with post filters. This may be too simplistic?

core/src/main/java/io/substrait/relation/ProtoRelConverter.java
```diff
@@ -0,0 +1,249 @@
...
+    var virtualTable = rel.getVirtualTable();
+    List<Expression.StructLiteral> structLiterals = new ArrayList<>(virtualTable.getValuesCount());
+    for (var struct : virtualTable.getValuesList()) {
+      structLiterals.add(
+          ImmutableExpression.StructLiteral.builder()
+              .fields(struct.getFieldsList().stream().map(ProtoExpressionConverter::from).toList())
+              .build());
+    }
+    var converter =
+        new ProtoExpressionConverter(lookup, extensions, newNamedScan(rel).getRecordType());
```
JamesRTaylor, line 142: This will create a namedStruct from the baseSchema. Is that the right type to pass through?

jacques-n: I wouldn't expect this record type to get used. I'd probably pass in an empty struct type. If we try to index into it, I'd hope we'd throw an exception. There is no input to a virtual table so proto expression converter shouldn't need a root type in this case.

JamesRTaylor: I created a static EMPTY_TYPE and use that instead. I also enforce that rootType is non null.

core/src/main/java/io/substrait/relation/ProtoRelConverter.java
jacques-n: Does calling from twice concatenate? I wouldn't have expected that.

jacques-n (approved):
LGTM. Thanks @JamesRTaylor !

Issue #13, opened 2022-05-06
Add back task of init substrait submodule to fix build issue.


jinfengni:
The commit 83abf918c8 removed the submodule init task. As such,  ./gradlew build would fail due to missing source codes in substrait repo.

```
 error: package io.substrait.proto does not exist
import io.substrait.proto.JoinRel;
```

https://github.com/substrait-io/substrait-java/commit/83abf918c8db778551c8b4b8f638db457ba04b28

jacques-n:
@curino , can you pick this up? It looks like your patch accidentally undid some other changes. Thanks!

curino:
Ops sorry about that... Happy to look at it, I had a messy merge and restarted with a clean clone copying the changes  but something must have gone wrong. My week looks busy, if I am too slow and this is blocking others feel free to revert the patch and I will look at it later.

nastra:
I cherry-picked it back with 5a660ad5701ebc7cdec7c4840aac0f34d64822d0

PR #15, opened 2022-05-09
Reverting inadverted gradle clash

curino:


curino:
Relates to #13. I think the manual step required by f2049838ce4c1d69df3974ea032a461b5bea8430 through off my setup. The checked version seems to compile fine in codespaces (still messy on my box, but I will refresh the setup)

curino:
Ok, it seems to work locally too. I still have to do the manual `git submodule update --init --recursive` related to f2049838ce4c1d69df3974ea032a461b5bea8430 which is a bit annoying. If the intent was to make it automatic it does not seem to work (at least in idea/cli on a new git clone).

nastra:
see related comments in https://github.com/substrait-io/substrait-java/commit/f2049838ce4c1d69df3974ea032a461b5bea8430#commitcomment-73290209

PR #19, opened 2022-05-12
Support subquery in SqlToSubStrait conversion.

jinfengni:
Add support to convert query containing the following pattern:
 1) exists (subquery)
 2) unique (subquery)
 3) inpredicate
 4) scalar subquery

13 tpc-h queries can run.


isthmus/src/main/java/io/substrait/isthmus/expression/RexExpressionConverter.java
```diff
@@ -2,30 +2,45 @@
...
-  public RexExpressionConverter(List<CallConverter> callConverters) {
+  public RexExpressionConverter(
+      SubstraitRelVisitor relVisitor, List<CallConverter> callConverters) {
     this.callConverters = callConverters;
+    this.relVisitor = relVisitor;
   }
 
   public RexExpressionConverter() {
     this.callConverters = CallConverters.DEFAULTS;
+    this.relVisitor = null; // TODO
```
JamesRTaylor, line 43: What's the TODO here?

isthmus/src/test/java/io/substrait/isthmus/SubqueryPlanTest.java
```diff
@@ -0,0 +1,117 @@
...
+    var creates = Arrays.stream(values).filter(t -> !t.trim().isBlank()).toList();
+    Plan plan =
+        s.execute(
+            "select l_partkey from lineitem where exists (select o_orderdate from orders where o_orderkey = l_orderkey)",
+            creates);
+    String planText = JsonFormat.printer().includingDefaultValueFields().print(plan);
+
+    assertTrue(planText.contains("setPredicate"));
+    assertTrue(planText.contains("PREDICATE_OP_EXISTS"));
+    assertTrue(planText.contains("\"stepsOut\": 1"));
```
JamesRTaylor, line 27: These tests are fine for now, but how about adding a TODO for a roundtrip test once the ProtoRelConverter is committed and updated to support subqueries?

isthmus/src/test/java/io/substrait/isthmus/SubqueryPlanTest.java
```diff
@@ -0,0 +1,117 @@
...
+    var creates = Arrays.stream(values).filter(t -> !t.trim().isBlank()).toList();
+    Plan plan = s.execute(sql, creates);
+    String planText = JsonFormat.printer().includingDefaultValueFields().print(plan);
+
+    assertTrue(planText.contains("setPredicate"));
+    assertTrue(planText.contains("PREDICATE_OP_EXISTS"));
+    assertTrue(planText.contains("PREDICATE_OP_UNIQUE"));
+
+    assertTrue(planText.contains("\"stepsOut\": 2"));
+    assertTrue(planText.contains("\"stepsOut\": 1"));
```
JamesRTaylor, line 71: Nice test!

isthmus/src/test/java/io/substrait/isthmus/SubqueryPlanTest.java
```diff
@@ -0,0 +1,117 @@
...
+    String sql = asString("subquery/nested_scalar_subquery.sql");
+    var creates = Arrays.stream(values).filter(t -> !t.trim().isBlank()).toList();
+    Plan plan = s.execute(sql, creates);
+    String planText = JsonFormat.printer().includingDefaultValueFields().print(plan);
+
+    assertTrue(planText.contains("\"stepsOut\": 2"));
+    assertTrue(planText.contains("\"stepsOut\": 1"));
+    assertTrue(planText.contains("\"scalar\""));
+
+    // System.out.println(JsonFormat.printer().includingDefaultValueFields().print(plan));
```
JamesRTaylor, line 115: Remove

JamesRTaylor:
This is awesome! Great work, @jinfengni .

isthmus/src/main/java/io/substrait/isthmus/expression/RexExpressionConverter.java
jinfengni: This is for the case where RexExpressionConverter is used for Rex expression without nesting subquery.  I changed TODO to a comment.

isthmus/src/test/java/io/substrait/isthmus/SubqueryPlanTest.java
jinfengni: added TODO.

isthmus/src/test/java/io/substrait/isthmus/SubqueryPlanTest.java
jinfengni: removed.

jinfengni:
Thanks a lot for your review! I addressed the comments. @JamesRTaylor

isthmus/src/main/java/io/substrait/isthmus/OuterReferenceResolver.java
```diff
@@ -0,0 +1,88 @@
...
+
+import java.util.HashMap;
+import java.util.IdentityHashMap;
+import java.util.Map;
+import org.apache.calcite.rel.RelNode;
+import org.apache.calcite.rel.core.CorrelationId;
+import org.apache.calcite.rel.logical.LogicalFilter;
+import org.apache.calcite.rex.*;
+
+/** Resolve correlated variable and get Depth map for RexFieldAccess */
```
jacques-n, line 11: My gut is that you did a bunch of research to figure out all of this. Maybe put a more in-depth comment here (and elsewhere as appropriate)? The steps out thing especially seemed like something that was initially confusing. I'm guessing if someone comes in here later, they're going to need help to fix a bug.

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelVisitor.java
```diff
@@ -32,13 +34,15 @@ public class SubstraitRelVisitor extends RelNodeVisitor<Rel, RuntimeException> {
   private final RexVisitor<Expression> converter;
   private final AggregateFunctionConverter aggregateFunctionConverter;
 
+  Map<RexFieldAccess, Integer> fieldAccessDepthMap;
```
jacques-n, line 37: Why not private? Usually it is good to avoid package protected fields.

core/src/main/java/io/substrait/expression/proto/ExpressionProtoConverter.java
```diff
@@ -12,9 +13,11 @@ public class ExpressionProtoConverter implements ExpressionVisitor<Expression, R
       org.slf4j.LoggerFactory.getLogger(ExpressionProtoConverter.class);
 
   private final FunctionLookup lookup;
+  private final RelConverter relConverter;
 
-  public ExpressionProtoConverter(FunctionLookup lookup) {
+  public ExpressionProtoConverter(FunctionLookup lookup, RelConverter relConverter) {
```
jacques-n, line 18: Do we need to be as specific as RelConverter here or would a narrower interface be sufficient?

isthmus/src/main/java/io/substrait/isthmus/OuterReferenceResolver.java
```diff
@@ -0,0 +1,88 @@
...
+      if (!nestedDepth.containsKey(id)) {
+        nestedDepth.put(id, 0);
+      }
+    }
+    filter.getCondition().accept(rexVisitor);
+    return super.visit(filter);
+  }
+
+  @Override // FIXME: join / project with expressions (possible contain subquery)
+  public RelNode visitOther(RelNode other) throws RuntimeException {
```
jacques-n, line 48: Can we do something here so that if we hit those scenarios, we make sure to throw as opposed to ignore or lose the data?

isthmus/src/main/java/io/substrait/isthmus/expression/RexExpressionConverter.java
```diff
@@ -3,29 +3,42 @@
...
+import org.apache.calcite.sql.fun.SqlStdOperatorTable;
 
 public class RexExpressionConverter implements RexVisitor<Expression> {
   static final org.slf4j.Logger logger =
       org.slf4j.LoggerFactory.getLogger(RexExpressionConverter.class);
 
   private final List<CallConverter> callConverters;
+  private final SubstraitRelVisitor relVisitor;
+
+  public RexExpressionConverter(SubstraitRelVisitor relVisitor, CallConverter... callConverters) {
```
jacques-n, line 24: Can we use something narrower than SubstraitRelVisitor as the argument here?

isthmus/src/main/java/io/substrait/isthmus/expression/RexExpressionConverter.java
```diff
@@ -80,11 +93,49 @@ public Expression visitRangeRef(RexRangeRef rangeRef) {
...
+    }
+    throw new UnsupportedOperationException(
+        "RexFieldAccess for other than RexCorrelVariable not supported");
   }
 
   @Override
   public Expression visitSubQuery(RexSubQuery subQuery) {
+    Rel rel = relVisitor.apply(subQuery.rel);
+
+    if (subQuery.getOperator() == SqlStdOperatorTable.EXISTS) {
```
jacques-n, line 112: This looks like it might be better as a switch statement. (?)

isthmus/src/test/java/io/substrait/isthmus/SubqueryPlanTest.java
```diff
@@ -0,0 +1,117 @@
...
+    SqlToSubstrait s = new SqlToSubstrait();
+    String[] values = asString("tpch/schema.sql").split(";");
+    var creates = Arrays.stream(values).filter(t -> !t.trim().isBlank()).toList();
+    Plan plan =
+        s.execute(
+            "select l_partkey from lineitem where exists (select o_orderdate from orders where o_orderkey = l_orderkey)",
+            creates);
+    String planText = JsonFormat.printer().includingDefaultValueFields().print(plan);
+
+    assertTrue(planText.contains("setPredicate"));
```
jacques-n, line 27: I'd prefer to see structured verification as opposed to text verification here. Any reason not to?

jacques-n (commented):
Mostly looks good. Only a few modifications. Mainly around reducing coupling by using narrower interfaces. Thanks!

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelVisitor.java
jinfengni: change to `private`.

core/src/main/java/io/substrait/expression/proto/ExpressionProtoConverter.java
jinfengni: Changed to use `ExpressionVisitor<Expression, RuntimeException>` here.

isthmus/src/main/java/io/substrait/isthmus/OuterReferenceResolver.java
jinfengni: I checked Project case (Join case will be a filter condition, which has been handled).  

For Filter's case, `variablesSet`  will be set in SqltoRelConverter's `convertWhere`.  However,  for `Project`, SqlToRelConverter will not set  variablesSet().  It delays to `SubQueryRemoveRule`. Therefore,  the code will throw Unsupported exception if a `Project` expression contain a RexSubquery. 

The `RelNode` API `getVariablesSet` has comment that it is only set for `org.apache.calcite.rel.core.Correlate`. But looks like Calcite's code will set the variablesSet for `Filter`, but not for `Project`. 

I added a UT to verify Exception is thrown for the case of `Project`.

isthmus/src/main/java/io/substrait/isthmus/expression/RexExpressionConverter.java
jinfengni: Here I have to use SubstraitRelVisitor as we need to get the `steps_out` value (which requires a view of the entire RelNode Tree). The RexExpression (only have access to a rex expression tree) does not have that knowledge.

jacques-n: Couldn't this information carried down on the traversal (rather than walking back up the tree once you hit the subquery)? Seems like that would reduce the number of times you're traversing the tree (if I understand this correctly). In that case, we'd probably carry something more akin to a map with correlation to distance from current location.

isthmus/src/main/java/io/substrait/isthmus/expression/RexExpressionConverter.java
jinfengni: switch statement requires compile time constant. Here `SqlStdOperatorTable.EXISTS` is run-time constant. Seems I could not change to a switch statement.

isthmus/src/test/java/io/substrait/isthmus/SubqueryPlanTest.java
jinfengni: I changed to structured verification.  The text verification is easier to write, but fail to check the plan structure.

isthmus/src/main/java/io/substrait/isthmus/OuterReferenceResolver.java
jinfengni: I put explanation how the code logic works in `OuterReferenceResolver.md` file, with an example and how the steps work. 

Previously, I tried to put in-depth comment. But the java formatter changed the text-based tree graph, and make it hard to read.

isthmus/src/main/java/io/substrait/isthmus/expression/RexExpressionConverter.java
jinfengni: The `fieldAccessDepthMap` is built on the initial traversal in `SubstraitRelVisitor.popFieldAccessDepthMap`.  In `RexExpressionConverter`, it use the `SubstraitRelVisitor's map to get `steps_out`.  So, in total, there are 2 times of tree traversal in `RelNode -> substrait Rel` conversion.

jacques-n: ok. Let's keep in mind this may be a place to find perf later if needed.

jacques-n (approved):
Looks good, thanks @jinfengni !

PR #25, opened 2022-05-21
Fix bug causing aggregate function invocation to be dropped

JamesRTaylor:
Found and fixed bug causing aggregate function invocation to be dropped on roundtrip. Also added better test for aggregate function invocation.

### DIFF
--- a/core/src/main/java/io/substrait/relation/RelChildVisitor.java
+++ b/core/src/main/java/io/substrait/relation/RelChildVisitor.java
@@ -1,0 +1,282 @@
+package io.substrait.relation;
+
+import io.substrait.expression.Expression;
+import io.substrait.expression.ExpressionVisitor;
+import io.substrait.expression.FieldReference;
+
+/**
+ * Class used to visit all child relations from a root relation. The traversal will include
+ * relations inside of subquery expressions.
+ */
+public class RelChildVisitor implements RelVisitor<Rel, RuntimeException> {
+  @Override
+  public Aggregate visit(Aggregate aggregate) {
+    aggregate.getInput().accept(this);
+    return aggregate;
+  }
+
+  @Override
+  public EmptyScan visit(EmptyScan emptyScan) {
+    return emptyScan;
+  }
+
+  @Override
+  public Fetch visit(Fetch fetch) {
+    fetch.getInput().accept(this);
+    return fetch;
+  }
+
+  @Override
+  public Filter visit(Filter filter) {
+    filter.getInput().accept(this);
+    visitExpression(filter.getCondition());
+    return filter;
+  }
+
+  @Override
+  public Join visit(Join join) {
+    join.getLeft().accept(this);
+    join.getRight().accept(this);
+    if (join.getCondition().isPresent()) {
+      visitExpression(join.getCondition().get());
+    }
+    if (join.getPostJoinFilter().isPresent()) {
+      visitExpression(join.getPostJoinFilter().get());
+    }
+    return join;
+  }
+
+  @Override
+  public NamedScan visit(NamedScan namedScan) {
+    return namedScan;
+  }
+
+  @Override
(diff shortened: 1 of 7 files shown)


JamesRTaylor:
FYI, @jinfengni in case this was getting in your way.

core/src/main/java/io/substrait/relation/RelChildVisitor.java
```diff
@@ -0,0 +1,275 @@
+package io.substrait.relation;
+
+import io.substrait.expression.Expression;
+import io.substrait.expression.ExpressionVisitor;
+import io.substrait.expression.FieldReference;
+
+public class RelChildVisitor<OUTPUT, EXCEPTION extends Exception>
```
jacques-n, line 7: Can you add javadocs? Also, it seems the methods return nulls which suggests that we shouldn't have the OUTPUT parameter in the RelVisitor declaration. Correct?

isthmus/src/test/java/io/substrait/isthmus/ProtoPlanConverterTest.java
```diff
@@ -15,15 +16,46 @@ private void assertProtoRelRoundrip(String query) throws IOException, SqlParseEx
...
+    SqlToSubstrait s = new SqlToSubstrait();
+    String[] values = asString("tpch/schema.sql").split(";");
+    var creates = Arrays.stream(values).filter(t -> !t.trim().isBlank()).toList();
+    return s.execute(query1, creates);
+  }
+
   @Test
   public void aggregate() throws IOException, SqlParseException {
-    assertProtoRelRoundrip("select count(DISTINCT L_ORDERKEY),sum(L_ORDERKEY) from lineitem");
+    assertProtoRelRoundrip("select count(L_ORDERKEY),sum(L_ORDERKEY) from lineitem");
```
jacques-n, line 36: We should add proto roundtrip tests to the core module. Relying on the SQL to Substrait code substantially increases this kind of bug.

core/src/main/java/io/substrait/relation/RelChildVisitor.java
JamesRTaylor: Added class level javadoc. Will add more after my next commit as I plan to use the visitor to copy-on-write for simple substitution replacement scenarios.

isthmus/src/test/java/io/substrait/isthmus/ProtoPlanConverterTest.java
JamesRTaylor: Added core test for aggregate function invocation roundtrip.

JamesRTaylor:
Thanks for the review, @jacques-n. I've updated the PR based on your feedback.

core/src/main/java/io/substrait/relation/RelChildVisitor.java
```diff
@@ -0,0 +1,282 @@
...
+          }
+
+          @Override
+          public Expression.StructLiteral visit(Expression.StructLiteral expr) {
+            return expr;
+          }
+
+          @Override
+          public Expression.Switch visit(Expression.Switch expr) {
+            expr.defaultClause().accept(this);
```
jacques-n, line 199: If this is a rewrite, shouldn't we compose a new switch statement (and others below such as ifthen and scalarfunctioninvocation) so that any mutation remains the same? Otherwise, it's weird to be returning something from the visitor.

JamesRTaylor: Yes, but I was planning on doing that in my next commit. Didn't want to merge it with this one which is two commits: one for a bug fix (plus new test) and one for a deep traversal visitor (not yet rewrite).

jacques-n (approved):
Looks good, thanks! Look forward to follow-up for traversal.

PR #26, opened 2022-05-23
Add table lookup entry point to SqlToSubstrait

rymurr:
This uses a function to look up a table schema without a create table statement. I've intentionally kept the interface free from calcite objects so as to not export them as part of the isthmus public API.

I am not an expert at calcite so I am not sure if this is the best way to do it. Also, note, my fix adds state to the converter. Happy to adjust based on input.

### DIFF
--- a/isthmus/src/main/java/io/substrait/isthmus/SqlToSubstrait.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/SqlToSubstrait.java
@@ -56,17 +56,17 @@ import org.apache.calcite.sql2rel.StandardConvertletTable;
 /** Take a SQL statement and a set of table definitions and return a substrait plan. */
 public class SqlToSubstrait {
 
-  private final CalciteSchema rootSchema = CalciteSchema.createRootSchema(false);
-  private final RelDataTypeFactory factory = new JavaTypeFactoryImpl();
-  private final CalciteConnectionConfig config =
-      CalciteConnectionConfig.DEFAULT.set(CalciteConnectionProperty.CASE_SENSITIVE, "false");
-  private final CalciteCatalogReader catalogReader =
-      new CalciteCatalogReader(rootSchema, List.of(), factory, config);
-  private final SqlValidator validator =
-      Validator.create(factory, catalogReader, SqlValidator.Config.DEFAULT);
 
   public Plan execute(String sql, Function<List<String>, Map<String, Type>> tableLookup)
       throws SqlParseException {
+    CalciteSchema rootSchema = CalciteSchema.createRootSchema(false);
+    RelDataTypeFactory factory = new JavaTypeFactoryImpl();
+    CalciteConnectionConfig config =
+        CalciteConnectionConfig.DEFAULT.set(CalciteConnectionProperty.CASE_SENSITIVE, "false");
+    CalciteCatalogReader catalogReader =
+        new CalciteCatalogReader(rootSchema, List.of(), factory, config);
+    SqlValidator validator =
+        Validator.create(factory, catalogReader, SqlValidator.Config.DEFAULT);
     SqlParser parser = SqlParser.create(sql, SqlParser.Config.DEFAULT);
     var parsed = parser.parseQuery();
     Set<SqlIdentifier> ids = new HashSet<>();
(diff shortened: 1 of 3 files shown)


jinfengni:
Seems to me we need `CalciteSchema` that dynamically loads subschema/tables on demand when the validator requests to validate an identifier either for subschema/table.   The `tableLookUp` function should be provided to such dynamic CalciteSchema.  

This pr first finds all the sql identifier in the parse tree, and then try to match/find the identifier in `tableLookUp`. However, an identifier `abc` could be either a subschema name, or table name or column name. We would not know which kind it is, unless we have validator's context.

rymurr:
Thanks a lot for the feedback @jinfengni and @jacques-n

I have updated to use a catalog. Do you think this is a better approach?

isthmus/src/main/java/io/substrait/isthmus/SqlToSubstrait.java
```diff
@@ -49,18 +54,44 @@
 /** Take a SQL statement and a set of table definitions and return a substrait plan. */
 public class SqlToSubstrait {
 
+  public Plan execute(String sql, Function<List<String>, Map<String, Type>> tableLookup)
```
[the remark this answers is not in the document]
rymurr: I would have preferred that as well but the substrait `Struct` type doesn't have field names, just types. So I can't construct a Calcite Struct type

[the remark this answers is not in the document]
jacques-n: Oh, you need to use a NamedStruct. Did we formalize that in the context of the java pojos? If no, let's do that. See the proto paradigm of the structre here: [NamedStruct](https://github.com/substrait-io/substrait/blob/00acfb52f12b5ddf696bb58d5f7509d1a6218171/proto/substrait/type.proto#L192)

isthmus/src/main/java/io/substrait/isthmus/SqlToSubstrait.java
```diff
@@ -49,18 +53,41 @@
...
+    RelDataTypeFactory factory = new JavaTypeFactoryImpl();
+    Function<List<String>, Table> lookup =
+        id -> {
+          NamedStruct table = tableLookup.apply(id);
+          if (table == null) {
+            return null;
+          }
+          List<RelDataType> types =
+              table.struct().fields().stream()
+                  .map(f -> TypeConverter.convert(factory, f))
```
jacques-n, line 67: I don't think this will work correctly for nested structs, will it? The names list is a tree of names and we'd need to somehow pass that into TypeConverter.

isthmus/src/main/java/io/substrait/isthmus/SqlToSubstrait.java
```diff
@@ -49,18 +53,41 @@
...
+  public Plan execute(String sql, Function<List<String>, NamedStruct> tableLookup)
+      throws SqlParseException {
+    RelDataTypeFactory factory = new JavaTypeFactoryImpl();
+    Function<List<String>, Table> lookup =
+        id -> {
+          NamedStruct table = tableLookup.apply(id);
+          if (table == null) {
+            return null;
+          }
+          List<RelDataType> types =
```
jacques-n, line 65: Does this work for a multi-level table. For example ryan.tpch.lineitem?

isthmus/src/main/java/org/apache/calcite/jdbc/LookupCalciteSchema.java
```diff
@@ -0,0 +1,48 @@
...
+      Schema schema,
+      String name,
+      Function<List<String>, Table> lookup) {
+    super(parent, schema, name);
+    this.lookup = lookup;
+  }
+
+  @Override
+  protected @Nullable CalciteSchema getImplicitSubSchema(String schemaName, boolean caseSensitive) {
+    if (cache.computeIfAbsent(path(schemaName), lookup) != null) {
```
jacques-n, line 27: I don't know the impl of `path()`. Is this ensure that if we have `foo.bar.t1` and `bar.foo.t1`, nothing get confused?

rymurr: yes. Path returns the name of each subSchema plus the table name. This ensures that we are looking up a table in the backend by its full path

isthmus/src/main/java/io/substrait/isthmus/SqlToSubstrait.java
rymurr: no it didn't. Forgot to override a method.

isthmus/src/main/java/io/substrait/isthmus/SqlToSubstrait.java
rymurr: yup, good catch. Fixed

rymurr:
Thanks for the review @jacques-n caught a few bugs and fixed them in the most recent iteration.

jacques-n:
This looks reasonable to me. Merging.

PR #28, opened 2022-05-26
Convert substrait plan to Sql.

jinfengni:
- SubstraitRelNodeConverter : convert pojo substrait Rel node
to Calcite RelNode.
- ExpressionRexConverter: convert pojo substrait expression
to Calcite RexNode.
- Extend CalciteCallTest and CalciteLiteralTest by doing round trip
conversion and comparision

Support filter/project/scan with scalar function, literal of integer,
string, date, time, timestamp.


isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
```diff
@@ -0,0 +1,126 @@
...
+    List<RexNode> rexList =
+        project.getExpressions().stream()
+            .map(
+                expr ->
+                    expr.accept(
+                        new ExpressionRexConverter(
+                            relOptCluster.getTypeFactory(), scalarFunctionConverter)))
+            .toList();
+
+    return LogicalProject.create(
```
jacques-n, line 110: I think we should probably be using a RelBuilder rathern invoking these static constructors individually. Or do you think that doesn't make sense?

isthmus/src/main/java/io/substrait/isthmus/SubstraitToSql.java
```diff
@@ -0,0 +1,85 @@
...
+import org.apache.calcite.sql.validate.SqlValidator;
+import org.apache.calcite.sql2rel.SqlToRelConverter;
+
+public class SubstraitToSql extends SqlConverterBase {
+  private final RelDataTypeFactory factory;
+  private final CalciteCatalogReader catalogReader;
+  private final RelOptCluster relOptCluster;
+  private final SqlValidator validator;
+
+  SubstraitToSql() {
```
jacques-n, line 33: Can we separate this into two steps? Substrait to Calcite and Calcite to SQL? That way people who want to only use the first part, don't have to reimplement things.

isthmus/src/main/java/io/substrait/isthmus/SubstraitToSql.java
```diff
@@ -0,0 +1,85 @@
...
+    CalciteSchema rootSchema = CalciteSchema.createRootSchema(false);
+    SqlToRelConverter.Config converterConfig =
+        SqlToRelConverter.config().withTrimUnusedFields(true).withExpand(false);
+    this.factory = new JavaTypeFactoryImpl();
+
+    CalciteConnectionConfig config =
```
jacques-n, line 39: This feels duplicative of other places. Can we factor out?

isthmus/src/main/java/io/substrait/isthmus/expression/ExpressionRexConverter.java
```diff
@@ -0,0 +1,289 @@
...
+  public RexNode visit(Expression.ListLiteral expr) throws RuntimeException {
+    throw new UnsupportedOperationException(
+        this.getClass().getCanonicalName()
+            + " does not support "
+            + expr.getClass().getCanonicalName());
+  }
+
+  @Override
+  public RexNode visit(Expression.StructLiteral expr) throws RuntimeException {
+    throw new UnsupportedOperationException(
```
jacques-n, line 193: As mentioned in #27, let's use an abstract visitor here with fallback for the failure cases.

isthmus/src/test/java/io/substrait/isthmus/Substrait2SqlTestNoValidation.java
```diff
@@ -0,0 +1,45 @@
...
+import java.util.Arrays;
+import org.apache.calcite.rel.RelNode;
+import org.apache.calcite.rel.RelRoot;
+import org.apache.calcite.sql.SqlDialect;
+import org.junit.jupiter.api.Test;
+
+public class Substrait2SqlTestNoValidation extends PlanTestBase {
+  @Test
+  public void simpleTest() throws Exception {
+    String query = "select p_size + 100 from part where p_partkey > 100";
```
jacques-n, line 15: let's add some tests that are sql => substrait => calcite plan => substrait and then do comparison of the substrait plans (round trips).

isthmus/src/main/java/io/substrait/isthmus/expression/FunctionMappings.java
```diff
@@ -41,6 +46,24 @@ public class FunctionMappings {
                 s(SqlStdOperatorTable.COUNT, "count"),
                 s(SqlStdOperatorTable.AVG, "avg"))
             .build();
+
```
jacques-n, line 49: Rather than having two separate sets, maybe we can create an object that contains bidirectional mappings and then just have the object referenced bidirectionally? Feels like that would avoid double-declaring things.

isthmus/src/main/java/io/substrait/isthmus/expression/FunctionConverter.java
```diff
@@ -69,9 +73,48 @@ public FunctionConverter(
...
+      }
+    }
+
     this.signatures = matcherMap;
   }
 
+  protected Optional<SqlOperator> getSqlOperatorFromSubstraitFunc(
+      Expression.ScalarFunctionInvocation expr) {
+    var resolver = getTypeBasedResolver();
+    if (substraitFuncKeyToSqlOperatorMap.containsKey(expr.declaration().key())) {
```
jacques-n, line 90: (style/readability) better to invert this condition and inner clause so we don't indent a large block.

isthmus/src/main/java/io/substrait/isthmus/expression/FunctionConverter.java
```diff
@@ -69,9 +73,48 @@ public FunctionConverter(
...
+                  operator ->
+                      resolver.containsKey(operator)
+                          && resolver.get(operator).types().contains(outputTypeStr))
+              .toList();
+      // only one SqlOperator is possible
+      if (resolvedOperators.size() == 1) {
+        return Optional.of(resolvedOperators.get(0));
+      } else if (resolvedOperators.size() > 1) {
+        throw new RuntimeException(
+            "Found two SqlOperators: " + resolvedOperators + " for ScalarFunction : " + expr);
```
jacques-n, line 110: +string format.

isthmus/src/main/java/io/substrait/isthmus/expression/FunctionConverter.java
```diff
@@ -69,9 +73,48 @@ public FunctionConverter(
...
+      }
+    }
+    return Optional.empty();
+  }
+
+  protected abstract Map<SqlOperator, FunctionMappings.TypeBasedResolver> getTypeBasedResolver();
```
jacques-n, line 116: Wondering whether we should expose a more concrete function here instead of the internals. (?)

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
```diff
@@ -0,0 +1,126 @@
...
+            .getCondition()
+            .accept(
+                new ExpressionRexConverter(
+                    relOptCluster.getTypeFactory(), scalarFunctionConverter));
+    return LogicalFilter.create(input, filterCondition);
+  }
+
+  @Override
+  public RelNode visit(Join join) throws RuntimeException {
+    throw new UnsupportedOperationException("Join is not supported in SubstraitToSql!");
```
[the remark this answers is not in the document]
jinfengni: I added AbStractRelVisitor / AbstractExpressionVisitor with fallback method.  Since #30 has the proposed abstract class, will rebase once #30 is merged.

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
```diff
@@ -0,0 +1,126 @@
...
+  public RelNode visit(NamedScan namedScan) throws RuntimeException {
+    RelDataType rowType =
+        TypeConverter.convert(
+            relOptCluster.getTypeFactory(),
+            namedScan.getRecordType(),
+            namedScan.getInitialSchema().names());
+    CalciteSchema.TableEntry entry =
+        catalogReader.getRootSchema().getTable(namedScan.getNames().get(0), false);
+    if (entry == null) {
+      throw new RuntimeException("Table " + namedScan.getNames() + " not found in catalog");
```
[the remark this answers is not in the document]
jinfengni: done

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
jinfengni: The current code only supports scan/project/filter, and the static constructors are relatively simple. 

I switched to RelBuilder to construct RelNode. However, I have to remove the checking of [1].  The RelBuilder seems to create a catalogreader and a new type system, different from the one when tables are parsed/added. 


1. https://github.com/substrait-io/substrait-java/blob/main/isthmus/src/main/java/io/substrait/isthmus/SqlToSubstrait.java#L255-L256

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
```diff
@@ -0,0 +1,126 @@
...
+    throw new UnsupportedOperationException("Join is not supported in SubstraitToSql!");
+  }
+
+  @Override
+  public RelNode visit(NamedScan namedScan) throws RuntimeException {
+    RelDataType rowType =
+        TypeConverter.convert(
+            relOptCluster.getTypeFactory(),
+            namedScan.getRecordType(),
+            namedScan.getInitialSchema().names());
```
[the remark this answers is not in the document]
jinfengni: After switched to using RelBuilder,  looks like I do not have to build row type, as it would be retrieved from relOptSchema when RelBuilder creates a `LogicalScan`. So, the above code is removed.

isthmus/src/main/java/io/substrait/isthmus/SubstraitToSql.java
jinfengni: I have two steps:  
 - `substraitRelToCalciteRel` : convert from substrait Rel to Calcite Rel
 - toSql : Calcite Rel to string SQL.

isthmus/src/main/java/io/substrait/isthmus/expression/FunctionConverter.java
jinfengni: changed.

isthmus/src/main/java/io/substrait/isthmus/expression/FunctionConverter.java
jinfengni: done

isthmus/src/test/java/io/substrait/isthmus/Substrait2SqlTestNoValidation.java
jinfengni: Add the sql => substrait => calcite plan => substrait in `Substrait2SqlTest`. 

For query with implicit cast involving literal, such as l_partkey > 100,  the first substrait plan will contain a CAST(100 as bigint, the second substrait plan will have 100 as i64 literal.  I have not located the exact places where the CAST is removed.  So, for now, the testcases do not contain cases with implicit cast over literals. 

```
Cast{type=I64{nullable=false}, input=I32Literal{nullable=false, value=100}

I64Literal{nullable=false, value=100}], outputType=Bool{nullable=false}
```

isthmus/src/main/java/io/substrait/isthmus/expression/FunctionMappings.java
jinfengni: The existing mapping of `SCALAR_SIGS` is SqlOperator -> a **category of substrait functions**. 

```
s(SqlStdOperatorTable.PLUS, "add")
s(SqlStdOperatorTable.DATETIME_PLUS, "add"
```

For instance, "add" will expand into add:i64_i64, add:i32_i32, .... in `FunctionConverter.signatures`.  When convert from calcite SqlOperator to substrait function,  we could use argument types to find the matched substrait function. 

When convert from substrait function to SqlOperator, using the above mapping,  a add:i64_i64 would map to `SqlStdOperatorTable.PLUS` and `SqlStdOperatorTable.DATETIME_PLUS`.   The new added `SCALAR_RESOLVER` is to provide a way to different based on the output type.   Calcite code seems also switch to `DATETIME_PLUS` based on output type [1].

If we want to avoid this two separate sets, we need have mapping from SqlOperator <-->. **individual substrait function** in the form of  `add:i64_i64_i64`, rather than a map to a category of functions.

1. https://github.com/apache/calcite/blob/main/core/src/main/java/org/apache/calcite/sql2rel/StandardConvertletTable.java#L1040-L1044

jinfengni:
@jacques-n , thanks a lot for your feedback. I changed code based on your comments. Can you please take another look?

isthmus/src/main/java/io/substrait/isthmus/SubstraitToSql.java
jinfengni: I moved the duplicative code into SqlConverterBase.  The two class `SqlToSubstrait` and `SubstraitToSql` will share those common codes.

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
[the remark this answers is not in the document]
jinfengni: I rebased on top of #30, and use the two Abstract visitors.

isthmus/src/main/java/io/substrait/isthmus/SqlConverterBase.java
```diff
@@ -0,0 +1,187 @@
...
+import org.apache.calcite.sql.parser.SqlParser;
+import org.apache.calcite.sql.parser.SqlParserPos;
+import org.apache.calcite.sql.parser.ddl.SqlDdlParserImpl;
+import org.apache.calcite.sql.validate.SqlValidator;
+import org.apache.calcite.sql.validate.SqlValidatorCatalogReader;
+import org.apache.calcite.sql.validate.SqlValidatorImpl;
+import org.apache.calcite.sql2rel.SqlToRelConverter;
+import org.apache.calcite.util.Pair;
+
+public class SqlConverterBase {
```
jacques-n, line 40: Should this be exposed outside this package? Seems like maybe the only users should be other things in this package.

isthmus/src/main/java/io/substrait/isthmus/SqlToSubstrait.java
```diff
@@ -78,23 +48,13 @@ public Plan execute(String sql, Function<List<String>, NamedStruct> tableLookup)
...
-          rootSchema.add(t.getName(), t);
-        }
-      }
-    }
+    var pair = registerCreateTables(tables);
+    return executeInner(sql, factory, pair.left, pair.right);
+  }
 
-    return executeInner(sql, factory, validator, catalogReader);
+  public RelRoot sqlToRelNode(String sql, List<String> tables) throws SqlParseException {
```
jacques-n, line 55: Given the shortening of these two functions, let's collapse them into one.

isthmus/src/main/java/io/substrait/isthmus/SqlToSubstrait.java
```diff
@@ -153,113 +84,30 @@ private Plan executeInner(
...
-    }
-
-    @Override
-    public RelDataType getRowType(RelDataTypeFactory typeFactory) {
-      if (factory != typeFactory) {
-        throw new IllegalStateException("Different type factory than previously used.");
-      }
-      return type;
-    }
+  private RelRoot sqlToRelNode(
```
jacques-n, line 87: I might collapse this with methods above.

isthmus/src/main/java/io/substrait/isthmus/expression/ExpressionRexConverter.java
```diff
@@ -0,0 +1,159 @@
...
+  public ExpressionRexConverter(
+      RelDataTypeFactory typeFactory, ScalarFunctionConverter scalarFunctionConverter) {
+    this.typeFactory = typeFactory;
+    this.rexBuilder = new RexBuilder(typeFactory);
+    this.scalarFunctionConverter = scalarFunctionConverter;
+  }
+
+  @Override
+  public RexNode visit(Expression.NullLiteral expr) throws RuntimeException {
+    return null;
```
jacques-n, line 32: This looks wrong. I think you should be making a typed null literal here, no?

isthmus/src/main/java/io/substrait/isthmus/expression/ExpressionRexConverter.java
```diff
@@ -0,0 +1,159 @@
...
+  public RexNode visit(Expression.BinaryLiteral expr) throws RuntimeException {
+    // Calcite RexLiteral only takes ByteString
+    return rexBuilder.makeLiteral(
+        new ByteString(expr.value().toByteArray()),
+        TypeConverter.convert(typeFactory, expr.getType()));
+  }
+
+  @Override
+  public RexNode visit(Expression.TimeLiteral expr) throws RuntimeException {
+    // Expression.TimeLiteral is Micros, while RexLiteral assumes Milliseconds for time type
```
jacques-n, line 85: This comment and the one below are misleading. It sounds like we're losing resolution but we aren't.

isthmus/src/main/java/io/substrait/isthmus/expression/ExpressionRexConverter.java
```diff
@@ -0,0 +1,159 @@
...
+    int milliS = (int) (TimeUnit.MICROSECONDS.toMillis(expr.value()));
```
jacques-n, line 86: Are you sure this is right? I'm thinking that millis should be on a day boundary followed by nanos for anything within a day. E.g. `TimeUnit.DAYS.toMillis(TimeUnit.MICROSECONDS.toDays(expr.value()))` instead of what you have.

isthmus/src/main/java/io/substrait/isthmus/expression/ExpressionRexConverter.java
```diff
@@ -0,0 +1,159 @@
...
+      return rexInputRef;
+    }
+
+    return visitFallback(expr);
+  }
+
+  @Override
+  public RexNode visitFallback(Expression expr) {
+    throw new UnsupportedOperationException(
+        String.format(
```
jacques-n, line 155: You might want to include the actual expression, not just the class. For example, a fieldReference that includes an array index wouldn't be obvious from this message (I don't think).

isthmus/src/main/java/io/substrait/isthmus/SqlConverterBase.java
```diff
@@ -0,0 +1,187 @@
...
+  protected final RelDataTypeFactory factory;
```
jacques-n, line 41: A lot of stuff is protected and this class is public. Are both of these things necessary?

isthmus/src/main/java/io/substrait/isthmus/expression/FunctionConverter.java
```diff
@@ -69,9 +73,51 @@ public FunctionConverter(
...
+      }
+    }
+
     this.signatures = matcherMap;
   }
 
+  protected Optional<SqlOperator> getSqlOperatorFromSubstraitFunc(
+      Expression.ScalarFunctionInvocation expr) {
+    var resolver = getTypeBasedResolver();
+    if (!substraitFuncKeyToSqlOperatorMap.containsKey(expr.declaration().key())) {
```
jacques-n, line 90: for all the returns below, can we add a verify step that output derivation is consistent between what we determined in substrait plan and what is in calcite plan?

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
```diff
@@ -0,0 +1,103 @@
...
+  private final CalciteCatalogReader catalogReader;
+
+  private final SimpleExtension.ExtensionCollection extensions;
+
+  private final ScalarFunctionConverter scalarFunctionConverter;
+
+  private final RelBuilder relBuilder;
+
+  public SubstraitRelNodeConverter(
+      SimpleExtension.ExtensionCollection extensions,
```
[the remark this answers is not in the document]
jinfengni: I moved SqlParser.Config to base class, and pass SqlParser.Config wherever needed. Both `SqlToSubstrait` and `SubstraitToSql` use the same version of parser config.

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
```diff
@@ -0,0 +1,103 @@
...
+import org.apache.calcite.plan.RelOptCluster;
+import org.apache.calcite.plan.RelTraitDef;
+import org.apache.calcite.prepare.CalciteCatalogReader;
+import org.apache.calcite.rel.RelNode;
+import org.apache.calcite.rex.RexNode;
+import org.apache.calcite.sql.parser.SqlParser;
+import org.apache.calcite.tools.Frameworks;
+import org.apache.calcite.tools.RelBuilder;
+
+public class SubstraitRelNodeConverter extends AbstractRelVisitor<RelNode, RuntimeException> {
```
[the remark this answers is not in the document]
jinfengni: added javadoc.

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
```diff
@@ -0,0 +1,103 @@
...
+    RelNode input = filter.getInput().accept(this);
+    RexNode filterCondition =
+        filter
+            .getCondition()
+            .accept(
+                new ExpressionRexConverter(
+                    relOptCluster.getTypeFactory(), scalarFunctionConverter));
+    return relBuilder.push(input).filter(filterCondition).build();
+
+    // return LogicalFilter.create(input, filterCondition);
```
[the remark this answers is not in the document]
jinfengni: done.

isthmus/src/main/java/io/substrait/isthmus/SqlConverterBase.java
jinfengni: right. it's only accessed within the package. Removed 'public' access.

isthmus/src/main/java/io/substrait/isthmus/expression/ExpressionRexConverter.java
jinfengni: The comment is not right. I changed the comment.

isthmus/src/main/java/io/substrait/isthmus/expression/ExpressionRexConverter.java
jinfengni: You are right. There is a bug.  We need a `second` boundary followed by nanos for any fractional second part.

```
    long microSec = expr.value();
    long seconds = TimeUnit.MICROSECONDS.toSeconds(microSec);
    int fracSecondsInNano =
        (int) (TimeUnit.MICROSECONDS.toNanos(microSec) - TimeUnit.SECONDS.toNanos(seconds));
    TimeString timeString =
        TimeString.fromMillisOfDay((int) TimeUnit.SECONDS.toMillis(seconds))
            .withNanos(fracSecondsInNano);
```

Previously, I used 'miliSeconds' as boundary. But turns out that Calcit's TimeString will remove the fractional milliseconds part, if we call `withNanos`.

I added two new UT `tTimeWithMicroSecond`, `tTimeWithNanoSecond`. The second is just to verify Calcite's time literal truncates to microseconds precision, the same precision as Substrait.

1. https://github.com/apache/calcite/blob/main/core/src/main/java/org/apache/calcite/util/TimeString.java#L87-L93

jinfengni: Also, I fixed the similar bug in TimeStampLiteral, changing to use seconds boundary. And add new UT testcase for such coverage.

isthmus/src/main/java/io/substrait/isthmus/expression/ExpressionRexConverter.java
jinfengni: I added actual expression in the Exception message.

isthmus/src/main/java/io/substrait/isthmus/SqlConverterBase.java
jinfengni: remove public/protected.  `SqlConvererBase` is only used in the package.

isthmus/src/main/java/io/substrait/isthmus/expression/FunctionConverter.java
jinfengni: I added a checking after `getSqlOperatorFromSubstraitFunc` returned with a SqlOperator.  Check SqlOperator's returntype, by using it's `SqlReturnTypeInference`, and infer the return type from list of operands.  The return type is compared with substrait function's return type.

jinfengni:
@jacques-n , Thank you for your feedback. I changed the code according to your comments. Can you please take another look?

isthmus/src/main/java/io/substrait/isthmus/expression/FunctionMappings.java
jinfengni: Yes, I'll think about it. I tried to use the new added way to do a return-type based verification, to choose the calcite SqlOperator.  But a datetime + operator's return type inference would not work for numeric +'s operands binding, that approach does not work. So, the checking is only applied `after` we pick up the operator, but could not use `before` among a set of candidate sql operators.

jacques-n (commented):
It seems like a couple review comments are not addressed. Correct?

isthmus/src/main/java/io/substrait/isthmus/expression/ExpressionRexConverter.java
jinfengni: Changed to return a typed null literal.  Also changed ut to cover the substrait -> rex for typed null case.

isthmus/src/main/java/io/substrait/isthmus/expression/FunctionConverter.java
jinfengni: Are you suggesting we change `getTypeBasedResolver` to a private function of `FunctionConverter`? Not sure if I understand it.  I changed it to private, and return `OPERATOR_RESOLVER` for both scalar and aggregate operator.

jinfengni: Also, after identify convert a substrait function invocation to a SqlOperator, using the above `getTypeBasedResolver`,  previously I added a check if the SqlOperator's returnTypeInference will derive to the same return type, by creating a `RexCallBinding`.  Turns out that approach will not work.  Calcite's SqlOperator's returnTypeInference seems to only work for SqlCallBinding, not for all `RexCallBinding`.   When I worked on interval literal and functions, I found that DateTimeSubstract or DateTimePlus operator assumes 3 arguments for SqlCall, and 2 arguments for RexCall.   The `returnTypeInference` requires 2-rd argment, which is only applicable to SqlCall.  

As such, I have to remove the newly added checking in `ExpressionRexConverter`.

1. https://github.com/apache/calcite/blob/main/core/src/main/java/org/apache/calcite/sql/fun/SqlDatetimeSubtractionOperator.java#L37-L38
2. https://github.com/apache/calcite/blob/main/core/src/main/java/org/apache/calcite/sql/fun/SqlDatetimeSubtractionOperator.java#L50

jinfengni:
I went to "File changed" tab, and saw two comments that I did not address.  I changed the code and put responses to the comments.  Have not figured out a quick way to find the comments that I had not addressed.

jacques-n (approved):

PR #29, opened 2022-05-26
Generate extension uris and functions from pojos instead of storing

JamesRTaylor:
Fix bug in Plan proto conversion caused by storing extension uris and functions in Plan pojo and using when generating proto instead of regenerating.

### DIFF
--- a/core/src/main/java/io/substrait/plan/Plan.java
+++ b/core/src/main/java/io/substrait/plan/Plan.java
@@ -1,8 +1,6 @@
 package io.substrait.plan;
 
 import io.substrait.proto.AdvancedExtension;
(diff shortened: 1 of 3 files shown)


jacques-n (approved):
It would be good to add a test to make sure we don't regress on this. Since it is a clear bug, I'm going to merge this as is.

JamesRTaylor:
FYI, I have a test in https://github.com/substrait-io/substrait-java/pull/27 over [here](https://github.com/substrait-io/substrait-java/pull/27/files#diff-9fc3786a9b17c5b0542315424f153c6bc31976544859771859989d99d1c8f247), but I forgot to move it into this PR.

PR #31, opened 2022-05-26
Add publishToLocalMaven command

rymurr:
this allows for the java modules to be published to a local maven repository (eg `~/.m2/repository` for integration with other local projects.

### DIFF
--- a/isthmus/build.gradle.kts
+++ b/isthmus/build.gradle.kts
@@ -1,9 +1,12 @@
 plugins {
+  `maven-publish`
   id("java")
   id("idea")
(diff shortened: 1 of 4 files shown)


isthmus/build.gradle.kts
```diff
@@ -1,10 +1,13 @@
 plugins {
+  `maven-publish`
```
jacques-n, line 2: Is backticks really the right thing? (here and elsewhere)?

rymurr: I was wondering the same. That's how it's defined in the docs though it doesn't work otherwise.

PR #34, opened 2022-06-01
[ISTHMUS] Add substrait to sql support for join and aggregate.

jinfengni:


isthmus/src/test/java/io/substrait/isthmus/Substrait2SqlTest.java
```diff
@@ -42,7 +42,78 @@ public void simpleTestDecimal() throws Exception {
...
+
+  @Test
+  public void simpleTestAgg() throws Exception {
+    test(
+        "select l_partkey, count(l_tax), COUNT(distinct l_discount) from lineitem group by l_partkey");
+  }
+
+  @Test
+  public void simpleTestAggNoGB() throws Exception {
+    test("select count(l_tax), count(distinct l_discount) from lineitem");
```
JamesRTaylor, line 65: Can we add a test that uses `approx_count_distinct`? Something like this:
```
test("select approx_count_distinct(l_discount) from lineitem");
```

isthmus/src/test/java/io/substrait/isthmus/Substrait2SqlTest.java
```diff
@@ -42,7 +42,78 @@ public void simpleTestDecimal() throws Exception {
...
+    // do not verify pojo in round trip test, because of implicit cast(1 as decimal) is removed in
+    // round trip
+    // conversion.
+    test(
+        "select\n"
+            + "  l_returnflag,\n"
+            + "  l_linestatus,\n"
+            + "  sum(l_quantity) as sum_qty,\n"
+            + "  sum(l_extendedprice) as sum_base_price,\n"
+            + "  sum(l_extendedprice * (1 - l_discount)) as sum_disc_price,\n"
```
JamesRTaylor, line 96: If you use `1.0` instead of `1` or if you add the cast to decimal, does the roundtrip equality succeed?

isthmus/src/test/java/io/substrait/isthmus/Substrait2SqlTest.java
```diff
@@ -58,7 +129,10 @@ private void test(String query) throws Exception {
     Rel pojoRel2 =
         SubstraitRelVisitor.convert(RelRoot.of(relnodeRoot, SqlKind.SELECT), EXTENSION_COLLECTION);
```
JamesRTaylor, line 130: With fix https://github.com/substrait-io/substrait-java/pull/35, does it work to use `ProtoPlanConverter` instead?

isthmus/src/test/java/io/substrait/isthmus/Substrait2SqlTest.java
jinfengni: good idea. Using 1.0 will avoid the implicit cast difference in the round trip conversion.

isthmus/src/test/java/io/substrait/isthmus/Substrait2SqlTest.java
jinfengni: for  `select approx_count_distinct(l_discount) from lineitem"`,  from sql -> calcite rel, we have approximate/distinct for count() function, but after calcite rel -> substrait rel , the `approximate` flag is lost during `SubstraitRelVisitor`.

Looks like we have to fix the issue calcite rel -> substrait rel.

JamesRTaylor: Does it work going the other direction (from substrait plan -> sql)? You can generate a substrait plan using `count(distinct l_discount)` and then run this code to replace them with `approx_count_distinct`:
```
    ReplaceCountDistinctWithApprox action = new ReplaceCountDistinctWithApprox();
    Plan newPlan = action.modify(oldPlan).orElse(oldPlan);
```
Finally, see if your code to go from substrait plan to sql writes the correct sql with the `approx_count_distinct` call.

JamesRTaylor:
LGTM

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
```diff
@@ -3,19 +3,27 @@
...
+import io.substrait.isthmus.expression.AggregateFunctionConverter;
 import io.substrait.isthmus.expression.ExpressionRexConverter;
 import io.substrait.isthmus.expression.ScalarFunctionConverter;
-import io.substrait.relation.AbstractRelVisitor;
-import io.substrait.relation.Filter;
-import io.substrait.relation.NamedScan;
-import io.substrait.relation.Project;
-import io.substrait.relation.Rel;
+import io.substrait.proto.AggregateFunction;
+import io.substrait.relation.*;
```
JamesRTaylor, line 10: Minor nit - remove wildcard import. You may want to increase default in IntelliJ which annoyingly defaults to this after three imports from the same package.

isthmus/src/test/java/io/substrait/isthmus/SimplePlansTest.java
```diff
@@ -26,8 +26,7 @@ public void cd() throws IOException, SqlParseException {
     String[] values = asString("tpch/schema.sql").split(";");
     var creates = Arrays.stream(values).filter(t -> !t.trim().isBlank()).toList();
     // creates.forEach(System.out::println);
-    s.execute(
-        "select l_partkey, sum(distinct L_ORDERKEY) from lineitem group by l_partkey ", creates);
+    s.execute("select l_partkey, sum(L_ORDERKEY) from lineitem group by l_partkey ", creates);
```
JamesRTaylor, line 29: Why this change? Revert if not necessary.

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
jinfengni: Removed wildcard import. Thanks for the tips in Intellij setting.

isthmus/src/test/java/io/substrait/isthmus/SimplePlansTest.java
jinfengni: probably changed during debug & testing. 

reverted.

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
```diff
@@ -89,22 +109,128 @@ public RelNode visit(NamedScan namedScan) throws RuntimeException {
...
-                expr ->
-                    expr.accept(
-                        new ExpressionRexConverter(
-                            relOptCluster.getTypeFactory(), scalarFunctionConverter)))
+                gr ->
+                    gr.getExpressions().stream()
+                        .map(expr -> expr.accept(expressionRexConverter))
+                        .toList())
             .toList();
+    List<RexNode> groupExprs =
```
jacques-n, line 150: It feels like the grouping set dynamics here are wrong. Do you have a test case to verify the correct conversion?

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
```diff
@@ -89,22 +109,128 @@ public RelNode visit(NamedScan namedScan) throws RuntimeException {
...
+        groupExprLists.stream().flatMap(Collection::stream).collect(Collectors.toList());
+    RelBuilder.GroupKey groupKey = relBuilder.groupKey(groupExprs, groupExprLists);
 
-    return relBuilder.push(child).project(rexList).build();
+    List<AggregateCall> aggregateCalls =
+        aggregate.getMeasures().stream().map(m -> fromMeasure(m)).toList();
```
jacques-n, line 155: For this kind of thing, using method handle `.map(this::fromMeasure).toList()` is more readable.

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
jacques-n: Also, how do we map the dynamics of grouping set identifiers between the Calcite and Substrait? (In Substrait we identify the grouping set the result is from by referencing it via ordinal).

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
```diff
@@ -89,22 +109,128 @@ public RelNode visit(NamedScan namedScan) throws RuntimeException {
...
+        aggregateFunctionConverter.getSqlOperatorFromSubstraitFunc(
+            measure.getFunction().declaration().key(), measure.getFunction().outputType());
+    if (!operator.isPresent()) {
+      throw new UnsupportedOperationException(
+          String.format(
+              "Unable to find binding for call %s", measure.getFunction().declaration().name()));
+    }
+    List<Integer> argIndex = new ArrayList<>();
+    for (RexNode arg : arguments) {
+      if (!(arg instanceof RexInputRef)) {
```
jacques-n, line 174: I think we need to fix this. In the case that there are compound expressions, I believe the rewrite to Calcite needs to be to insert a project with the calculation and refer to it from the aggregate function.

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
```diff
@@ -89,22 +109,128 @@ public RelNode visit(NamedScan namedScan) throws RuntimeException {
...
+        throw new UnsupportedOperationException(
+            String.format(
+                "aggregation function only support column as argument. % is not support", arg));
+      }
+      argIndex.add(((RexInputRef) arg).getIndex());
+    }
+
+    boolean distinct =
+        measure.getFunction().invocation()
+            == AggregateFunction.AggregationInvocation.AGGREGATION_INVOCATION_DISTINCT;
```
jacques-n, line 184: Seems like we should put a check here to ensure invocation is not some unexpected type.

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
```diff
@@ -89,22 +109,128 @@ public RelNode visit(NamedScan namedScan) throws RuntimeException {
...
+
+    SqlAggFunction aggFunction;
+    RelDataType returnType =
+        TypeConverter.convert(relOptCluster.getTypeFactory(), measure.getFunction().getType());
+
+    if (operator.get() == SqlStdOperatorTable.COUNT) {
```
jacques-n, line 190: This looks like it should be turned into something. I also don't really understand it. It looks like you're saying if a == foo, set b to foo. Why not just set b to foo in the cases a is one of the defined items.

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
```diff
@@ -89,22 +109,128 @@ public RelNode visit(NamedScan namedScan) throws RuntimeException {
...
+      throw new UnsupportedOperationException(
+          String.format(
+              "Unsupported agg operator % for agg function % ",
+              operator.get(), measure.getFunction().declaration().name()));
+    }
+
+    int filterArg = -1;
+    if (measure.getPreMeasureFilter().isPresent()) {
+      RexNode filter = measure.getPreMeasureFilter().get().accept(expressionRexConverter);
+      if (!(filter instanceof RexInputRef)) {
```
jacques-n, line 206: as above, feels like this should be moved to project.

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
jinfengni: `Substrait2SqlTest::simpleTestGroupingSets` has tests covering query with grouping sets.  including `()`. It also has query with `grouping sets` + `group by`.

Substrait uses ordinal, while Calcite uses bitsets.  When converting from substrait to calcite, ordinals are converted into RexInputReference.  The list of RexInputReference in `groupExprs` will converted into indexes in `bitsets`, when we call:

 `RelBuilder.groupKey(Iterable<? extends RexNode> nodes,
      Iterable<? extends Iterable<? extends RexNode>> nodeLists)`

jinfengni: I just added one group sets test, same column appear in the group sets multiple times. It seems the conversion work correctly.

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
jinfengni: changed.

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
jinfengni: Right, when Calcite converts from sql to RelNode, it would insert a project for compound expression.  The conversion of substrait -> calcite should process in a similar way, by adding a project.

However, for now, the substrait plan comes from Calcite RelNode, which has a project for compound expression.  In future, we would add substrait plan utility similar to Calcite's RelBuild/RexBuilder to generate arbitrary substrait plan/expr,  we are required to handle compound expression.  Without such substrait plan utility, it seems not straightforward to write a test with a substrait plan with compound expression in the agg call's argument.    

Shall we open a issue to track this gap and leave it as a TODO?

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
jinfengni: If invocation is other type,  `distinct` is `false`.   For what reason, we need put such check?

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
jinfengni: I changed to:
```
   if (operator.get() instanceof SqlAggFunction) {
      aggFunction = (SqlAggFunction) operator.get();
    }
```

`operator` is `SqlOperator`,  a base class.  I could not directly assign it to `aggFunction`.

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
jinfengni: Similar to compound expression for `argument`.  Maybe as a TODO?

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
jacques-n: That's fine but we should state in the exception that this just isn't implemented yet (as opposed to being impossible).

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
jacques-n: The problem is someone may introduce another branch of the enum in the future. If they do, this path will blindly ignore it. (Basically, the default case of switch).

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
jacques-n: As above, let's make sure the error message makes clear that this is a gap in implementation, not capability.

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
jinfengni: Changed the message in the exception to "something is not implemented yet". 

Also, change the exception  from `UnsupportedOperationException`  to `IllegalArgumentException`, when the operator binding is not successful, or the operator is not the expected type.

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
jinfengni: Changed to use switch statement with default cases.

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
jinfengni: Changed error message to "something is not implemented yet".

jacques-n:
This looks good to me. Does this look good to you @JamesRTaylor ?

JamesRTaylor:
LGTM too, @jacques-n

PR #35, opened 2022-06-01
Hack to get expected rootType for expression conversion

JamesRTaylor:


core/src/main/java/io/substrait/expression/FieldReference.java
```diff
@@ -214,7 +214,10 @@ private static FieldReference of(
       if (i == 0) {
         var last = segments.get(0);
         reference =
-            type == null ? last.constructOnExpression(expression) : last.constructOnRoot(type);
+            type == null
+                ? last.constructOnExpression(expression)
+                : last.constructOnRoot(
+                    ((Type.Struct) type).fields().get(((StructField) last).offset()));
```
JamesRTaylor, line 220: We don't know the root type in advance, so we pass in the type of the input. The type of the field reference can then be computed based on indexing into the fields. Unclear what to do if `type` is not a struct.

JamesRTaylor:
cc. @jinfengni

jinfengni:
LGTM.

core/src/main/java/io/substrait/expression/FieldReference.java
```diff
@@ -149,7 +149,12 @@ public FieldReference constructOnExpression(Expression expr) {
 
     @Override
     public FieldReference constructOnRoot(Type type) {
```
jacques-n, line 151: Let's modify this to take `Type.Struct`. That way it works cleanly with `constructOnRoot(myRel.getRecordType())`

core/src/main/java/io/substrait/expression/FieldReference.java
```diff
@@ -149,7 +149,12 @@ public FieldReference constructOnExpression(Expression expr) {
...
-      return FieldReference.newRootStructReference(offset(), type);
+      if (!(type instanceof Type.Struct)) {
+        throw new IllegalArgumentException(
+            "Root type expected to be Type.Struct but is " + type.getClass());
+      }
+      return FieldReference.newRootStructReference(
```
jacques-n, line 156: Let's add a check on the offset() being within the range of available fields. It would be more helpful to give a more meaningful error here than aioob with a generic message.

core/src/main/java/io/substrait/expression/FieldReference.java
JamesRTaylor: Done

core/src/main/java/io/substrait/expression/FieldReference.java
JamesRTaylor: Added check

JamesRTaylor:
Thanks for the review, @jacques-n . I've updated the PR.

jacques-n (approved):
LGTM. Thanks for changes and identifying the issue.

PR #39, opened 2022-06-03
[ISTHMUS] Support approx_count_distinct in substrait to sql conversion 

jinfengni:


isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
```diff
@@ -1,21 +1,30 @@
...
+import io.substrait.isthmus.expression.AggregateFunctionConverter;
 import io.substrait.isthmus.expression.ExpressionRexConverter;
 import io.substrait.isthmus.expression.ScalarFunctionConverter;
-import io.substrait.relation.AbstractRelVisitor;
-import io.substrait.relation.Filter;
-import io.substrait.relation.NamedScan;
-import io.substrait.relation.Project;
-import io.substrait.relation.Rel;
+import io.substrait.proto.AggregateFunction;
+import io.substrait.relation.*;
```
JamesRTaylor, line 11: Minor nit - remove wildcard import. You may want to increase default in IntelliJ which annoyingly defaults to this after three imports from the same package.

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
```diff
@@ -89,22 +104,133 @@ public RelNode visit(NamedScan namedScan) throws RuntimeException {
...
+      throw new UnsupportedOperationException(
+          String.format(
+              "Unable to find binding for call %s", measure.getFunction().declaration().name()));
+    }
+    List<Integer> argIndex = new ArrayList<>();
+    for (RexNode arg : arguments) {
+      if (!(arg instanceof RexInputRef)) {
+        throw new UnsupportedOperationException(
+            String.format(
+                "aggregation function only support column as argument. % is not support", arg));
```
JamesRTaylor, line 172: Prepend "Unsupported " here so it says "Unsupported aggregation function ..."

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
```diff
@@ -89,22 +104,133 @@ public RelNode visit(NamedScan namedScan) throws RuntimeException {
...
+      throw new UnsupportedOperationException(
+          String.format(
+              "Unsupported agg operator % for agg function % ",
+              operator.get(), measure.getFunction().declaration().name()));
+    }
+
+    int filterArg = -1;
+    if (measure.getPreMeasureFilter().isPresent()) {
+      RexNode filter = measure.getPreMeasureFilter().get().accept(expressionRexConverter);
+      if (!(filter instanceof RexInputRef)) {
```
JamesRTaylor, line 206: Consider factoring this check out to a helper method as it's repeated above.

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
```diff
@@ -89,22 +104,133 @@ public RelNode visit(NamedScan namedScan) throws RuntimeException {
...
+        returnType,
+        null);
   }
 
   @Override
   public RelNode visitFallback(Rel rel) throws RuntimeException {
     throw new UnsupportedOperationException(
         String.format(
-            "Rel $ of type %s not handled by visitor type %s.",
+            "Rel % of type %s not handled by visitor type %s.",
```
JamesRTaylor, line 233: What does "%" do as opposed to "%s" in format string? Just curious

isthmus/src/test/java/io/substrait/isthmus/RelCopyOnWriteVisitorTest.java
```diff
@@ -105,6 +103,29 @@ public void replaceCountDistincts() throws IOException, SqlParseException {
...
+    assertPlanRoundrip(newPlan);
+
+    // convert newPlan back to sql
+    var pojoRel = newPlan.getRoots().get(0).getInput();
+    String[] values = asString("tpch/schema.sql").split(";");
+    var creates = Arrays.stream(values).filter(t -> !t.trim().isBlank()).toList();
+    RelNode relnodeRoot = new SubstraitToSql().substraitRelToCalciteRel(pojoRel, creates);
+    String newSql = SubstraitToSql.toSql(relnodeRoot);
+    assertTrue(newSql.toUpperCase().contains("APPROX_COUNT_DISTINCT"));
+    System.out.println(newSql);
```
JamesRTaylor, line 126: Remove System.out.println

isthmus/src/test/java/io/substrait/isthmus/RelCopyOnWriteVisitorTest.java
```diff
@@ -105,6 +103,29 @@ public void replaceCountDistincts() throws IOException, SqlParseException {
...
+    assertEquals(0, new CountCountDistinct().getCountDistincts(newPlan));
+    assertPlanRoundrip(newPlan);
+
+    // convert newPlan back to sql
+    var pojoRel = newPlan.getRoots().get(0).getInput();
+    String[] values = asString("tpch/schema.sql").split(";");
+    var creates = Arrays.stream(values).filter(t -> !t.trim().isBlank()).toList();
+    RelNode relnodeRoot = new SubstraitToSql().substraitRelToCalciteRel(pojoRel, creates);
+    String newSql = SubstraitToSql.toSql(relnodeRoot);
+    assertTrue(newSql.toUpperCase().contains("APPROX_COUNT_DISTINCT"));
```
JamesRTaylor, line 125: Stronger assert would be good her. Couple of things here would be nice:
- Compare equality of original SQL where you replace `count(distinct <col>)` with `approx_count_distinct(<col>)`
- Convert newSql back to Plan using buildPlanFromQuery and then to proto, and compare to newPlan converted to proto.

isthmus/src/test/java/io/substrait/isthmus/Substrait2SqlTest.java
```diff
@@ -58,7 +146,10 @@ private void test(String query) throws Exception {
     Rel pojoRel2 =
         SubstraitRelVisitor.convert(RelRoot.of(relnodeRoot, SqlKind.SELECT), EXTENSION_COLLECTION);
 
-    Assertions.assertEquals(pojoRel, pojoRel2);
+    if (assertPojo) {
```
JamesRTaylor, line 149: If never using, remove this condition and argument. Best if we don't need it (though we might eventually)

JamesRTaylor:
A few minor comments, but this looks very good to me.

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
jinfengni: done

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
jinfengni: done.

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
jinfengni: Thanks for catching this bug.  My eyes were not sharp enough to recognize these typos.  Changed to `%s` in all error messages.

isthmus/src/test/java/io/substrait/isthmus/RelCopyOnWriteVisitorTest.java
jinfengni: done

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
jinfengni: Factor the common codes into a private method.

isthmus/src/test/java/io/substrait/isthmus/RelCopyOnWriteVisitorTest.java
jinfengni: For the first equality checking of SQL:
   - I feel comparing text is not a very stable way, since any format changing in Calcite's  RelToSql class could lead to testcase failure. 
   - Also, the converted `newSQL` contains a different field name `$f0` for the expression. That issue will be addressed later on.  Before that is fixed, could not compare SQL directly. 
 
For the second checking of substrait pojo/proto plan : 
- This pr adds support of `approx_count_distinct()` in substrait rel -> sql.  But the other direction (`sql` -> `substrait rel`) does not have support of `approx_count_distinct()`.  That probably need a separate PR, as I feel it's not straightforward to fix it (requires changing the function resolution logic). 
- So, for now, I could not add the checking of pojo/proto plan.  I put TODO to add new checking once 'sql' -> substrait work properly. 
-

jinfengni:
@JamesRTaylor thanks a lot for your review! I modified the pr based on your feedback.  Can you please take a look? Thanks!

jacques-n:
This PR appears to have multiple changes in it, not just what is in the title. Can you fix?

jinfengni:
I changed pr's title to reflect the changes made in this pr. 

Also, after changing to use Calcite's APPROX_COUNT_DISTINCT, and change the code of converting substrait function to Calcite SqlOperator,  we could convert back to `APPROX_COUNT_DISTINCT` directly; no need for special processing.  So, clean up some of earlier implementation.

JamesRTaylor:
LGTM

JamesRTaylor:
Would be good to rebase after merge of dependent PR and force push to get rid of merge commit.

jinfengni:
I have a fix to address `approx_count_distinct` conversion in sql -> substrait.  The simple idea is to replace Calcite's `count` + distinct/approximate with `approx_count_distinct` before converting into substrait function. I'll rebase and clean up the code, and submit for review.   

@jacques-n , please do not merge this PR for now.  Thanks!

isthmus/src/test/java/io/substrait/isthmus/PlanTestBase.java
```diff
@@ -51,4 +58,35 @@ protected void assertPlanRoundrip(Plan plan) throws IOException, SqlParseExcepti
...
+
+      // 3. Calcite Rel -> substrait rel
+      Rel pojoRel2 =
+          SubstraitRelVisitor.convert(
+              RelRoot.of(relnodeRoot, SqlKind.SELECT), EXTENSION_COLLECTION);
+
+      Assertions.assertEquals(pojoRel, pojoRel2);
+
+      // 4. Calcite Rel -> sql
+      String convertedSql = SubstraitToSql.toSql(relnodeRoot);
```
JamesRTaylor, line 88: Is this conversion just to see if you get an exception or not? Would it be possible to go back from sql to proto rel or pojo rel and compare it's the same?

isthmus/src/main/java/io/substrait/isthmus/SubstraitToSql.java
```diff
@@ -66,16 +66,22 @@ public RelNode substraitRelToCalciteRel(
...
   }
 
-  private static final SqlDialect DEFAULT_DIALECT =
-      new SqlDialect(
-          SqlDialect.EMPTY_CONTEXT
-              .withDatabaseProduct(SqlDialect.DatabaseProduct.SNOWFLAKE)
-              .withDatabaseProductName("SNOWFLAKE")
-              .withIdentifierQuoteString(null)
-              .withNullCollation(NullCollation.HIGH));
+  private static final SqlDialect.Context DEFAULT_DIALECT_CONTEXT =
```
JamesRTaylor, line 69: Just curious on what the implications of these defaults are. Does this determine whether column name references are double quoted? What about the case sensitive case? A detail, but eventually we'll need to deal with that. Also, with SQL dialect set to SNOWFLAKE, does `supportsApproxCountDistinct()` still return false? That seems like a bug, no? Maybe some comments would be good here.

isthmus/src/main/java/io/substrait/isthmus/expression/AggregateFunctionConverter.java
```diff
@@ -76,7 +78,13 @@ public Optional<AggregateFunctionInvocation> convert(
       AggregateCall call,
       Function<RexNode, Expression> topLevelConverter) {
 
-    FunctionFinder m = signatures.get(call.getAggregation());
+    // replace COUNT() + distinct == true and approximate == true with APPROX_COUNT_DISTINCT
+    // before converting into substrait function
+    SqlAggFunction aggFunction = call.getAggregation();
+    if (aggFunction == SqlStdOperatorTable.COUNT && call.isDistinct() && call.isApproximate()) {
+      aggFunction = SqlStdOperatorTable.APPROX_COUNT_DISTINCT;
```
JamesRTaylor, line 85: Does this mean we can go from sql -> plan for approx_count_distinct?

JamesRTaylor:
This looks really good. Just a couple of questions.

isthmus/src/test/java/io/substrait/isthmus/PlanTestBase.java
jinfengni: Previously, this conversion is to return a sql and assert Sql contains certain substring. Later, I changed to return RelNode, and verify RelNode tree contains certain expression.  

I tried to go back from sql to pojo rel and do the compare. The problem is `toSql` may loose some information. Consider the following query:
```
select p_size  from part where p_partkey > cast(100 as bigint)
``` 

In the  first round sql -> RelNode -> substrait rel ->  Calcite rel,  the condition is essentially represented as
`p_partkey > 100:bigint`.  However, from calcite rel -> sql, the condition becomes `p_partkey > 100`.  If we convert such sql back to substrait rel, the expression would contains a `cast`. Although they should return a same resultset, the comparison of pojo rel would fail.

In another words,  `toSql()` produces semantic equivalent yet different query.  Probably,  I should remove this `toSql` step.

isthmus/src/main/java/io/substrait/isthmus/SubstraitToSql.java
jinfengni: `.withIdentifierQuoteString` will determine how to quote an identifier. Passing a `null` means no quoted identifier. Passing `"` would double quote all identifiers, including table/column names. 

Calcite has a `SnowflakeDialect`[1]. It does set `supportsApproxCountDistinct` to true. However, it specifies double quote for all identifier, which would lead problem to table/column searching in catalog. 

I changed the code to just use Calcite's `EMPTY_CONTEXT` to construct a sql dilect, since `EMPTY_CONTEXT` defaults to use null as IdentifierQuoteString. Same as SnowflakeDialect, I set `supportsApproxCountDistinct` for the constructed sql dialect. 

1. https://github.com/apache/calcite/blob/main/core/src/main/java/org/apache/calcite/sql/dialect/SnowflakeSqlDialect.java#L28-L29

isthmus/src/main/java/io/substrait/isthmus/expression/AggregateFunctionConverter.java
jinfengni: Yes, now we can go sql -> substrait for approx_count_distinct.  Previously, in sql -> substrait, the code will convert `approx_count_distinct` to `count`, which is not correct.

JamesRTaylor:
LGTM

JamesRTaylor (approved):

jacques-n (approved):
LGTM.

PR #42, opened 2022-06-04
[ISTHMUS] Add --multistatement option to allow multiple sql statements

JamesRTaylor:
As a byproduct of having a multi-statement option, trailing semicolons will be ignored.

### DIFF
--- a/isthmus/src/main/java/io/substrait/isthmus/SqlToSubstrait.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/SqlToSubstrait.java
@@ -3,7 +3,6 @@ package io.substrait.isthmus;
 import io.substrait.expression.proto.FunctionCollector;
 import io.substrait.proto.Plan;
 import io.substrait.proto.PlanRel;
-import io.substrait.relation.Rel;
 import io.substrait.relation.RelProtoConverter;
 import io.substrait.type.NamedStruct;
 import java.util.List;
@@ -24,6 +23,37 @@ import org.apache.calcite.sql2rel.StandardConvertletTable;
 
 /** Take a SQL statement and a set of table definitions and return a substrait plan. */
 public class SqlToSubstrait extends SqlConverterBase {
+
+  public enum StatementBatching {
+    SINGLE_STATEMENT,
+    MULTI_STATEMENT
+  }
+
+  private final Options options;
+
+  public SqlToSubstrait() {
+    this.options = new Options();
+  }
+
+  public SqlToSubstrait(Options options) {
+    this.options = options;
+  }
+
+  public static class Options {
+    private final StatementBatching statementBatching;
+
+    public Options() {
+      statementBatching = StatementBatching.SINGLE_STATEMENT;
+    }
+
+    public Options(StatementBatching statementBatching) {
+      this.statementBatching = statementBatching;
+    }
+
+    public StatementBatching getStatementBatching() {
+      return this.statementBatching;
+    }
+  }
 
   public Plan execute(String sql, Function<List<String>, NamedStruct> tableLookup)
       throws SqlParseException {
@@ -52,12 +82,14 @@ public class SqlToSubstrait extends SqlConverterBase {
     return executeInner(sql, factory, pair.left, pair.right);
   }
 
-  public RelRoot sqlToRelNode(String sql, List<String> tables) throws SqlParseException {
+  // Package protected for testing
+  List<RelRoot> sqlToRelNode(String sql, List<String> tables) throws SqlParseException {
     var pair = registerCreateTables(tables);
-    return sqlToRelNode(sql, factory, pair.left, pair.right);
+    return sqlToRelNode(sql, pair.left, pair.right);
   }
 
-  public RelRoot sqlToRelNode(String sql, Function<List<String>, NamedStruct> tableLookup)
+  // Package protected for testing
+  List<RelRoot> sqlToRelNode(String sql, Function<List<String>, NamedStruct> tableLookup)
       throws SqlParseException {
     Function<List<String>, Table> lookup =
         id -> {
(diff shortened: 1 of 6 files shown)


jacques-n:
A plan can have multiple separate trees. What do you think about having an optional cli flag that allows multiple statements and puts them in separate trees in the plan?

jinfengni:
I also feel it makes sense to have another option to specify whether it's for single or multiple statements in `execute()` method call.   

Per substrait spec, a `Plan` could have multiple `PlanRel`.  I'm not sure if that's a good use case here.  How are we going to deal with extension urls used by different queries? What to do with the multiple result sets at the execution engine?

JamesRTaylor:
Thanks for the review, @jacques-n and @jinfengni. I've updated the PR to include a new CLI flag to support multiple statements. As a byproduct, terminating semicolons are ignored now.

jacques-n:
My assumption is that trailing semi-colon is allowed always. Multiple independent statements are only allowed if -m is passed. Is that correct?

JamesRTaylor:
> My assumption is that trailing semi-colon is allowed always. Multiple independent statements are only allowed if -m is passed. Is that correct?

Yes, exactly. I have tests for this.

isthmus/src/test/java/io/substrait/isthmus/SimplePlansTest.java
```diff
@@ -1,59 +1,53 @@
...
-            "select * from lineitem l, orders o WHERE o.o_orderkey = l.l_orderkey  and L_ORDERKEY > 10",
-            ddlStmts));
+  public void trailingSemicolon() throws IOException, SqlParseException {
+    assertProtoPlanRoundrip("select * from lineitem WHERE L_ORDERKEY > 10;");
   }
 
-  private void print(Plan plan) {
+  @Test
+  public void multiStatement() throws IOException, SqlParseException {
     try {
```
jacques-n, line 35: I believe there is a new lambda canonical way to do exception checking using junit.

isthmus/src/test/java/io/substrait/isthmus/SimplePlansTest.java
```diff
@@ -1,59 +1,53 @@
...
+      assertTrue(e.getMessage().contains("SQL must contain only a single statement"));
     }
+    assertProtoPlanRoundrip(
+        "select l_orderkey from lineitem; select l_partkey from lineitem WHERE L_ORDERKEY > 20;",
+        new SqlToSubstrait(
+            new SqlToSubstrait.Options(SqlToSubstrait.StatementBatching.MULTI_STATEMENT)));
+  }
+
+  @Test
+  public void joinWithMultiDDLInOneString() throws IOException, SqlParseException {
```
jacques-n, line 49: I don't understand why this is called joinWithMultiDDLInOneString.

JamesRTaylor: The name didn't change. It's just the diff that got a bit confused because the `multiStatement()` appears before it. I can reorder methods if you want.

isthmus/src/test/java/io/substrait/isthmus/SimplePlansTest.java
JamesRTaylor: This tests both that it fails and succeeds without vs with flag.

JamesRTaylor: But I can have two tests if you think that's better

jacques-n: I'm not arguing for two separate tests. Just saying that something like this would be more idiomatic:

```
assertThrows(UnsupportedOperationException.class, () -> {
  assertProtoPlanRoundrip("select l_orderkey from lineitem; select l_partkey from lineitem WHERE L_ORDERKEY > 20;");
}, "SQL must contain only a single statement");
```

isthmus/src/test/java/io/substrait/isthmus/SimplePlansTest.java
jinfengni: The name `joinWithMultiDDLInOneString` is a test I added for the scenario where one `--create` string contains multiple create_table DDLs.  It's not for multiple queries per input string.

isthmus/src/main/java/io/substrait/isthmus/SqlToSubstrait.java
```diff
@@ -84,36 +116,35 @@ private Plan executeInner(
...
+    var relProtoConverter = new RelProtoConverter(functionCollector);
+    sqlToRelNode(sql, validator, catalogReader)
+        .forEach(
+            root -> {
+              plan.addRelations(
+                  PlanRel.newBuilder()
+                      .setRoot(
+                          io.substrait.proto.RelRoot.newBuilder()
+                              .setInput(
+                                  SubstraitRelVisitor.convert(root, EXTENSION_COLLECTION)
```
jinfengni, line 130: Here we use "one fails all" policy. That is, failure of one query's conversion would fail all the queries. This is fine for executing a transaction with multi-queries. In other scenarios, we may want to keep the successful conversions, and record the failed queries in result.  Probably something to consider in the future.

isthmus/src/test/java/io/substrait/isthmus/SimplePlansTest.java
JamesRTaylor: Updated

isthmus/src/test/java/io/substrait/isthmus/SimplePlansTest.java
JamesRTaylor: Reordered to help with diff

isthmus/src/main/java/io/substrait/isthmus/SqlToSubstrait.java
JamesRTaylor: Good point. Added TODO here.

JamesRTaylor:
Thanks for the reviews, @jacques-n and @jinfengni. I've updated PR based on your feedback.

jacques-n (approved):

Issue #40, opened 2022-06-03
[ISTHMUS] support  order by, limit offset clause in substrait to sql conversion


jinfengni:
Currently, the converter from substrait rel plan to calcite relnode plan does not support SortRel / FetchRel, and would not produce sql with order by clause, limit offset clause.

This ticket is to track this missing functionality.

weijietong:
I'll work on this.

JamesRTaylor:
Fixed by https://github.com/substrait-io/substrait-java/pull/50

PR #50, opened 2022-06-06
[ISTHMUS]support order by, limit offset clause in substrait to sql conversion

weijietong:
fix issue #40

### DIFF
--- a/isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
@@ -2,6 +2,7 @@ package io.substrait.isthmus;
 
 import static io.substrait.isthmus.SqlToSubstrait.EXTENSION_COLLECTION;
 
+import io.substrait.expression.Expression;
 import io.substrait.function.SimpleExtension;
 import io.substrait.isthmus.expression.AggregateFunctionConverter;
 import io.substrait.isthmus.expression.ExpressionRexConverter;
@@ -9,13 +10,16 @@ import io.substrait.isthmus.expression.ScalarFunctionConverter;
 import io.substrait.relation.AbstractRelVisitor;
 import io.substrait.relation.Aggregate;
 import io.substrait.relation.Cross;
+import io.substrait.relation.Fetch;
 import io.substrait.relation.Filter;
 import io.substrait.relation.Join;
 import io.substrait.relation.NamedScan;
 import io.substrait.relation.Project;
 import io.substrait.relation.Rel;
+import io.substrait.relation.Sort;
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.Collections;
 import java.util.List;
 import java.util.stream.Collectors;
 import java.util.stream.IntStream;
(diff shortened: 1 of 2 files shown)


isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
```diff
@@ -100,6 +107,57 @@ public RelNode visit(Project project) throws RuntimeException {
     return relBuilder.push(child).project(rexList).build();
   }
 
+  @Override
+  public RelNode visit(Sort sort) throws RuntimeException {
+    RelNode child = sort.getInput().accept(this);
+    List<RelFieldCollation> relFieldCollations =
+        sort.getSortFields().stream().map(sortField -> toRelFieldCollation(sortField)).toList();
+    if (relFieldCollations.isEmpty()) {
```
JamesRTaylor, line 115: Is this if statement necessary?

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
```diff
@@ -100,6 +107,57 @@ public RelNode visit(Project project) throws RuntimeException {
...
+        nullDirection = RelFieldCollation.NullDirection.FIRST;
+        fieldDirection = RelFieldCollation.Direction.DESCENDING;
+      }
+      case DESC_NULLS_LAST -> {
+        nullDirection = RelFieldCollation.NullDirection.LAST;
+        fieldDirection = RelFieldCollation.Direction.DESCENDING;
+      }
+      case CLUSTERED -> fieldDirection = RelFieldCollation.Direction.CLUSTERED;
+
+      default -> throw new RuntimeException("Unexpected Expression.SortDirection enum!");
```
JamesRTaylor, line 156: Add sortDirection to exception message

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
```diff
@@ -100,6 +107,57 @@ public RelNode visit(Project project) throws RuntimeException {
...
+    }
+    return relBuilder.push(child).sort(RelCollations.of(relFieldCollations)).build();
+  }
+
+  @Override
+  public RelNode visit(Fetch fetch) throws RuntimeException {
+    RelNode child = fetch.getInput().accept(this);
+    var optCount = fetch.getCount();
+    long count = -1L;
+    if (optCount.isPresent()) {
```
jacques-n, line 126: You can avoid all of this using `optCount.orElse(-1L)`

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
```diff
@@ -100,6 +107,57 @@ public RelNode visit(Project project) throws RuntimeException {
...
+      count = optCount.getAsLong();
+    }
+    var offset = fetch.getOffset();
+    return relBuilder.push(child).limit((int) offset, (int) count).build();
```
jacques-n, line 130: Let's put a check above this that throws if these truncations are going to cause a lost of value (as opposed to silently truncating).

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
weijietong: yes, it is to solve the `select xxx from xxx limit 100` case which does not have the `order by` clause

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
weijietong: done

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
weijietong: that's elegant! done.

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
weijietong: done

JamesRTaylor (approved):

jacques-n:
This needs to be rebased before merging.

JamesRTaylor:
LGTM. Thanks for the fix, @weijietong.

Issue #51, opened 2022-06-07
Add FuncArg type


hbutani:
- Update `Scalar` and `Aggregate` `FunctionInvocation` to be made of `FuncArg`s
- Update Proto, Calcite converters to build & consume `FuncArg`s

See Issue #37 for background on this issue.

PR #55, opened 2022-06-09
Introduce FunctionArg Type to model function arguments

hbutani:
Fix for substrait-io/substrait-java#51

## Representing and mapping Function arguments
- `FunctionArg` is a marker interface
- subtypes are `Expression`, `Type`, and  the new `EnumArg` class

### EnumArg class
- captured as the `Option` value and a reference to the `Enum`
- Reference to the `Enum` is a reference to an `SimpleExtension.EnumArgument` instance.
  - TODO: should also include the `FunctionAnchor`

## Changes to Function invocation structure
- `ScalarFunctionInvocation`, `AggregateFunctionInvocation`
  - `arguments` is now  of type `List<FunctionArg>`
  - currently retained `List<Expression> exprArguments()` which is a view on `arguments`
    - seems useful. Need feedback..
    - TODO: memonize

## Proto mapping
- **To Proto**
  - map substrait `Expression` and `Type` instances to proto and then wrap them in a `FunctionArgument`
  - map an `EnumArg` instance to a `FunctionArgument.Enum` that only holds the option value.
- **From Proto**
  - map `FunctionArgument`s wrapping `Type` and `Expression` to substrait `Type and `Expression` respectively
  - for an `FunctionArgument.Enum` value at a particular parameter index in the function invocation
    - look within possible `SimpleExtension.EnumArgument` and pick the first one that has the option value
      in its options list.

## Other changes in core
- `RelCopyOnWriteVisitor`
  - TODO: incorporate `FunctionArg` in trees
  - for now replaced `expr.arguments()` with `expr.exprArguments()`
- `ExpressionCreator`
  - factory methods `scalarFunction`, `aggregateFunction` now take `FunctionArg` parameters.

## Calcite mapping
- **EnumConverter**
  - Encapsulate mapping calcite Enums to substrait EnumArgs.
  - provide unctions for
  - convert `EnumArg` to `RexLiteral`
  - convert `RexLiteral` to `EnumArg` within the context of a `SimpleExtension.Function`
- **To Calcite**
  - `ExpressionRexConverter` implements `FunctionArg.FuncArgVisitor`
    - For `Expression` FunctionArg just call `accept`
    - For `Type` FunctionArg throws `UnsupportedException`
      - TODO: how to map Type method parameters to calcite `RexCall`?
    - For `EnumArg` FunctionArg
      - call the `EnumConverter`
    - visit for a `Expression.ScalarFunctionInvocation`
      - map the `expr.arguments` by calling `acceptFuncArgVis`
- **From Calcite**
  - `FunctionFinder:attemptMatch`
    - the `RexLiteral` of `Enum` values still first converted to substrait `String` Literal
    - but then try to match against a function variant with enum arguments.
    - if a match is found, convert operand Expressions into `FunctionArg` and call `generateBinding`
  -


core/src/main/java/io/substrait/expression/EnumArg.java
```diff
@@ -0,0 +1,24 @@
...
+
+import io.substrait.function.SimpleExtension;
+import org.immutables.value.Value;
+
+@Value.Immutable
+/**
+ * Capture the enum value and the associated `SimpleExtension.EnumArgument` TODO: probably better to
+ * capture `SimpleExtension.Function` also.
+ */
+public interface EnumArg extends FunctionArg {
```
JamesRTaylor, line 11: Would `EnumLiteral` be a better name and wouldn't it be simpler to have `EnumArg extends Expression`? Then you wouldn't need `exprArguments()` above and you could add something like this to `ExpressionVisitor`:
```
  R visit(Expression.EnumLiteral expr) throws E;
```

core/src/main/java/io/substrait/relation/ProtoRelConverter.java
```diff
@@ -2,10 +2,7 @@
 
 import static io.substrait.expression.proto.ProtoExpressionConverter.EMPTY_TYPE;
 
-import io.substrait.expression.AggregateFunctionInvocation;
-import io.substrait.expression.Expression;
-import io.substrait.expression.FunctionLookup;
-import io.substrait.expression.ImmutableExpression;
+import io.substrait.expression.*;
```
JamesRTaylor, line 5: Revert the wildcard for imports

core/src/main/java/io/substrait/relation/RelProtoConverter.java
```diff
@@ -1,19 +1,11 @@
...
 import io.substrait.expression.proto.ExpressionProtoConverter;
 import io.substrait.expression.proto.FunctionCollector;
-import io.substrait.proto.AggregateFunction;
-import io.substrait.proto.AggregateRel;
-import io.substrait.proto.FetchRel;
-import io.substrait.proto.FilterRel;
-import io.substrait.proto.JoinRel;
-import io.substrait.proto.ProjectRel;
-import io.substrait.proto.ReadRel;
+import io.substrait.proto.*;
```
JamesRTaylor, line 7: Revert wildcard

core/src/main/java/io/substrait/expression/EnumArg.java
hbutani: - Didn't make them Expressions because Enums don't have an associated `Type`. But can  extend Expression and model them as `StringType`. Do you anticipate any issues with this?
- Currently Enum's are tied to a `Function`; that is why I called them `EnumArg`. I was going to investigate moving EnumArg inside `Function`.
- I was wondering about the thought process here.. Alternate could be having an `EnumType` and then Functions could have `args` of the `EnumType`.

core/src/main/java/io/substrait/expression/doc-files/FunctionArg.md
```diff
@@ -0,0 +1,54 @@
...
+    - TODO: memonize
+
+## Proto mapping
+- **To Proto**
+  - map substrait `Expression` and `Type` instances to proto and then wrap them in a `FunctionArgument`
+  - map an `EnumArg` instance to a `FunctionArgument.Enum` that only holds the option value.
+- **From Proto**
+  - map `FunctionArgument`s wrapping `Type` and `Expression` to substrait `Type and `Expression` respectively
+  - for an `FunctionArgument.Enum` value at a particular parameter index in the function invocation
+    - look within possible `SimpleExtension.EnumArgument` and pick the first one that has the option value
```
JamesRTaylor, line 24: Does an Enum work similar to a Function in that its definition (i.e. name and possible values) is externalized and referenced by position through an anchor? I'd expect the String literal value to be captured plus a reference by index to the Enum definition. Something like that. I don't quite understand the necessity of all the search code you've written.

core/src/main/java/io/substrait/relation/ProtoRelConverter.java
hbutani: Sorry `intelliJ` default setting of conversion to import.* after 5 imports.

core/src/main/java/io/substrait/expression/doc-files/FunctionArg.md
```diff
@@ -0,0 +1,54 @@
+## Representing and mapping Function arguments
+- `FunctionArg` is a marker interface
+- subtypes are `Expression`, `Type`, and  the new `EnumArg` class
```
JamesRTaylor, line 3: What's an example of a FunctionArg that is a Type? Is that when the argument requires the name of a type? Can that be modeled as an Enum whose possible values must be a Type (i.e. the possible values don't need to be enumerated)?

core/src/main/java/io/substrait/expression/doc-files/FunctionArg.md
hbutani: In `algebra.proto` `FunctionArgument.Enum` only captures the enum value.
I took this as a design constraint.
So mapping when mapping to substrait have to also account for optional arguments.
Hence the search logic.

core/src/main/java/io/substrait/expression/doc-files/FunctionArg.md
hbutani: I don't have an example.
But `Argument` in `Function` has a notion of `TypeArgument`. I am assuming this is modeling generic function invocations.

core/src/main/java/io/substrait/expression/EnumArg.java
JamesRTaylor: I like the idea of Enum extending Expression and having an EnumType.

core/src/main/java/io/substrait/expression/doc-files/FunctionArg.md
JamesRTaylor: Don't we have the type of the Enum also (based on the type of the expression)?

core/src/main/java/io/substrait/expression/doc-files/FunctionArg.md
JamesRTaylor: Not sure what this is either. Maybe @jinfengni  or @jacques-n know?

core/src/main/java/io/substrait/expression/doc-files/FunctionArg.md
hbutani: This is the comment in `algebra.proto`
```
// deprecated: enum literals are only sensible in the context of
    // function arguments, for which FunctionArgument should now be
    // used
    Enum enum = 10 [deprecated = true];
```

core/src/main/java/io/substrait/expression/FunctionArg.java
```diff
@@ -0,0 +1,101 @@
...
+     * arg3`
+     *
+     * <p>TODO: move this to `SimpleExtension.Function`, or at least an inner function in
+     * `possiblePositions`
+     *
+     * @param args
+     * @param argIdx
+     * @return
+     */
+    private int possiblePositions(List<SimpleExtension.Argument> args, int argIdx) {
```
jacques-n, line 80: I think there is a misunderstanding here of the nature of optional enum arguments. They must always be declared when using a function. (There should be no ambiguity on the consumption of the plan as to what each argument means.) The optionality is to declare an enum that has no value.

For example, if you have a function that was `foo(Optional Enum, Required Enum, Value, Value, Type)` you always have to pass the optional enum. You can either provide a value or no value. If you provide no value, the default is used. The argument must always be passed, however.

Part of the misunderstanding is that isthmus currently fails to behave correctly I think...

core/src/main/java/io/substrait/expression/AggregateFunctionInvocation.java
```diff
@@ -4,13 +4,21 @@
...
 import org.immutables.value.Value;
 
 @Value.Immutable
 public abstract class AggregateFunctionInvocation {
   public abstract SimpleExtension.AggregateFunctionVariant declaration();
 
-  public abstract List<Expression> arguments();
+  public abstract List<FunctionArg> arguments();
+
+  public final List<Expression> exprArguments() {
```
jacques-n, line 16: Let's avoid adding this function. I don't think we should need it.

core/src/main/java/io/substrait/expression/FunctionArg.java
hbutani: Thanks, this helps. Yes, I took optional to me possibly not present in the invocation. Will fix.

hbutani:
A quick google search seems to indicate Postgres(https://www.postgresql.org/docs/9.1/datatype-enum.html) and Mysql(https://dev.mysql.com/doc/refman/5.7/en/enum.html) support enum types.

Capturing that the `datetime portion` argument in `extract` as an enum argument is nice; as there is no notion of enum types.  As you said it cannot appear anywhere an expression can...

Down the line if we want to add `enum types`, modeling `datetime portion` argument in `extract` as enum type seem to provide enough semantic information.  The `datetime` option can be modeled as a built-in enum type.

Having both the notion of `enum types` and `enum args` would be confusing, i think...

Raising this here,  because addressing this issue later may be painful...
But will proceed with current design.

jacques-n:
> Having both the notion of enum types and enum args would be confusing, i think...

It's a good point. It feels more like a naming problem than a logical problem. We original were calling these things "options". I'm not inclined to change. What are your thoughts around this potential future confusion @cpcloud ?

cpcloud:
Is there a context in which both enum types and enum arguments are used together? If they're always used separately I don't think it's confusing, but I think it is if they need to be talked about/thought about in the same context.

jacques-n:
> Is there a context in which both enum types and enum arguments are used together? If they're always used separately I don't think it's confusing, but I think it is if they need to be talked about/thought about in the same context.

They would never be used interchangeably. You could potentially have arguments of both types in a single function but they would have known positions. (An enum datatype would need to be an value type function argument). For example:

`function(EnumArg, EnumArg, EnumType, EnumType)`

hbutani:
> There is an important distinction here between the enum type and the enum argument....

Thanks, this is very helpful. Intuitively I could sense there is a difference, but couldn't pin it down.
I think the same questions will come up again...
This should be documented, will add your explanation to the javadoc as a start.

PR #38, opened 2022-06-02
[ISTHMUS] Add aggregation function MIN/MAX support

weijietong:
add aggregation function min/max support, so the tpc-h Q2,Q15 could run success. Delivers #36

PR #56, opened 2022-06-10
[ISTHMUS]support the window function translation from sql to substrait

### LINKED ISSUE #48: [ISTHMUS] Add support bi-drectional calcite/substrait support for window functions

weijietong:
To fix #48 , the work will be split into two PRs. This the first one to support window function translation from sql to substrait direction.


isthmus/src/main/java/io/substrait/isthmus/expression/RexExpressionConverter.java
```diff
@@ -73,7 +92,115 @@ public Expression visitLiteral(RexLiteral literal) {
...
+              inputRel,
+              inputType,
+              aggCall,
+              t -> {
+                RexNode r = (RexNode) t;
+                return r.accept(this);
+              });
+      if (funcInvocationOpt.isPresent()) {
+        functionInvocation = (AbstractFunctionInvocation) funcInvocationOpt.get();
+        if (functionInvocation instanceof AggregateFunctionInvocation) {
```
jinfengni, line 128: In stead of using `instanceof`, maybe an  API that return `functionType` to differentiate among `scalar function`, `aggregation function`, and `window agg function`?

isthmus/src/main/java/io/substrait/isthmus/expression/RexExpressionConverter.java
```diff
@@ -73,7 +92,115 @@ public Expression visitLiteral(RexLiteral literal) {
...
+          var windowFunc =
+              ImmutableExpression.WindowFunction.builder().function(windowFuncInvocation).build();
+          builder.windowFunction(windowFunc).hasNormalAggregateFunction(false);
+        } else {
+          throw new RuntimeException(
+              String.format("Unexpected AbstractFunctionInvocation :%s", functionInvocation));
+        }
+      }
+    }
+    if (functionInvocation == null) {
```
jinfengni, line 143: Maybe move this `if` block before line 126 for the case `functionInvocationOpt` is empty, for better code readability.

isthmus/src/main/java/io/substrait/isthmus/expression/RexExpressionConverter.java
```diff
@@ -73,7 +92,115 @@ public Expression visitLiteral(RexLiteral literal) {
...
+        .type(TypeConverter.convert(over.getType()))
+        .build();
+  }
+
+  private WindowBound toWindowBound(RexWindowBound rexWindowBound) {
+    if (rexWindowBound.isCurrentRow()) {
+      return ImmutableWindowBound.CurrentRowWindowBound.builder().build();
+    }
+    if (rexWindowBound.isUnbounded()) {
+      var direction =
```
jinfengni, line 166: line 166 - 169 code seems to be same as line 173 - 175. Can we refactor to remove the duplicate?

isthmus/src/main/java/io/substrait/isthmus/expression/RexExpressionConverter.java
```diff
@@ -73,7 +92,115 @@ public Expression visitLiteral(RexLiteral literal) {
...
+        switch (rexDirection) {
+          case ASCENDING -> rexFieldCollation.getNullDirection()
+                  == RelFieldCollation.NullDirection.LAST
+              ? Expression.SortDirection.ASC_NULLS_LAST
+              : Expression.SortDirection.ASC_NULLS_FIRST;
+          case DESCENDING -> rexFieldCollation.getNullDirection()
+                  == RelFieldCollation.NullDirection.LAST
+              ? Expression.SortDirection.DESC_NULLS_LAST
+              : Expression.SortDirection.DESC_NULLS_FIRST;
+          default -> throw new RuntimeException(
```
jinfengni, line 197: Is it better to use `IllegalArgumentException` here?

isthmus/src/main/java/io/substrait/isthmus/expression/RexExpressionConverter.java
```diff
@@ -73,7 +92,115 @@ public Expression visitLiteral(RexLiteral literal) {
...
+          var aggFuncInvocation = (AggregateFunctionInvocation) functionInvocation;
+          var aggMeasure = Aggregate.Measure.builder().function(aggFuncInvocation).build();
+          builder.aggregateFunction(aggMeasure).hasNormalAggregateFunction(true);
+        } else if (functionInvocation instanceof WindowFunctionInvocation) {
+          var windowFuncInvocation = (WindowFunctionInvocation) functionInvocation;
+          var windowFunc =
+              ImmutableExpression.WindowFunction.builder().function(windowFuncInvocation).build();
+          builder.windowFunction(windowFunc).hasNormalAggregateFunction(false);
+        } else {
+          throw new RuntimeException(
```
jinfengni, line 138: Is it good to use `RuntimeException`?

isthmus/src/main/java/io/substrait/isthmus/expression/RexExpressionConverter.java
```diff
@@ -153,4 +280,20 @@ public Expression visitLocalRef(RexLocalRef localRef) {
   public Expression visitPatternFieldRef(RexPatternFieldRef fieldRef) {
     throw new UnsupportedOperationException("RexPatternFieldRef not supported");
   }
+
+  public RelNode getInputRel() {
```
jinfengni, line 284: What's the purpose of code block L24 - 297, for a `RexExpressionConverter`?

isthmus/src/test/java/io/substrait/isthmus/OneWaySimpleTest.java
```diff
@@ -0,0 +1,36 @@
...
+
+import java.io.IOException;
+import java.util.List;
+
+public class OneWaySimpleTest extends PlanTestBase {
+  @Test
+  public void windowFunctionsSql2Sub() throws IOException, SqlParseException {
+    SqlToSubstrait sqlToSubstrait = new SqlToSubstrait();
+    List<String> ddlStmts = Lists.newArrayList(asString("tpch/schema.sql"));
+    print(
```
jinfengni, line 18: The test did not verify the output? 

Also, will it be good to add more example of window function, to cover more usage cases?

core/src/main/java/io/substrait/expression/AbstractFunctionInvocation.java
```diff
@@ -0,0 +1,24 @@
...
+import io.substrait.type.Type;
+import java.util.List;
+
+public abstract class AbstractFunctionInvocation<T extends SimpleExtension.Function, I> {
+
+  public abstract T declaration();
+
+  public abstract List<Expression> arguments();
+
+  public abstract Expression.AggregationPhase aggregationPhase();
```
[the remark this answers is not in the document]
weijietong: The substrait's specification declares some window functions like `sum` could be decomposable into multiple stages. So it's feasible to declare this common aggregation phase method.

As functions like sum are both aggregate functions  and window functions, I don't know whether there's a better name to it.

isthmus/src/main/java/io/substrait/isthmus/expression/RexExpressionConverter.java
```diff
@@ -73,7 +92,115 @@ public Expression visitLiteral(RexLiteral literal) {
...
+            approximate,
+            over.ignoreNulls(),
+            argList,
+            filterArg,
+            null,
+            RelCollations.EMPTY,
+            over.getType(),
+            sqlAggFunction.getName());
+    var builder = Expression.Window.builder();
+    AbstractFunctionInvocation functionInvocation = null;
```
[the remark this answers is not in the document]
weijietong: The same as the above, window functions could be some aggregate functions and the specific window functions.

isthmus/src/main/java/io/substrait/isthmus/expression/RexExpressionConverter.java
weijietong: can't imagine how to ignore the `instanceof`,or just wrap it.

isthmus/src/main/java/io/substrait/isthmus/expression/RexExpressionConverter.java
weijietong: Maybe that will not catch some failed cases.

isthmus/src/main/java/io/substrait/isthmus/expression/RexExpressionConverter.java
weijietong: done

isthmus/src/test/java/io/substrait/isthmus/OneWaySimpleTest.java
weijietong: benefit line 119, for the invocator to set or reset the window expression's related relnode and type

isthmus/src/main/java/io/substrait/isthmus/expression/RexExpressionConverter.java
weijietong: benefit line 119, for the invocator to set or reset the window expression's related relnode and type

isthmus/src/test/java/io/substrait/isthmus/OneWaySimpleTest.java
weijietong: wait for the substrait to merge the window function PR, I think this test case it temp, after the reverse translation done, will be refactored to the roundtrip test case.

isthmus/src/main/java/io/substrait/isthmus/expression/RexExpressionConverter.java
weijietong: have no idea

isthmus/src/test/java/io/substrait/isthmus/OneWaySimpleTest.java
jinfengni: The one test you have is window aggregate function? would it make sense to add at least a window non-aggregate function, such as lag(), rank(), etc? This is to exercise the new code you added.

core/src/main/java/io/substrait/expression/AbstractFunctionInvocation.java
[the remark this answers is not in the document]
jinfengni: What I'm confused is whether `sum` is processed as `WindowFunctionInvocation`, or `AggregateionFunctionInvocation`, in the case of `sum(s_nationkey) over (partition by s_address order by s_phone desc)' ?

isthmus/src/main/java/io/substrait/isthmus/expression/ExpressionRexConverter.java
```diff
@@ -213,6 +213,12 @@ public RexNode visit(FieldReference expr) throws RuntimeException {
     return visitFallback(expr);
   }
 
+  @Override
+  public RexNode visit(Expression.Window expr) throws RuntimeException {
+    // todo:to construct the RexOver
```
jinfengni, line 218: If Substrait -> Rex conversion is not implemented yet, better not to Override this method. In stead, let it call `visitFallBack`.

core/src/main/java/io/substrait/expression/AbstractFunctionInvocation.java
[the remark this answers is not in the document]
weijietong: as `WindowFunctionInvocation`

isthmus/src/test/java/io/substrait/isthmus/OneWaySimpleTest.java
weijietong: yes

wait for this [substrait PR](https://github.com/substrait-io/substrait/pull/224) to be merged,but I tested those window functions at local. Maybe I could add those test case at this PR,but would have a bad pr experience like this [one](https://github.com/substrait-io/substrait-java/pull/38).

isthmus/src/main/java/io/substrait/isthmus/expression/ExpressionRexConverter.java
weijietong: good catch, done.

isthmus/src/main/java/io/substrait/isthmus/expression/RexExpressionConverter.java
```diff
@@ -73,7 +92,115 @@ public Expression visitLiteral(RexLiteral literal) {
...
+    for (var converter : functionConverters) {
+      var funcInvocationOpt =
+          converter.convert(
```
jinfengni, line 118: In stead of letting `converter.convert()` return an `AbstractFunctionInvocation`,  will it be good to return a Substrait `Expression`?  Each `FunctionConverter` knows the type of function, either aggregate function, or window function, and it could build the appropriate Substrait expression.  In that way, there seems no need for this `AbstractFunctionInvocation`. 

 
FYI, the existing interface `CallConverter` works in that way.

core/src/main/java/io/substrait/expression/AbstractFunctionInvocation.java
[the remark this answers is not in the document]
jinfengni: See my comment in `RexExpressionConverter.java`.  I feel we do not have to introduce this `AbstractFunctionInvocation`.  In stead, we could let `FunctionConverter.convert` to return a substrait expression.

isthmus/src/main/java/io/substrait/isthmus/expression/RexExpressionConverter.java
weijietong: I have tried to define `AggregateFunctionInvocation` and `WindowFunctionInvocation` as `Expression`. Not seen any conveniently. By doing so ,when we traverse the expression tree,what will these invocation expression be translated to?  The `ScalarFunctionInvocation` has a corresponding `proto` expression that make sense.  Anyway we also need to distinguish which type of function invocation expression it is to set the `WindowExpression`.

jinfengni: I'm not saying that we should define `AggregateFunctionInvocation` or `WindowFunctionInvocation` as `Expression`.  

`CallConverter.convert()` returns an `Expression`.  For each subclass, it would return a `ScalarFunctionInvocation` for `ScalarFunctionConverter`, or a `IfThenStatement` for `Case`.   My thinking is let each `FunctionConverter` to decide what kind of `FunctionInvocation` / `Expression` to create.  For instance, `AggregateFunctionConverter` could create a `AggrgateFunctionInvocation`, while the `WindowFunctionConverter` would create a `WindowFunctionInvocation`. Same thing for future new type of Function Converter.  

The current PR's convert() would return an `AbstractFunctionInvocation`, and check the instance type to decide which FunctionInvocation is created.  

1. https://github.com/substrait-io/substrait-java/blob/main/isthmus/src/main/java/io/substrait/isthmus/CallConverter.java#L11
2. https://github.com/substrait-io/substrait-java/blob/main/isthmus/src/main/java/io/substrait/isthmus/expression/ScalarFunctionConverter.java#L63
3.  https://github.com/substrait-io/substrait-java/blob/main/isthmus/src/main/java/io/substrait/isthmus/expression/CallConverters.java#L41

isthmus/src/main/java/io/substrait/isthmus/expression/WindowFunctionConverter.java
```diff
@@ -0,0 +1,243 @@
...
+import org.apache.calcite.rex.RexSlot;
+import org.apache.calcite.rex.RexWindowBound;
+
+public class WindowFunctionConverter
+    extends FunctionConverter<
+        SimpleExtension.WindowFunctionVariant,
+        WindowFunctionInvocation,
+        WindowFunctionConverter.WrappedAggregateCall> {
+
+  private AggregateFunctionConverter aggregateFunctionConverter;
```
jinfengni, line 41: Maybe add a comment explaining that aggregate function call could appear as a Window Function, and as such we wrap an `AggregateFunctionConverter` inside a `WindowFunctionConverter`.

jinfengni (approved):
LGTM.

jinfengni:
@jacques-n , can you take a look at this PR?

jacques-n (approved):
Thanks for pulling this together @weijietong !

PR #59, opened 2022-06-16
Ignore SqlKeyConstraints while parsing table definitions

### LINKED ISSUE #58: SqlToSubstrait conversion fails on encountering key constraints
SQL create-table statements often contain key constraints, for e.g. TPC-DS table primary key declarations like below: 
```
create table store_sales (
    ss_item_sk                integer               not null,
    ...
    ss_net_profit             decimal(7,2)                  ,
    primary key (ss_item_sk));
```
 These constraint declaration are valid per SQL standard. Currently the SQL to Substrait fails when it encounters such constraints with the following error:
```
java.lang.ClassCastException: class org.apache.calcite.sql.ddl.SqlDdlNodes$1 cannot be cast to class org.apache.calcite.sql.ddl.SqlColumnDeclaration (org.apache.calcite.sql.ddl.SqlDdlNodes$1 and org.apache.calcite.sql.ddl.SqlColumnDeclaration are in unnamed module of loader 'app')
	at io.substrait.isthmus.SqlConverterBase.parseCreateTable(SqlConverterBase.java:123)
	at io.substrait.isthmus.SqlConverterBase.registerCreateTables(SqlConverterBase.java:86)
	at io.substrait.isthmus.SqlToSubstrait.execute(SqlToSubstrait.java:81)
```

Instead of expecting users to omit constraints from input table definitions, key constraints should be ignored so that the plan generation does not fail.

ashvina:
SQL create-table statements often contain key constraints, for e.g. table primary key declarations. These constraint declaration are valid per SQL standard. Currently the SQL to Substrait conversion fails on encountering constraints. This change ignores such declarations when present.
Closes #58


jacques-n (approved):
LGTM. Thanks @ashvina !

PR #60, opened 2022-06-19
feat: add support for set operations

### LINKED ISSUE #57: [ISTHMUS] Add support for set operations
Add support for set operations UNION and MINUS in SQL -> Substrait Proto -> Substrait POJO. This would involve adding implementation of methods in SubstraitRelVisitor, RelCopyOnWriteVisitor, ProtoRelConverter, and RelProtoConverter. Queries such as these should roundtrip between SQL, Substrait Proto, and Substrait POJO:

```
select p_partkey as partkey from part where p_partkey > cast(100 as bigint) 
union select l_partkey as partkey from lineitem where l_orderkey > cast(100 as bigint);
select p_partkey as partkey from part where p_partkey > cast(100 as bigint) 
minus select l_partkey as partkey from lineitem where l_orderkey > cast(100 as bigint);
```

guykhazma:
Fixes #57.


core/src/main/java/io/substrait/relation/Set.java
```diff
@@ -0,0 +1,54 @@
...
+                }
+            }
+
+            throw new IllegalArgumentException("Unknown operation: " + proto);
+        }
+    }
+
+    @Override
+    protected Type.Struct deriveRecordType() {
+        return getInputs().get(0).getRecordType();
```
guykhazma, line 43: Is it safe to assume that we have at least one input?
If not, how should I handle a case when all relations are empty?

jacques-n:
@JamesRTaylor or @jinfengni , can one of you give @guykhazma and early review?

core/src/main/java/io/substrait/relation/RelCopyOnWriteVisitor.java
```diff
@@ -102,6 +102,20 @@ public Optional<Rel> visit(Join join) throws RuntimeException {
             .build());
   }
 
+  @Override
+  public Optional<Rel> visit(Set set) throws RuntimeException {
+    var inputs =
+        set.getInputs().stream()
+            .map(inputRel -> inputRel.accept(this))
+            .flatMap(Optional::stream)
+            .toList();
```
JamesRTaylor, line 111: There are two use cases for the RelCopyOnWriteVisitor:

1. To traverse the tree to detect something. For example, count how many occurrences of COUNT DISTINCT there are across the entire tree.
2. To rewrite parts of the subtree. For example, to replace any occurrence of COUNT DISTINCT with a call to APPROX_COUNT_DISTINCT instead.

For the default read-only traversal case of (1), each visitor method returns Optional.empty() as an indication to the caller that nothing has changed. That way the caller knows not to create a new instance of the Rel. Only for the rewrite case of (2) will a new Set relation be created. 

In support of this pattern, there's a `RelCopyOnWriteVisitor.transformList` static function to help. For an example usage, see [here](https://github.com/substrait-io/substrait-java/blob/5872ceaaef67b8cb774017a50e0f2920fe0a56f3/isthmus/src/test/java/io/substrait/isthmus/RelCopyOnWriteVisitorTest.java#L228-L234). You'll want to do something similar to only create the new Set if one of the inputs changed.

guykhazma: Thanks for the explanation.
I have fixed the function to return Optional.empty() only if all of the inputs didn't change.
Note that I preferred doing two passes over the input list since doing it in one pass in functional programming using streams would have introduced a side effect which I would rather avoid as it's not best practice.
I expect the list of inputs to be reasonably small so a two pass won't be an issue.

isthmus/src/test/java/io/substrait/isthmus/utils/Utils.java
```diff
@@ -0,0 +1,52 @@
+package io.substrait.isthmus.utils;
+
+import static org.junit.jupiter.params.provider.Arguments.arguments;
+
+import io.substrait.relation.Set;
+import java.util.Arrays;
+import java.util.stream.Stream;
+import org.junit.jupiter.params.provider.Arguments;
+
+public class Utils {
```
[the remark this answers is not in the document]
guykhazma: done

isthmus/src/test/java/io/substrait/isthmus/utils/Utils.java
```diff
@@ -0,0 +1,52 @@
...
+          case INTERSECTION_PRIMARY -> "INTERSECT";
+          case INTERSECTION_MULTISET -> "INTERSECT ALL";
+          case UNION_DISTINCT -> "UNION";
+          case UNION_ALL -> "UNION ALL";
+          case UNKNOWN -> throw new UnsupportedOperationException(
+              "Unknown set operation is not supported");
+        };
+
+    StringBuilder query = new StringBuilder();
+    query.append("select p_partkey as partkey from part where p_partkey > cast(100 as bigint)\n");
```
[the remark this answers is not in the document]
guykhazma: I have changed the sql to use multiple columns and an expression, it's only for testing purposes as the sql itself doesn't have real meaning

core/src/main/java/io/substrait/relation/RelCopyOnWriteVisitor.java
```diff
@@ -115,6 +116,21 @@ public Optional<Rel> visit(Join join) throws RuntimeException {
...
+    }
+    // replace all optional values with the input values
+    var setInputs =
+        IntStream.range(0, set.getInputs().size())
+            .mapToObj(i -> visitedInputs.get(i).orElse(set.getInputs().get(i)))
+            .toList();
+    return Optional.of(
+        ImmutableSet.builder().from(set).inputs(setInputs).setOp(set.getSetOp()).build());
+  }
+
```
JamesRTaylor, line 133: Since all other traversals of expressions and relations are already going through transformList, I'd be inclined to have this one do the same. Maybe worthwhile to rename transformList or add a comment? Here's what it would look like using that function:
```
    return transformList(set.getInputs(), t -> t.accept(this))
        .map(u -> ImmutableSet.builder().from(set).inputs(u).setOp(set.getSetOp()).build());

```

JamesRTaylor:
Looking very good, @guykhazma. Just a couple more comments.

core/src/main/java/io/substrait/relation/RelCopyOnWriteVisitor.java
guykhazma: thanks for pointing this out, it's indeed better to re-use the transformList function, I missed it by mistake.

JamesRTaylor:
Everything looks great, @guykhazma. There's some approval thing that's preventing the standard build from running, but once that's fixed you're going to get a formatting violation. Run `./gradlew :isthmus:spotlessApply` and then check in the re-formatted code and we should be good to go.

JamesRTaylor:
One other minor thing - we don't use the same commit message naming scheme as with `substrait`, so no need to prefix commit message with `feat:`

guykhazma:
@JamesRTaylor I have fixed the styling, sorry forgot to push the commit the fixed that. thanks.

JamesRTaylor (approved):
LGTM

jacques-n (approved):
LGTM. Thanks @guykhazma !

Issue #61, opened 2022-06-24
[ISTHMUS] Null value support in filter


Deegue:
Hi, it seems we don't support null data in filter currently, e.g
table : 
`CREATE TABLE table_test(col_a BIGINT)`
query : 
`SELECT col_a FROM table_test WHERE col_a IS NOT NULL` or
`SELECT col_a FROM table_test WHERE col_a != NULL`

Any help will be appreciated!

jacques-n:
I'm guessing that we haven't yet added a Calcite to Substrait function mapping. Would you be open to propose a patch to add that in? We already have the function defined [here](https://github.com/substrait-io/substrait/blob/9b0e21f5f3d8334a6a291ca06fcc2bd04d6007be/extensions/functions_comparison.yaml#L61).

Deegue:
> I'm guessing that we haven't yet added a Calcite to Substrait function mapping. Would you be open to propose a patch to add that in? We already have the function defined [here](https://github.com/substrait-io/substrait/blob/9b0e21f5f3d8334a6a291ca06fcc2bd04d6007be/extensions/functions_comparison.yaml#L61).

Thanks, I submit a PR and could you take a look at it?

JamesRTaylor:
Duplicate of https://github.com/substrait-io/substrait-java/issues/66

PR #64, opened 2022-06-27
Add TPC-DS smoke tests

hbutani:
- Issue #49 
-  Schema using `org.apache.calcite.adapter.tpcds.TpcdsSchema` from `org.apache.calcite:calcite-plus:1.28.0`
-  For queries started with `net.hydromatic.tpcds.query.Query` and then fixed generation issues
  using Spark SQL tpcds benchmark queries.

- Applied the [windowing pull request](https://github.com/weijietong/substrait-java/tree/issue_48)
- Added `SUBSTRING` to `io.substrait.isthmus.expression.FunctionMappings`
- See `resources/tpcds/README.md` for more details; [tpcds_investigate branch for query generation and testing process](https://github.com/hbutani/substrait-java/tree/tpcds_investigate)

**Queries passing:**
```
[1, 3, 4, 6, 7, 8, 10, 11, 13, 14, 15, 16, 18, 19, 21, 22, 23, 25, 26, 28, 29, 30, 31, 32, 33, 34, 35, 37, 38, 41,
 42, 43, 45, 46, 48, 50, 52, 54, 55, 56, 58, 59, 60, 61, 62, 64, 65, 68, 69, 71, 73, 74, 79, 81, 82, 83, 85, 87, 88,
 90, 92, 94, 95, 96, 99]
```

**Queries failing:**
  - No support for reference to Aggregate expr in windowing expressions for example
    in `q12` `sum(sum(ws_ext_sales_price)) over (partition by i_class)`
    - `[12, 20, 47, 51, 53, 57, 63, 89, 98]`
  - No support for `is null` or `is not null` check
    - `[40, 44, 49, 75, 76, 77, 78, 80, 93, 97]`
  - Error on `rank()` call
    - `[36, 67, 86]`
  - Missing functions: round, STDDEV_SAMP, UPPER, LIKE
    - `[2, 17, 24, 39, 91]`
  - Concat: vargs behavior, implicit casting
    - `[5, 54, 84]`
  - Unsupported subquery nested in Project relational operator
    - `[9]`
  - Support for GROUPING
    - `[27]`
  - Cannot apply '+' to arguments of type `<DATE> + <INT>`
    - `[72]`

PR #74, opened 2022-06-28
[ISTHMUS] Add null function mapping and OpType

Deegue:
Issue: [ISTHMUS] Null value support in filter #61 

IIUC, we just need to add the mapping to Calcite.


jacques-n:
The optype change doesn't belong here and seems unecessary. OpType has to do with function type resolution, not user operations.

Can you add some test cases for the is null and is not null patterns to make sure this is transforming right? I belive Calcite may have an atypical object tree for those operations (not just simple function calls).

Deegue:
> The optype change doesn't belong here and seems unecessary. OpType has to do with function type resolution, not user operations.
> 
> Can you add some test cases for the is null and is not null patterns to make sure this is transforming right? I belive Calcite may have an atypical object tree for those operations (not just simple function calls).

Thanks for your review, added some test cases and removed unnecessary code.

JamesRTaylor:
Do anymore of the tpc-ds queries pass with this change?

jikunshang:
> Do anymore of the tpc-ds queries pass with this change?

from description from PR #64,  seems these queries are related 
  - No support for `is null` or `is not null` check
    - [40, 44, 49, 75, 76, 77, 78, 80, 93, 97]
    
can update this [file ](https://github.com/substrait-io/substrait-java/blob/main/isthmus/src/test/java/io/substrait/isthmus/TpcdsQueryNoValidation.java#L44) to enable these queries to check.

BTW, anyone can approve running CI workflows?

JamesRTaylor:
After your first commit, CI workflow will run.

Deegue:
Thanks @JamesRTaylor @jikunshang , any more comments? Gentle ping @jacques-n

JamesRTaylor:
> can update this [file ](https://github.com/substrait-io/substrait-java/blob/main/isthmus/src/test/java/io/substrait/isthmus/TpcdsQueryNoValidation.java#L44)to enable these queries to check.

@jikunshang - that was my thinking as well. Please update this PR with the queries from tpc-ds that pass based on this change.

Deegue:
I tested `[40, 44, 49, 75, 76, 77, 78, 80, 93, 97]`, the result is:
**Queries passing:**
`[40, 75, 76, 77, 93, 97]`

**Queries failing:**
`[44, 49, 78, 80]`

image

Failed reason for Q44:
`java.lang.RuntimeException: Not found the corresponding window aggregate function:RANK`

Failed reason for Q49:
`java.lang.RuntimeException: Not found the corresponding window aggregate function:RANK`

Failed reason for Q78:
`java.lang.IllegalArgumentException: Unable to convert call ROUND(i32, i32).`

Failed reason for Q80:
`java.lang.IllegalArgumentException: Unable to convert call ||(char<5>, char<16>).`

cc @JamesRTaylor

JamesRTaylor:
Thanks for testing, @Deegue. Please add the passing tests to be run in the tpc-ds smoke test. I'll update the open issues with the new information about the failures you mentioned. I think we'll be good-to-go after that.

Deegue:
Thanks @JamesRTaylor , passing tests added.

JamesRTaylor (approved):
Looks great! Thanks so much for the contribution, @Deegue.

JamesRTaylor:
@Deegue - looks like you need to run spotless to fix some formatting issues:
```
Run './gradlew :isthmus:spotlessApply' to fix these violations.
```

Issue #18, opened 2022-05-11
Out-of-spec plans returned by Isthmus


jvanstraten:
While testing the [validator](https://github.com/substrait-io/substrait-validator/pull/1) against Isthmus output for TPC-H, I found a number of issues where Isthmus disagrees with the spec as written on the website. The ones I found are:

 - The grouping set index column appended by the aggregate relation isn't removed from the output, resulting in excess columns both functionally and in the RelRoot column name list. Failing example: `select l_returnflag, count(*) from lineitem group by l_returnflag`. The easy solution is just to generate an emit clause for all aggregate relations.
 - Decimal literals are generated with 38 rather than 16 bytes. Failing example: `select 0.1`
 - Negative decimal literals are not sign extended all the way to the MSB. Failing example: `select -0.1`, which yields out-of-bounds value 22.5 for P=2 S=1.
 - For whatever reason, some decimal literals are just emitted as zero. Failing example: `select 000000.000001`.
 - The emitted function signatures don't follow the table [here](https://substrait.io/extensions/#function-signature-compound-names) for the abbreviated names. For example, `decimal` is used instead of `dec`, and `any1` is used in place of just `any`. Failing example: literally anything with a function, such as `select 0.1 + 1.0`, which generates `add:opt_decimal_decimal` instead of `add:opt_dec_dec`.
 - Virtual tables receive a schema representing a single column of structs containing the fields. Failing example: `select 1` is given the schema `STRUCT<STRUCT<i32>>`. The type itself is also invalid because of this, because only one field name is attached when two would be needed for that type. It should be `NSTRUCT<EXPR$0: i32>` here.
 - Arguably this is not a spec violation (the spec never mentions how this should be done), but using anchor value 0 is IMO a really bad idea, because on the reference side it's indistinguishable from not referencing anything. For type variations it would actually *be* an issue, because "no variation" would be indistinguishable from the type variation with anchor 0, and if any optional references are ever added for the other anchor types it would become a problem for them too. It's also confusing when reading a plan where the reference is "optimized out" due to having the default value, and it's easy to forget to add a reference because nothing can check for this error. Simply starting the counter from 1 instead of 0 prevents all this and barely costs anything (specifically, it costs ~0.00000003% of the anchor code space, and a couple bytes in the binary serialization).

I haven't touched anything java for about a decade, so I'm probably not the right person to propose fixes for these things.

[...]

PR #76, opened 2022-07-08
Fix issue #18

hbutani:
sub-issue:
Virtual tables receive a schema representing a single column of structs containing the fields.
Failing example: select 1 is given the schema STRUCT<STRUCT<i32>>.

### DIFF
--- a/core/src/main/java/io/substrait/relation/VirtualTableScan.java
+++ b/core/src/main/java/io/substrait/relation/VirtualTableScan.java
@@ -3,7 +3,6 @@ package io.substrait.relation;
 import io.substrait.expression.Expression;
 import io.substrait.type.NamedStruct;
 import io.substrait.type.Type;
-import io.substrait.type.TypeCreator;
 import java.util.List;
 import org.immutables.value.Value;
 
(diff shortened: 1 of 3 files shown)


core/src/main/java/io/substrait/relation/VirtualTableScan.java
```diff
@@ -14,11 +13,26 @@ public abstract class VirtualTableScan extends AbstractReadRel {
...
 
+  /**
+   *
+   * <li>non-empty rowset
+   * <li>non-null field-names
+   * <li>no null rows
+   * <li>row shape must match field-list
+   */
+  @Value.Check
+  protected void check() {
```
JamesRTaylor, line 24: Just curious when is this called?

JamesRTaylor (approved):
LGTM

JamesRTaylor:
Would be good if we broke up #18 into separate issues.

core/src/main/java/io/substrait/relation/VirtualTableScan.java
hbutani: https://immutables.github.io/immutable.html#precondition-check-method
Its called from the `validate` method in the generated `ImmutableVirtualTableScan`

jacques-n (approved):
LGTM. Thanks!

PR #77, opened 2022-07-08
Fix issue #18

hbutani:
sub-issue:
The emitted function signatures don't follow the spec.
For example, decimal is used instead of dec, and any1 is used in place of just any.


JamesRTaylor (approved):
LGTM

jacques-n (approved):

PR #84, opened 2022-08-03
Upgrade substrait to 0.9.0

JamesRTaylor:
* Make intermediate field optional in function variant
    * Add optional description field to function arg
    * Add newly passing TPC-DS tests to test


jacques-n (approved):
Thanks @JamesRTaylor !

PR #85, opened 2022-08-04
[Isthmus]enable tpch queries tpch_smoke.sh

jikunshang:


jikunshang:
`1 workflow awaiting approval` @jacques-n  can you approve it to enable CI workflow? thanks!

isthmus/src/test/resources/tpch/queries/15.sql
```diff
@@ -1,4 +1,3 @@
--- converted to CTE since DDL is not part of Substrait.
```
JamesRTaylor, line 1: My preference would be to figure out why comments are causing a problem

jikunshang: My guess is Isthmus will treat the dash here as input parameter and it throws `Missing required parameter: '<sql>'` message.

JamesRTaylor:
We can commit this patch is you remove your changes around getting rid of the comments (i.e. just keep the change that registers the missing proto function for reflection), and adjust the passing tests.

jikunshang:
> We can commit this patch is you remove your changes around getting rid of the comments (i.e. just keep the change that registers the missing proto function for reflection), and adjust the passing tests.

updated.

isthmus/src/test/resources/tpch/queries/15.sql
JamesRTaylor: Please file an issue for the comment in the sql file breaking when using native image

JamesRTaylor:
Thanks for the contribution, @jikunshang

PR #88, opened 2022-08-17
feat: add schema for tpcds queries

richtia:
Adding the schema for TPC-DS queries

### DIFF
--- a/isthmus/src/test/resources/tpcds/schema.sql
+++ b/isthmus/src/test/resources/tpcds/schema.sql
@@ -1,0 +1,544 @@
+create table customer_address
+(
+    ca_address_sk             integer               not null,
+    ca_address_id             char(16)              not null,
+    ca_street_number          char(10)                      ,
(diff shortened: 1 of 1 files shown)


jacques-n:
@richtia , any chance you want to also add at least a smoke test that uses this? It seems lonely right now.

JamesRTaylor:
@richtia - thanks for the PR. Here's some feedback:

- I like the idea of documenting the tpcds schema and being consistent with the tpch convention in terms of the location of the schema file. We actually use Calcite to define the tpch and tpcds schema, but having this documented is definitely useful too.
- Remove the duplicate definition of this schema [here](https://github.com/substrait-io/substrait-java/blob/main/isthmus/src/test/resources/keyconstraints_schema.sql).
- Modify the test that uses `keyconstraints_schema.sql` to use your new schema instead.

vbarua:
Closing this as stale.

PR #92, opened 2022-09-23
Convert 'or' function call to 'singleOrList' exression

chaojun-zhang:
Convert 'or' call to substrait singleOrList in case of calcite optimize the 'in' expression as 'or' expression.

### DIFF
--- a/isthmus/src/main/java/io/substrait/isthmus/expression/CallConverters.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/expression/CallConverters.java
@@ -68,6 +68,43 @@ public class CallConverters {
         return ExpressionCreator.ifThenStatement(defaultResult, caseConditions);
       };
 
+    /**
+     * Convert 'or' call to substrait singleOrList in case of calcite optimize the in expression as or expression.
+     */
+    public static SimpleCallConverter SINGLE_OR_LIST = (call, visitor) -> {
+        if (call.getKind() != SqlKind.OR) {
+          return null;
+        }
+
+        List<RexNode> options = new ArrayList<>();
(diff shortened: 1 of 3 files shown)


chaojun-zhang:
@jacques-n can you take a look this PR?

jacques-n:
I don't think this is the best modification to solve this problem. I think we're actually hitting a Calcite "optimization" which is converting in clause into an or statement (maybe?). I do see the SqlStdOperatorTable has an IN operator so we should figure out why we don't have that here. It could be that we need to turn off some heuristic optimization that Calcite is doing. Have you tried with a larger in clause and still see Calcite convert to an or expression?

chaojun-zhang:
> I don't think this is the best modification to solve this problem. I think we're actually hitting a Calcite "optimization" which is converting in clause into an or statement (maybe?). I do see the SqlStdOperatorTable has an IN operator so we should figure out why we don't have that here. It could be that we need to turn off some heuristic optimization that Calcite is doing. Have you tried with a larger in clause and still see Calcite convert to an or expression?

 Yes, reading code [here](https://github.com/apache/calcite/blob/main/core/src/main/java/org/apache/calcite/sql2rel/SqlToRelConverter.java#L1196)  seems the  in sub query threshold is default to 20  , we can turn off by simple specify the threshold config as 0, but calcite will  optimized the plan  as joinRel with logicalValue right side 
 
 
image

JamesRTaylor:
@chaojun-zhang - wanted to circle back to this. Given Jacques' comment, do you still think this is the correct approach. If so, would you mind rebasing your patch when you have a chance?

PR #95, opened 2022-10-19
Downgrade core to java 8

baibaichen:
I currently work on [gluten](https://github.com/oap-project/gluten), which translates SparkPlan into a Substrait plan.  Although Java 8 is at the end of its life, as we know,  Spark currently mainly runs on JVM 8.  And hence, gluten has to focus on Java 8 in its first release. Jabel may be fine, but it needs to modify the byte code, which introduces some risks.

Based on https://github.com/substrait-io/substrait-java/pull/94, this patch downgrades the core module into Java 8.
1. remove` var`
2. Using java 8 `switch case`, which is a litter Cumbersome.
3. Replace `o instanceof Type.Boolean t` with explicitly cast

Using existed unit test is enoguth.

cc @jacques-n  @FelixYBW

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -42,9 +42,12 @@ allprojects {
     javaLauncher.set(javaToolchains.launcherFor { languageVersion.set(JavaLanguageVersion.of(11)) })
   }
   tasks.withType<JavaCompile> {
-    sourceCompatibility = "17"
-    options.release.set(11)
-
+    if (project.name != "core") {
+      sourceCompatibility = "17"
+      options.release.set(11)
+    } else {
+      sourceCompatibility = "8"
+    }
     dependsOn(submodulesUpdate)
   }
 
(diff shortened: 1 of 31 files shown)


jacques-n:
Can you propose an alternative that leverages jabel? I'd like to see the difference in terms of code changes. Targeting source code to java 8 feels like the wrong choice.

baibaichen:
I'm not very familiar with jabel, it may take a while

FelixYBW:
> as to focus on Java 8 in its first release. Ja

SInce we are going to add the spark plan conversion here, and java8 is still very widely used to run Spark. So it will be great if we can support java 8 directly. Once spark removed the java8 support, we may upgrade version. 

Using jable needs us to compile the substrait-java seperately, then convert the bytecode, then call from Spark, it's not a convinent way.

jacques-n:
Going to prefer #97 over this. Closing for now.

PR #96, opened 2022-10-22
Add ReadRel LocalFiles

baibaichen:
Currently substrait-java doesn't implement `LocalFiles` ReadRel,  this patch adds this missing `Rel`.  

## How to test

Add a new UT `LocalFilesRoundtripTest`

### DIFF
--- a/core/src/test/java/io/substrait/type/proto/LocalFilesRoundtripTest.java
+++ b/core/src/test/java/io/substrait/type/proto/LocalFilesRoundtripTest.java
@@ -1,0 +1,114 @@
+package io.substrait.type.proto;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+import io.substrait.expression.ExpressionCreator;
+import io.substrait.expression.FieldReference;
+import io.substrait.expression.ImmutableFieldReference;
+import io.substrait.expression.proto.FunctionCollector;
+import io.substrait.function.SimpleExtension;
+import io.substrait.proto.ReadRel;
+import io.substrait.relation.LocalFiles;
+import io.substrait.relation.ProtoRelConverter;
+import io.substrait.relation.RelProtoConverter;
+import io.substrait.relation.files.FileOrFiles;
+import io.substrait.relation.files.ImmutableFileFormat;
+import io.substrait.relation.files.ImmutableFileOrFiles;
+import io.substrait.type.ImmutableNamedStruct;
+import io.substrait.type.Type;
+import io.substrait.type.TypeCreator;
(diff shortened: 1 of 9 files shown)


baibaichen:
cc @jacques-n

jacques-n (changes requested):
This generally looks good. I think it probably makes sense to make each of the formats empty classes instead of enums so that we can introduce options on each of those formats later without having breaking changes. Can you make that change?

baibaichen:
> This generally looks good. I think it probably makes sense to make each of the formats empty classes instead of enums so that we can introduce options on each of those formats later without having breaking changes. Can you make that change?

Sure, let me first verify it in the gluten and see whther it's ok or not, it could take me a while.

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
```diff
@@ -112,6 +113,11 @@ public RelNode visit(NamedScan namedScan) throws RuntimeException {
     return relBuilder.scan(namedScan.getNames()).build();
   }
 
+  @Override
+  public RelNode visit(LocalFiles localFiles) throws RuntimeException {
+    throw new UnsupportedOperationException("LocalFiles is not supported");
```
jacques-n, line 118: Why are you throwing here?

jacques-n (commented):
Can you please add roundtrip tests to this?

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
baibaichen: > Why are you throwing here?

Do you  know which Calcite rel to use for `LocalFiles` ?

core/src/test/java/io/substrait/type/proto/LocalFilesRoundtripTest.java
```diff
@@ -0,0 +1,114 @@
...
+import io.substrait.relation.files.ImmutableFileFormat;
+import io.substrait.relation.files.ImmutableFileOrFiles;
+import io.substrait.type.ImmutableNamedStruct;
+import io.substrait.type.Type;
+import io.substrait.type.TypeCreator;
+import java.io.IOException;
+import java.util.Arrays;
+import org.junit.jupiter.api.Test;
+
+public class LocalFilesRoundtripTest {
```
baibaichen, line 25: @jacques-n 

Do you mean this?

jacques-n: Yes, apparently missed that on first pass. Thanks

jacques-n (approved):

PR #97, opened 2022-10-25
Downgrade core to java 8 via Jabel

baibaichen:
This patch does the same thing as https://github.com/substrait-io/substrait-java/pull/95, but using jabel


jacques-n:
Seems like small enough refactoring to accept. Let's see long term how much Java8 this is a burden and reevaluate based on that.

Thanks for the patch @baibaichen !

PR #101, opened 2022-11-08
refactor: Add FeatureBoard for configuring Isthmus

ashvina:
This change adds a FeatureBoard, a collection of configuration and flags to control the behavior of SQL to Substrait conversion and vice versa. The original intent was to add a config for Calcite's conformance mode. Expecting more such changes in future, this PR adds a generic FeatureBoard which should provide sufficient hints to user (leveraging command-line help messages) and easier configuration using properties file (see picocli PropertiesDefaultProvider).

FeatureBoard replaces Options and is inspired by Feature Toggle design pattern. All configs are available in one place, instead of Options which required config hunting in multiple files. Also some Options were not accessible from command line. 

This change is backward compatible.


jacques-n:
LGTM. Thanks @ashvina !

PR #106, opened 2022-11-23
feat: support parsing of SQL queries with APPLY

ashvina:
This change adds support for parsing of SQL queries with APPLY (join with
correlated subquery), and to build OuterReferences map of correlated variables
present in the query's join predicates. The OuterRefs will be used while
constructing Substrait plans to bind correlated variables. The change also
adds few example queries which depend on APPLY / LATERAL operators.

This change still does not map calcite-correlated-join to Substrait, as the
spec for APPLY is still not approved. As such, while the parsing of calcite
query plans will succeed after this change, the unit tests and run time
conversion will continue to fail in the final step of building the
Substrait plan. Additional changes are needed to support APPLY.

Refs #substrait-io/substrait/issues/357

### DIFF
--- a/isthmus/src/main/java/io/substrait/isthmus/SubstraitRelVisitor.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/SubstraitRelVisitor.java
@@ -207,8 +207,8 @@ public class SubstraitRelVisitor extends RelNodeVisitor<Rel, RuntimeException> {
         switch (correlate.getJoinType()) {
           case INNER -> Join.JoinType.INNER; // corresponds to CROSS APPLY join
           case LEFT -> Join.JoinType.LEFT; // corresponds to OUTER APPLY join
-          default -> throw new UnsupportedOperationException(
-              "Unsupported correlated join type: " + correlate.getJoinType());
+          default -> throw new IllegalArgumentException(
+              "Invalid correlated join type: " + correlate.getJoinType());
         };
     return super.visit(correlate);
   }
--- a/isthmus/src/main/java/io/substrait/isthmus/SubstraitRelVisitor.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/SubstraitRelVisitor.java
@@ -61,6 +61,7 @@ import org.immutables.value.Value;
 @SuppressWarnings("UnstableApiUsage")
 @Value.Enclosing
 public class SubstraitRelVisitor extends RelNodeVisitor<Rel, RuntimeException> {
+
   static final org.slf4j.Logger logger =
       org.slf4j.LoggerFactory.getLogger(SubstraitRelVisitor.class);
   private static final FeatureBoard FEATURES_DEFAULT = ImmutableFeatureBoard.builder().build();
@@ -196,6 +197,19 @@ public class SubstraitRelVisitor extends RelNodeVisitor<Rel, RuntimeException> {
 
   @Override
   public Rel visit(LogicalCorrelate correlate) {
+    // left input of correlated-join is similar to the left input of a logical join
+    apply(correlate.getLeft());
+
+    // right input of correlated-join is similar to a correlated sub-query
+    apply(correlate.getRight());
+
+    var joinType =
+        switch (correlate.getJoinType()) {
+          case INNER -> Join.JoinType.INNER; // corresponds to CROSS APPLY join
+          case LEFT -> Join.JoinType.LEFT; // corresponds to OUTER APPLY join
+          default -> throw new UnsupportedOperationException(
+              "Unsupported correlated join type: " + correlate.getJoinType());
+        };
     return super.visit(correlate);
   }
 
--- a/isthmus/src/test/java/io/substrait/isthmus/ApplyJoinPlanTest.java
+++ b/isthmus/src/test/java/io/substrait/isthmus/ApplyJoinPlanTest.java
@@ -1,27 +1,74 @@
 package io.substrait.isthmus;
 
+import java.util.Map;
 import org.apache.calcite.adapter.tpcds.TpcdsSchema;
+import org.apache.calcite.rel.RelRoot;
+import org.apache.calcite.rex.RexFieldAccess;
 import org.apache.calcite.sql.parser.SqlParseException;
+import org.apache.calcite.sql.parser.SqlParser;
 import org.apache.calcite.sql.validate.SqlConformanceEnum;
 import org.junit.jupiter.api.Assertions;
 import org.junit.jupiter.api.Test;
 
 public class ApplyJoinPlanTest {
 
+  private static RelRoot getCalcitePlan(SqlToSubstrait s, TpcdsSchema schema, String sql)
+      throws SqlParseException {
+    var pair = s.registerSchema("tpcds", schema);
+    var converter = s.createSqlToRelConverter(pair.left, pair.right);
+    SqlParser parser = SqlParser.create(sql, s.parserConfig);
+    var root = s.getBestExpRelRoot(converter, parser.parseQuery());
+    return root;
+  }
+
+  private static void validateOuterRef(
+      Map<RexFieldAccess, Integer> fieldAccessDepthMap, String refName, String colName) {
+    var rexField =
+        fieldAccessDepthMap.keySet().stream()
+            .filter(f -> f.getReferenceExpr().toString().equals(refName))
+            .findFirst();
+    Assertions.assertTrue(rexField.isPresent());
+    Assertions.assertEquals(colName, rexField.get().getField().getKey());
+  }
+
+  private static Map<RexFieldAccess, Integer> buildOuterFieldRefMap(RelRoot root) {
+    final OuterReferenceResolver resolver = new OuterReferenceResolver();
+    var fieldAccessDepthMap = resolver.getFieldAccessDepthMap();
+    Assertions.assertEquals(0, fieldAccessDepthMap.size());
+    resolver.apply(root.rel);
+    return fieldAccessDepthMap;
+  }
+
   @Test
-  public void lateralJoinQuery() {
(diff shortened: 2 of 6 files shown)


jacques-n:
@jinfengni and @weijietong , can you both review this?

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelVisitor.java
```diff
@@ -196,6 +197,21 @@ public Rel visit(LogicalJoin join) {
 
   @Override
   public Rel visit(LogicalCorrelate correlate) {
+    // left input of correlated-join is similar to the left input of a logical join
+    var left = apply(correlate.getLeft());
```
weijietong, line 201: remove unused vars.

ashvina: Thanks for the review. For now I have removed the unused variables. This PR is part a feature developed in iterations. The currently unused variables would be needed for building Substrait plan, planned for the upcoming PR. I have not removed the join-type variable because it helps in clarifying the expected types till the next PR is merged.

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelVisitor.java
```diff
@@ -196,6 +197,19 @@ public Rel visit(LogicalJoin join) {
...
+    apply(correlate.getLeft());
+
+    // right input of correlated-join is similar to a correlated sub-query
+    apply(correlate.getRight());
+
+    var joinType =
+        switch (correlate.getJoinType()) {
+          case INNER -> Join.JoinType.INNER; // corresponds to CROSS APPLY join
+          case LEFT -> Join.JoinType.LEFT; // corresponds to OUTER APPLY join
+          default -> throw new UnsupportedOperationException(
```
jinfengni, line 210: Could `JoinType` be `RIGHT` ?   Why do we through `UnsupportedOperationException` for other join type?

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelVisitor.java
```diff
@@ -196,6 +197,19 @@ public Rel visit(LogicalJoin join) {
 
   @Override
   public Rel visit(LogicalCorrelate correlate) {
+    // left input of correlated-join is similar to the left input of a logical join
+    apply(correlate.getLeft());
+
+    // right input of correlated-join is similar to a correlated sub-query
+    apply(correlate.getRight());
+
+    var joinType =
```
jinfengni, line 206: Is `joinType` an unused variable?

isthmus/src/test/java/io/substrait/isthmus/ApplyJoinPlanTest.java
```diff
@@ -0,0 +1,69 @@
...
+        """
+        SELECT ss_sold_date_sk, ss_item_sk, ss_customer_sk
+        FROM store_sales CROSS JOIN LATERAL
+          (select i_item_sk from item where item.i_item_sk = store_sales.ss_item_sk)""";
+
+    SqlToSubstrait s = new SqlToSubstrait();
+    Assertions.assertThrows(
+        UnsupportedOperationException.class,
+        () -> s.execute(sql, "tpcds", schema),
+        "Lateral join is not supported");
```
jinfengni, line 25: With this `not supported` exception as the expected result, it's hard to see how the code change works.

If Substrait's spec for APPLY is not ready yet, is it possible to verify the first step's output, instead of going through all the steps from SQL -> Substrait? Another choice is to wait until the spec of APPLY is ready.

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelVisitor.java
ashvina: `APPLY` is a nested-loop join, it evaluates the `RIGHT` expression for each record in the `LEFT` relation. Hence, I don't think other join-types, including `RIGHT`, are valid for `APPLY`.

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelVisitor.java
ashvina: Yes, I believe leaving this unused-variable in the code for now helps in code readability. I mentioned the following in response to another comment ([here](https://github.com/substrait-io/substrait-java/pull/106#discussion_r1032851122)): 
> I have not removed the join-type variable because it helps in clarifying the expected types till the next PR is merged.

Given this features is developed in iterations, I believe it is best to not remove this variable as it will be needed in the next PR. It is not dead code.

isthmus/src/test/java/io/substrait/isthmus/ApplyJoinPlanTest.java
ashvina: I agree. And I like the suggestion to expose internal data structures to validate the first step. I will try to create it.

Please note that the `Calcite-plan` for `APPLY` query already exists. As the Calcite-plan is external to Substrait, parsing it is deterministic and is independent of the `APPLY` spec in Substrait. As such, I believe this PR should precede calcite-to-Substrait. In-fact this was requested in one of the community meetings as it helps in understanding how `APPLY` is coded in Calcite.

But I agree, the tests in the PR do not confirm if the intended data structures in Substrait are correctly updated. I will try to incorporate it. Thanks for the suggestion.

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelVisitor.java
ashvina: For reference please refer to [SQL dialect recognized by Calcite’s default SQL parser](https://calcite.apache.org/docs/reference.html). Below is text relevant to the PR. A `JOIN` can be `left or right` but `APPLY` can only be `OUTER` which maps to `LEFT`.
> tableExpression [ NATURAL ] [ { LEFT | RIGHT | FULL } [ OUTER ] ] JOIN tableExpression [ joinCondition ]
   tableExpression [ CROSS | OUTER ] APPLY tableExpression

isthmus/src/test/java/io/substrait/isthmus/ApplyJoinPlanTest.java
jinfengni: Totally agree.  It helps to verify the result of first step in the test, as it will verify that the internal data is updated as expected. Thanks for the effort.

ashvina: I have pushed a change to validate the first step (calcite plan parsing). I also added the calcite plan corresponding to the query for reference. Thanks.

isthmus/src/test/java/io/substrait/isthmus/ApplyJoinPlanTest.java
```diff
@@ -0,0 +1,125 @@
...
+  private static RelRoot getCalcitePlan(SqlToSubstrait s, TpcdsSchema schema, String sql)
+      throws SqlParseException {
+    var pair = s.registerSchema("tpcds", schema);
+    var converter = s.createSqlToRelConverter(pair.left, pair.right);
+    SqlParser parser = SqlParser.create(sql, s.parserConfig);
+    var root = s.getBestExpRelRoot(converter, parser.parseQuery());
+    return root;
+  }
+
+  private static void validateOuterRef(
```
jinfengni, line 24: In addition to validate `$cor0.SS_ITEM_SK` is in the map, is it possible to verify the depth of the correlated variable?  

Also, the test seems to have a one level nested correlated variable (depth = 1).  Does it make sense to have multi-level correlated variable (i.e. depth = 2) and verify it works as expected?

isthmus/src/test/java/io/substrait/isthmus/ApplyJoinPlanTest.java
jinfengni: The new tests look great. I just have one minor comment. Thank you for responding with new code change.

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelVisitor.java
jinfengni: Make sense to me.  

If I understand correctly, the `switch` is to check if `JoinType` is indeed Inner or Left. For other types, something is wrong in the passed in `LogicalCorrelate`.  Does it make sense to use `UnsupportedOperationException` here, or maybe use `assert` to make sure the condition is met?

ashvina: I think this is a good point. In case of `LogicanlCorrelate` and other join type is invalid. How about throwing `IllegalArgumentException` if any other join type is passed?

isthmus/src/test/java/io/substrait/isthmus/ApplyJoinPlanTest.java
ashvina: Thanks for suggesting it in the first place. I appreciate your thorough review.

isthmus/src/test/java/io/substrait/isthmus/ApplyJoinPlanTest.java
ashvina: Thanks again for the suggestion. In the latest update I have added test for nested-apply-join and added depth verification too.

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelVisitor.java
jinfengni: Yes, `IllegalArgumentException` makes sense for invalid input.

jinfengni (approved):
LGTM. Thanks for working on this feature!

weijietong (approved):

PR #111, opened 2022-11-25
bugfix: quote signing key in sanity script

westonpace:


westonpace:
CI failures unrelated.  This is a trivial change and I'm going to push it so I can run the sanity script on the github secret.

PR #115, opened 2022-12-30
feat: add FailureBehavior enum to Cast immutable

gatesn:


JamesRTaylor:
@gatesn - thanks for the PR. The change looks fine, but I think it needs to be rebased to get rid of the error we're seeing. Also, there's a linter for the commit message, so if you amend your commit message to match the title of your PR, it should be ok:
```
feat: add FailureBehavior enum to Cast immutable
```

gatesn:
I think that should be fixed up now

PR #117, opened 2023-01-10
feat!: Set RelBuilder directly in SubstraitRelNodeConverter

vbarua:
Adds a new constructor for `SubstraitRelNodeConverter` that allows for directly setting the `RelBuilder`. This allows users to directly configure the `RelBuilder` and to apply optimizations successfully to the plans generated by the converter.

The first commit in these changes introduces the new constructor. The following commits are all quality of life refactors (in my opinion). If they are not desirable I'm happy to remove them.

# Context
I'm trying to leverage the work in Isthmus to convert a Substrait relational tree to a Calcite relational tree, and then apply optimizations. I'm using the `SubstraitRelNodeConverter` to achieve the conversion, which works great, however there is an issue that occurs when trying to optimize the converted plan.

# Problem
The problem stems from the following code in Calcite:
https://github.com/apache/calcite/blob/4bebdb07c2f45a95c9a4fdf81e9bcfbdd11a15de/core/src/main/java/org/apache/calcite/plan/volcano/VolcanoPlanner.java#L1246-L1249

The issue is that the RelBuilder constructed in 
https://github.com/substrait-io/substrait-java/blob/63be6ea1f4858c59a3d7a5eaaf0937ac53cc4df4/isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java#L73-L79
creates a new VolcanoPlanner, which is then used to the create the RelNodes in the converted plan, which get registered with that planner. The way we've set up our system, we already have VolcanoPlanner that we're using for optimization (with all of our own configs) and trying to use that to the optimize the converted plan triggers the assertion.

Generally as well, not being able to configure the RelBuilder directly also results in issues when using non-standard traits.

### DIFF
--- a/isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
@@ -27,7 +27,7 @@ import java.util.stream.Collectors;
 import java.util.stream.IntStream;
 import org.apache.calcite.plan.RelOptCluster;
 import org.apache.calcite.plan.RelTraitDef;
-import org.apache.calcite.prepare.CalciteCatalogReader;
+import org.apache.calcite.prepare.Prepare;
 import org.apache.calcite.rel.RelCollations;
 import org.apache.calcite.rel.RelFieldCollation;
 import org.apache.calcite.rel.RelNode;
@@ -49,7 +49,7 @@ import org.apache.calcite.tools.RelBuilder;
 public class SubstraitRelNodeConverter extends AbstractRelVisitor<RelNode, RuntimeException> {
 
   private final RelOptCluster relOptCluster;
-  private final CalciteCatalogReader catalogReader;
+  private final Prepare.CatalogReader catalogReader;
 
   private final SimpleExtension.ExtensionCollection extensions;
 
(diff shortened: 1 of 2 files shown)


isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
```diff
@@ -48,57 +50,67 @@
...
+    this.aggregateFunctionConverter =
+        new AggregateFunctionConverter(extensions.aggregateFunctions(), typeFactory);
+
+    this.expressionRexConverter =
+        new ExpressionRexConverter(
+            typeFactory, scalarFunctionConverter, aggregateFunctionConverter);
+  }
+
+  @Deprecated
   public SubstraitRelNodeConverter(
```
vbarua, line 80: I would opt to remove this constructor entirely, but I have not at this point because I believe that would be a breaking change. As such I have marked is as deprecated for now.

jacques-n: I'm okay with removing this. It's a breaking change but we're still earlier. Just add a breaking change note to the PR description

jacques-n (commented):
Update looks good. I say remove the old constructor and add a commit message that there is that breaking change. We're still early and keeping things clean now will make life easier in the future.

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
vbarua: Went ahead and remove this constructor. Rebased to make the history nicer as well.

vbarua:
I updated the final commit to use `BREAKING CHANGE` rather than `refactor!: <message>` because the latter was making the linter unhappy.

vbarua:
@jacques-n I've gone ahead and bumped the node-version to 18 in the workflow configuration. I think/hope that should fix the ci error of
```
[semantic-release]: node version >=18 is required. Found v16.18.1.
```

jacques-n:
Thanks @vbarua

PR #126, opened 2023-01-31
fix: decimal parsing reversed

vbarua:
The Decimal type is specified as `Decimal<P,S>` on https://substrait.io/types/type_classes/

The parser was configured to parse the first parameter as scale, and the second as precision.
```
Decimal isnull='?'? Lt scale=numericParameter Comma precision=numericParameter Gt #decimal
```

This PR fixes this and adds additional test coverage.


JamesRTaylor (approved):
Thanks for the fix

PR #131, opened 2023-02-16
fix: incorrect mapping of floating point + and - ops

vbarua:
* Fixes an issue preventing conversion of floating point + and - ops from Substrait to Calcite.
* Adds additional arithmetic operators available in the `SqlStdOperatorTable`.
* Adds tests for the above.
* Refactors `PlanTestBase` utils to allow for custom schema usage.


JamesRTaylor (approved):
Looks great. Nice tests!

PR #132, opened 2023-03-06
feat: support output order remapping

vbarua:
* Makes fields in SubstraitRelNodeConverter protected to allow for re-use when extending it with custom behaviours.
* Adds deriveRecordType to Cross to make it consistent with other Rels.
* Adds SubstraitBuilder, which encodes a DSL to make it easier to build plans programmatically.
* Adds SubstraitToCalcite, to convert directly from Substrait to Calcite. Includes functionality to extract a schema from the Substrait Rel directly.
* Bumps the junit-jupiter version for access to the assertInstanceOf method.
* Adds tests to verify the layout of Rels when converted w/ both Direct and Remap output ordering.
* Adds support for the Remap output type to SubstraitRelNodeConverter.


core/src/main/java/io/substrait/dsl/SubstraitBuilder.java
```diff
@@ -0,0 +1,288 @@
...
+import io.substrait.type.Type;
+import io.substrait.type.TypeCreator;
+import java.util.Arrays;
+import java.util.List;
+import java.util.Optional;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+public class SubstraitBuilder {
```
vbarua, line 32: See SubstraitRelNodeConverter for an example of how the methods in this class are used.

core/src/main/java/io/substrait/relation/ProtoRelConverter.java
```diff
@@ -336,15 +336,7 @@ private Join newJoin(JoinRel rel) {
   private Rel newCross(CrossRel rel) {
     Rel left = from(rel.getLeft());
     Rel right = from(rel.getRight());
-    Type.Struct leftStruct = left.getRecordType();
-    Type.Struct rightStruct = right.getRecordType();
-    Type.Struct unionedStruct = Type.Struct.builder().from(leftStruct).from(rightStruct).build();
-    return Cross.builder()
-        .left(left)
-        .right(right)
-        .deriveRecordType(unionedStruct)
```
vbarua, line 345: Follows from adding `deriveRecordType` to `Cross`.

core/src/main/java/io/substrait/relation/RelCopyOnWriteVisitor.java
```diff
@@ -163,7 +163,6 @@ public Optional<Rel> visit(Cross cross) throws RuntimeException {
             .from(cross)
             .left(left.orElse(cross.getLeft()))
             .right(right.orElse(cross.getRight()))
-            .deriveRecordType(unionedStruct)
```
vbarua, line 166: Follows from adding `deriveRecordType` to `Cross`.

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelVisitor.java
```diff
@@ -185,12 +185,7 @@ public Rel visit(LogicalJoin join) {
     if (joinType == Join.JoinType.INNER
         && TRUE.equals(condition)
         && featureBoard.crossJoinPolicy().equals(KEEP_AS_CROSS_JOIN)) {
-      return Cross.builder()
-          .left(left)
-          .right(right)
-          .deriveRecordType(
```
vbarua, line 191: Follows from adding `deriveRecordType` to `Cross`.

isthmus/src/test/java/io/substrait/isthmus/PlanTestBase.java
```diff
@@ -23,10 +24,20 @@
...
+    } catch (IOException e) {
+      throw new RuntimeException(e);
+    }
+  }
+
   protected final RelCreator creator = new RelCreator();
   protected final RelBuilder builder = creator.createRelBuilder();
   protected final RexBuilder rex = creator.rex();
-  protected final RelDataTypeFactory type = creator.type();
+  protected final RelDataTypeFactory typeFactory = creator.typeFactory();
```
vbarua, line 40: I'm using this in my new tests, but it wasn't used before. Updated the named to `typeFactory` because I found `type` a confusing name for this.

isthmus/src/test/java/io/substrait/isthmus/SubstraitRelNodeConverterTest.java
```diff
@@ -0,0 +1,242 @@
...
+
+  void assertRowMatch(RelDataType actual, List<Type> expected) {
+    Type type = TypeConverter.convert(actual);
+    assertInstanceOf(Type.Struct.class, type);
+    Type.Struct struct = (Type.Struct) type;
+    assertEquals(expected, struct.fields());
+  }
+
+  @Nested
+  class Aggregate {
```
vbarua, line 49: Using Nested classes here with the idea that more Rel specific tests can/will be added in the future.

core/src/main/java/io/substrait/relation/Cross.java
```diff
@@ -1,10 +1,21 @@
...
+import io.substrait.type.Type;
+import io.substrait.type.TypeCreator;
+import java.util.stream.Stream;
 import org.immutables.value.Value;
 
 @Value.Immutable
 public abstract class Cross extends BiRel {
 
+  @Override
+  protected Type.Struct deriveRecordType() {
```
vbarua, line 12: I've realised that adding this removes the `deriveRecordType` method from the `ImmutableCross.Builder`, meaning this is technically a breaking change.

isthmus/src/main/java/io/substrait/isthmus/SubstraitToCalcite.java
```diff
@@ -0,0 +1,119 @@
...
+import org.apache.calcite.tools.Frameworks;
+import org.apache.calcite.tools.RelBuilder;
+
+/**
+ * Converts between Substrait {@link Rel}s and Calcite {@link RelNode}s.
+ *
+ * <p>Can be extended to customize the {@link RelBuilder} and {@link SubstraitRelNodeConverter} used
+ * in the conversion.
+ */
+public class SubstraitToCalcite {
```
vbarua, line 26: Thinking about this some more, though I'm pretty happy with this I don't know if I'm "happy as public API forever" kind of happy. Could make this protected and keep it internal only for now.

core/src/main/java/io/substrait/dsl/SubstraitBuilder.java
adamkennedy: I like the idea of moving towards an easier to write wrapper for generating plans, but any "usability" API would almost certainly need to come with proper docs, ideally containing short usage examples. An undocumented usability API, or one only documented by having to read the source code of other code, is going to be of limited use.

isthmus/src/main/java/io/substrait/isthmus/SubstraitToCalcite.java
lwhite1: I think if you're uncertain, it might be best to be conservative and leave it protected until someone needs it.

JamesRTaylor (approved):
Nice improvement. Thanks for the fixes too.

isthmus/src/main/java/io/substrait/isthmus/SubstraitToCalcite.java
vbarua: This got merged before I circled back to hiding this.

Looking at https://substrait.io/spec/versioning/, because we're pre-1.0 I'm more comfortable with having it in. It is actually helpful to have as downstream user and it does make it easier for folks experiment with converting substrait plans directly to Calcite.

core/src/main/java/io/substrait/dsl/SubstraitBuilder.java
vbarua: This got merged before I circled back to hiding this.

Looking at https://substrait.io/spec/versioning/, because we're pre-1.0 I'm more comfortable with having it in. It is super useful for writing tests, both in substream-java and downstream.

> but any "usability" API would almost certainly need to come with proper docs, ideally containing short usage examples.

100% something to have in place before 1.0.

PR #133, opened 2023-03-10
proposal: any type handling

vbarua:
This PR is split into two commits:
* The first commits adds code for handling `google.protobuf.Any` types in the AdvanceExtension message, ReadRel.ExtensionTable message and ExtensionSingleRel messages.
* The second commit includes tests for the ProtoRelConverter, including a demonstration of how it would be overridden by the user to handle `google.protobuf.Any` types.

## Summary

`com.protobuf.Any` types are handled by methods in the `ProtoRelConverter` that are intended to be overriden by end users. For example:

```java
public class ProtoRelConverter {
  ...  
  protected Extension.Optimization optimizationFromAdvancedExtension(com.google.protobuf.Any any)
  ...
}
```
can be overriden by end users to convert the Any into their own Java equivalent. 

`Extension.Optimization` is a tag interface to be able to distinguish between various types of Any data (instead of having `Object` everywhere).

```java
/** Contains tag interfaces for handling {@link com.google.protobuf.Any} types within Substrait. */
public class Extension {

  private interface ToProto {
    com.google.protobuf.Any toProto();
  }

  public interface Optimization extends ToProto {}
 ...
```

### DIFF
--- a/core/src/main/java/io/substrait/relation/ProtoRelConverter.java
+++ b/core/src/main/java/io/substrait/relation/ProtoRelConverter.java
@@ -2,6 +2,7 @@ package io.substrait.relation;
 
 import static io.substrait.expression.proto.ProtoExpressionConverter.EMPTY_TYPE;
 
+import com.google.protobuf.Any;
 import io.substrait.expression.AggregateFunctionInvocation;
 import io.substrait.expression.Expression;
 import io.substrait.expression.FunctionArg;
@@ -9,8 +10,10 @@ import io.substrait.expression.FunctionLookup;
 import io.substrait.expression.ImmutableExpression;
 import io.substrait.expression.proto.ProtoExpressionConverter;
 import io.substrait.function.SimpleExtension;
+import io.substrait.io.substrait.extension.AdvancedExtension;
 import io.substrait.proto.AggregateRel;
 import io.substrait.proto.CrossRel;
+import io.substrait.proto.ExtensionSingleRel;
 import io.substrait.proto.FetchRel;
 import io.substrait.proto.FilterRel;
 import io.substrait.proto.JoinRel;
@@ -18,6 +21,8 @@ import io.substrait.proto.ProjectRel;
 import io.substrait.proto.ReadRel;
 import io.substrait.proto.SetRel;
 import io.substrait.proto.SortRel;
+import io.substrait.relation.extensions.EmptyDetail;
+import io.substrait.relation.extensions.EmptyOptimization;
 import io.substrait.relation.files.FileOrFiles;
 import io.substrait.relation.files.ImmutableFileFormat;
 import io.substrait.relation.files.ImmutableFileOrFiles;
@@ -77,8 +82,11 @@ public class ProtoRelConverter {
       case CROSS -> {
         return newCross(rel.getCross());
       }
+      case EXTENSION_SINGLE -> {
+        return newExtensionSingle(rel.getExtensionSingle());
+      }
       default -> {
-        // TODO: add support for EXTENSION_SINGLE, EXTENSION_MULTI, EXTENSION_LEAF
+        // TODO: add support for EXTENSION_MULTI, EXTENSION_LEAF
         throw new UnsupportedOperationException("Unsupported RelTypeCase of " + relType);
       }
     }
@@ -91,6 +99,8 @@ public class ProtoRelConverter {
       return newNamedScan(rel);
     } else if (rel.hasLocalFiles()) {
       return newLocalFiles(rel);
+    } else if (rel.hasExtensionTable()) {
+      return newExtensionTable(rel);
     } else {
       return newEmptyScan(rel);
     }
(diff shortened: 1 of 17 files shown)


core/src/main/java/io/substrait/relation/ProtoRelConverter.java
```diff
@@ -137,12 +147,26 @@ private EmptyScan newEmptyScan(ReadRel rel) {
         .build();
   }
 
+  private ExtensionSingleInput newExtensionSingle(ExtensionSingleRel rel) {
+    var input = from(rel.getInput());
+    var builder =
+        ExtensionSingleInput.builder()
+            .input(input)
+            .remap(optionalRelmap(rel.getCommon()))
+            .extension(optionalAdvancedExtension(rel.getCommon()));
```
JamesRTaylor, line 156: If there was more than one type of SingleRel extension, how would those be stored? I suppose that could be encapsulated in the interpretation of the protobuf.Any data?

isthmus/src/test/java/io/substrait/isthmus/SubstraitRelNodeConverterTest.java
```diff
@@ -236,4 +243,58 @@ public void emit() {
...
+        return new CustomSubstraitRelNodeConverter(extensions, typeFactory, relBuilder);
+      }
+    }
+
+    @Test
+    public void defaultValue() {
+      Plan.Root root = b.root(b.namedScan(List.of("foo"), List.of(), List.of(), advancedExtension));
+
+      var customConverter = new CustomSubstraitToCalcite(extensions, typeFactory);
+      var relNode = customConverter.convert(root.getInput());
```
JamesRTaylor, line 297: Would it be possible to add some kind of assertion here?

core/src/main/java/io/substrait/relation/ProtoRelConverter.java
```diff
@@ -77,8 +82,11 @@ public Rel from(io.substrait.proto.Rel rel) {
       case CROSS -> {
         return newCross(rel.getCross());
       }
+      case EXTENSION_SINGLE -> {
+        return newExtensionSingle(rel.getExtensionSingle());
+      }
       default -> {
-        // TODO: add support for EXTENSION_SINGLE, EXTENSION_MULTI, EXTENSION_LEAF
+        // TODO: add support for EXTENSION_MULTI, EXTENSION_LEAF
```
JamesRTaylor, line 89: I see some methods implemented for these other cases below (EXTENSION_MULTI, EXTENSION_LEAF), but are they not completely implemented?

core/src/test/java/io/substrait/relation/utils/StringHolder.java
```diff
@@ -0,0 +1,37 @@
...
+import com.google.protobuf.Any;
+import io.substrait.relation.Extension;
+import java.util.Objects;
+
+/**
+ * For use in {@link io.substrait.relation.ProtoRelConverterTest}
+ *
+ * <p>Used to verify serde of {@link com.google.protobuf.Any} fields in the spec
+ */
+public class StringHolder implements Extension.Enhancement, Extension.Optimization {
```
JamesRTaylor, line 12: Just curious if an extension protobuf.Any was partly composed of already supported operators (or other know objects), do you think that the processing could still take place smoothly?

JamesRTaylor:
This is really cool! Thanks so much for putting it together, @vbarua. What do you think would be a good real world use case for this feature?

isthmus/src/test/java/io/substrait/isthmus/SubstraitRelNodeConverterTest.java
vbarua: For sure. I want to add some more testing once we're aligned on an approach.

core/src/test/java/io/substrait/relation/utils/StringHolder.java
vbarua: I don't think I can speak to a general case as that's going to depend on how you've structured your message. This effectively just gives you a hook where you have access to the underlying Any. It should be flexibly enough to allow you convert your data into whatever you need. Something like:
```java
  @Override
  protected Extension.Enhancement enhancementFromAdvancedExtension(Any any) {
    try {
      var unpackedMessage = any.unpack(YourMessage.class)
       <further processing into your own custom representation>
    } catch (InvalidProtocolBufferException e) {
      throw new RuntimeException(e);
    }
  }


```

vbarua:
@JamesRTaylor 
> What do you think would be a good real world use case for this feature?

Well, some of the stuff we want to use it for:
* **ExtensionTable**: We have a couple of internal systems we fetch data from. Our planner system is seperate from our execution system, and we want to use Extension Tables to embed query messages into the ReadRel. Effectively, our planner can already convert Substrait to our internal query messages, but our executor can't and we'd prefer not to duplicate the logic.
* **ExtensionSingleRel**: We have an in-house executor which has some custom physical relations, so we need to be able to model them in Substrait to use them.
* **AdvancedExtensions**: We're looking at adding some custom metadata for debug purposes to ensure that the plans generated by our planner don't violate invariants in our executor.

vbarua:
Superceded by https://github.com/substrait-io/substrait-java/pull/135

PR #135, opened 2023-03-22
feat: google.protobuf.Any handling

vbarua:
Adds support for handling the `google.protobuf.Any` data in the spec. This includes `Any` data in:
* `AdvancedExtension` messages. Both in `RelCommon` and directly on Rels.
* `ReadRel.ExtensionTable` messages.
* `ExtensionLeafRel`, `ExtensionSingleRel` and `ExtensionMultiRel` messages.

This is achieved by adding extension hooks to the `ProtoRelConverter`, which users can override to provide custom conversions from `Any` data to POJOs.

This code includes tests for:
* The default behaviour of the `ProtoRelConverter` in the face of various pieces of `Any` data.
* Round trip behaviour for an extended `ProtoRelConverter` with custom data handling.

### DIFF
--- a/core/src/main/java/io/substrait/relation/Extension.java
+++ b/core/src/main/java/io/substrait/relation/Extension.java
@@ -7,33 +7,29 @@ import java.util.List;
 /** Contains tag interfaces for handling {@link com.google.protobuf.Any} types within Substrait. */
 public class Extension {
 
-  private interface ToProto {
-    com.google.protobuf.Any toProto();
-  }
-
   /**
    * Optimization associated with an {@link io.substrait.proto.AdvancedExtension}
    *
    * <p>An optimization is helpful information that don't influence semantics. May be ignored by a
    * consumer.
    */
-  public interface Optimization extends ToProto {}
+  public interface Optimization extends ToProto<com.google.protobuf.Any> {}
 
   /**
    * Enhancement associated with an {@link io.substrait.proto.AdvancedExtension}
    *
    * <p>An enhancement alter semantics. Cannot be ignored by a consumer.
    */
-  public interface Enhancement extends ToProto {}
+  public interface Enhancement extends ToProto<com.google.protobuf.Any> {}
 
-  public interface LeafRelDetail extends ToProto {
+  public interface LeafRelDetail extends ToProto<com.google.protobuf.Any> {
     /**
      * @return the record layout for the associated {@link ExtensionLeaf} relation
      */
     Type.Struct deriveRecordType();
   }
 
-  public interface SingleRelDetail extends ToProto {
+  public interface SingleRelDetail extends ToProto<com.google.protobuf.Any> {
     /**
      * @param input to the associated {@link ExtensionSingle} relation
      * @return the record layout for the associated {@link ExtensionSingle} relation
@@ -41,7 +37,7 @@ public class Extension {
     Type.Struct deriveRecordType(Rel input);
   }
 
-  public interface MultiRelDetail extends ToProto {
+  public interface MultiRelDetail extends ToProto<com.google.protobuf.Any> {
     /**
      * @param inputs to the associated {@link ExtensionMulti} relation
      * @return the record layout for the associated {@link ExtensionMulti} relation
@@ -49,7 +45,7 @@ public class Extension {
     Type.Struct deriveRecordType(List<Rel> inputs);
   }
 
-  public interface ExtensionTableDetail extends ToProto {
+  public interface ExtensionTableDetail extends ToProto<com.google.protobuf.Any> {
     /**
      * @return the table schema for the associated {@link ExtensionTable} relation
      */
--- a/core/src/main/java/io/substrait/relation/Extension.java
+++ b/core/src/main/java/io/substrait/relation/Extension.java
@@ -10,4 +10,10 @@ public class Extension {
   public interface Optimization extends ToProto {}
 
   public interface Enhancement extends ToProto {}
+
+  public interface LeafRelDetail extends ToProto {}
+
+  public interface SingleRelDetail extends ToProto {}
+
+  public interface MultiRelDetail extends ToProto {}
 }
--- a/core/src/main/java/io/substrait/relation/Extension.java
+++ b/core/src/main/java/io/substrait/relation/Extension.java
@@ -1,5 +1,7 @@
 package io.substrait.relation;
 
+import io.substrait.type.Type;
+
 /** Contains tag interfaces for handling {@link com.google.protobuf.Any} types within Substrait. */
 public class Extension {
 
@@ -11,11 +13,18 @@ public class Extension {
 
   public interface Enhancement extends ToProto {}
 
-  public interface LeafRelDetail extends ToProto {}
+  public interface LeafRelDetail extends ToProto {
 
-  public interface SingleRelDetail extends ToProto {}
+    Type.Struct deriveRecordType();
+  }
 
-  public interface MultiRelDetail extends ToProto {}
+  public interface SingleRelDetail extends ToProto {
+    Type.Struct deriveRecordType(Rel input);
+  }
+
+  public interface MultiRelDetail extends ToProto {
+    Type.Struct deriveRecordType(Rel... inputs);
+  }
 
   public interface ExtensionTableDetail extends ToProto {}
 }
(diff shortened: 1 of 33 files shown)


vbarua:
cc: @JamesRTaylor

core/src/main/java/io/substrait/relation/Extension.java
```diff
@@ -0,0 +1,47 @@
...
+import java.util.List;
+
+/** Contains tag interfaces for handling {@link com.google.protobuf.Any} types within Substrait. */
+public class Extension {
+
+  private interface ToProto {
+    com.google.protobuf.Any toProto();
+  }
+
+  public interface Optimization extends ToProto {}
```
JamesRTaylor, line 14: What's the distinction between Optimization and Enhancement?

core/src/main/java/io/substrait/relation/HasExtension.java
```diff
@@ -0,0 +1,17 @@
+package io.substrait.relation;
+
+import io.substrait.io.substrait.extension.AdvancedExtension;
+import java.util.Optional;
+
+/**
+ * Used to indicate the presence of an {@link AdvancedExtension} directly on a relation
+ *
+ * <p>This exists outside of {@link Rel} because there are relations that do not have advanced
+ * extensions. See {@link ExtensionLeaf}, {@link ExtensionSingle}, {@link ExtensionMulti}).
```
JamesRTaylor, line 10: Which relations don't support advanced extensions? Is it only `io.substrait.relation.Set`? Would it simplify things to build this into Rel instead and throw if attempts are made to access for Rels that don't support it?

core/src/main/java/io/substrait/relation/Extension.java
mbrobbel: Explained here: https://github.com/substrait-io/substrait/blob/45b4e483ff1fca3c3e4d0f71e6e55436c6d7638a/proto/substrait/extensions/extensions.proto#L72-L81

vbarua: I've added docs to both `Optimization` and `Enhancement` w/ that information.

core/src/main/java/io/substrait/relation/Extension.java
```diff
@@ -0,0 +1,57 @@
+package io.substrait.relation;
+
+import io.substrait.type.NamedStruct;
+import io.substrait.type.Type;
+import java.util.List;
+
+/** Contains tag interfaces for handling {@link com.google.protobuf.Any} types within Substrait. */
+public class Extension {
+
+  private interface ToProto {
```
jacques-n, line 10: I'd move this out of Extension and probably apply it consistently throughout the codebase. (the move out could be done here, application elsewhere as a separate piece of work).

core/src/main/java/io/substrait/relation/Extension.java
```diff
@@ -0,0 +1,57 @@
...
+    com.google.protobuf.Any toProto();
+  }
+
+  /**
+   * Optimization associated with an {@link io.substrait.proto.AdvancedExtension}
+   *
+   * <p>An enhancement alter semantics. Cannot be ignored by a consumer.
```
jacques-n, line 17: wrong text for optimization

core/src/main/java/io/substrait/relation/HasExtension.java
```diff
@@ -0,0 +1,17 @@
...
+ */
+public interface HasExtension {
+  /**
+   * @return the {@link AdvancedExtension} associated directly with the relation, if present
+   */
+  Optional<AdvancedExtension> getRelExtension();
```
jacques-n, line 16: how about we move this to general location since other things besides rels can have extensions (and rename to getExtension)

core/src/main/java/io/substrait/relation/Project.java
```diff
@@ -8,7 +8,7 @@
 import org.immutables.value.Value;
 
 @Value.Immutable
-public abstract class Project extends SingleInputRel {
+public abstract class Project extends SingleInputRel implements HasExtension {
```
jacques-n, line 11: Why not just make `Rel extends HasExtension`?

vbarua: `ExtensionSingle`, `ExtensionLeaf` and `ExtensionMulti` all ultimately inherent from `Rel` as well, however they don't have extensions associated with them. 

I originally put `getRelExtension` directly in `Rel`, but moved it into the `HasExtension` to work around this.

core/src/main/java/io/substrait/relation/Extension.java
vbarua: Updated

core/src/main/java/io/substrait/relation/HasExtension.java
vbarua: To add to this a bit.

The POJO representation effectively merges the contents of `message RelCommon` and the different `message Rel` variants into a single Java object. Effectively we end up with a single object containing two AdvancedExtensions from different sources.

I made the names verbose to help distinguish these, but if you think that `rel.getExtension` and `rel.getCommonExtension` are clear enough I'm happy to change it.

core/src/main/java/io/substrait/relation/Extension.java
jacques-n: I suggest that we modify the interface to ToProto<T extends Message>. We have the toProto pattern in a lot of locations throughout the java codebase.

core/src/main/java/io/substrait/relation/HasExtension.java
jacques-n: I don't have a super strong opinion but it seems like I'd be inclined to have consistency across objects about how to get extensions. For example, it's annoying if it's Plan.getPlanExtension and Rel.getRelExtension as opposed to both of those simply being called getPlanExtension().

In general terms, the way we're using Any is a little weird here. Protobuf has a registry concept. I'm almost more inclined to switch this around and have it be getExtension(Registry) which then returns a list of extensions found as direct objects. This would resolve the double extension pattern we have on Rel/RelCommon and abstract away the use of Any types.

core/src/main/java/io/substrait/relation/Project.java
jacques-n: Hmm.. not strongly held but if we adjusted per above the registry/list above, the method for extension rels would simply return the common extension in those cases (and not both) since the common extension is independent of specific extension.

jacques-n (commented):
Most of my comments aren't strongly held. I think it would be good to rename the getRelExtension() method. The refactor to use a registry seems good but I don't hold it strongly. LMK your thoughts.

core/src/main/java/io/substrait/relation/Extension.java
vbarua: Gotcha. The `T extends Message` was what I was missing. It wasn't immediately clear to me how the existing `ToProto` interface could work for serializing to different message types. Parametrizing it with a generic makes sense. I've made the change.

core/src/main/java/io/substrait/relation/HasExtension.java
vbarua: Gotcha, so have `getExtension` as the standard accessor, and `getCommonExtension` as a special case for Rels because they collapse two different messages with extension into one class. I like it, I've made the change.

> In general terms, the way we're using Any is a little weird here. Protobuf has a registry concept. I'm almost more inclined to switch this around and have it be getExtension(Registry) which then returns a list of extensions found as direct objects.

I'm less inclined to make this change because I think being able to tell the provenance of the extension is useful. I can imagine a consumer which allows different sets of extensions in RelCommon than on Rels. Even without that, from the user side, it's nice to be able to tell the provenance of the extension from the message, which something like `getExtension(Registry): List<AdvancedExtension>` would throw away.

Additionally, at the relation layer I'm also trying to hide the protobuf layer. I do actually imagine that a common way to implement a custom ProtoRelConverter will be to use the registry mechanism to decode the `any` messages, but by the time we're working with data at the relation layer we (IMO) shouldn't have any protobuf code ideally.

vbarua:
@jacques-n Made some changes you suggested and replied to your comments. Let me know what you think.

jacques-n (approved):

PR #139, opened 2023-04-04
test: add rounding functions mapping and test

richtia:


isthmus/src/main/java/io/substrait/isthmus/expression/FunctionMappings.java
```diff
@@ -34,6 +34,7 @@ public class FunctionMappings {
                 s(SqlStdOperatorTable.ASIN, "asin"),
                 s(SqlStdOperatorTable.ACOS, "acos"),
                 s(SqlStdOperatorTable.ATAN, "atan"),
+                s(SqlStdOperatorTable.ATAN2, "atan2"),
```
vbarua, line 37: Could you add this function to 
https://github.com/substrait-io/substrait-java/blob/dd069f69e9ddd7e7e668188ab535174181e61d02/isthmus/src/test/java/io/substrait/isthmus/ArithmeticFunctionTest.java#L57-L62

so we have a test for it.

richtia: added!

isthmus/src/test/java/io/substrait/isthmus/RoundingFunctionTest.java
```diff
@@ -0,0 +1,33 @@
...
+  void floor(String column) throws Exception {
+    String query = String.format("SELECT floor(%s) FROM numbers", column);
+    assertSqlSubstraitRelRoundTrip(query, CREATES);
+  }
+
+  @ParameterizedTest
+  @ValueSource(strings = {"decimal"})
+  void round(String column) throws Exception {
+    String query = String.format("SELECT round(%s, 1) FROM numbers", column);
+    assertSqlSubstraitRelRoundTrip(query, CREATES);
```
vbarua, line 31: The version of substrait that's submoduled in substrait-java doesn't have `round` defined.

https://github.com/substrait-io/substrait/blob/6a74f40f0d835a447a7175b20dd430c7c39c5d11/extensions/functions_rounding.yaml

isthmus/src/test/java/io/substrait/isthmus/RoundingFunctionTest.java
```diff
@@ -0,0 +1,33 @@
...
+
+import java.util.List;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.ValueSource;
+
+public class RoundingFunctionTest extends PlanTestBase {
+
+  static List<String> CREATES =
+      List.of(
+          "CREATE TABLE numbers (i8 TINYINT, i16 SMALLINT, i32 INT, i64 BIGINT, fp32 FLOAT, fp64 DOUBLE, decimal DECIMAL(5,2)");
```
vbarua, line 11: Calcite doesn't like this CREATE statement because `decimal` is a keyword. Probably easiest to pick a different name.

substrait
vibhatha, line 1: shouldn't we avoid committing the `submodule`?

richtia: hmm...is there a problem with that?  I was under the impression that I was just bumping the version to be the latest?  @vbarua had suggested that.  I could be doing it incorrectly though.

vibhatha: Oh I understand, @richtia I could be wrong here, but would `git submodule update --init` would bump it? Not sure though, 
cc @vbarua

vbarua: I think this is just Github's UI being helpful/unhelpful. Looking at it locally, @richtia's changes are bumping this correctly to 5e99f0b, which is the latest substrait commit.

Screen Shot 2023-04-17 at 8 40 16 AM

isthmus/src/test/java/io/substrait/isthmus/RoundingFunctionTest.java
```diff
@@ -0,0 +1,33 @@
...
+  @ParameterizedTest
+  @ValueSource(strings = {"fp32", "fp64"})
+  void floor(String column) throws Exception {
+    String query = String.format("SELECT floor(%s) FROM numbers", column);
+    assertSqlSubstraitRelRoundTrip(query, CREATES);
+  }
+
+  @ParameterizedTest
+  @ValueSource(strings = {"fp32", "fp64"})
+  void round(String column) throws Exception {
```
vbarua, line 29: minor: It looks like `round` is defined for i8, i16, i32 and i64 as well. Could you add these to the value source for completeness.

vbarua (approved):
One more minor request, but overall LGTM. Thanks for updating the spec.

PR #140, opened 2023-04-14
feat: support for custom scalar and aggregate fns

vbarua:
The existing SubstraitRelNodeConverter was not extensible enough to allow for conversion of custom scalar and aggregate functions from Substrait to Calcite and back.

These changes:
*


jacques-n (approved):
Nice update. Thanks @vbarua !

PR #141, opened 2023-04-19
test: add mapping and test for logarithmic functions

richtia:
PR to add logarithmic functions mapping and tests


vbarua (approved):
Mappings look good to me

PR #142, opened 2023-04-25
feat: add comparison function mappings

richtia:
PR to add comparison functions mapping

### DIFF
--- a/isthmus/src/main/java/io/substrait/isthmus/expression/FunctionMappings.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/expression/FunctionMappings.java
@@ -60,7 +60,10 @@ public class FunctionMappings {
                 s(SqlStdOperatorTable.SUBSTRING, "substring"),
                 s(SqlStdOperatorTable.CONCAT, "concat"),
                 s(SqlStdOperatorTable.LOWER, "lower"),
-                s(SqlStdOperatorTable.UPPER, "upper"))
+                s(SqlStdOperatorTable.UPPER, "upper"),
+                s(SqlStdOperatorTable.BETWEEN),
+                s(SqlStdOperatorTable.IS_NOT_DISTINCT_FROM, "is_not_distinct_from"),
+                s(SqlStdOperatorTable.COALESCE, "coalesce"))
             .build();
 
     AGGREGATE_SIGS =


isthmus/src/test/java/io/substrait/isthmus/ComparisonFunctionTest.java
```diff
@@ -0,0 +1,29 @@
...
+public class ComparisonFunctionTest extends PlanTestBase {
+
+  static List<String> CREATES =
+      List.of("CREATE TABLE numbers (c1 TINYINT, c2 TINYINT, c3 TINYINT)");
+
+  @Test
+  void between() throws Exception {
+    String query = "SELECT c1 FROM numbers WHERE c1 BETWEEN 1 AND 100";
+    assertSqlSubstraitRelRoundTrip(query, CREATES);
+  }
```
vbarua, line 16: This test doesn't actually exercise `between` because during query parsing, Calcite converts BETWEEN into an equivalent using two comparisons and an AND.
```
LogicalProject(C1=[$0])
  LogicalFilter(condition=[AND(>=($0, 1), <=($0, 100))])
    LogicalTableScan(table=[[NUMBERS]])
```

isthmus/src/test/java/io/substrait/isthmus/ComparisonFunctionTest.java
```diff
@@ -0,0 +1,29 @@
...
+
+  @Test
+  void coalesce() throws Exception {
+    String query = "SELECT coalesce(NULL, c1) FROM numbers";
+    assertSqlSubstraitRelRoundTrip(query, CREATES);
+  }
```
vbarua, line 22: This test doesn't exercise `coalesce` because Calcite converts it to a CASE expression.
```
LogicalProject(EXPR$0=[CASE(IS NOT NULL($0), $0, IS NOT NULL($1), $1, $2)])
  LogicalTableScan(table=[[NUMBERS]])
```

isthmus/src/test/java/io/substrait/isthmus/ComparisonFunctionTest.java
```diff
@@ -0,0 +1,29 @@
...
+
+  @Test
+  void is_not_distinct_from() throws Exception {
+    String query = "SELECT * FROM numbers l JOIN numbers r ON l.c1 IS NOT DISTINCT FROM r.c2";
+    assertSqlSubstraitRelRoundTrip(query, CREATES);
+  }
```
vbarua, line 28: This test doesn't exercise `IS NOT DISTINCT FROM` because Calcite converts to a series of OR and AND expression.
```
LogicalProject(EXPR$0=[OR(AND(IS NULL($0), IS NULL($1)), IS TRUE(=($0, $1)))])
  LogicalTableScan(table=[[NUMBERS]])
```

isthmus/src/test/java/io/substrait/isthmus/ComparisonFunctionTest.java
```diff
@@ -0,0 +1,29 @@
+package io.substrait.isthmus;
+
+import java.io.*;
+import java.util.List;
+import org.junit.jupiter.api.Test;
+
+public class ComparisonFunctionTest extends PlanTestBase {
```
vbarua, line 7: These are somewhat cursed to test via SQL roundtrip, because Calcite is too smart sometimes.

1. If we want to test Substrait <-> Calcite trips (which is what the function mappings are for), that can be achieved by building either the Substrait or Calcite plan trees directly, and then checking for that roundtrip.

2. If we actually want to test the SQL -> Calcite -> Substrait -> Calcite trip, we'll need to investigate configuring Calcite to not optimised/convert those functions away.

I don't think either of these are worth spending too much time though.

richtia: Thanks for the review!  I'll remove these tests for now and just have the mapping in this PR in order to get them into Isthmus.  We can look more into adding tests in a subsequent PR.

PR #144, opened 2023-04-25
feat: support proto <-> pojo custom type conversion

vbarua:
Adds machinery for reading type extensions and handling their conversion in POJOs to proto form and back.

All changes were motivated by the example in TestTypeRoundtrip, which:
* Reads user defined types from an extension
* Reads function using those types from the extension
* Builds a POJO using those types and functions.
* Converts the POJO -> Proto -> POJO and checks that the two POJOs are the same.

The majority of changes stem come from:
* Adding the new `io.substrait.type.Type.UserDefined` to the type visitors.
* Adding code to resolve and lookup types within a plan context.
* Updating static usages of TypeProtoConverter to instance specific usages to perform said resolution and lookup. TypeProtoConverter can no longer be static as resolution and lookup are plan specific.

There a number of classes w/ TODOs to apply renames and/or moves. I held off on applying them in this PR to keep the diff smaller and also check w/ the community to see if they were reasonable before investing time in them.

### DIFF
--- a/core/src/main/java/io/substrait/function/SimpleExtension.java
+++ b/core/src/main/java/io/substrait/function/SimpleExtension.java
@@ -13,7 +13,6 @@ import com.fasterxml.jackson.dataformat.yaml.YAMLFactory;
 import com.fasterxml.jackson.datatype.jdk8.Jdk8Module;
 import io.substrait.expression.Expression;
 import io.substrait.type.Deserializers;
-import io.substrait.type.Type;
 import io.substrait.type.TypeExpressionEvaluator;
 import io.substrait.util.Util;
 import java.io.IOException;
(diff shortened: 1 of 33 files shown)


core/src/main/java/io/substrait/function/SimpleExtension.java
```diff
@@ -516,10 +524,43 @@ WindowFunctionVariant resolve(String uri, String name, String description) {
...
+
+    private final Supplier<TypeAnchor> anchorSupplier =
+        Util.memoize(() -> TypeAnchor.of(uri(), name()));
+  }
+
   @JsonDeserialize(as = ImmutableSimpleExtension.FunctionSignatures.class)
   @JsonSerialize(as = ImmutableSimpleExtension.FunctionSignatures.class)
   @Value.Immutable
   public abstract static class FunctionSignatures {
+    // TODO: Rename to ExtensionSignatures ???
```
vbarua, line 559: I want to move away from FunctionSignature as this now has types in it as well, but I'm not sure if ExtensionSignature is a good name either. I'm open to suggestions if anyone has a better alternative.

vbarua:
Running tests using `./gradlew clean test --no-build-cache` let me identify the CI issue. Fixed in b6ecf312cde386aa7261baa85868bb877fe7bc3c

jacques-n (approved):

PR #146, opened 2023-05-08
bump calcite to 1.34.0

MasseGuillaume:
https://central.sonatype.com/artifact/org.apache.calcite/calcite-core


jacques-n (approved):
thanks for the update!

PR #149, opened 2023-05-23
feat!: handle user-defined types in Isthmus

vbarua:
BREAKING CHANGE: TypeConverter no longer uses static methods
BREAKING CHANGE: SimpleExtension.MAPPER has been replaced
  with SimpleExtension.objectMapper(String namespace)


vbarua:
# Change Summary

Updated the `TypeConverter` to accept a `UserTypeMapper`, an interface which implements type conversions between Substrait and Calcite. The existing code made heavy use of the static `TypeConverter.convert` method, which didn't work well for allowing users to include their own `UserTypeMapper`. I've updated all call sites to use an instance of `TypeConverter`, and most of the this code in this PR is for propagating the `TypeConverter` through the system.

As part of this change, I renamed `convert` to `toSubstrait` and `toCalcite` for improved clarity (the original method had an overload for each side of the conversion).

Additionally, I've simplified the resolution of URIs for types by injecting the URI directly into the ObjectMapper. This came up because types in FuncArgs were being created without URIs, and propagating the resolution to them was harder than just getting rid of it entirely.

core/src/main/java/io/substrait/extension/SimpleExtension.java
```diff
@@ -798,10 +800,7 @@ public static ExtensionCollection buildExtensionCollection(
                 extensionSignatures.windows().stream()
                     .flatMap(t -> t.resolve(namespace))
                     .collect(java.util.stream.Collectors.toList()))
-            .addAllTypes(
-                extensionSignatures.types().stream()
-                    .map(t -> t.resolve(namespace))
-                    .collect(java.util.stream.Collectors.toList()))
```
vbarua, line 804: Eventually, I would like to remove the need for function resolution as well.

isthmus/src/main/java/io/substrait/isthmus/TypeConverter.java
```diff
@@ -100,43 +129,53 @@ private static Type convert(RelDataType type, List<String> names) {
...
       default -> throw new UnsupportedOperationException(
           String.format("Unable to convert the type " + type.toString()));
     };
   }
 
-  public static RelDataType convert(RelDataTypeFactory type, TypeExpression typeExpression) {
-    return convert(type, typeExpression, null);
+  public RelDataType toCalcite(
+      RelDataTypeFactory relDataTypeFactory, TypeExpression typeExpression) {
+    return toCalcite(relDataTypeFactory, typeExpression, null);
```
vbarua, line 151: This takes a `RelDataTypeFactory`, rather than bundling it in the `TypeConverter` itself, because not everywhere where you would want to build a `TypeConverter` has easy access to one of these factories, but everywhere you want to call `toCalcite` does.

isthmus/src/test/java/io/substrait/isthmus/utils/UserTypeFactory.java
```diff
@@ -0,0 +1,67 @@
...
+import io.substrait.type.TypeCreator;
+import org.apache.calcite.rel.type.RelDataType;
+import org.apache.calcite.rel.type.RelDataTypeImpl;
+import org.apache.calcite.sql.type.SqlTypeName;
+
+/**
+ * Utility factory for creating user-defined types implementing {@link RelDataTypeImpl}, correctly.
+ * Used for creating types as part of testing.
+ */
+public class UserTypeFactory {
```
vbarua, line 13: There's a lot of boilerplate to implementing a `RelDataType` in Calcite, so I made a factory for them (for testing only).

jacques-n (approved):
Confession, this is a bit of a rubberstamp. Feels like it is better to get this in and iterate if we figure out it needs improvements later. It has been hanging open too long.

vbarua:
> Feels like it is better to get this in and iterate if we figure out it needs improvements later.

On my end, I'm more than happy to make any fixes and improvements as the need arises 🙇

PR #150, opened 2023-06-05
bug: fix vist of nested struct

MasseGuillaume:
This test case will hit:

https://github.com/substrait-io/substrait-java/blob/4749dcad7eec5325414b0b08d7c2f0fdfc058378/isthmus/src/main/java/io/substrait/isthmus/expression/RexExpressionConverter.java#L128-L129

since `fieldAccess.getReferenceExpr()` is a `RexFieldAccess`

### DIFF
--- a/isthmus/src/test/java/io/substrait/isthmus/NestedStructQueryTest.java
+++ b/isthmus/src/test/java/io/substrait/isthmus/NestedStructQueryTest.java
@@ -1,0 +1,427 @@
+package io.substrait.isthmus;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+import com.google.protobuf.TextFormat;
+import io.substrait.dsl.SubstraitBuilder;
+import io.substrait.plan.ProtoPlanConverter;
+import io.substrait.proto.Expression;
+import io.substrait.proto.Plan;
+import java.io.IOException;
+import java.util.Arrays;
+import java.util.Map;
+import org.apache.calcite.rel.type.RelDataType;
+import org.apache.calcite.rel.type.RelDataTypeFactory;
+import org.apache.calcite.schema.Schema;
+import org.apache.calcite.schema.Table;
+import org.apache.calcite.schema.impl.AbstractSchema;
+import org.apache.calcite.schema.impl.AbstractTable;
+import org.apache.calcite.sql.parser.SqlParseException;
+import org.apache.calcite.sql.type.SqlTypeName;
+import org.apache.calcite.util.Pair;
+import org.junit.jupiter.api.Test;
+
+public class NestedStructQueryTest extends PlanTestBase {
+  private class TypeHelper {
+    private final RelDataTypeFactory factory;
+
+    public TypeHelper(RelDataTypeFactory factory) {
+      this.factory = factory;
+    }
+
+    RelDataType struct(String field, RelDataType value) {
+      return factory.createStructType(Arrays.asList(Pair.of(field, value)));
+    }
+
+    RelDataType struct2(String field1, RelDataType value1, String field2, RelDataType value2) {
+      return factory.createStructType(
+          Arrays.asList(Pair.of(field1, value1), Pair.of(field2, value2)));
+    }
+
+    RelDataType struct3(
+        String field1,
+        RelDataType value1,
+        String field2,
+        RelDataType value2,
+        String field3,
+        RelDataType value3) {
+      return factory.createStructType(
+          Arrays.asList(Pair.of(field1, value1), Pair.of(field2, value2), Pair.of(field3, value3)));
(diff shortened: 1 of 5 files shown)


MasseGuillaume:
Ready for review!

isthmus/src/test/java/io/substrait/isthmus/NestedStructQueryTest.java
```diff
@@ -0,0 +1,427 @@
...
+      assertEquals(expectedExpression, obtainedExpression);
+
+      ProtoPlanConverter converter = new ProtoPlanConverter();
+      io.substrait.plan.Plan plan2 = converter.from(plan);
+      assertPlanRoundrip(plan2);
+    } catch (IOException e) {
+      throw new RuntimeException(e);
+    } catch (SqlParseException e) {
+      throw new RuntimeException(e);
+    }
```
vbarua, line 94: Other tests using `assertPlanRoundtrip` don't try and catch the exceptions, rather they allow the methods/tests to throw. I think this is cleaner, because in practice the tests will fail if an exception is thrown.

isthmus/src/test/java/io/substrait/isthmus/NestedStructQueryTest.java
```diff
@@ -0,0 +1,427 @@
...
+        new AbstractSchema() {
+          @Override
+          protected Map<String, Table> getTableMap() {
+            return Map.of("my_table", table);
+          }
+        };
+
+    try {
+      final SqlToSubstrait sqlToSubstrait = new SqlToSubstrait();
+      final SubstraitBuilder builder = new SubstraitBuilder(extensions);
```
vbarua, line 80: This builder is unused

isthmus/src/test/java/io/substrait/isthmus/NestedStructQueryTest.java
```diff
@@ -0,0 +1,427 @@
...
+    RelDataType struct3(
+        String field1,
+        RelDataType value1,
+        String field2,
+        RelDataType value2,
+        String field3,
+        RelDataType value3) {
+      return factory.createStructType(
+          Arrays.asList(Pair.of(field1, value1), Pair.of(field2, value2), Pair.of(field3, value3)));
+    }
```
vbarua, line 50: This method is unused.

isthmus/src/test/java/io/substrait/isthmus/NestedStructQueryTest.java
```diff
@@ -0,0 +1,427 @@
...
+
+  @Test
+  public void testList2() {
+    final Table table =
+        new AbstractTable() {
+          @Override
+          public RelDataType getRowType(RelDataTypeFactory factory) {
+            var helper = new TypeHelper(factory);
+
+            return helper.struct2(
```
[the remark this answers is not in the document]
vbarua: This is a good trick. Could you rename the tests itself from `testList2` to something like `testNestedList`, which I think better reflects what you're actually testing.

isthmus/src/test/java/io/substrait/isthmus/NestedStructQueryTest.java
```diff
@@ -0,0 +1,427 @@
...
+              }
+              root_reference: {}
+            }
+        """;
+
+    test(table, query, expectedExpressionText);
+  }
+
+  @Test
+  public void testProtobufDoc() throws SqlParseException {
```
vbarua, line 360: What does this test verify that the other tests don't already?

isthmus/src/test/java/io/substrait/isthmus/NestedStructQueryTest.java
```diff
@@ -0,0 +1,427 @@
...
+                        field: 0 # a
+                      }
+                    }
+                  }
+                }
+              }
+            }
+            root_reference: {}
+          }
+        """;
```
vbarua, line 213: As per discussions in Slack, the references chains should start with the outermost reference (i.e. `a`) not the innermost one.

https://github.com/substrait-io/substrait/pull/520

isthmus/src/test/java/io/substrait/isthmus/NestedStructQueryTest.java
MasseGuillaume: It's more an example from the doc: https://github.com/substrait-io/substrait/blob/main/proto/substrait/algebra.proto#L960-L966.

I could remove the previous step and keep this one.

MasseGuillaume:
@jacques-n if you can review this PR, your feedback would be greatly appreciated.

jacques-n (approved):
LGTM as well. I spent more time on the test cases than actual code.

PR #154, opened 2023-06-21
[feat] more ergonomic converter extensions

vbarua:


vbarua:
These are small ergonomics changes for Isthmus. 

As a user of Isthmus, when extending `SubstraitToCalcite` we find ourselves needing access to the extensionCollection, typeFactory and typeConverter in our own class.

When extending SubstraitRelNodeConverter.java, it would be helpful to be able to use the existing remap processing code (rather than having to duplicate it).

PR #157, opened 2023-07-13
fix: Left/right/outer joins should have nullable fields.

carlyeks:
During a left, right, or outer join, some set of fields from input become nullable because they are optionally matched in the output. For a left join, the right fields become nullable. For a right join, the left fields become nullable. For an outer join, both sets become nullable.

This test fails with assertions enabled in Calcite code because the left join should have nullable right types.

### DIFF
--- a/isthmus/src/test/java/io/substrait/isthmus/SubstraitRelNodeConverterTest.java
+++ b/isthmus/src/test/java/io/substrait/isthmus/SubstraitRelNodeConverterTest.java
@@ -5,6 +5,7 @@ import static org.junit.jupiter.api.Assertions.assertInstanceOf;
 
 import io.substrait.dsl.SubstraitBuilder;
 import io.substrait.plan.Plan;
+import io.substrait.relation.Join.JoinType;
 import io.substrait.relation.Rel;
 import io.substrait.relation.Set.SetOp;
 import io.substrait.type.Type;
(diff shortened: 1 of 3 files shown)


isthmus/src/test/java/io/substrait/isthmus/SubstraitRelNodeConverterTest.java
```diff
@@ -150,6 +151,63 @@ public void emit() {
...
+
+    @Test
+    public void leftJoin() {
+      final List<Type> joinTableType = List.of(R.STRING, R.FP64, R.BINARY);
+      final Rel joinTable = b.namedScan(List.of("join"), List.of("a", "b", "c"), joinTableType);
+
+      Plan.Root root =
+          b.root(
+              b.project(r -> b.fieldReferences(r, 0, 1, 3),
+                  b.remap(0, 1, 3),
```
vbarua, line 163: One small change. Could you switch these remaps to `6, 7, 8`.

Semantically, if you're referencing fields `0, 1, 3` the project is redundant in these plans, whereas `6, 7, 8` would correspond to the references added in the project so the project would not be redundant. `6,7,8` still triggers the underlying assertion error.

carlyeks: I made that change and also ran `./gradlew spotlessApply`.

vbarua (approved):
Thanks for the fix!

PR #159, opened 2023-07-17
feat: Add SingleOrList support to the Isthmus converter

carlyeks:
The expression should use the `makeInCall`, which will be translated into Calcite as a `SEARCH` expression.

Added a new test to SubstraitRelNodeConverterTest which fails without this change.

### DIFF
--- a/isthmus/src/test/java/io/substrait/isthmus/SimplePlansTest.java
+++ b/isthmus/src/test/java/io/substrait/isthmus/SimplePlansTest.java
@@ -25,6 +25,11 @@ public class SimplePlansTest extends PlanTestBase {
   }
 
   @Test
+  public void in() throws IOException, SqlParseException {
+    assertProtoPlanRoundrip("select * from lineitem WHERE L_ORDERKEY IN (10, 20)");
+  }
+
+  @Test
   public void joinWithMultiDDLInOneString() throws IOException, SqlParseException {
     assertProtoPlanRoundrip(
         "select * from lineitem l, orders o WHERE o.o_orderkey = l.l_orderkey  and L_ORDERKEY > 10");
--- a/isthmus/src/test/java/io/substrait/isthmus/SubstraitRelNodeConverterTest.java
+++ b/isthmus/src/test/java/io/substrait/isthmus/SubstraitRelNodeConverterTest.java
@@ -131,17 +131,6 @@ public class SubstraitRelNodeConverterTest extends PlanTestBase {
       var relNode = converter.convert(root.getInput());
       assertRowMatch(relNode.getRowType(), R.I32, N.STRING);
     }
-
-    @Test
-    public void singleOrList() {
-      Plan.Root root =
-          b.root(
-              b.filter(
-                  input -> b.singleOrList(b.fieldReference(input, 0), b.i32(5), b.i32(10)),
-                  commonTable));
-      var relNode = converter.convert(root.getInput());
-      assertRowMatch(relNode.getRowType(), commonTableType);
-    }
   }
 
   @Nested
--- a/isthmus/src/test/java/io/substrait/isthmus/SubstraitRelNodeConverterTest.java
+++ b/isthmus/src/test/java/io/substrait/isthmus/SubstraitRelNodeConverterTest.java
@@ -131,6 +131,17 @@ public class SubstraitRelNodeConverterTest extends PlanTestBase {
       var relNode = converter.convert(root.getInput());
       assertRowMatch(relNode.getRowType(), R.I32, N.STRING);
     }
+
+    @Test
+    public void singleOrList() {
+      Plan.Root root =
+          b.root(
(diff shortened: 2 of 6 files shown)


isthmus/src/test/java/io/substrait/isthmus/SimplePlansTest.java
```diff
@@ -24,6 +24,11 @@ public void filter() throws IOException, SqlParseException {
     assertProtoPlanRoundrip("select * from lineitem WHERE L_ORDERKEY > 10");
   }
 
+  @Test
+  public void in() throws IOException, SqlParseException {
+    assertProtoPlanRoundrip("select * from lineitem WHERE L_ORDERKEY IN (10, 20)");
+  }
```
vbarua, line 30: This test ends up creating the following Calcite plan
```
LogicalProject(L_ORDERKEY=[$0], L_PARTKEY=[$1], L_SUPPKEY=[$2], L_LINENUMBER=[$3], L_QUANTITY=[$4], L_EXTENDEDPRICE=[$5], L_DISCOUNT=[$6], L_TAX=[$7], L_RETURNFLAG=[$8], L_LINESTATUS=[$9], L_SHIPDATE=[$10], L_COMMITDATE=[$11], L_RECEIPTDATE=[$12], L_SHIPINSTRUCT=[$13], L_SHIPMODE=[$14], L_COMMENT=[$15])
  LogicalFilter(condition=[OR(=($0, 10), =($0, 20))])
    LogicalTableScan(table=[[LINEITEM]])
```

So while this is a useful test, it doesn't test the SEARCH path.

isthmus/src/test/java/io/substrait/isthmus/SubstraitRelNodeConverterTest.java
```diff
@@ -131,6 +131,17 @@ public void emit() {
...
     }
+
+    @Test
+    public void singleOrList() {
+      Plan.Root root =
+          b.root(
+              b.filter(
+                  input -> b.singleOrList(b.fieldReference(input, 0), b.i32(5), b.i32(10)),
+                  commonTable));
+      var relNode = converter.convert(root.getInput());
```
vbarua, line 142: The plan here does get created with a SEARCH
```
LogicalFilter(condition=[SEARCH($0, Sarg[5, 10])])
  LogicalTableScan(table=[[example]])
```

So it does test the SEARCH path, but only from Substrait -> Calcite. We should add a check that for going from Calcite -> Substrait. 

Also, I'm not sure if this is the best place for this test. All the other tests in the file effectively check the record layout of various relations, whereas we're interested verifying the conversion of the SingleOrList _expression_ which won't affect the record layout and isn't tied to a specific relation.

I would advocate for having something like `ExpressionRoundtripTest` and adding a roundtrip test inside of that. There are some other expressions that get created differently depending on whether you use SQL as input or Substrait as input, so I imagine we will add more to a test file like that.

isthmus/src/test/java/io/substrait/isthmus/SimplePlansTest.java
carlyeks: We're testing that we handle the in() function instead of crashing in this test.

isthmus/src/test/java/io/substrait/isthmus/SubstraitRelNodeConverterTest.java
carlyeks: `IN` statements won't roundtrip cleanly when dealing with Calcite. A `SEARCH` expression gets translated into a series of `OR` statements using `RexUtil.expandSearch`, which is how Isthmus tries to [translate `SEARCH`](https://github.com/substrait-io/substrait-java/blob/main/isthmus/src/main/java/io/substrait/isthmus/expression/CallConverters.java#L79-L89).

I added a test that does ensure that we still generate valid Substrait from this plan, but I'm not sure `ExpressionRoundtripTest` is the right name. The test is really for similar (and semantically equivalent) queries, but not necessarily of identical expressions.

vbarua: > IN statements won't roundtrip cleanly when dealing with Calcite.
> ...
> I'm not sure ExpressionRoundtripTest is the right name.

Good point. I'm open to a different/better name if you have one.

vbarua (approved):
Thanks for adding this! Implementation looks reasonable and the checks make me confident that:
* SQL -> Substrait works
* Substrait -> Calcite -> Substrait works

isthmus/src/test/java/io/substrait/isthmus/SubstraitRelNodeConverterTest.java
carlyeks: I've renamed to ExpressionConvertabilityTest, since we're testing the ability to convert to and from Calcite here.

PR #161, opened 2023-08-03
fix: handle custom extensions through expressions

vbarua:


vbarua:
`ProtoExpressionConverter` was creating a new default `ProtoRelConverter` to translate rels contained in expressions. This caused problems when those Rels contain extensions, because the default `ProtoRelConverter` instance did not know how to handle them.

These changes pass a `ProtoRelConverter` to the `ProtoExpressionConverter`, which it uses to translate expression.

core/src/main/java/io/substrait/expression/Expression.java
```diff
@@ -668,6 +668,10 @@ abstract static class InPredicate implements Subquery {
 
     public abstract List<Expression> needles();
 
+    public Type getType() {
+      return Type.REQUIRED.BOOLEAN;
+    }
```
vbarua, line 673: Like the `SetPredicate` the `InPredicate` expression always returns a boolean.

core/src/main/java/io/substrait/expression/proto/ProtoExpressionConverter.java
```diff
@@ -173,27 +174,19 @@ public Expression from(io.substrait.proto.Expression expr) {
...
-                    .from(expr.getSubquery().getInPredicate().getHaystack());
+            var rel = protoRelConverter.from(expr.getSubquery().getInPredicate().getHaystack());
             var needles =
                 expr.getSubquery().getInPredicate().getNeedlesList().stream()
                     .map(e -> this.from(e))
                     .collect(java.util.stream.Collectors.toList());
-            yield ImmutableExpression.InPredicate.builder()
-                .haystack(rel)
-                .needles(needles)
-                .type(rel.getRecordType())
```
vbarua, line 195: The return type of the InPredicate should just be a boolean.

core/src/test/java/io/substrait/TestBase.java
```diff
@@ -0,0 +1,17 @@
...
+
+  protected static final SimpleExtension.ExtensionCollection defaultExtensionCollection;
+
+  static {
+    try {
+      defaultExtensionCollection = SimpleExtension.loadDefaults();
+    } catch (IOException e) {
+      throw new RuntimeException(e);
+    }
+  }
```
vbarua, line 16: This chunk of code showed up repeatedly in a number of tests.

isthmus/src/test/java/io/substrait/isthmus/SubqueryPlanTest.java
```diff
@@ -176,8 +176,6 @@ public void notInPredicateCorrelatedSubquery() throws IOException, SqlParseExcep
             .getScalarFunction()
             .getArguments(0)
             .getValue()
-            .getCast()
-            .getInput()
```
vbarua, line 180: As the `InPredicate` didn't return a boolean, a cast was being added here. This is gone now.

isthmus/src/main/java/io/substrait/isthmus/expression/RexExpressionConverter.java
```diff
@@ -155,11 +155,7 @@ public Expression visitSubQuery(RexSubQuery subQuery) {
       for (RexNode inOperand : subQuery.getOperands()) {
         needles.add(inOperand.accept(this));
       }
-      return Expression.InPredicate.builder()
-          .needles(needles)
-          .haystack(rel)
-          .type(typeConverter.toSubstrait(subQuery.rel.getRowType()))
```
vbarua, line 161: The type here should have been boolean.

core/src/test/java/io/substrait/type/proto/ExtensionRoundtripTest.java
```diff
@@ -232,4 +227,68 @@ void cross() {
...
+  @Nested
+  class ExtensionThroughExpression {
+    // There are some expression that can contains relations.
+    // Check that custom extensions in these relations can be handled.
+
+    Rel baseTable =
+        b.namedScan(
+            Stream.of("test_table").collect(Collectors.toList()),
+            Stream.of("test_column").collect(Collectors.toList()),
+            Stream.of(TypeCreator.REQUIRED.I64).collect(Collectors.toList()));
```
patientstreetlight, line 240: Minor, but I'm curious why this bothers creating a stream and collecting to list rather than

```
List.of("test_table"),
List.of("test_column"),
...
```

or is substrait-java targeting an older jdk version which doesn't have `List.of()`?

patientstreetlight (approved):
I just had a minor style comment/question, but I think the overall change looks good, and very welcome.

core/src/test/java/io/substrait/type/proto/ExtensionRoundtripTest.java
patientstreetlight: Java 8 also has support for `Collections.singletonList("test_table")`

vbarua: > is substrait-java targeting an older jdk version which doesn't have List.of()?

Exactly that 
> Java 8 also has support for Collections.singletonList("test_table")

TIL. I might use that in the future.

PR #164, opened 2023-08-09
refactor(pojos)!: avoid using raw proto enums in POJO layer

vbarua:
BREAKING CHANGE: various public functions that took the
AggregateFunction.AggregationInvocation proto now take the POJO
equivalent Expression.AggregationInvocation.


vbarua:
`AggregateFunction.AggregationInvocation` was the only protobuf enum that didn't have a POJO equivalent.
I added one for consistency while I was looking into additional support for window functions.

EpsilonPrime (approved):

PR #165, opened 2023-08-10
feat: support reading Substrait plans with Window Functions 

vbarua:


core/src/main/java/io/substrait/expression/proto/ExpressionProtoConverter.java
```diff
@@ -465,57 +471,68 @@ public Expression visit(io.substrait.expression.Expression.Window expr) throws R
...
-        WindowBound.UnboundedWindowBound unboundedWindowBound =
-            (WindowBound.UnboundedWindowBound) windowBound;
-        boolean isPreceding = unboundedWindowBound.direction() == WindowBound.Direction.PRECEDING;
-        var unboundedProto = Expression.WindowFunction.Bound.Unbounded.getDefaultInstance();
-        if (isPreceding) {
-          var preceding = Expression.WindowFunction.Bound.Preceding.newBuilder().build();
-          bound =
-              Expression.WindowFunction.Bound.newBuilder()
-                  .setUnbounded(unboundedProto)
-                  .setPreceding(preceding)
```
vbarua, line 505: This was problematic for two reasons:
1. `unbounded` and `proceeding` are both part of the same one of, so both should not be set.
2. The default offset for `preceding` is set to 0 here, which is dissalowed per the spec (offset must be a positive integer). I noticed this issue when I added the `LiteralToWindowBoundOffset` converter which checked for the positivity of the offset.

core/src/main/java/io/substrait/expression/proto/ExpressionProtoConverter.java
```diff
@@ -465,57 +471,68 @@ public Expression visit(io.substrait.expression.Expression.Window expr) throws R
...
+      case CURRENT_ROW -> Expression.WindowFunction.Bound.newBuilder()
+          .setCurrentRow(Expression.WindowFunction.Bound.CurrentRow.getDefaultInstance())
+          .build();
       case BOUNDED -> {
         WindowBound.BoundedWindowBound boundedWindowBound =
             (WindowBound.BoundedWindowBound) windowBound;
-        var offset = boundedWindowBound.offset();
-        boolean isPreceding = boundedWindowBound.direction() == WindowBound.Direction.PRECEDING;
-        io.substrait.expression.Expression.I32Literal offsetLiteral =
-            (io.substrait.expression.Expression.I32Literal) offset;
```
vbarua, line 482: The cast of the didn't always succeed. Reading plans in from protos, the offset is encoded as a long which is converted to an I64Literal. Additionally, depending on how Calcite encodes the offset as well, these could potentially be I8 or I16. The possibilities are handled in the LiteralToWindowBoundOffset visitor.

core/src/main/java/io/substrait/expression/proto/ProtoExpressionConverter.java
```diff
@@ -109,7 +111,54 @@ public Expression from(io.substrait.proto.Expression expr) {
...
+        var wf = ImmutableExpression.WindowFunction.builder().function(wfi).build();
+        yield Expression.Window.builder()
+            .windowFunction(wf)
+            .hasNormalAggregateFunction(false)
+            .type(protoTypeConverter.from(windowFunction.getOutputType()))
+            .partitionBy(partitionExprs)
+            .orderBy(sortFields)
+            .lowerBound(lowerBound)
+            .upperBound(upperBound)
             .build();
```
vbarua, line 162: The fact that `WindowFunctionInvocation` and `Window` both share so many of the same fields seem like a bit of a code smell. I think it's possible to reduce the duplication, but decided to leave that off for a different change.

core/src/main/java/io/substrait/expression/WindowBound.java
```diff
@@ -26,19 +29,27 @@ public BoundedKind boundedKind() {
...
   }
 
   @Value.Immutable
   abstract static class BoundedWindowBound implements WindowBound {
 
     @Override
     public BoundedKind boundedKind() {
-      return BoundedKind.UNBOUNDED;
+      return BoundedKind.BOUNDED;
     }
```
vbarua, line 44: This was a bug which caused round-trip failures.

core/src/main/java/io/substrait/expression/proto/ExpressionProtoConverter.java
```diff
@@ -420,7 +421,8 @@ public Expression visit(io.substrait.expression.Expression.Window expr) throws R
         expr.partitionBy().stream()
             .map(e -> e.accept(this))
             .collect(java.util.stream.Collectors.toList());
-    var builder = Expression.WindowFunction.newBuilder();
+    var outputType = expr.getType().accept(typeProtoConverter);
```
vbarua, line 424: The missing output type caused round-trip failures.

vbarua:
The changes in this PR allows substrait-java to read in Substrait plans containing window functions (the existing machinary only allowed for generating these plans).

Various roundtrip tests are included in WindowFunctionTests. These tests helped my identify some bugs in the existing code to convert POJO window bounds to PROTO window bounds. I've simplified the handling of these as well.

core/src/main/java/io/substrait/expression/proto/ProtoExpressionConverter.java
```diff
@@ -109,7 +112,73 @@ public Expression from(io.substrait.proto.Expression expr) {
...
+      }
+      case WINDOW_FUNCTION -> {
+        var windowFunction = expr.getWindowFunction();
+        var functionReference = windowFunction.getFunctionReference();
+        SimpleExtension.WindowFunctionVariant functionVariant;
+        try {
+          functionVariant = lookup.getWindowFunction(functionReference, extensions);
+        } catch (RuntimeException e) {
+          // TODO: Ideally we shouldn't need to catch a RuntimeException to be able to attempt our
+          // second lookup
```
vbarua, line 126: Not as part of this PR, but I think it would be good to move substrait-java and isthmus away from throwing RuntimeExceptions and towards a hierarchy of checked exceptions to improve library economics.

jacques-n:
@jinfengni , can you please review?

vbarua:
Note: this work depends on the refactor in https://github.com/substrait-io/substrait-java/pull/164

vbarua:
PR to add this behaviour to the spec is https://github.com/substrait-io/substrait/pull/540

EpsilonPrime (approved):
Looks good to me.

isthmus/src/test/java/io/substrait/isthmus/WindowFunctionTest.java
```diff
@@ -0,0 +1,143 @@
...
+          String.format("select min(O_SHIPPRIORITY) over (%s) from ORDERS", overClaus));
+    }
+
+    @Test
+    void unboundedFollowing() throws IOException, SqlParseException {
+      /*
+      LogicalProject(EXPR$0=[MAX($7) OVER (PARTITION BY $1 ORDER BY $4 ROWS BETWEEN CURRENT ROW AND UNBOUNDED FOLLOWING)])
+        LogicalTableScan(table=[[ORDERS]])
+      */
+      var overClaus =
```
EpsilonPrime, line 89: Clause?

PR #167, opened 2023-08-18
feat: implement fractional second intervals, with tests

arkanovicz:
This PR :

- adds fractional second intervals to substrait-java
- fixes isthmus for INTERVAL_DAY conversion and add other day-time intervals
- adds literal conversion tests to isthmus

Fractional seconds where added to the Substrait specifications [in this PR from May 2022](https://github.com/substrait-io/substrait/pull/199).

### DIFF
--- a/isthmus/src/main/java/io/substrait/isthmus/expression/ExpressionRexConverter.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/expression/ExpressionRexConverter.java
@@ -47,6 +47,14 @@ public class ExpressionRexConverter extends AbstractExpressionVisitor<RexNode, R
           -1,
           SqlParserPos.QUOTED_ZERO);
 
+  private static final SqlIntervalQualifier DAY_SECOND_INTERVAL =
+      new SqlIntervalQualifier(
+          org.apache.calcite.avatica.util.TimeUnit.DAY,
+          -1,
+          org.apache.calcite.avatica.util.TimeUnit.SECOND,
+          3, // Calcite only supports millisecond at the moment
+          SqlParserPos.QUOTED_ZERO);
+
   public ExpressionRexConverter(
       RelDataTypeFactory typeFactory,
       ScalarFunctionConverter scalarFunctionConverter,
@@ -176,6 +184,20 @@ public class ExpressionRexConverter extends AbstractExpressionVisitor<RexNode, R
   public RexNode visit(Expression.IntervalYearLiteral expr) throws RuntimeException {
     return rexBuilder.makeIntervalLiteral(
         new BigDecimal(expr.years() * 12 + expr.months()), YEAR_MONTH_INTERVAL);
+  }
+
+  private static final long MICROS_IN_DAY = TimeUnit.DAYS.toMicros(1);
+
+  @Override
+  public RexNode visit(Expression.IntervalDayLiteral expr) throws RuntimeException {
+    return rexBuilder.makeIntervalLiteral(
+        // Current Calcite behavior is to store milliseconds since Epoch
+        // microseconds version: new BigDecimal(expr.days() * MICROS_IN_DAY + expr.seconds() *
+        // 100000L + expr.microseconds()), DAY_SECOND_INTERVAL);
+        new BigDecimal(
+            (expr.days() * MICROS_IN_DAY + expr.seconds() * 1_000_000L + expr.microseconds())
+                / 1000L),
+        DAY_SECOND_INTERVAL);
   }
 
   @Override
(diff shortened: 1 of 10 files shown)


EpsilonPrime (approved):

isthmus/src/test/java/io/substrait/isthmus/CalciteLiteralTest.java
```diff
@@ -194,6 +194,29 @@ void tIntervalYearMonthWithPrecision() {
...
+                + 500); // '3-5:7:9.500' day to second (6)
+    RexLiteral intervalDaySecond =
+        rex.makeIntervalLiteral(
+            bd,
+            new SqlIntervalQualifier(
+                org.apache.calcite.avatica.util.TimeUnit.DAY,
+                -1,
+                org.apache.calcite.avatica.util.TimeUnit.SECOND,
+                3,
+                SqlParserPos.QUOTED_ZERO));
```
vbarua, line 215: I was taking a look and noticed that we've defined the same intervals in 3 places effectively:
1. https://github.com/substrait-io/substrait-java/blob/1a4b7b1d729f87b9ccd310aef976568f3d867590/isthmus/src/main/java/io/substrait/isthmus/TypeConverter.java#L45-L48
2. https://github.com/substrait-io/substrait-java/blob/1a4b7b1d729f87b9ccd310aef976568f3d867590/isthmus/src/main/java/io/substrait/isthmus/expression/ExpressionRexConverter.java#L42-L48
3. Now in these tests.

I think it would be good to define them in one place, make them public and re-use them. I think `SubstraitTypeSystem` would be a good home for them.

isthmus/src/test/java/io/substrait/isthmus/CalciteLiteralTest.java
```diff
@@ -194,6 +194,29 @@ void tIntervalYearMonthWithPrecision() {
         convertedRex.getValueAs(BigDecimal.class).longValue());
   }
 
+  @Test
+  void tIntervalDaySecond() {
+    // Calcite stores milliseconds since Epoch, so test only millisecond precision
+    BigDecimal bd =
```
vbarua, line 200: minor: could you name this something like `milliseconds`, mostly to make the unit involved crystal-clear (it took me a bit longer that I would like to admit to figure that out from the code).

isthmus/src/main/java/io/substrait/isthmus/expression/ExpressionRexConverter.java
```diff
@@ -178,6 +186,20 @@ public RexNode visit(Expression.IntervalYearLiteral expr) throws RuntimeExceptio
...
   }
 
+  private static final long MICROS_IN_DAY = TimeUnit.DAYS.toMicros(1);
+
+  @Override
+  public RexNode visit(Expression.IntervalDayLiteral expr) throws RuntimeException {
+    return rexBuilder.makeIntervalLiteral(
+        // Current Calcite behavior is to store milliseconds since Epoch
+        // microseconds version: new BigDecimal(expr.days() * MICROS_IN_DAY + expr.seconds() *
+        // 100000L + expr.microseconds()), DAY_SECOND_INTERVAL);
```
vbarua, line 196: Is that Calcite behaviour part of the type system? What would be involved in allow microsecond precision as per the spec (if you happen to know).

isthmus/src/test/java/io/substrait/isthmus/CalciteLiteralTest.java
arkanovicz: I factorized a bit, but since the tests do tweak precision they still create some specific SqlIntervalQualifier objects.

isthmus/src/main/java/io/substrait/isthmus/expression/ExpressionRexConverter.java
arkanovicz: Both Avatica and Calcite use Epoch milliseconds from timestamps and milliseconds for day-time intervals. This is a huge refactoring which will probably imply a major version change if it happens.

vbarua: We can do that as a future change then.

isthmus/src/test/java/io/substrait/isthmus/CalciteLiteralTest.java
vbarua: Factorizing out only the really common ones makes sense.

vbarua (approved):
Looks good. Thanks for the contribution!

PR #170, opened 2023-08-28
refactor: unify aggregate and window functions in window handling 

vbarua:
feat: allow window function operands beyond just column references
fix: invocation was not set when building WindowFunction proto

BREAKING CHANGES:
* Window and WindowFunction have been merged into WindowFunctionInvocation
* WindowFunctionInvocation is now an inner class of expression
* WindowFunctionInvocation now implements Expression.
* Expression visitor now visit WindowFunctionInvocation
* WindowFunctionConvert constructor requires fewer arguments
* windowCreator has additional fields
* POJO WindowBound offset must now be a long, better reflecting spec


core/src/main/java/io/substrait/expression/Expression.java
```diff
@@ -684,41 +717,6 @@ public <R, E extends Throwable> R accept(ExpressionVisitor<R, E> visitor) throws
...
-    public <R, E extends Throwable> R accept(ExpressionVisitor<R, E> visitor) throws E {
-      return visitor.visit(this);
-    }
-  }
-
-  @Value.Immutable
-  public abstract static class WindowFunction {
-    public abstract WindowFunctionInvocation getFunction();
-
-    public abstract Optional<Expression> getPreMeasureFilter();
```
vbarua, line 718: The `preMeasureFilter` field was not used anywhere, and the [WindowFunction](https://github.com/substrait-io/substrait/blob/51765cccd92fa0014e9cf077cc8cc952731f33d0/proto/substrait/algebra.proto#L847) message cannot actually encode a pre-measure filter (for now).

I opted to remove it entirely. Support can be added in once the spec supports it.

core/src/main/java/io/substrait/expression/WindowBound.java
```diff
@@ -45,7 +45,7 @@ public BoundedKind boundedKind() {
 
     public abstract Direction direction();
 
-    public abstract Expression offset();
+    public abstract long offset();
```
vbarua, line 48: In the [spec](https://github.com/substrait-io/substrait/blob/51765cccd92fa0014e9cf077cc8cc952731f33d0/proto/substrait/algebra.proto#L933C5-L948), offset must be `int64`.

The POJO layer should reflect this restriction.

core/src/main/java/io/substrait/expression/WindowFunctionInvocation.java
```diff
@@ -1,33 +0,0 @@
-package io.substrait.expression;
-
-import io.substrait.extension.SimpleExtension;
-import io.substrait.type.Type;
-import java.util.List;
-import java.util.Map;
-import org.immutables.value.Value;
-
-@Value.Immutable
-public abstract class WindowFunctionInvocation {
```
vbarua, line 10: WindowFunctionInvocation is a / implements `Expression`, much like ScalarFunctionInvocation.

Moved this to an inner class of Expression for consistency.

core/src/main/java/io/substrait/extension/SimpleExtension.java
```diff
@@ -796,20 +804,43 @@ public static ExtensionCollection load(String namespace, InputStream stream) {
...
+                        // Set WindowFunctionVariant fields
+                        .decomposability(afi.decomposability())
+                        .intermediate(afi.intermediate())
+                        // Aggregate Functions used in Windows have WindowType Streaming
+                        .windowType(SimpleExtension.WindowType.STREAMING)
+                        .build());
+
+    List<WindowFunctionVariant> allWindowFunctionVariants =
+        Stream.concat(windowFunctionVariants, windowAggFunctionVariants)
+            .collect(Collectors.toList());
```
vbarua, line 837: The spec is being loosed to allow usage of aggregate functions as window function.
https://github.com/substrait-io/substrait/pull/540

`substrait-java` already supported this behaviour, but the implementation is simpler/cleaner if aggregate functions are added as valid window functions here.

core/src/main/java/io/substrait/expression/Expression.java
```diff
@@ -684,41 +717,6 @@ public <R, E extends Throwable> R accept(ExpressionVisitor<R, E> visitor) throws
     }
   }
 
-  @Value.Immutable
-  abstract static class Window implements Expression {
-    @Nullable
-    public abstract Aggregate.Measure aggregateFunction();
-
-    @Nullable
-    public abstract WindowFunction windowFunction();
```
vbarua, line 693: Instead of having separate handling for window calls with window functions, and window calls with aggregate functions, this PR updates the extension handling to treat all aggregate functions as valid window functions and a such removes the need for this distinction.

isthmus/src/main/java/io/substrait/isthmus/expression/RexExpressionConverter.java
```diff
@@ -7,13 +7,24 @@
 import io.substrait.isthmus.TypeConverter;
 import io.substrait.relation.Rel;
 import io.substrait.type.StringTypeVisitor;
-import io.substrait.type.Type;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
 import java.util.stream.Collectors;
-import org.apache.calcite.rel.RelNode;
-import org.apache.calcite.rex.*;
```
vbarua, line 16: Avoiding * imports

isthmus/src/main/java/io/substrait/isthmus/expression/WindowFunctionConverter.java
```diff
@@ -4,172 +4,98 @@
...
+        partitionExprs,
+        lowerBound,
+        upperBound,
         arguments);
   }
 
-  public Expression.Window convert(
-      RelNode input,
-      Type.Struct inputType,
+  public Optional<Expression.WindowFunctionInvocation> convert(
```
vbarua, line 84:
```java
public Optional<Expression.WindowFunctionInvocation> convert
```

This output types brings `convert` inline with both ScalarFunctionConverter and AggregateFunctionConverter.

isthmus/src/main/java/io/substrait/isthmus/expression/WindowFunctionConverter.java
```diff
@@ -4,172 +4,98 @@
...
-            sqlAggFunction,
-            over.isDistinct(),
-            approximate,
-            over.ignoreNulls(),
-            argList,
-            filterArg,
-            null,
-            RelCollations.EMPTY,
-            over.getType(),
-            sqlAggFunction.getName());
```
vbarua, line 110: The usage of this API meant that calls like `ntile(4)` were not supported, a `4` would be encoded as a `RexLiteral` which is not a RexSlot. Switching away from this handling enables those queries, and in fact I was able to enable the test for the `ntile` query in `WindowFunctionTest`.

EpsilonPrime (approved):
Good set of changes.  I especially like how bounds and operands are much cleaner (as opposed to invoking a potentially unrelated builder in place).

PR #171, opened 2023-08-30
ci: switch graalvm provider

vbarua:


vbarua:
The current graalvm setup action (https://github.com/DeLaGuardo/setup-graalvm) has been archived.

They recommend https://github.com/graalvm/setup-graalvm instead.

EpsilonPrime (approved):
Thanks for figuring this out!

PR #172, opened 2023-08-30
feat: window function calcite support

vbarua:
feat: support for window bounds type

BREAKING CHANGE:
* windowFunction expression creator now requires window bound type param
* the WindowBound POJO representation has been reworked to use visitation and more closely match the spec
* ExpressionRexConvert now requires a WindowFunctionConverter

### DIFF
--- a/core/src/main/java/io/substrait/expression/WindowBound.java
+++ b/core/src/main/java/io/substrait/expression/WindowBound.java
@@ -5,58 +5,62 @@ import org.immutables.value.Value;
 @Value.Enclosing
 public interface WindowBound {
 
-  public BoundedKind boundedKind();
+  interface WindowBoundVisitor<R, E extends Throwable> {
+    R visit(Preceding preceding);
 
-  enum BoundedKind {
-    UNBOUNDED,
-    BOUNDED,
-    CURRENT_ROW
+    R visit(Following following);
+
+    R visit(CurrentRow currentRow);
+
+    R visit(Unbounded unbounded);
   }
 
-  enum Direction {
-    PRECEDING,
-    FOLLOWING
-  }
+  <R, E extends Throwable> R accept(WindowBoundVisitor<R, E> visitor);
 
-  public static CurrentRowWindowBound CURRENT_ROW =
-      ImmutableWindowBound.CurrentRowWindowBound.builder().build();
+  CurrentRow CURRENT_ROW = ImmutableWindowBound.CurrentRow.builder().build();
+  Unbounded UNBOUNDED = ImmutableWindowBound.Unbounded.builder().build();
 
   @Value.Immutable
-  abstract static class UnboundedWindowBound implements WindowBound {
-    @Override
-    public BoundedKind boundedKind() {
-      return BoundedKind.UNBOUNDED;
-    }
-
-    public abstract Direction direction();
-
-    public static ImmutableWindowBound.UnboundedWindowBound.Builder builder() {
-      return ImmutableWindowBound.UnboundedWindowBound.builder();
-    }
-  }
-
-  @Value.Immutable
-  abstract static class BoundedWindowBound implements WindowBound {
-
-    @Override
-    public BoundedKind boundedKind() {
-      return BoundedKind.BOUNDED;
-    }
-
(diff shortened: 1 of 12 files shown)


core/src/main/java/io/substrait/expression/WindowBound.java
```diff
@@ -5,58 +5,62 @@
...
-  }
+    R visit(Following following);
+
+    R visit(CurrentRow currentRow);
 
-  enum Direction {
-    PRECEDING,
-    FOLLOWING
+    R visit(Unbounded unbounded);
   }
```
vbarua, line 16: Direction doesn't have an equivalent in the Substrait spec itself. Calcite has a concept of `UNBOUNDED_PRECEDING` and `UNBOUNDED_FOLLOWING`, but those can be reduced to singular `Unbounded` WindowBound when converting from Calcite to Pojo.

core/src/main/java/io/substrait/expression/WindowBound.java
```diff
@@ -5,58 +5,62 @@
 @Value.Enclosing
 public interface WindowBound {
```
vbarua, line 6: Rather than encoding the WindowBound type using a combination of BoundedKind and Direction, I've switched this to 4 classes:
1. Preceding
2. Following
3. Current
4. Unbounded

and introduced a WindowBoundVisitor to handle dispatch when converting from these to various other representations.

This representation of Window Bounds:
1. More closely matches the spec.
2. Reduces the need for casting of WindowBound interfaces to concrete types.
3. Reduces cognitive complexity (in my opinion) as users no longer need to reason about 6 different enum combinations (3 bounded kinds * 2 directions), and instead have 4 concrete classes.

core/src/main/java/io/substrait/expression/proto/ExpressionProtoConverter.java
```diff
@@ -440,8 +440,8 @@ public Expression visit(io.substrait.expression.Expression.WindowFunctionInvocat
                         .build())
             .collect(java.util.stream.Collectors.toList());
 
-    Expression.WindowFunction.Bound upperBound = toBound(expr.upperBound());
-    Expression.WindowFunction.Bound lowerBound = toBound(expr.lowerBound());
+    Expression.WindowFunction.Bound upperBound = expr.upperBound().accept(TO_BOUND_VISITOR);
+    Expression.WindowFunction.Bound lowerBound = expr.lowerBound().accept(TO_BOUND_VISITOR);
```
EpsilonPrime, line 444: So you're putting an identical bounds visitor into both lowerBound and upperBound but call getLowerBound or getUpperBound as appropriate?  Why not have only one variable for storing bounds?

EpsilonPrime (approved):

EpsilonPrime (approved):

core/src/main/java/io/substrait/expression/proto/ExpressionProtoConverter.java
vbarua: I've cleaned this up a bit to hide the visit invocation behind a static method:
```java
    Expression.WindowFunction.Bound lowerBound = BoundConverter.convert(expr.lowerBound());
    Expression.WindowFunction.Bound upperBound = BoundConverter.convert(expr.upperBound());
```

To your other point, are you suggesting have something like a single `class WindowBounds` which contains both the lower and upper bound? The main reason for them to separate is to better match the [protos](https://github.com/substrait-io/substrait/blob/51765cccd92fa0014e9cf077cc8cc952731f33d0/proto/substrait/algebra.proto#L261-L263).

Another thing as well, is that the POJO <-> Proto conversion used to care about direction which is why there was a lower bound and upper bound variant. However, the substrait spec don't care about this, it was a Calcite detail. By pulling the direction out of the POJOs, I've limited the code that cares about direction to the POJO -> Calcite conversion in ExpressionRexConverter.

EpsilonPrime: Thanks for the changes, it looks less weird now.  Also agreed that the protos are oddly defined (by having multiple semantics in one field).

PR #177, opened 2023-09-12
feat(calcite): support reading in list and map literals

vbarua:
feat: inject ExpressionRexConverter into SubstraitRelNodeConverter


vbarua:
Let's us read Substrait List and Map literals into Calcite.

Additionally, allows for the injection of the ExpressionRexConverter to allow users extra control over conversion if they need it.

EpsilonPrime (approved):

Issue #137, opened 2023-03-23
Visit Core Calcite relations types in SubstraitRelVisitor 


vbarua:
## Context
As part of performing optimizations in Calcite, we convert from Logical relation to Physical equivalents

```java
class PhysicalFilter extends Filter
```

This is relatively common pattern when using Calcite
* [CassandraFilter](https://github.com/apache/calcite/blob/c28d1dcbc34e748b7bea9712ef6bcf43793a91e8/cassandra/src/main/java/org/apache/calcite/adapter/cassandra/CassandraFilter.java)
* [EnumerableFilter](https://github.com/apache/calcite/blob/c28d1dcbc34e748b7bea9712ef6bcf43793a91e8/core/src/main/java/org/apache/calcite/adapter/enumerable/EnumerableFilter.java)

## Problem
Currently the [SubstraitRelVisitor](https://github.com/substrait-io/substrait-java/blob/dd069f69e9ddd7e7e668188ab535174181e61d02/isthmus/src/main/java/io/substrait/isthmus/SubstraitRelVisitor.java) visits Logical relations directly, which means it can't be re-used easily to convert these physical relation types.

For example because
https://github.com/substrait-io/substrait-java/blob/dd069f69e9ddd7e7e668188ab535174181e61d02/isthmus/src/main/java/io/substrait/isthmus/SubstraitRelVisitor.java#L138-L142

we have to write our own visit like
```java
 @Override 
 public Rel visit(PhysicalFilter filter) { 
   var condition = toExpression(filter.getCondition()); 
   return Filter.builder().condition(condition).input(apply(filter.getInput())).build(); 
 } 
```
and duplicate the code.

It would be nicer to have to have SubstraitRelVisitor visit the common interface/abstract class types:
```java
@Override 
public Rel visit(Filter filter) { 
    var condition = toExpression(filter.getCondition()); 
    return Filter.builder().condition(condition).input(apply(filter.getInput())).build(); 
} 
```
which would allow us to re-use this logic, and then also let us do things like this when extending the SubstraitRelVisitor

```java
public Rel visit(PhysicalFilter filter) { 
    return Filter.builder().from(super.visit(this)).advancedExtension(...).build()
} 
```

I'm happy to work on this if it seems reasonable.

jacques-n:
Yes, I agree that this seems better.

PR #178, opened 2023-09-12
feat!: visit over core substrait types

vbarua:


vbarua:
Addresses: https://github.com/substrait-io/substrait-java/issues/137

This makes it easier to re-use the `SubstraitRelVisitor` when converting from Calcite back to Substrait.

I'm not sure if this is a breaking change, but it's a big enough change in functionality that I'm marking it as such.

EpsilonPrime (approved):

PR #179, opened 2023-09-14
feat: use Immutables for Argument subclasses

vbarua:
allows for equality comparisons of Argument subclasses

BREAKING CHANGE: EnumArgument, TypeArgument and ValueArgument are now abstract


vbarua:
I was trying to compare plans and noticed that the Argument subclasses didn't implement equals. Updated them to use Immutable like everything else in the file uses.

Determined required fields based on:
https://github.com/substrait-io/substrait/blob/f2bfe15585943a137fafa560401e0cf0266c0650/text/simple_extensions_schema.yaml#L91-L123

EpsilonPrime (approved):

PR #180, opened 2023-09-15
feat(calcite): dedicated Substrait MAX, MIN, SUM, SUM0 and AVG functions

vbarua:
fix(calcite): address type inference mismatch in arithmetic aggregate functions

feat: util methods on TypeCreator for setting type nullability
feat: builder utils for arithmetic aggregate functions
feat(calcite): support sum0 sql function
test: static fields in Type interface caused NPEs in tests 

BREAKING CHANGE: Isthmus no longer uses Calcite built-in MAX, MIN, SUM, SUM0 and AVG functions
BREAKING CHANGE: removed REQUIRED and NULLABLE fields from Type interface

### DIFF
--- a/core/src/main/java/io/substrait/type/Type.java
+++ b/core/src/main/java/io/substrait/type/Type.java
@@ -11,9 +11,6 @@ import org.immutables.value.Value;
 public interface Type extends TypeExpression, ParameterizedType, NullableType, FunctionArg {
   static final org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(Type.class);
 
-  public static final TypeCreator REQUIRED = TypeCreator.REQUIRED;
-  public static final TypeCreator NULLABLE = TypeCreator.NULLABLE;
-
   public static TypeCreator withNullability(boolean nullable) {
     return nullable ? TypeCreator.NULLABLE : TypeCreator.REQUIRED;
   }
--- a/isthmus/src/main/java/io/substrait/isthmus/calcite/SubstraitOperatorTable.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/calcite/SubstraitOperatorTable.java
@@ -1,0 +1,73 @@
+package io.substrait.isthmus.calcite;
+
+import io.substrait.isthmus.AggregateFunctions;
+import java.util.List;
+import java.util.Set;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+import org.apache.calcite.sql.SqlFunctionCategory;
+import org.apache.calcite.sql.SqlIdentifier;
+import org.apache.calcite.sql.SqlKind;
+import org.apache.calcite.sql.SqlOperator;
+import org.apache.calcite.sql.SqlOperatorTable;
+import org.apache.calcite.sql.SqlSyntax;
+import org.apache.calcite.sql.fun.SqlStdOperatorTable;
+import org.apache.calcite.sql.util.SqlOperatorTables;
+import org.apache.calcite.sql.validate.SqlNameMatcher;
+import org.checkerframework.checker.nullness.qual.Nullable;
+
+/**
+ * Overrides SQL operator lookups to return Substrait specific functions variants (e.g. {@link
+ * AggregateFunctions#MAX}} when they are available.
+ */
+public class SubstraitOperatorTable implements SqlOperatorTable {
+
+  public static SubstraitOperatorTable INSTANCE = new SubstraitOperatorTable();
+
+  private static final SqlOperatorTable SUBSTRAIT_OPERATOR_TABLE =
+      SqlOperatorTables.of(
+          List.of(
+              AggregateFunctions.MAX,
+              AggregateFunctions.MIN,
+              AggregateFunctions.AVG,
+              AggregateFunctions.SUM,
+              AggregateFunctions.SUM0));
+
+  // kinds for which Substrait specific operators are provided
+  private static final Set<SqlKind> OVERRIDE_KINDS =
+      Set.of(SqlKind.MAX, SqlKind.MIN, SqlKind.AVG, SqlKind.SUM, SqlKind.SUM0);
+
+  private static final SqlOperatorTable STANDARD_OPERATOR_TABLE = SqlStdOperatorTable.instance();
+
+  private static final List<SqlOperator> OPERATOR_LIST =
+      Stream.concat(
+              SUBSTRAIT_OPERATOR_TABLE.getOperatorList().stream(),
+              // filter out the kinds that have been overriden from the standard operator table
+              STANDARD_OPERATOR_TABLE.getOperatorList().stream()
+                  .filter(op -> !OVERRIDE_KINDS.contains(op.kind)))
+          .collect(Collectors.toUnmodifiableList());
+
+  @Override
+  public void lookupOperatorOverloads(
+      SqlIdentifier opName,
+      @Nullable SqlFunctionCategory category,
+      SqlSyntax syntax,
+      List<SqlOperator> operatorList,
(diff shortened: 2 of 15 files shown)


vbarua:
The Substrait functions MIN, MAX, AVG and SUM _ALWAYS_ return a nullable version of the input type (per the spec). In most cases, the default Calcite definitions of these functions behave the same way, _EXCEPT_ in some cases where they infer a `NOT NULL` return type. 

This results in errors like:
```
type mismatch:
aggCall type:
TINYINT
inferred type:
```

This PR addresses these issues by creating variants of Calcite function definitions with less clever return type inference.

This also fixes a similar issue with SUM0 which always returns a NOT NULL I64  per the spec, but the Calcite definition returns a NOT NULL version of the _input_ numeric type.

This results in errors like:
```
type mismatch:
aggCall type:
BIGINT NOT NULL
inferred type:
TINYINT NOT NULL
```

vbarua:
The changes for the above comment, which fixed Substrait to Calcite conversions, broke Calcite to Substrait conversions.

Effectively, in a query like
```sql
SELECT MAX(column) FROM foo
```
Calcite would parse the MAX function into its standard `SqlMinMaxAggFunction` representation instead of the newer `SubstraitSqlMinMaxAggFunction`. This caused the Calcite to Substrait conversion to fail because `AGGREGATE_SIGS` contained a mapping from `AggregateFunctions.MAX` to `"max"`, and not one from `SqlStdOperatorTable.MAX`.

The solution I chose for this was to introduce a `SubstraitOperatorTable` that preferentially selects Substrait function variants when they are available.

EpsilonPrime (approved):

core/src/main/java/io/substrait/type/Type.java
```diff
@@ -11,9 +11,6 @@
 public interface Type extends TypeExpression, ParameterizedType, NullableType, FunctionArg {
   static final org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(Type.class);
 
-  public static final TypeCreator REQUIRED = TypeCreator.REQUIRED;
-  public static final TypeCreator NULLABLE = TypeCreator.NULLABLE;
```
vbarua, line 15: In tests, these fields would consistently end up as null. See failures in:
https://github.com/substrait-io/substrait-java/actions/runs/6203665019/job/16844629387

This started happening once I added additional code to the TypeCreator, which makes me suspect some weird initialisation race conditions.

IntellIJ was also hinted me in that direction:
```
Modifier 'static' is redundant for interface fields
```

Moving to using TypeCreator.REQUIRED and TypeCreator.NULLABLE seems to have resolved the issue.

EpsilonPrime:
lgtm

core/src/main/java/io/substrait/type/Type.java
vbarua: Talking through this with a co-worker, `TypeCreator` references `Type` and `Type` references `TypeCreator`, which may also affect the initialization.

isthmus/src/test/java/io/substrait/isthmus/ArithmeticFunctionTest.java
```diff
@@ -16,62 +16,97 @@ void arithmetic(String c) throws Exception {
     String query =
         String.format(
             "SELECT %s + %s, %s - %s, %s * %s, %s / %s FROM numbers", c, c, c, c, c, c, c, c);
-    assertSqlSubstraitRelRoundTrip(query, CREATES);
+    assertFullRoundTrip(query, CREATES);
```
vbarua, line 19: `assertFullRoundTrip` is a strong assert. Will eventually remove all instances of `assertSqlSubstraitRelRoundTrip`

EpsilonPrime (approved):

isthmus/src/main/java/io/substrait/isthmus/calcite/SubstraitOperatorTable.java
```diff
@@ -0,0 +1,73 @@
...
+import org.apache.calcite.sql.validate.SqlNameMatcher;
+import org.checkerframework.checker.nullness.qual.Nullable;
+
+/**
+ * Overrides SQL operator lookups to return Substrait specific functions variants (e.g. {@link
+ * AggregateFunctions#MAX}} when they are available.
+ */
+public class SubstraitOperatorTable implements SqlOperatorTable {
+
+  public static SubstraitOperatorTable INSTANCE = new SubstraitOperatorTable();
```
patientstreetlight, line 25: minor: consider giving this class a private constructor if it's not meant to be instantiated by other classes.

isthmus/src/main/java/io/substrait/isthmus/calcite/SubstraitOperatorTable.java
```diff
@@ -0,0 +1,73 @@
...
+      SqlOperatorTables.of(
+          List.of(
+              AggregateFunctions.MAX,
+              AggregateFunctions.MIN,
+              AggregateFunctions.AVG,
+              AggregateFunctions.SUM,
+              AggregateFunctions.SUM0));
+
+  // kinds for which Substrait specific operators are provided
+  private static final Set<SqlKind> OVERRIDE_KINDS =
```
patientstreetlight, line 37: Couple minor things:
* consider using an [EnumSet](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/util/EnumSet.html) here
* I wonder if it would be easier to maintain this list if it could be automatically computed somehow from `SUBSTRAIT_OPERATOR_TABLE` - would that be possible?

isthmus/src/main/java/io/substrait/isthmus/calcite/SubstraitOperatorTable.java
```diff
@@ -0,0 +1,73 @@
...
+
+  private static final SqlOperatorTable STANDARD_OPERATOR_TABLE = SqlStdOperatorTable.instance();
+
+  private static final List<SqlOperator> OPERATOR_LIST =
+      Stream.concat(
+              SUBSTRAIT_OPERATOR_TABLE.getOperatorList().stream(),
+              // filter out the kinds that have been overriden from the standard operator table
+              STANDARD_OPERATOR_TABLE.getOperatorList().stream()
+                  .filter(op -> !OVERRIDE_KINDS.contains(op.kind)))
+          .collect(Collectors.toUnmodifiableList());
```
patientstreetlight, line 48: TIL about `Collectors.toUnmodifiableList()`.  Seems like a good choice here 👍

isthmus/src/main/java/io/substrait/isthmus/calcite/SubstraitOperatorTable.java
```diff
@@ -0,0 +1,73 @@
...
+  @Override
+  public void lookupOperatorOverloads(
+      SqlIdentifier opName,
+      @Nullable SqlFunctionCategory category,
+      SqlSyntax syntax,
+      List<SqlOperator> operatorList,
+      SqlNameMatcher nameMatcher) {
+    SUBSTRAIT_OPERATOR_TABLE.lookupOperatorOverloads(
+        opName, category, syntax, operatorList, nameMatcher);
+    if (!operatorList.isEmpty()) {
```
patientstreetlight, line 59: I know this is Calcite, and not anything you can control, but having `lookupOperatorOverloads()` modify its input list is not my favorite 😞

patientstreetlight (approved):
Overall looks good to me!

I just had some really minor comments you can take if you'd like.

isthmus/src/main/java/io/substrait/isthmus/calcite/SubstraitOperatorTable.java
vbarua: TIL about EnuSet. Done

Issue #162, opened 2023-08-04
Does substrait-java core rely on full antlr4 at runtime?


adamkennedy:
The current substrait core module has a dependency on antlr4 at run-time, which pulls in a lot of extra dependencies in the form of things like the icu library.

Could this be changed to depend on only antlr4-runtime instead?

jacques-n:
We have a relatively simple grammar. I don't think I ever evaluated alternatives. I would expect antlr runtime to work. Want to post a patch trying it out?

vbarua:
Handled in https://github.com/substrait-io/substrait-java/pull/182

PR #182, opened 2023-09-20
build: use antlr4-runtime instead of antlr

guiyanakuang:
To avoid introducing unnecessary indirect dependencies. Resolve issue #162


vbarua:
I'm seeing build failures on this, but I don't think they're related to your changes. I'm going to investigate them today.

vbarua (approved):
Thanks for the change!

The only failure in CI now is a lint check due to commit message formatting, but that will go away when I squash the commit as part of the merge.

guiyanakuang:
@vbarua . Thanks for the quick review and merge of PR.

PR #183, opened 2023-09-20
feat: extend literal conversion support

vbarua:


vbarua:
Adds Calcite conversion support for VarCharLiteral and FixedBinaryLiteral.

Additionally, forces
* StrLiteral to be converted to VARCHAR
* BinaryLiteral to be converted to VARBINARY

Before this, they were converted to
* CHAR(<length>)
* BINARY(<length>)

respectively. If those types are desired, Substrait has the FixedChar and FixedBinary types. 

Calcite being a bit to smart when converting these two types results in failures during query planning.

EpsilonPrime (approved):

PR #184, opened 2023-09-21
fix: `TypeParam` supports optional nullable suffix

### LINKED ISSUE #169: unable to specify return type of any1?
Given a scalar function definition with nullable `any` return, the following definition works:
```
scalar_functions:
- name: "array_index"
    description: "returns the element in the array at index, or NULL if index is out of bounds"
    impls:
      - args:
          - name: array
            value: list<any>
          - name: index
            value: i64
        return: any?
```

whilst the equivalent definition

```
%YAML 1.2
---
scalar_functions:
- name: "array_index"
    description: "returns the element in the array at index, or NULL if index is out of bounds"
    impls:
      - args:
          - name: array
            value: list<any1>
          - name: index
            value: i64
        return: any1?
```

fails with
```
Caused by: com.fasterxml.jackson.databind.JsonMappingException: Unable to parse string any1?
```

guiyanakuang:
This PR aims to make TypeParam support optional nullable suffix. To fix #169.

Added unit test TypeExtensionTest#arrayIndexScalarFunction

### DIFF
--- a/core/src/test/java/io/substrait/extension/TypeExtensionTest.java
+++ b/core/src/test/java/io/substrait/extension/TypeExtensionTest.java
@@ -1,5 +1,6 @@
 package io.substrait.extension;
 
+import static io.substrait.type.TypeCreator.REQUIRED;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 
 import io.substrait.dsl.SubstraitBuilder;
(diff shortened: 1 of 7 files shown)


core/src/test/java/io/substrait/type/parser/TestTypeParser.java
```diff
@@ -114,7 +114,7 @@ private void compoundTests(ParseToPojo.Visitor v) {
   private <T> void parameterizedTests(ParseToPojo.Visitor v) {
     test(v, pn.listE(pr.parameter("K")), "List?<K>");
     test(v, pr.structE(r.I8, r.I16, n.I8, pr.parameter("K")), "STRUCT<i8, i16, i8?, K>");
-    test(v, pn.parameter("any"), "any");
+    test(v, pn.parameter("any"), "any?");
```
vbarua, line 117: minor: I think it would be good to test both nullable and not nullable any
```java
    test(v, pr.parameter("any"), "any");
    test(v, pn.parameter("any"), "any?");
```

core/src/test/java/io/substrait/extension/TypeExtensionTest.java
```diff
@@ -76,4 +77,30 @@ void roundtripCustomType() {
     var planReturned = protoPlanConverter.from(protoPlan);
     assertEquals(plan, planReturned);
   }
+
+  @Test
+  void arrayIndexScalarFunction() {
```
vbarua, line 82: minor: what do you think naming this something `roundtripNumberedAnyTypes`, which I think better describes what we're checking (i.e. that substrait-java can read in types like `any1` and `any1?`).

vbarua (commented):
Left two minor comments @guiyanakuang 

If you can take a look, I'll be happy to merge this tomorrow.

core/src/test/java/io/substrait/extension/TypeExtensionTest.java
guiyanakuang: Resolved. Thanks for the suggestion.

vbarua (approved):
Thanks for the changes!

Issue #185, opened 2023-09-21
stale type in FieldReference after RelCopyOnWrite modifications


vbarua:
The [FieldReference](https://github.com/substrait-io/substrait-java/blob/6e82f397a168f3811593fe8248753d70d4f0c04f/core/src/main/java/io/substrait/expression/FieldReference.java) POJO includes a field to store its type, which is set based on the input rel when it is created.

When the plan tree is modified by the RelCopyOnWrite visitor, existing field refs are copied over as they are. This includes the stored type. If the visitation _changes_ the output type of a relation, FieldReferences to changed columns _are not_ updated.

## Code Example
The test below illustrates this issue:
```java
@Test
  void staleFieldRefStateAfterCopyOnWriteVisitation() {
    var b = new SubstraitBuilder(extensions);
    TypeCreator R = TypeCreator.of(false);
    TypeCreator N = TypeCreator.of(true);

    var plan =
        b.project(
            input -> List.of(b.fieldReference(input, 0)),
            // original named scan has type R.I64
            b.namedScan(List.of("example"), List.of("n"), List.of(R.I64)));

    var newPlanOption =
        plan.accept(
            new RelCopyOnWriteVisitor() {
              @Override
              public Optional<Rel> visit(NamedScan namedScan) throws RuntimeException {
                // replace the original named scan with one of type N.I64
                return Optional.of(b.namedScan(List.of("example"), List.of("n"), List.of(N.I64)));
              }
            });

    // decompose the new plan
    assertTrue(newPlanOption.isPresent());
    var newPlan = newPlanOption.get();
    assertInstanceOf(Project.class, newPlan);
    var newProject = (Project) newPlan;
    assertInstanceOf(NamedScan.class, newProject.getInput());
    var newNamedScan = (NamedScan) newProject.getInput();

    // the new named scan emits single column of type N.I64, as expected
    assertEquals(NamedStruct.of(List.of("n"), R.struct(N.I64)), newNamedScan.getInitialSchema());

    // retrieve the field reference
    var exprs = newProject.getExpressions();
    assertTrue(exprs.size() == 1);
    var expr = exprs.get(0);
    assertInstanceOf(FieldReference.class, expr);
    var fieldRef = (FieldReference) expr;

    // The field reference type should correspond to the type of the new named scan.
    // Instead, the type is R.I64 which is from the ORIGINAL named scan
    assertEquals(fieldRef.type(), N.I64); // <- FAILS
  }
```

## Fixing
[...]

PR #187, opened 2023-10-04
feat: initial HashJoinRel support

vibhatha:
This PR includes the base work for integrating HashEquiJoin.


CHANGELOG.md
```diff
@@ -8,7 +8,7 @@ Release Notes
 
 * StrLiteral is no longer converted to CHAR(<length>)
 
-fix(isthmus): convert BinaryLiteral to VARBINARY 
+fix(isthmus): convert BinaryLiteral to VARBINARY
```
danepitkin, line 11: nit: maybe revert whitespace change?

core/src/main/java/io/substrait/relation/AbstractJoin.java
```diff
@@ -0,0 +1,9 @@
+package io.substrait.relation;
+
+import io.substrait.expression.Expression;
+import java.util.Optional;
+
+public abstract class AbstractJoin extends BiRel implements HasExtension {
+
+  public abstract Optional<Expression> getPostJoinFilter();
```
danepitkin, line 8: I just want to add that a NestedLoopJoin does not have a post join filter, so it wouldn't be able to use this abstract class. Maybe it's not needed after all?

danepitkin (commented):
Nice work!

CHANGELOG.md
vibhatha: I got a formatting issue, that's why I added it, but let me check it again.

core/src/main/java/io/substrait/relation/AbstractJoin.java
vibhatha: Ah good point. I will update the PR.

CHANGELOG.md
vibhatha: @danepitkin I just went through NLJ PR to Substrait-Java, and I see something similar.

https://github.com/substrait-io/substrait-java/actions/runs/6475823135/job/17583520911#step:4:1

core/src/main/java/io/substrait/relation/physical/HashJoin.java
```diff
@@ -0,0 +1,79 @@
...
+
+  public abstract Optional<List<FieldReference>> getRightKeys();
+
+  public abstract JoinType getJoinType();
+
+  @Override
+  protected Type.Struct deriveRecordType() {
+    Stream<Type> leftTypes =
+        switch (getJoinType()) {
+          case RIGHT, OUTER -> getLeft().getRecordType().fields().stream()
```
danepitkin, line 27: Should we add RIGHT_SEMI and RIGHT_ANTI? (Same with LEFT below)

CHANGELOG.md
danepitkin: ohhh, is that what my error is?? Good catch, thank you

vibhatha: It seems like that, but how could it be there in the first place?

core/src/main/java/io/substrait/relation/physical/HashJoin.java
vibhatha: Good catch. no columns from the opposite are added.

core/src/test/java/io/substrait/type/proto/ExtensionRoundtripTest.java
```diff
@@ -174,6 +175,21 @@ void join() {
     verifyRoundTrip(rel);
   }
 
+  @Test
+  void hashJoin() {
+    int[] left_keys = {};
```
EpsilonPrime, line 180: Should we have some tests with non-empty keys to make sure that's working correctly?

CHANGELOG.md
vbarua: This is extra weird because the changelog is generated programatically 🤔 

Ah, the whitespace is in the commit message itself
https://github.com/substrait-io/substrait-java/commit/6e82f397a168f3811593fe8248753d70d4f0c04f

I think it's fine to fix it like this.

core/src/main/java/io/substrait/relation/physical/HashJoin.java
```diff
@@ -0,0 +1,85 @@
...
+import java.util.Optional;
+import java.util.stream.Stream;
+import org.immutables.value.Value;
+
+@Value.Immutable
+public abstract class HashJoin extends BiRel implements HasExtension {
+
+  public abstract Optional<List<FieldReference>> getLeftKeys();
+
+  public abstract Optional<List<FieldReference>> getRightKeys();
```
vbarua, line 21: I would argue that both `getLeftKeys` and `getRightKeys` should be required fields for a HashJoin (i.e. `List<FieldReference>`) as users should be forced to set them. As is, a user can construct a HashJoin with only left keys or only right keys.

Admittedly they can do this in the proto as well as the fields are not marked as required, but we can choose to be stricter in the library to guide their hand. If a user really wants a hash join on no keys, they should set the keys to empty lists.

core/src/main/java/io/substrait/relation/RelProtoConverter.java
```diff
@@ -228,6 +235,37 @@ public Rel visit(ExtensionTable extensionTable) throws RuntimeException {
...
+
+    hashJoin
+        .getRightKeys()
+        .ifPresent(
+            keys -> {
+              for (int i = 0; i < keys.size(); i++) {
+                builder.setRightKeys(i, toProto(keys.get(i)));
+              }
+            });
+
```
vbarua, line 264: We should also set the post join condition

core/src/main/java/io/substrait/relation/RelCopyOnWriteVisitor.java
```diff
@@ -166,6 +168,29 @@ public Optional<Rel> visit(Cross cross) throws RuntimeException {
...
+  @Override
+  public Optional<Rel> visit(HashJoin hashJoin) throws RuntimeException {
+    var left = hashJoin.getLeft().accept(this);
+    var right = hashJoin.getRight().accept(this);
+    var leftKeys = hashJoin.getLeftKeys();
+    var rightKeys = hashJoin.getRightKeys();
+    var postFilter = hashJoin.getPostJoinFilter().flatMap(t -> visitExpression(t));
+    if (allEmpty(left, right, leftKeys, rightKeys, postFilter)) {
+      return Optional.empty();
+    }
```
vbarua, line 180: The purpose of the RelCopyOnWriteVisitor is to produce a copy of a plan tree ONLY if some part of the tree has changed.

Here, `left`, `right` and `postFilter` are all empty if they have not been changed by calling `accept` or `visitExpression`, so it makes sense to include them in the `allEmpty` check 

`leftKeys` and `rightKeys` just return optional if they are not set (possibly another reason to have them be required in HashJoin). As is, they shouldn't be in the allEmpty check.

There is a deeper issue in that field references possibly _should_ be recomputed as part of the RelCopyOnWriteVisitor (see https://github.com/substrait-io/substrait-java/issues/185), but I think that's outside the scope of the PR. Noting it here so it gets linked to the issue.

core/src/test/java/io/substrait/type/proto/ExtensionRoundtripTest.java
vbarua: +1 to using non-empty keys.

core/src/main/java/io/substrait/relation/RelProtoConverter.java
vbarua: Heh, the Join converter is also missing the post-join condition 😅

core/src/main/java/io/substrait/dsl/SubstraitBuilder.java
```diff
@@ -165,6 +166,28 @@ private Join join(
         .build();
   }
 
+  public HashJoin hashJoin(
+      int[] leftKeys, int[] rightKeys, HashJoin.JoinType joinType, Rel left, Rel right) {
```
vbarua, line 170: I would avoid using bare arrays here as it makes this hard to use in my opinion. With something like:
```java
  public HashJoin hashJoin(
      List<Integer> leftKeys, List<Integer> rightKeys, HashJoin.JoinType joinType, Rel left, Rel right)
```
users can inline the key lists like
```java
b.hashJoin(
  Collections.EMPTY_LIST,
  Collections.EMPTY_LIST,
  HashJoin.JoinType.INNER, commonTable, commonTable))
```
or
```java
b.hashJoin(
  List.of(1,3),
  List.of(2,4),
  HashJoin.JoinType.INNER, commonTable, commonTable)
```
assuming they have access to newer versions of Java which _we_ don't within the core library, but we do with the isthmus side of things.

vbarua (changes requested):
Overall this looks really good. Left some comments.

core/src/main/java/io/substrait/relation/RelCopyOnWriteVisitor.java
vibhatha: Shall we create an issue? And make sure this is addressed.

core/src/main/java/io/substrait/dsl/SubstraitBuilder.java
vibhatha: Ah I was thinking a better way 😃
I will update.

vibhatha:
@vbarua Thank you for the review. I will address the comments.

core/src/main/java/io/substrait/relation/RelCopyOnWriteVisitor.java
vbarua: I think https://github.com/substrait-io/substrait-java/issues/185 should be enough. It captures the overall issue.

core/src/test/java/io/substrait/type/proto/ExtensionRoundtripTest.java
vibhatha: @vbarua @EpsilonPrime 
Looking into the `commonTable` initialization, we do have empty lists to create it, so if we use non-empty lists for keys, we may need to introduce a commonTable with non-empty initialization? Am I following this incorrectly?


```java
final Rel commonTable =
      b.namedScan(Collections.emptyList(), Collections.emptyList(), Collections.emptyList());
```

core/src/main/java/io/substrait/relation/physical/HashJoin.java
vibhatha: This is a good point. I think I didn't think about it well in the first go. +1 for this change. 
I will add it.

core/src/main/java/io/substrait/relation/RelProtoConverter.java
vibhatha: I will add it too.

core/src/main/java/io/substrait/relation/RelProtoConverter.java
```diff
@@ -228,6 +237,33 @@ public Rel visit(ExtensionTable extensionTable) throws RuntimeException {
...
+    List<FieldReference> rightKeys = hashJoin.getRightKeys();
+
+    if (leftKeys.size() != rightKeys.size()) {
+      throw new RuntimeException("Number of left and right keys must be equal.");
+    }
+
+    for (int idx = 0; idx < hashJoin.getLeftKeys().size(); idx++) {
+      builder.setLeftKeys(idx, toProto(leftKeys.get(idx)));
+      builder.setRightKeys(idx, toProto(rightKeys.get(idx)));
+    }
```
vibhatha, line 259: @vbarua can I do better here? 
Also validation of left and right key sizes is a bit doubtful.

core/src/main/java/io/substrait/relation/physical/HashJoin.java
vibhatha: I updated the code.

core/src/main/java/io/substrait/relation/RelProtoConverter.java
vibhatha: Updated the code.

core/src/main/java/io/substrait/dsl/SubstraitBuilder.java
vibhatha: Updated the code.

core/src/main/java/io/substrait/relation/RelProtoConverter.java
vbarua: The validation seems reasonable to me. The only thing I would suggest to improve it would be to provide additional context. Something like:
```
The number of left and right keys must be equal for HashJoin. Received left %s and right %s keys.
```

For setting the keys themselves, I think
```java
    builder.addAllLeftKeys(leftKeys.stream().map(this::toProto).collect(Collectors.toList()));
    builder.addAllRightKeys(rightKeys.stream().map(this::toProto).collect(Collectors.toList()));
```
is a bit cleaner.

core/src/main/java/io/substrait/relation/RelProtoConverter.java
vbarua: Thanks for fixing it in both places 🙇

core/src/test/java/io/substrait/type/proto/ExtensionRoundtripTest.java
vbarua: > we may need to introduce a commonTable with non-empty initialization? 

This is correct. You can use the `b.namedScan` method to create a table with columns, something like
```java
b.namedScan(List.of("example"), List.of("a"), List.of(R.STRING)
```

One small structural thing. The tests in `ExtensionRoundtripTest.java` verify that _extensions_ roundtrip. The test that you've added already does this perfectly well.

For adding a test using non-empty keys, could I ask that you introduce something like `JoinRountripTest.java` (in a similar vein to `AggregateRoundtripTest.java`) and do it there. That will also be a good place to add nested join tests later.

core/src/main/java/io/substrait/relation/ProtoRelConverter.java
```diff
@@ -490,6 +495,35 @@ private Set newSet(SetRel rel) {
...
+    Type.Struct unionedStruct = Type.Struct.builder().from(leftStruct).from(rightStruct).build();
+    var converter = new ProtoExpressionConverter(lookup, extensions, unionedStruct, this);
+    var builder =
+        HashJoin.builder()
+            .left(left)
+            .right(right)
+            .leftKeys(
+                rel.getLeftKeysList().stream().map(converter::from).collect(Collectors.toList()))
+            .rightKeys(
+                rel.getRightKeysList().stream().map(converter::from).collect(Collectors.toList()))
```
vbarua, line 512: I _think_ there might be a bug here. The offsets of the right key list are relative to the right input. However, here the converter is created using the left _and_ right inputs. To illustrate, given two tables like:
```
t1: (a: i64, b: fp64, c:string)
t2: (x: fp64, y: string, z: i64)
```
let's say we want to join on
```
left keys: (0, 1)
right keys: (2, 0)
```
the converter has the following root type
```
(a: i64, b: fp64, c:string, x: fp64, y: string, z: i64)
```
so the left keys `(0,1)` refer to the correct inputs `(a, b)` but the right keys `(2, 0)` refer to `(c, a)` not `(z, x)`.

It would be good to have a test case like this.

core/src/main/java/io/substrait/relation/RelCopyOnWriteVisitor.java
```diff
@@ -166,6 +168,29 @@ public Optional<Rel> visit(Cross cross) throws RuntimeException {
...
+    return Optional.of(
+        ImmutableHashJoin.builder()
+            .from(hashJoin)
+            .left(left.orElse(hashJoin.getLeft()))
+            .right(right.orElse(hashJoin.getRight()))
+            .leftKeys(leftKeys)
+            .rightKeys(rightKeys)
+            .postJoinFilter(
+                Optional.ofNullable(
+                    postFilter.orElseGet(() -> hashJoin.getPostJoinFilter().orElse(null))))
```
vbarua, line 190: Because postFilter is already an Optional, we can use the `or` method instead of `Optional.ofNullable` here like:
```java
            .postJoinFilter(postFilter.or(hashJoin::getPostJoinFilter))
```

core/src/main/java/io/substrait/relation/RelProtoConverter.java
vibhatha: This is perfect. I will update.

core/src/test/java/io/substrait/type/proto/ExtensionRoundtripTest.java
vibhatha: @vbarua I will defintely add a test class as you have susggested here. 👍

core/src/main/java/io/substrait/relation/ProtoRelConverter.java
vibhatha: @vbarua you are correct. Thanks for catching this. It did lack the non-key test and it uncovered this issue. 

I added a converter separately for left and right, and then kept the union-based one for the post join filter. 

Also there are redundant parts in the newly added join test class with respect to the extension roundrip tests. Shall we use another abstraction to keep the common utils?

core/src/main/java/io/substrait/relation/RelProtoConverter.java
vibhatha: Updated.

vibhatha:
@vbarua ready for another round of review and left a few questions too 🙂

core/src/main/java/io/substrait/relation/RelCopyOnWriteVisitor.java
vbarua: You're not missing anything, I forget that we're on Java 8 sometimes so we don't get access to all the nice Optional util methods 🥲

core/src/main/java/io/substrait/relation/ProtoRelConverter.java
```diff
@@ -490,6 +496,39 @@ private Set newSet(SetRel rel) {
     return builder.build();
   }
 
+  private Rel newHashJoin(HashJoinRel rel) {
+    Rel left = from(rel.getLeft());
+    Rel right = from(rel.getRight());
+    var leftKeys = rel.getLeftKeysList();
+    var rightKeys = rel.getRightKeysList();
+    var rightOffSetKeys =
+        Stream.concat(leftKeys.stream(), rightKeys.stream()).collect(Collectors.toList());
```
vbarua, line 505: `rightOffSetKeys` isn't used, could you remove it.

core/src/test/java/io/substrait/type/proto/JoinRoundtripTest.java
```diff
@@ -0,0 +1,78 @@
...
+      AdvancedExtension.builder()
+          .enhancement(new StringHolder("COMMON ENHANCEMENT"))
+          .optimization(new StringHolder("COMMON OPTIMIZATION"))
+          .build();
+
+  final AdvancedExtension relExtension =
+      AdvancedExtension.builder()
+          .enhancement(new StringHolder("REL ENHANCEMENT"))
+          .optimization(new StringHolder("REL OPTIMIZATION"))
+          .build();
```
vbarua, line 56: We don't need to build and test common and rel extensions here, the one test we have in ExtensionRoundtripTest verifies it.

core/src/test/java/io/substrait/type/proto/JoinRoundtripTest.java
```diff
@@ -0,0 +1,78 @@
...
+  void verifyRoundTrip(Rel rel) {
+    io.substrait.proto.Rel protoRel = relProtoConverter.toProto(rel);
+    Rel relReturned = protoRelConverter.from(protoRel);
+    assertEquals(rel, relReturned);
+  }
+
+  @Test
+  void hashJoin() {
+    List<Integer> leftEmptyKeys = Arrays.asList(0, 1);
+    List<Integer> rightEmptyKeys = Arrays.asList(2, 0);
```
vbarua, line 67: minor: `leftEmptyKeys` -> `leftKeys` and `rightEmptyKeys` -> `rightKeys`

The keys are no longer empty 🙂

core/src/main/java/io/substrait/relation/ProtoRelConverter.java
vbarua: > Also there are redundant parts in the newly added join test class with respect to the extension roundrip tests. Shall we use another abstraction to keep the common utils?

I think it would be worth doing at some point, but I don't think it needs to be part of this PR. I'd rather keep the changeset small.

vbarua (commented):
Overall change looks good. Left some final comments. I think we can get this in next week 🙂

core/src/test/java/io/substrait/type/proto/JoinRoundtripTest.java
vibhatha: Oh 😲

vibhatha:
@vbarua I will make these changes and thank you for the quick response 🙏

vibhatha:
@vbarua addressed the reviews 🙂

vbarua (approved):
LGMT! Thanks for contribution 🙇

vibhatha:
Appreciate your support :pray:

PR #188, opened 2023-10-10
feat: add NestedLoopJoin rel

danepitkin:
Requires https://github.com/substrait-io/substrait/pull/561 to be merged and released.


core/src/main/java/io/substrait/dsl/SubstraitBuilder.java
```diff
@@ -19,6 +19,7 @@
 import io.substrait.relation.Filter;
 import io.substrait.relation.Join;
 import io.substrait.relation.NamedScan;
+import io.substrait.relation.NestedLoopJoin;
```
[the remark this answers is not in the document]
danepitkin: Ah yes, I can move it over. It does add quite a few additional imports to files, are physical operators special enough to warrant a subfolder in your opinion?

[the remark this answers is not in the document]
EpsilonPrime: @vbarua Should we create a subdirectory or not?  If not we need to update the HashJoin  PR.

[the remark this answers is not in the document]
vbarua: The HashJoin PR already introduces the `physical` subdirectory, which seems reasonable to me.

> It does add quite a few additional imports to files

I think that's a reasonable tradeoff for extra clarity.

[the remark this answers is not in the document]
danepitkin: Sounds good. I'll move it in this PR as well

vbarua:
https://github.com/substrait-io/substrait/pull/561 has been merged. After it's released over the weekend, we can update the submodule to point to it (preferably as it's own PR).

vbarua:
I went ahead and updated substrait-java with the latest version of substrait, which includes the NestedLoopJoin changes.

core/src/main/java/io/substrait/relation/ProtoRelConverter.java
```diff
@@ -527,6 +532,35 @@ private Rel newHashJoin(HashJoinRel rel) {
...
+        NestedLoopJoin.builder()
+            .left(left)
+            .right(right)
+            .condition(
+                // defaults to true if the join expression is unassigned, resulting in a cartesian
+                // join
+                Optional.of(
+                    rel.hasExpression()
+                        ? converter.from(rel.getExpression())
+                        : Expression.BoolLiteral.builder().value(true).build()))
```
danepitkin, line 552: One thing to note for NLJ is that the spec says the Join Expression is optional, but defaults to True which becomes a cartesian join. I've added the code to default to True here.

danepitkin (commented):
Thanks for updating to the latest version of Substrait! I reviewed the HashJoin PR and tried to implement everything that is applicable for NLJ.

core/src/test/java/io/substrait/type/proto/JoinRoundtripTest.java
```diff
@@ -41,10 +43,13 @@ public class JoinRoundtripTest extends TestBase {
           Arrays.asList("d", "e", "f"),
           Arrays.asList(R.FP64, R.STRING, R.I64));
 
-  void verifyRoundTrip(Rel rel) {
+  Rel roundTrip(Rel rel) {
     io.substrait.proto.Rel protoRel = relProtoConverter.toProto(rel);
-    Rel relReturned = protoRelConverter.from(protoRel);
-    assertEquals(rel, relReturned);
+    return protoRelConverter.from(protoRel);
+  }
```
vibhatha, line 49: Looks better. :+1:

danepitkin: Thanks! I split up roundtrip functionality from verification-of-roundtrip so I could test inequality for NLJ more easily.

core/src/test/java/io/substrait/type/proto/ExtensionRoundtripTest.java
```diff
@@ -196,6 +197,19 @@ void hashJoin() {
     verifyRoundTrip(relWithoutKeys);
   }
 
+  @Test
+  void nested_loop_join() {
```
vbarua, line 201: minor: the Java expectation is `camelCase` for function names.

core/src/main/java/io/substrait/relation/physical/NestedLoopJoin.java
```diff
@@ -0,0 +1,80 @@
...
+import io.substrait.type.Type;
+import io.substrait.type.TypeCreator;
+import java.util.Optional;
+import java.util.stream.Stream;
+import org.immutables.value.Value;
+
+@Value.Immutable
+public abstract class NestedLoopJoin extends BiRel implements HasExtension {
+
+  public abstract Optional<Expression> getCondition();
```
vbarua, line 17: What do you think about making this
```java
public abstract Expression getCondition();
```

The spec indicates that
```proto
  // optional, defaults to true (a cartesian join)
  Expression expression = 4;
```

but we can be stricter within the POJO layer and require it. You're already inserting a boolean literal condition if the protobuf doesn't contain one https://github.com/substrait-io/substrait-java/pull/188/files#r1372502462

core/src/test/java/io/substrait/type/proto/ExtensionRoundtripTest.java
danepitkin: Thank you! This is my python subconscious taking over.. haha

core/src/main/java/io/substrait/relation/physical/NestedLoopJoin.java
danepitkin: I can go ahead and change the spec, since I just added it recently.

danepitkin: Hmm, it's worth noting that `JoinRel` uses ```public abstract Optional<Expression> getCondition();```, which is required. Maybe it is best to implement them similarly?

danepitkin: I made the change in the latest commit if it is the preferred choice!

vbarua: The JoinRel in the protobuf spec doesn't have any information about whether condition is required or not:
https://github.com/substrait-io/substrait/blob/b3071bc9cd77cf916568641c83056a285f8123be/proto/substrait/algebra.proto#L156-L160

I'm guessing that's why it's Optional currently. I think it would make sense to make the JoinRel condition be required as well. That can be a separate change though.

core/src/main/java/io/substrait/relation/ProtoRelConverter.java
```diff
@@ -527,6 +532,29 @@ private Rel newHashJoin(HashJoinRel rel) {
...
+    Rel right = from(rel.getRight());
+    Type.Struct leftStruct = left.getRecordType();
+    Type.Struct rightStruct = right.getRecordType();
+    Type.Struct unionedStruct = Type.Struct.builder().from(leftStruct).from(rightStruct).build();
+    var converter = new ProtoExpressionConverter(lookup, extensions, unionedStruct, this);
+    var builder =
+        NestedLoopJoin.builder()
+            .left(left)
+            .right(right)
+            .condition(converter.from(rel.getExpression()))
```
vbarua, line 546: I think we still want your original code that checks if the condition is null and then sets the condition to True if it is. The `expression` field in the protobuf isn't marked as required, so we can't assume it's present when reading the protos in.

core/src/test/java/io/substrait/type/proto/JoinRoundtripTest.java
```diff
@@ -57,4 +58,15 @@ void hashJoin() {
...
     verifyRoundTrip(relWithoutKeys);
   }
+
+  @Test
+  void nestedLoopJoin() {
+    Rel rel =
+        NestedLoopJoin.builder()
+            .from(
+                b.nestedLoopJoin(
+                    __ -> b.bool(true), NestedLoopJoin.JoinType.INNER, leftTable, rightTable))
```
vbarua, line 68: The test in ExtensionRoundtripTest.java use a bool true condition as well. Could you use a more interesting condition here like a key equality comparison between the two tables (ie. `leftTable.a = rightTable.f`) so that we can have a test that verifies that the condition expressions are being converted correctly.

vbarua (commented):
Thanks for the updates. Left some small comments.

core/src/main/java/io/substrait/relation/ProtoRelConverter.java
vbarua: How I'm thinking about this is that this is _effectively_ required, because when it is null it should be treated as True. But we can handle that behaviour at our serialisation boundary and have it be properly required internally, which makes it easier to work with within our code.

core/src/test/java/io/substrait/type/proto/JoinRoundtripTest.java
danepitkin: Good point! Will update.

core/src/main/java/io/substrait/relation/ProtoRelConverter.java
danepitkin: Thanks for the explanation! +1

danepitkin:
I can repro the failure locally. I'm still digging into it, it seems the test case is surfacing an actual bug.

core/src/test/java/io/substrait/type/proto/JoinRoundtripTest.java
```diff
@@ -31,4 +32,18 @@ void hashJoin() {
...
     verifyRoundTrip(relWithoutKeys);
   }
+
+  @Test
+  void nestedLoopJoin() {
+    Rel rel =
+        NestedLoopJoin.builder()
+            .from(
+                b.nestedLoopJoin(
+                    __ -> b.equal(b.fieldReference(leftTable, 0), b.fieldReference(rightTable, 2)),
```
vbarua, line 42: Ah! This is subtle and weird (and something to improve in the builder actually).

The condition on the join is evaluated relative to the joins record shape, which is the union of the left and right table records. It would look like:
```
(a, b, c, d, e, f)
```

`b.fieldReference(rightTable, 2)` here ends up referring to `c`

danepitkin: Ah that makes sense! Thank you

vbarua: I'm realising there's nothing else in the code that uses this, which is why it hasn't come up before.

danepitkin: For the test case, should I create a unioned named table to refer to? `5` is our of bounds when creating the Rel, but in bounds after the roundtrip

danepitkin: Hmm, we have `newInputRelReference(int index, List<Rel> rels)`, which might handle this. I'll dig into this more myself.

core/src/main/java/io/substrait/dsl/SubstraitBuilder.java
```diff
@@ -286,6 +311,16 @@ public List<FieldReference> fieldReferences(Rel input, int... indexes) {
...
 
+  public FieldReference fieldReference(List<Rel> inputs, int index) {
+    return ImmutableFieldReference.newInputRelReference(index, inputs);
+  }
+
+  public List<FieldReference> fieldReferences(List<Rel> inputs, int... indexes) {
+    return Arrays.stream(indexes)
+        .mapToObj(index -> fieldReference(inputs, index))
+        .collect(java.util.stream.Collectors.toList());
+  }
```
vbarua, line 322: Good call on these, they will be quite helpful.

vbarua (approved):
Looks good! Thanks for adding this.

danepitkin:
Thank you so much for the help!

PR #189, opened 2023-10-19
fix: map switch expression to a Calcite CASE statement

carlyeks:
The switch expression POJO in substrait-java doesn't contain the match expression, so I've updated the definition to include that.

To map into Calcite, I construct a CASE statement. This ends up being translated into an IfThen expression when round-tripping Substrait into Calcite.

### DIFF
--- a/isthmus/src/test/java/io/substrait/isthmus/ExpressionConvertabilityTest.java
+++ b/isthmus/src/test/java/io/substrait/isthmus/ExpressionConvertabilityTest.java
@@ -1,5 +1,6 @@
 package io.substrait.isthmus;
 
+import static io.substrait.isthmus.expression.CallConverters.CASE;
 import static io.substrait.isthmus.expression.CallConverters.CREATE_SEARCH_CONV;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 
@@ -21,6 +22,7 @@ import org.junit.jupiter.api.Test;
 
 /** Tests which test that an expression can be converted to and from Calcite expressions. */
 public class ExpressionConvertabilityTest extends PlanTestBase {
+
   static final TypeCreator R = TypeCreator.of(false);
   static final TypeCreator N = TypeCreator.of(true);
 
@@ -61,7 +63,6 @@ public class ExpressionConvertabilityTest extends PlanTestBase {
                         SimpleExtension.loadDefaults().scalarFunctions(), typeFactory)));
     var to = new ExpressionProtoConverter(new ExtensionCollector(), null);
     assertEquals(
-        expression.accept(to),
         b.scalarFn(
                 "/functions_boolean.yaml",
                 "or:bool",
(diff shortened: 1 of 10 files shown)


EpsilonPrime (approved):
Looks solid to me.  I'll let others have a chance to take a look at it though.

isthmus/src/test/java/io/substrait/isthmus/SubstraitRelNodeConverterTest.java
```diff
@@ -285,4 +285,24 @@ public void emit() {
...
+                          b.switchExpression(
+                              b.fieldReference(input, 0),
+                              List.of(b.switchClause(b.i32(0), b.fieldReference(input, 3))),
+                              b.bool(false))),
+                  b.remap(4),
+                  commonTable));
+      var relNode = converter.convert(root.getInput());
+      assertRowMatch(relNode.getRowType(), N.BOOLEAN);
+    }
+  }
```
vbarua, line 307: This test feel out of place in this file, given that it's testing an expression conversion rather an a rel conversion. It also seem redundant as your other test is more thorough.

isthmus/src/test/java/io/substrait/isthmus/ExpressionConvertabilityTest.java
```diff
@@ -78,6 +79,54 @@ public void singleOrList() throws IOException {
...
+                        b.scalarFn(
+                            "/functions_comparison.yaml",
+                            "equal:any_any",
+                            R.BOOLEAN,
+                            b.fieldReference(commonTable, 0),
+                            b.i32(10)),
+                        b.i32(2))),
+                b.i32(3))
+            .accept(to),
+        expression.accept(to));
```
vbarua, line 130: It took me a while to figure out what this test is doing. If I understand this correctly, we can't just do a roundtrip test here because Calcite changes case switch statement to a series of case when expressions. What you're doing is:
1. Build a substrait plan with the case expression.
2. Convert it to Calcite
3. Pull out the expression from the filter.
4. Manually create the expected substrait equivalent as the first argument to the assert.
5. Compare the protobuf forms of both expressions.

Could you pull out 4 and assigned it to a variable like `expectedExpression`, along with a comment indicating how Calcite is changing the original expression to a new equivalent.

vbarua (commented):
Overall looks good, just some small comments about tests.

isthmus/src/test/java/io/substrait/isthmus/ExpressionConvertabilityTest.java
vbarua: Have a much better idea for this which I'm implementing.

isthmus/src/test/java/io/substrait/isthmus/ExpressionConvertabilityTest.java
```diff
@@ -44,40 +58,50 @@ public void mapLiteral() throws IOException, SqlParseException {
...
+
+      // cannot roundtrip test switchExpression because Calcite simplifies the representation
+      assertExpressionEquality(
+        b.ifThen(
+            List.of(
+                b.ifClause(b.equal(b.fieldReference(commonTable, 0), b.i32(5)), b.i32(1)),
+                b.ifClause(b.equal(b.fieldReference(commonTable, 0), b.i32(10)), b.i32(2))),
+            b.i32(3)),
+        expression);
+  }
```
vbarua, line 99: These tests were creating a full plan and converting it, but we only actually want/need to verify the _expression_ conversion.

Removed the plan related noise by converting the expression directly. Also used a bunch of our nice utils (including one I added) to make the tests more readable.

vbarua (approved):
Core implementation looks good. I went ahead and updated some of the tests, mostly to simplify them.

Issue #128, opened 2023-02-01
[Isthmus] Convert SQL expression to Substrait `ExtendedExpression`


ianmcook:
https://github.com/substrait-io/substrait/pull/405 added an `ExtendedExpression` message to represent standalone expressions (expressions that are not part of plans). #127 would add support for this in the Substrait Java library. It would also be nice if Isthmus could convert SQL expressions (not whole queries) to Substrait `ExtendedExpression` messages. This would help to open up more applications for Substrait as an interoperable representation for expressions.

danepitkin:
Looks like we just need to extend https://github.com/substrait-io/substrait-java/blob/main/isthmus/src/main/java/io/substrait/isthmus/SqlToSubstrait.java to work with `ExtendedExpressions`.

PR #191, opened 2023-10-24
feat: convert sql expression into proto extended expressions

davisusanibar:
The proposal involves receiving a SQL expression (e.g. column_a > 18), validating the SqlNode expression, and then converting it into RexNode. It then traverses RexNode to validate the type of expression, and finally creates a proto Extended Expression message.

These are the activities involved on this PR:

- [x] The objective of this PoC is to convert a SQL Expression into an Extended Expression by using the [RexNode convertExpression(...)](https://github.com/apache/calcite/blob/c6031ca15579b88bd392a56f4588fdd9a29d1071/core/src/main/java/org/apache/calcite/sql2rel/SqlToRelConverter.java#L2132) method and hardcoding some components of the logic. This draft PR cover that part.

These are the activities pending to solve on this PR:

- [ ] Need to create dynamically nameToNodeMap relation [Map<String, RexNode> nameToNodeMap](https://github.com/substrait-io/substrait-java/blob/333989ffa70efc18c6707c3cf670bac826925fe1/isthmus/src/main/java/io/substrait/isthmus/SqlToSubstrait.java#L286-L301)
- [ ] Need to cover support for project expression (curren support only filter expression)
- [ ] Need to set [function reference and addArguments](https://github.com/substrait-io/substrait-java/blob/333989ffa70efc18c6707c3cf670bac826925fe1/isthmus/src/main/java/io/substrait/isthmus/SqlToSubstrait.java#L141-L153) dynamically
- [ ] Need to create [schema](https://github.com/substrait-io/substrait-java/blob/333989ffa70efc18c6707c3cf670bac826925fe1/isthmus/src/main/java/io/substrait/isthmus/SqlToSubstrait.java#L170-L172) dynamically
- [ ] Need to add [extensions URI ](https://github.com/substrait-io/substrait-java/blob/333989ffa70efc18c6707c3cf670bac826925fe1/isthmus/src/main/java/io/substrait/isthmus/SqlToSubstrait.java#L179-L185)dynamically
- [ ] Need to add [extensions](https://github.com/substrait-io/substrait-java/blob/333989ffa70efc18c6707c3cf670bac826925fe1/isthmus/src/main/java/io/substrait/isthmus/SqlToSubstrait.java#L188-L197) dynamically

Please let me know if you have any ideas, improvements, or concerns so that I can evaluate them appropriately. I appreciate your assistance in advance.

To closes: https://github.com/substrait-io/substrait-java/issues/128

### DIFF
--- a/isthmus/src/main/java/io/substrait/isthmus/SqlToSubstrait.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/SqlToSubstrait.java
@@ -1,22 +1,44 @@
 package io.substrait.isthmus;
 
+import com.github.bsideup.jabel.Desugar;
 import com.google.common.annotations.VisibleForTesting;
 import io.substrait.extension.ExtensionCollector;
+import io.substrait.proto.Expression;
+import io.substrait.proto.Expression.ScalarFunction;
+import io.substrait.proto.ExpressionReference;
+import io.substrait.proto.ExtendedExpression;
+import io.substrait.proto.FunctionArgument;
 import io.substrait.proto.Plan;
 import io.substrait.proto.PlanRel;
+import io.substrait.proto.SimpleExtensionDeclaration;
+import io.substrait.proto.SimpleExtensionURI;
 import io.substrait.relation.RelProtoConverter;
 import io.substrait.type.NamedStruct;
+import io.substrait.type.Type;
+import io.substrait.type.TypeCreator;
+import io.substrait.type.proto.TypeProtoConverter;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 import java.util.function.Function;
 import org.apache.calcite.plan.hep.HepPlanner;
 import org.apache.calcite.plan.hep.HepProgram;
 import org.apache.calcite.prepare.CalciteCatalogReader;
 import org.apache.calcite.rel.RelRoot;
+import org.apache.calcite.rel.type.RelDataType;
 import org.apache.calcite.rel.type.RelDataTypeFactory;
+import org.apache.calcite.rex.RexCall;
+import org.apache.calcite.rex.RexInputRef;
+import org.apache.calcite.rex.RexLiteral;
+import org.apache.calcite.rex.RexNode;
 import org.apache.calcite.schema.Schema;
 import org.apache.calcite.sql.SqlNode;
 import org.apache.calcite.sql.parser.SqlParseException;
 import org.apache.calcite.sql.parser.SqlParser;
+import org.apache.calcite.sql.type.SqlTypeName;
 import org.apache.calcite.sql.validate.SqlValidator;
 import org.apache.calcite.sql2rel.SqlToRelConverter;
 import org.apache.calcite.sql2rel.StandardConvertletTable;
@@ -46,6 +68,12 @@ public class SqlToSubstrait extends SqlConverterBase {
   public Plan execute(String sql, String name, Schema schema) throws SqlParseException {
     var pair = registerSchema(name, schema);
     return executeInner(sql, factory, pair.left, pair.right);
+  }
+
+  public ExtendedExpression executeExpression(String expr, List<String> tables)
+      throws SqlParseException {
+    var pair = registerCreateTables(tables);
+    return executeInnerExpression(expr, pair.left, pair.right);
   }
 
   // Package protected for testing
@@ -91,6 +119,138 @@ public class SqlToSubstrait extends SqlConverterBase {
     return plan.build();
   }
 
+  private ExtendedExpression executeInnerExpression(
+      String sql, SqlValidator validator, CalciteCatalogReader catalogReader)
+      throws SqlParseException {
+    ExtendedExpression.Builder extendedExpressionBuilder = ExtendedExpression.newBuilder();
+    ExtensionCollector functionCollector = new ExtensionCollector();
+    sqlToRexNode(sql, validator, catalogReader)
+        .forEach(
+            rexNode -> {
+              // FIXME! Implement it dynamically for more expression types
+              ResulTraverseRowExpression result = TraverseRexNode.getRowExpression(rexNode);
+
+              // FIXME! Get output type dynamically:
+              // final static Map<String, Type> getTypeCreator = new HashMap<>(){{put("BOOLEAN",
+              // TypeCreator.of(true).BOOLEAN);}};
+              // getTypeCreator.get(rexNode.getType()).accept(...)
+              io.substrait.proto.Type output =
+                  TypeCreator.NULLABLE.BOOLEAN.accept(new TypeProtoConverter(functionCollector));
+
+              // FIXME! setFunctionReference, addArguments(index: 0, 1)
+              Expression.Builder expressionBuilder =
+                  Expression.newBuilder()
+                      .setScalarFunction(
+                          ScalarFunction.newBuilder()
+                              .setFunctionReference(1)
+                              .setOutputType(output)
+                              .addArguments(
+                                  0,
+                                  FunctionArgument.newBuilder().setValue(result.referenceBuilder()))
+                              .addArguments(
+                                  1,
+                                  FunctionArgument.newBuilder()
+                                      .setValue(result.expressionBuilderLiteral())));
+              ExpressionReference.Builder expressionReferenceBuilder =
+                  ExpressionReference.newBuilder()
+                      .setExpression(expressionBuilder)
+                      .addOutputNames(result.ref().getName());
+
+              // FIXME! Get schema dynamically
+              // (as the same for Plan with:
+              // TypeConverter.DEFAULT.toNamedStruct(rexNode.getType());)
+              List<String> columnNames =
+                  Arrays.asList("N_NATIONKEY", "N_NAME", "N_REGIONKEY", "N_COMMENT");
+              List<Type> dataTypes =
+                  Arrays.asList(
+                      TypeCreator.NULLABLE.I32,
+                      TypeCreator.NULLABLE.STRING,
+                      TypeCreator.NULLABLE.I32,
+                      TypeCreator.NULLABLE.STRING);
+              NamedStruct namedStruct =
+                  NamedStruct.of(
+                      columnNames, Type.Struct.builder().fields(dataTypes).nullable(false).build());
+
+              extendedExpressionBuilder
+                  .addReferredExpr(0, expressionReferenceBuilder)
+                  .setBaseSchema(namedStruct.toProto(new TypeProtoConverter(functionCollector)));
+
+              // Extensions URI FIXME! Populate/create this dynamically
+              HashMap<String, SimpleExtensionURI> extensionUris = new HashMap<>();
+              extensionUris.put(
+                  "key-001",
+                  SimpleExtensionURI.newBuilder()
+                      .setExtensionUriAnchor(1)
+                      .setUri("/functions_comparison.yaml")
+                      .build());
+
+              // Extensions FIXME! Populate/create this dynamically, maybe use rexNode.getKind()
+              ArrayList<SimpleExtensionDeclaration> extensions = new ArrayList<>();
+              SimpleExtensionDeclaration extensionFunctionLowerThan =
+                  SimpleExtensionDeclaration.newBuilder()
+                      .setExtensionFunction(
+                          SimpleExtensionDeclaration.ExtensionFunction.newBuilder()
+                              .setFunctionAnchor(1)
+                              .setName("gt:any_any")
+                              .setExtensionUriReference(1))
+                      .build();
+              extensions.add(extensionFunctionLowerThan);
+
+              extendedExpressionBuilder.addAllExtensionUris(extensionUris.values());
+              extendedExpressionBuilder.addAllExtensions(extensions);
+            });
+    return extendedExpressionBuilder.build();
+  }
+
+  static class TraverseRexNode {
+    static RexInputRef ref = null;
+    static Expression.Builder referenceBuilder = null;
+    static Expression.Builder expressionBuilderLiteral = null;
+
+    static ResulTraverseRowExpression getRowExpression(RexNode rexNode) {
+
+      switch (rexNode.getClass().getSimpleName().toUpperCase()) {
+        case "REXCALL":
+          for (RexNode rexInternal : ((RexCall) rexNode).operands) {
+            getRowExpression(rexInternal);
+          }
+          ;
+          break;
+        case "REXINPUTREF":
+          ref = (RexInputRef) rexNode;
+          referenceBuilder =
+              Expression.newBuilder()
+                  .setSelection(
+                      Expression.FieldReference.newBuilder()
+                          .setDirectReference(
+                              Expression.ReferenceSegment.newBuilder()
+                                  .setStructField(
+                                      Expression.ReferenceSegment.StructField.newBuilder()
+                                          .setField(ref.getIndex()))));
+          break;
+        case "REXLITERAL":
+          RexLiteral literal = (RexLiteral) rexNode;
+          expressionBuilderLiteral =
+              Expression.newBuilder()
+                  .setLiteral(
+                      Expression.Literal.newBuilder().setI32(literal.getValueAs(Integer.class)));
+          break;
+        default:
+          throw new AssertionError(
+              "Unsupported type for: " + rexNode.getClass().getSimpleName().toUpperCase());
+      }
+      ResulTraverseRowExpression result =
+          new ResulTraverseRowExpression(ref, referenceBuilder, expressionBuilderLiteral);
+      return result;
+    }
+  }
+
+  @Desugar
+  private record ResulTraverseRowExpression(
+      RexInputRef ref,
+      Expression.Builder referenceBuilder,
+      Expression.Builder expressionBuilderLiteral) {}
+
   private List<RelRoot> sqlToRelNode(
       String sql, SqlValidator validator, CalciteCatalogReader catalogReader)
       throws SqlParseException {
@@ -106,6 +266,42 @@ public class SqlToSubstrait extends SqlConverterBase {
             .collect(java.util.stream.Collectors.toList());
     return roots;
   }
+
+  private List<RexNode> sqlToRexNode(
+      String sql, SqlValidator validator, CalciteCatalogReader catalogReader)
+      throws SqlParseException {
+    SqlParser parser = SqlParser.create(sql, parserConfig);
+    SqlNode sqlNode = parser.parseExpression();
+    Result result = getResult(validator);
+    SqlNode validSQLNode =
+        validator.validateParameterizedExpression(
+            sqlNode,
+            result.nameToTypeMap()); // FIXME! It may be optional to include this validation
+    SqlToRelConverter converter = createSqlToRelConverter(validator, catalogReader);
+    RexNode rexNode = converter.convertExpression(validSQLNode, result.nameToNodeMap());
+
+    return Collections.singletonList(rexNode);
+  }
+
+  private static Result getResult(SqlValidator validator) {
+    // FIXME! Needs to be created dinamycally, this is for PoC purpose
+    HashMap<String, RexNode> nameToNodeMap = new HashMap<>();
+    nameToNodeMap.put(
+        "N_NATIONKEY",
+        new RexInputRef(0, validator.getTypeFactory().createSqlType(SqlTypeName.BIGINT)));
+    nameToNodeMap.put(
+        "N_REGIONKEY",
+        new RexInputRef(1, validator.getTypeFactory().createSqlType(SqlTypeName.BIGINT)));
+    final Map<String, RelDataType> nameToTypeMap = new HashMap<>();
+    for (Map.Entry<String, RexNode> entry : nameToNodeMap.entrySet()) {
+      nameToTypeMap.put(entry.getKey(), entry.getValue().getType());
+    }
+    Result result = new Result(nameToNodeMap, nameToTypeMap);
+    return result;
+  }
+
+  private @Desugar record Result(
+      HashMap<String, RexNode> nameToNodeMap, Map<String, RelDataType> nameToTypeMap) {}
 
   @VisibleForTesting
   SqlToRelConverter createSqlToRelConverter(
--- a/isthmus/src/test/java/io/substrait/isthmus/ExtendedExpressionTestBase.java
+++ b/isthmus/src/test/java/io/substrait/isthmus/ExtendedExpressionTestBase.java
@@ -1,0 +1,51 @@
+package io.substrait.isthmus;
+
+import com.google.common.base.Charsets;
+import com.google.common.io.Resources;
+import com.google.protobuf.InvalidProtocolBufferException;
+import com.google.protobuf.util.JsonFormat;
+import io.substrait.proto.ExtendedExpression;
+import java.io.IOException;
+import java.util.Arrays;
+import java.util.List;
+import org.apache.calcite.sql.parser.SqlParseException;
+
+public class ExtendedExpressionTestBase {
+  public static String asString(String resource) throws IOException {
+    return Resources.toString(Resources.getResource(resource), Charsets.UTF_8);
+  }
+
+  public static List<String> tpchSchemaCreateStatements() throws IOException {
+    String[] values = asString("tpch/schema.sql").split(";");
+    return Arrays.stream(values)
+        .filter(t -> !t.trim().isBlank())
+        .collect(java.util.stream.Collectors.toList());
+  }
+
+  protected ExtendedExpression assertProtoExtendedExpressionRoundrip(String query)
+      throws IOException, SqlParseException {
+    return assertProtoExtendedExpressionRoundrip(query, new SqlToSubstrait());
+  }
+
+  protected ExtendedExpression assertProtoExtendedExpressionRoundrip(String query, SqlToSubstrait s)
+      throws IOException, SqlParseException {
+    return assertProtoExtendedExpressionRoundrip(query, s, tpchSchemaCreateStatements());
+  }
+
+  protected ExtendedExpression assertProtoExtendedExpressionRoundrip(
+      String query, SqlToSubstrait s, List<String> creates) throws SqlParseException {
+    io.substrait.proto.ExtendedExpression protoExtendedExpression =
+        s.executeExpression(query, creates);
+
+    try {
+      String ee = JsonFormat.printer().print(protoExtendedExpression);
+      System.out.println("Proto Extended Expression: \n" + ee);
+
+      // FIXME! Implement test validation as the same as proto Plan implementation
+    } catch (InvalidProtocolBufferException e) {
+      throw new RuntimeException(e);
+    }
+
+    return protoExtendedExpression;
+  }
+}
--- a/isthmus/src/test/java/io/substrait/isthmus/SimpleExtendedExpressionsTest.java
+++ b/isthmus/src/test/java/io/substrait/isthmus/SimpleExtendedExpressionsTest.java
@@ -1,0 +1,13 @@
+package io.substrait.isthmus;
+
+import java.io.IOException;
+import org.apache.calcite.sql.parser.SqlParseException;
+import org.junit.jupiter.api.Test;
+
+public class SimpleExtendedExpressionsTest extends ExtendedExpressionTestBase {
+
+  @Test
+  public void filter() throws IOException, SqlParseException {
+    assertProtoExtendedExpressionRoundrip("N_NATIONKEY > 18");
+  }
+}
--- a/core/src/main/java/io/substrait/extension/ExtensionCollector.java
+++ b/core/src/main/java/io/substrait/extension/ExtensionCollector.java
@@ -1,5 +1,6 @@
 package io.substrait.extension;
 
+import io.substrait.proto.ExtendedExpression;
 import io.substrait.proto.Plan;
 import io.substrait.proto.SimpleExtensionDeclaration;
 import io.substrait.proto.SimpleExtensionURI;
@@ -51,6 +52,54 @@ public class ExtensionCollector extends AbstractExtensionLookup {
   }
 
   public void addExtensionsToPlan(Plan.Builder builder) {
+    var uriPos = new AtomicInteger(1);
+    var uris = new HashMap<String, SimpleExtensionURI>();
+
+    var extensionList = new ArrayList<SimpleExtensionDeclaration>();
+    for (var e : funcMap.forwardMap.entrySet()) {
+      SimpleExtensionURI uri =
+          uris.computeIfAbsent(
+              e.getValue().namespace(),
+              k ->
+                  SimpleExtensionURI.newBuilder()
+                      .setExtensionUriAnchor(uriPos.getAndIncrement())
+                      .setUri(k)
+                      .build());
+      var decl =
+          SimpleExtensionDeclaration.newBuilder()
+              .setExtensionFunction(
+                  SimpleExtensionDeclaration.ExtensionFunction.newBuilder()
+                      .setFunctionAnchor(e.getKey())
+                      .setName(e.getValue().key())
+                      .setExtensionUriReference(uri.getExtensionUriAnchor()))
+              .build();
+      extensionList.add(decl);
+    }
+    for (var e : typeMap.forwardMap.entrySet()) {
+      SimpleExtensionURI uri =
+          uris.computeIfAbsent(
+              e.getValue().namespace(),
+              k ->
+                  SimpleExtensionURI.newBuilder()
+                      .setExtensionUriAnchor(uriPos.getAndIncrement())
+                      .setUri(k)
+                      .build());
+      var decl =
+          SimpleExtensionDeclaration.newBuilder()
+              .setExtensionType(
+                  SimpleExtensionDeclaration.ExtensionType.newBuilder()
+                      .setTypeAnchor(e.getKey())
+                      .setName(e.getValue().key())
+                      .setExtensionUriReference(uri.getExtensionUriAnchor()))
+              .build();
+      extensionList.add(decl);
+    }
+
+    builder.addAllExtensionUris(uris.values());
+    builder.addAllExtensions(extensionList);
+  }
+
+  public void addExtensionsToExtendedExpression(ExtendedExpression.Builder builder) {
     var uriPos = new AtomicInteger(1);
     var uris = new HashMap<String, SimpleExtensionURI>();
 


isthmus/src/test/java/io/substrait/isthmus/SimpleExtendedExpressionsTest.java
```diff
@@ -0,0 +1,13 @@
...
+
+import java.io.IOException;
+import org.apache.calcite.sql.parser.SqlParseException;
+import org.junit.jupiter.api.Test;
+
+public class SimpleExtendedExpressionsTest extends ExtendedExpressionTestBase {
+
+  @Test
+  public void filter() throws IOException, SqlParseException {
+    assertProtoExtendedExpressionRoundrip("N_NATIONKEY > 18");
```
vbarua, line 11: I'm curious about this. Does the Calcite let you parse expressions by themselves outside the context of a query?

davisusanibar: This test `assertProtoExtendedExpressionRoundrip ` internally pass "the expression + tpchSchemaCreateStatements"

Then, there is a documented parse expression for example for [expr = "1 in (1,2,3)"](https://github.com/apache/calcite/blob/dc3f11bf9b8c7814eaaf7cbbf96a24b0c2b50456/core/src/test/java/org/apache/calcite/test/SqlToRelConverterTest.java#L4932-L4935).

It is also possible to parse expression passing internal definition using  [Map<String, RexNode> nameToNodeMap](https://github.com/apache/calcite/blob/c6031ca15579b88bd392a56f4588fdd9a29d1071/core/src/main/java/org/apache/calcite/sql2rel/SqlToRelConverter.java#L2153-L2155), ... my current approach is that I [just created that manually](https://github.com/substrait-io/substrait-java/blob/333989ffa70efc18c6707c3cf670bac826925fe1/isthmus/src/main/java/io/substrait/isthmus/SqlToSubstrait.java#L286-L301), my next step is to figure out how to populate/create `Map<String, RexNode> nameToNodeMap]` dynamically using schema definition provided.

```
  /**
   * Converts an expression from {@link SqlNode} to {@link RexNode} format,
   * mapping identifier references to predefined expressions.
   *
   * @param node          Expression to translate
   * @param nameToNodeMap map from String to {@link RexNode}; when an
   *                      {@link SqlIdentifier} is encountered, it is used as a
   *                      key and translated to the corresponding value from
   *                      this map
   * @return Converted expression
   */
  public RexNode convertExpression(
      SqlNode node,
      Map<String, RexNode> nameToNodeMap) {
```

isthmus/src/main/java/io/substrait/isthmus/SqlToSubstrait.java
```diff
@@ -1,18 +1,37 @@
...
+import com.github.bsideup.jabel.Desugar;
 import com.google.common.annotations.VisibleForTesting;
 import io.substrait.extension.ExtensionCollector;
+import io.substrait.extension.SimpleExtension;
+import io.substrait.isthmus.expression.RexExpressionConverter;
+import io.substrait.isthmus.expression.ScalarFunctionConverter;
+import io.substrait.proto.Expression;
+import io.substrait.proto.Expression.ScalarFunction;
+import io.substrait.proto.ExpressionReference;
+import io.substrait.proto.ExtendedExpression;
```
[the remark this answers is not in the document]
davisusanibar: Thank you @danepitkin for your review. Just adding an additional task to creare extended expression POJO classes.

.pre-commit-config.yaml
```diff
@@ -1,11 +1,11 @@
 repos:
 - repo: https://github.com/adrienverge/yamllint.git
-  rev: v1.26.0
```
vibhatha, line 3: @davisusanibar any specific reason why this upgrade is included in this PR?

isthmus/src/main/java/io/substrait/isthmus/SqlConverterBase.java
```diff
@@ -1,11 +1,11 @@
 package io.substrait.isthmus;
 
+import com.github.bsideup.jabel.Desugar;
 import io.substrait.extension.SimpleExtension;
 import io.substrait.isthmus.calcite.SubstraitOperatorTable;
 import io.substrait.type.NamedStruct;
 import java.io.IOException;
-import java.util.ArrayList;
-import java.util.List;
```
vibhatha, line 8: @davisusanibar I think it is better to minimize `.*` type imports.

isthmus/src/main/java/io/substrait/isthmus/SqlToSubstrait.java
```diff
@@ -1,18 +1,37 @@
...
 import io.substrait.proto.PlanRel;
+import io.substrait.proto.SimpleExtensionDeclaration;
+import io.substrait.proto.SimpleExtensionURI;
 import io.substrait.relation.RelProtoConverter;
 import io.substrait.type.NamedStruct;
-import java.util.List;
+import io.substrait.type.TypeCreator;
+import io.substrait.type.proto.TypeProtoConverter;
+import java.io.IOException;
+import java.util.*;
```
vibhatha, line 23: same here.

isthmus/src/test/java/io/substrait/isthmus/integration/ExtendedExpressionIntegrationTest.java
```diff
@@ -0,0 +1,162 @@
...
+
+  private static ByteBuffer getFilterExtendedExpression(String sqlExpression)
+      throws IOException, SqlParseException {
+    ExtendedExpression extendedExpression =
+        new SqlToSubstrait()
+            .executeSQLExpression(
+                sqlExpression, ExtendedExpressionTestBase.tpchSchemaCreateStatements());
+    System.out.println(
+        "JsonFormat.printer().print(getFilterExtendedExpression): "
+            + JsonFormat.printer().print(extendedExpression));
```
vibhatha, line 137: Shall we remove the these? Or maybe use a more useful text?

isthmus/src/test/java/io/substrait/isthmus/integration/ExtendedExpressionIntegrationTest.java
```diff
@@ -0,0 +1,162 @@
...
+        System.out.println(reader.getVectorSchemaRoot().contentToTSVString());
+      }
+      assertEquals(4, count);
+    } catch (Exception e) {
+      e.printStackTrace();
+      throw new RuntimeException(e);
+    }
+  }
+
+  private static ByteBuffer getFilterExtendedExpression(String sqlExpression)
```
vibhatha, line 129: Do we need separate methods like `getFilterExtendedExpression` and `getProjectExtendedExpression` ? 
Isn't the functionality same in both functions? Please correct me if I am wrong.

isthmus/src/test/java/io/substrait/isthmus/ExtendedExpressionTestBase.java
```diff
@@ -0,0 +1,51 @@
...
+  }
+
+  protected ExtendedExpression assertProtoExtendedExpressionRoundrip(
+      String query, SqlToSubstrait s, List<String> creates) throws SqlParseException {
+    io.substrait.proto.ExtendedExpression protoExtendedExpression =
+        s.executeSQLExpression(query, creates);
+
+    try {
+      String ee = JsonFormat.printer().print(protoExtendedExpression);
+      System.out.println("Proto Extended Expression: \n" + ee);
```
vibhatha, line 42: Do we need this println?

isthmus/src/test/java/io/substrait/isthmus/ExtendedExpressionTestBase.java
```diff
@@ -0,0 +1,51 @@
...
+
+      // FIXME! Implement test validation as the same as proto Plan implementation
```
vibhatha, line 44: I assume this is probably a todo?

isthmus/src/main/java/io/substrait/isthmus/SqlToSubstrait.java
```diff
@@ -39,19 +64,39 @@ public Plan execute(String sql, Function<List<String>, NamedStruct> tableLookup)
   }
 
   public Plan execute(String sql, List<String> tables) throws SqlParseException {
-    var pair = registerCreateTables(tables);
-    return executeInner(sql, factory, pair.left, pair.right);
+    var result = registerCreateTables(tables);
+    return executeInner(sql, factory, result.validator(), result.catalogReader());
```
vibhatha, line 68: @davisusanibar what is the motivation for this change? I feel like the previous was okay, but would like to know your thoughts.

isthmus/src/main/java/io/substrait/isthmus/SqlToSubstrait.java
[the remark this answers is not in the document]
davisusanibar: Hi @danepitkin PR just added for POJO extended expressions https://github.com/substrait-io/substrait-java/pull/206

.pre-commit-config.yaml
davisusanibar: My problem with commitlint-pre-commit-hook was resolved by updating to the last version and applying the same update to yamlint to keep the version updated.

isthmus/src/main/java/io/substrait/isthmus/SqlConverterBase.java
davisusanibar: Changed

isthmus/src/main/java/io/substrait/isthmus/SqlToSubstrait.java
davisusanibar: This case, I defer to `./gradlew spotlessApply`

isthmus/src/test/java/io/substrait/isthmus/integration/ExtendedExpressionIntegrationTest.java
```diff
@@ -0,0 +1,162 @@
...
+                NativeMemoryPool.getDefault(),
+                FileFormat.PARQUET,
+                resource.toURI().toString());
+        Dataset dataset = datasetFactory.finish();
+        Scanner scanner = dataset.newScan(options);
+        ArrowReader reader = scanner.scanBatches()) {
+      int count = 0;
+      while (reader.loadNextBatch()) {
+        count += reader.getVectorSchemaRoot().getRowCount();
+        System.out.println(reader.getVectorSchemaRoot().contentToTSVString());
```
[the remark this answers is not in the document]
davisusanibar: Deleted

isthmus/src/test/java/io/substrait/isthmus/integration/ExtendedExpressionIntegrationTest.java
davisusanibar: Deletef

isthmus/src/test/java/io/substrait/isthmus/integration/ExtendedExpressionIntegrationTest.java
davisusanibar: Thanks, now just maintain one method.

isthmus/src/test/java/io/substrait/isthmus/ExtendedExpressionTestBase.java
davisusanibar: Deleted

isthmus/src/test/java/io/substrait/isthmus/ExtendedExpressionTestBase.java
davisusanibar: Just implemented at https://github.com/substrait-io/substrait-java/pull/206

isthmus/src/main/java/io/substrait/isthmus/SqlToSubstrait.java
davisusanibar: I just rolled back.

davisusanibar:
The PR is ready for review. It therefore depends on https://github.com/substrait-io/substrait-java/pull/206

isthmus/src/main/java/io/substrait/isthmus/SqlToSubstrait.java
vbarua: spotless doesn't actually squash imports down like this. What IDE are you using? I know with IntelliJ it defaults to doing this when you autoimport. There's a way to turn it [off](https://stackoverflow.com/questions/3348816/intellij-never-use-wildcard-imports)

isthmus/src/main/java/io/substrait/isthmus/TypeConverter.java
```diff
@@ -56,6 +58,17 @@ public NamedStruct toNamedStruct(RelDataType type) {
     return NamedStruct.of(names, struct);
   }
 
+  public NamedStruct toNamedStruct(Map<String, RelDataType> nameToTypeMap) {
+    var names = Lists.<String>newArrayList();
+    var types = Lists.<Type>newArrayList();
```
vbarua, line 63: Can you use the ArrayList constructor directly?
```java
var names = new ArrayList<String>
```

The docs for the method state
```
Note: this method is now unnecessary and should be treated as deprecated. Instead, use the ArrayList [constructor](https://docs.oracle.com/javase/9/docs/api/java/util/ArrayList.html?is-external=true#%3Cinit%3E()) directly, taking advantage of ["diamond" syntax](http://goo.gl/iz2Wi).
```

isthmus/src/main/java/io/substrait/isthmus/SqlToSubstrait.java
```diff
@@ -48,6 +62,26 @@ public Plan execute(String sql, String name, Schema schema) throws SqlParseExcep
...
+  public ExtendedExpression executeSQLExpression(String sqlExpression, List<String> tables)
+      throws SqlParseException {
+    var result = registerCreateTablesForExtendedExpression(tables);
+    return executeInnerSQLExpression(
+        sqlExpression,
+        result.validator(),
+        result.catalogReader(),
+        result.nameToTypeMap(),
+        result.nameToNodeMap());
+  }
```
vbarua, line 83: What do you think about moving all of the logic associated with converting SQL _expressions_ into a dedicated class like `SqlExpressionToSubstrait`? That way each class has one responsibility and folks looking to convert sql or sql expressions can just look at the specific class for the conversion they need to understand what's happening under the hood.

davisusanibar: Changed

isthmus/src/main/java/io/substrait/isthmus/TypeConverter.java
davisusanibar: Changed

core/src/main/java/io/substrait/extendedexpression/ExtendedExpressionProtoConverter.java
```diff
@@ -0,0 +1,66 @@
...
+
+    for (io.substrait.extendedexpression.ExtendedExpression.ExpressionReference
+        expressionReference : extendedExpression.getReferredExpressions()) {
+      io.substrait.extendedexpression.ExtendedExpression.ExpressionTypeReference expressionType =
+          expressionReference.getExpressionType();
+      if (expressionType
+          instanceof io.substrait.extendedexpression.ExtendedExpression.ExpressionType) {
+        io.substrait.proto.Expression expressionProto =
+            expressionProtoConverter.visit(
+                (Expression.ScalarFunctionInvocation)
```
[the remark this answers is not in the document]
davisusanibar: Hi @vbarua ,

Just added changes to be more generic. Also added unit test to evidence that.

isthmus/src/main/java/io/substrait/isthmus/SqlExpressionToSubstrait.java
```diff
@@ -0,0 +1,114 @@
...
+
+  protected SqlExpressionToSubstrait(FeatureBoard features) {
+    super(features);
+  }
+
+  private final ScalarFunctionConverter functionConverter =
+      new ScalarFunctionConverter(EXTENSION_COLLECTION.scalarFunctions(), factory);
+
+  private final RexExpressionConverter rexExpressionConverter =
+      new RexExpressionConverter(functionConverter);
```
vbarua, line 38: Small API consideration for a little future-proofing. I don't recommend hardcoding the `EXTENSION_COLLECTION` like this because it makes it difficult for users to provide their own extensions and/or manipulate existing ones (which is a limitation of the current `SqlToSubstrait` code).

What I recommend is something like
```java
    protected final RexExpressionConverter rexConverter;

        this(FEATURES_DEFAULT, EXTENSION_COLLECTION);
    }

    public SqlExpressionToSubstrait(
            FeatureBoard features,
            SimpleExtension.ExtensionCollection extensions) {
        super(features);
        var scalarFunctionConverter = new ScalarFunctionConverter(extensions.scalarFunctions(), factory);
        this.rexConverter = new RexExpressionConverter(scalarFunctionConverter);
    }

```
which gives users a bit more control over extensions.

isthmus/src/main/java/io/substrait/isthmus/TypeConverter.java
```diff
@@ -56,6 +57,17 @@ public NamedStruct toNamedStruct(RelDataType type) {
     return NamedStruct.of(names, struct);
   }
 
+  public NamedStruct toNamedStruct(Map<String, RelDataType> nameToTypeMap) {
```
vbarua, line 60: I'm not sure if it's a good idea expose a public API like this. Because this takes a Map, the order of the fields in the final named struct is dependent on the iteration order of the Map.

isthmus/src/main/java/io/substrait/isthmus/SqlExpressionToSubstrait.java
```diff
@@ -0,0 +1,114 @@
...
+  /**
+   * Process to execute an SQL Expression to convert into an Extended expression protobuf message
+   *
+   * @param sqlExpression expression defined by the user
+   * @param tables of names of table needed to consider to load into memory for catalog, schema,
+   *     validate and parse sql
+   * @return extended expression protobuf message
+   * @throws SqlParseException
+   */
+  public ExtendedExpression executeSQLExpression(String sqlExpression, List<String> tables)
```
vbarua, line 49: What do you think about simply calling this method `convert`?

Strictly speaking it's not executing the SQL expression, and because it's now a method of the `SqlExpressionToSubstrait` class I think it would be redundant to call it `convertSqlExpression`.

Additionally, what do you think about changing `List<String> tables` to `List<String> createStatements` to better indicate the kinds of string we are expecting.

isthmus/src/main/java/io/substrait/isthmus/SqlConverterBase.java
```diff
@@ -103,6 +110,39 @@ Pair<SqlValidator, CalciteCatalogReader> registerCreateTables(List<String> table
...
+        new CalciteCatalogReader(rootSchema, List.of(), factory, config);
+    SqlValidator validator = Validator.create(factory, catalogReader, SqlValidator.Config.DEFAULT);
+    if (tables != null) {
+      for (String tableDef : tables) {
+        List<DefinedTable> tList = parseCreateTable(factory, validator, tableDef);
+        for (DefinedTable t : tList) {
+          rootSchema.add(t.getName(), t);
+          for (RelDataTypeField field : t.type.getFieldList()) {
+            nameToTypeMap.put(
+                field.getName(), field.getType()); // to validate the sql expression tree
```
vbarua, line 127: Does this handle multiple tables containing a field with the same name?

isthmus/src/main/java/io/substrait/isthmus/SqlConverterBase.java
```diff
@@ -103,6 +110,39 @@ Pair<SqlValidator, CalciteCatalogReader> registerCreateTables(List<String> table
...
+            nameToNodeMap.put(
```
vbarua, line 128: Does this handle multiple tables containing a field with the same name?

isthmus/src/main/java/io/substrait/isthmus/SqlExpressionToSubstrait.java
```diff
@@ -0,0 +1,114 @@
...
+
+  private RexNode sqlToRexNode(
+      String sql,
+      SqlValidator validator,
+      CalciteCatalogReader catalogReader,
+      Map<String, RelDataType> nameToTypeMap,
+      Map<String, RexNode> nameToNodeMap)
+      throws SqlParseException {
+    SqlParser parser = SqlParser.create(sql, parserConfig);
+    SqlNode sqlNode = parser.parseExpression();
```
vbarua, line 97: I did not know you could re-use the Calcite expression parser like this. That's handy!

isthmus/src/main/java/io/substrait/isthmus/SqlExpressionToSubstrait.java
```diff
@@ -0,0 +1,114 @@
...
+    SqlNode validSQLNode = validator.validateParameterizedExpression(sqlNode, nameToTypeMap);
+    SqlToRelConverter converter = createSqlToRelConverter(validator, catalogReader);
+    return converter.convertExpression(validSQLNode, nameToNodeMap);
+  }
+
+  @VisibleForTesting
```
vbarua, line 103: We don't seem to be using for testing. Can we remove the annotation?

vbarua (commented):
Not a full review yet, but I had some time for a first pass.

isthmus/src/main/java/io/substrait/isthmus/SqlConverterBase.java
davisusanibar: I'll test this scenario

isthmus/src/main/java/io/substrait/isthmus/SqlConverterBase.java
```diff
@@ -103,6 +110,39 @@ Pair<SqlValidator, CalciteCatalogReader> registerCreateTables(List<String> table
     return Pair.of(validator, catalogReader);
   }
 
+  Result registerCreateTablesForExtendedExpression(List<String> tables) throws SqlParseException {
```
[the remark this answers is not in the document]
davisusanibar: Changed

isthmus/src/main/java/io/substrait/isthmus/SqlConverterBase.java
```diff
@@ -103,6 +110,39 @@ Pair<SqlValidator, CalciteCatalogReader> registerCreateTables(List<String> table
...
+                    field.getIndex(), field.getType())); // to convert sql expression into RexNode
+          }
+        }
+      }
+    }
+    return new Result(validator, catalogReader, nameToTypeMap, nameToNodeMap);
+  }
+
+  @Desugar
+  public record Result(
```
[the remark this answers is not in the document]
davisusanibar: Changed

isthmus/src/main/java/io/substrait/isthmus/SqlExpressionToSubstrait.java
```diff
@@ -0,0 +1,114 @@
...
+import org.apache.calcite.sql.parser.SqlParseException;
+import org.apache.calcite.sql.parser.SqlParser;
+import org.apache.calcite.sql.validate.SqlValidator;
+import org.apache.calcite.sql2rel.SqlToRelConverter;
+import org.apache.calcite.sql2rel.StandardConvertletTable;
+
+public class SqlExpressionToSubstrait extends SqlConverterBase {
+
+  public SqlExpressionToSubstrait() {
+    this(null);
```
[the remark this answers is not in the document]
davisusanibar: Changed

isthmus/src/main/java/io/substrait/isthmus/SqlExpressionToSubstrait.java
davisusanibar: Changed

isthmus/src/main/java/io/substrait/isthmus/TypeConverter.java
davisusanibar: Changed

isthmus/src/main/java/io/substrait/isthmus/SqlExpressionToSubstrait.java
davisusanibar: Changed

isthmus/src/main/java/io/substrait/isthmus/SqlExpressionToSubstrait.java
davisusanibar: Changed

isthmus/src/main/java/io/substrait/isthmus/SqlConverterBase.java
davisusanibar: This scenario is not supported. It will be updated with the last filed type declared. Calcite API is not offering an option to mark field per this table if it is duplicated.

````
  /**
   * Converts an expression from {@link SqlNode} to {@link RexNode} format,
   * mapping identifier references to predefined expressions.
   *
   * @param node          Expression to translate
   * @param nameToNodeMap map from String to {@link RexNode}; when an
   *                      {@link SqlIdentifier} is encountered, it is used as a
   *                      key and translated to the corresponding value from
   *                      this map
   * @return Converted expression
   */
  public RexNode convertExpression(
      SqlNode node,
      Map<String, RexNode> nameToNodeMap) {
````

**Comment by davisusanibar** [comment]
Hi @vbarua thank you for the code review, the comments were addressed.

Additionally please if you could give some reason of the current error on the CI build (for some reason I am not able to replicate that locally):

````
> Task :core:generateTestProto NO-SOURCE

/home/runner/work/substrait-java/substrait-java/core/src/test/java/io/substrait/type/proto/TestTypeRoundtrip.java:54: warning: as of release 10, 'var' is a restricted type name and cannot be used for type declarations or as the element type of an array
    var converted = type.accept(typeProtoConverter);
        ^
> Task :core:compileTestJava FAILED
/home/runner/work/substrait-java/substrait-java/core/src/test/java/io/substrait/type/proto/LocalFilesRoundtripTest.java:24: warning: as of release 10, 'var' is a restricted type name and cannot be used for type declarations or as the element type of an array
:core:compileTestJava failure marker
````

**Comment by vbarua** [comment]
> Additionally please if you could give some reason of the current error on the CI build

I'm not entirely sure either unfortunately.  In some case Jabel doesn't seem to fire correctly and it doesn't transform code that it should be. It happens more than I would like, see https://github.com/substrait-io/substrait-java/issues/195

I tried making a small change in https://github.com/substrait-io/substrait-java/pull/191/commits/3835a4f35d41ac31dc9dfcd9d0fb4876602a002e to retry the build, which sometimes works, but apparently not in this case. If the CI check is failing after we've approved the PR we can figure it out then.

File: isthmus/src/main/java/io/substrait/isthmus/SqlConverterBase.java (hunk 19 above)
Comment (vbarua) [reply isthmus/src/main/java/io/substrait/isthmus/SqlConverterBase.java]: If Calcite can't handle this, then we should throw an exception if we try to build either of those maps with duplicate values. I think it's better to explode loudly than to have a last-write wins style validator.

Comment (vbarua) [reply isthmus/src/main/java/io/substrait/isthmus/SqlConverterBase.java]: One thing to check, does it work with qualified paths? Like if you have:
```sql
CREATE TABLE foo(a INT);
```
and you put the fully qualified identifier `foo.a` as the map key, can the expression `a` find it?

Then going further if you had
```sql
CREATE TABLE foo(a INT);
CREATE TABLE bar(a INT);
```
does Calcite reject the expression `a` as being ambiguous, but accept `foo.a`?

davisusanibar: I will add that validation and also test the last scenario you mentioned.

isthmus/build.gradle.kts
```diff
@@ -94,6 +95,8 @@ dependencies {
   implementation("org.immutables:value-annotations:2.8.8")
   annotationProcessor("org.immutables:value:2.8.8")
   testImplementation("org.apache.calcite:calcite-plus:${CALCITE_VERSION}")
+  testImplementation("org.apache.arrow:arrow-dataset:${ARROW_VERSION}")
+  testImplementation("org.apache.arrow:arrow-memory-netty:${ARROW_VERSION}")
```
danepitkin, line 99: This is a great integration test, but I think it would be better if we moved it to the Arrow repo instead of adding Arrow as a dependency to Isthmus. What do you think?

isthmus/src/main/java/io/substrait/isthmus/SqlExpressionToSubstrait.java
```diff
@@ -0,0 +1,165 @@
...
+      Map<String, RexNode> nameToNodeMap)
+      throws SqlParseException {
+    RexNode rexNode =
+        sqlToRexNode(sqlExpression, validator, catalogReader, nameToTypeMap, nameToNodeMap);
+    NamedStruct namedStruct = toNamedStruct(nameToTypeMap);
+
+    ImmutableExpressionReference expressionReference =
+        ImmutableExpressionReference.builder()
+            .expression(rexNode.accept(this.rexConverter))
+            .addOutputNames("new-column")
```
vibhatha, line 87: @davisusanibar just for the clarity, 
Should this parameter come from the user input or is this a temporary placeholder which would get updated later?

isthmus/src/main/java/io/substrait/isthmus/SqlExpressionToSubstrait.java
```diff
@@ -0,0 +1,165 @@
...
+      SqlValidator validator,
+      CalciteCatalogReader catalogReader,
+      Map<String, RelDataType> nameToTypeMap,
+      Map<String, RexNode> nameToNodeMap)
+      throws SqlParseException {
+    SqlParser parser = SqlParser.create(sql, parserConfig);
+    SqlNode sqlNode = parser.parseExpression();
+    SqlNode validSQLNode = validator.validateParameterizedExpression(sqlNode, nameToTypeMap);
+    SqlToRelConverter converter = createSqlToRelConverter(validator, catalogReader);
+    return converter.convertExpression(validSQLNode, nameToNodeMap);
```
vibhatha, line 113: nit: 

```suggestion
    SqlNode validSqlNode = validator.validateParameterizedExpression(sqlNode, nameToTypeMap);
    SqlToRelConverter converter = createSqlToRelConverter(validator, catalogReader);
    return converter.convertExpression(validSqlNode, nameToNodeMap);
```

isthmus/src/main/java/io/substrait/isthmus/SqlExpressionToSubstrait.java
```diff
@@ -0,0 +1,165 @@
...
+  SqlToRelConverter createSqlToRelConverter(
+      SqlValidator validator, CalciteCatalogReader catalogReader) {
+    return new SqlToRelConverter(
+        null,
+        validator,
+        catalogReader,
+        relOptCluster,
+        StandardConvertletTable.INSTANCE,
+        converterConfig);
+  }
```
vibhatha, line 125: nit: 
Do we need this method?

isthmus/src/main/java/io/substrait/isthmus/SqlExpressionToSubstrait.java
```diff
@@ -0,0 +1,165 @@
...
+
+  private Result registerCreateTablesForExtendedExpression(List<String> tables)
+      throws SqlParseException {
+    Map<String, RelDataType> nameToTypeMap = new LinkedHashMap<>();
+    Map<String, RexNode> nameToNodeMap = new HashMap<>();
```
vibhatha, line 130: @davisusanibar just for my curiosity, does the order matter for `nameToTypeMap`?

isthmus/src/main/java/io/substrait/isthmus/SqlExpressionToSubstrait.java
```diff
@@ -0,0 +1,165 @@
...
+    return new SqlToRelConverter(
+        null,
+        validator,
+        catalogReader,
+        relOptCluster,
+        StandardConvertletTable.INSTANCE,
+        converterConfig);
+  }
+
+  private Result registerCreateTablesForExtendedExpression(List<String> tables)
```
vibhatha, line 127: What would happen if we pass null here?

vibhatha (commented):
@davisusanibar added a few comments and a few questions for my curiosity. 
Nice work!!!

isthmus/src/main/java/io/substrait/isthmus/SqlConverterBase.java
davisusanibar: Hi Team,

I am trying to implement this, but I am seeing limitations on the current Calcite implementation of SQL Expressions for qualified paths. There are two issues I would like to register:

1. Is needed to add support for qualified path at `SqlNode validSQLNode = validator.validateParameterizedExpression(sqlNode, nameToTypeMap);`
2. Is also needed to support for qualified path at `converter.convertExpression(validSQLNode, nameToNodeMap);`

davisusanibar: Just filled (1) https://issues.apache.org/jira/browse/CALCITE-6195.

vbarua: Thanks for looking into this! If qualified paths aren't supported Calcite, then I don't think we should concern ourselves with them here. I think we still need to deal with ambiguous lookups, probably by rejecting them if there is no mechanism to specify one or the other with a qualified path.

isthmus/src/main/java/io/substrait/isthmus/SqlExpressionToSubstrait.java
davisusanibar: Let me test that

isthmus/src/main/java/io/substrait/isthmus/SqlExpressionToSubstrait.java
davisusanibar: Let me remember why I did it that way.

isthmus/src/main/java/io/substrait/isthmus/SqlExpressionToSubstrait.java
davisusanibar: Yes, it is needed.

isthmus/src/main/java/io/substrait/isthmus/SqlExpressionToSubstrait.java
davisusanibar: Certainly, this is intentionally assumed as a column name by default, but it could be changed in the future to be defined by the user. Let's wait and see how it is used.

isthmus/src/main/java/io/substrait/isthmus/SqlConverterBase.java
davisusanibar: Just added a logic to handle illegal argument for ambiguous lookups and also added unit test to validate that.

vibhatha:
> > Additionally please if you could give some reason of the current error on the CI build
> 
> I'm not entirely sure either unfortunately. In some case Jabel doesn't seem to fire correctly and it doesn't transform code that it should be. It happens more than I would like, see #195
> 
> I tried making a small change in [3835a4f](https://github.com/substrait-io/substrait-java/commit/3835a4f35d41ac31dc9dfcd9d0fb4876602a002e) to retry the build, which sometimes works, but apparently not in this case. If the CI check is failing after we've approved the PR we can figure it out then.

@vbarua would be a best practice to keep the usage of higher SDK features minimal and stick to basics at most cases?

isthmus/src/main/java/io/substrait/isthmus/SqlExpressionToSubstrait.java
davisusanibar: Okay, It is necessary to maintain the order of the inserted elements for `nameToTypeMap`, because we use that to recreate NamedStruct at `NamedStruct namedStruct = toNamedStruct(nameToTypeMap);`.

isthmus/build.gradle.kts
davisusanibar: Ok, let me delete that part.

isthmus/src/main/java/io/substrait/isthmus/SqlExpressionToSubstrait.java
davisusanibar: It may be possible, but an exception like this will be thrown: `org.apache.calcite.runtime.CalciteContextException: From line 1, column 1 to line 1, column 10: Unknown identifier 'L_ORDERKEY'`

isthmus/build.gradle.kts
davisusanibar: Just deleted

vibhatha:
> PD: PR related to [support qualified path on Calcite SQL Expression](https://github.com/apache/calcite/pull/3621) was created.
:fireworks: :+1:

isthmus/src/test/java/io/substrait/isthmus/ExtendedExpressionTestBase.java
```diff
@@ -0,0 +1,72 @@
...
+    return Arrays.stream(values)
+        .filter(t -> !t.trim().isBlank())
+        .collect(java.util.stream.Collectors.toList());
+  }
+
+  public static List<String> tpchSchemaCreateStatements() throws IOException {
+    return tpchSchemaCreateStatements("tpch/schema.sql");
+  }
+
+  protected ExtendedExpression assertProtoExtendedExpressionRoundrip(String query)
```
vibhatha, line 30: nit:
```suggestion
  protected ExtendedExpression assertProtoExtendedExpressionRoundtrip(String query)
```

isthmus/src/test/java/io/substrait/isthmus/ExtendedExpressionTestBase.java
```diff
@@ -0,0 +1,72 @@
...
+      throws IOException, SqlParseException {
+    return assertProtoExtendedExpressionRoundrip(query, new SqlExpressionToSubstrait());
```
vibhatha, line 32: nit:
```suggestion
    return assertProtoExtendedExpressionRoundtrip(query, new SqlExpressionToSubstrait());
```

isthmus/src/test/java/io/substrait/isthmus/ExtendedExpressionTestBase.java
```diff
@@ -0,0 +1,72 @@
...
+  }
+
+  protected ExtendedExpression assertProtoExtendedExpressionRoundrip(
+      String query, String schemaToLoad) throws IOException, SqlParseException {
+    return assertProtoExtendedExpressionRoundrip(
+        query, new SqlExpressionToSubstrait(), schemaToLoad);
+  }
+
+  protected ExtendedExpression assertProtoExtendedExpressionRoundrip(
```
vibhatha, line 41: nit:
```suggestion
  protected ExtendedExpression assertProtoExtendedExpressionRoundtrip(
```

isthmus/src/test/java/io/substrait/isthmus/ExtendedExpressionTestBase.java
```diff
@@ -0,0 +1,72 @@
...
+  public static List<String> tpchSchemaCreateStatements() throws IOException {
+    return tpchSchemaCreateStatements("tpch/schema.sql");
+  }
+
+  protected ExtendedExpression assertProtoExtendedExpressionRoundrip(String query)
+      throws IOException, SqlParseException {
+    return assertProtoExtendedExpressionRoundrip(query, new SqlExpressionToSubstrait());
+  }
+
+  protected ExtendedExpression assertProtoExtendedExpressionRoundrip(
```
vibhatha, line 35: nit:
```suggestion
  protected ExtendedExpression assertProtoExtendedExpressionRoundtrip(
```

isthmus/src/test/java/io/substrait/isthmus/ExtendedExpressionTestBase.java
```diff
@@ -0,0 +1,72 @@
...
+    return assertProtoExtendedExpressionRoundrip(
+        query, new SqlExpressionToSubstrait(), schemaToLoad);
+  }
+
+  protected ExtendedExpression assertProtoExtendedExpressionRoundrip(
+      String query, SqlExpressionToSubstrait s) throws IOException, SqlParseException {
+    return assertProtoExtendedExpressionRoundrip(query, s, tpchSchemaCreateStatements());
+  }
+
+  protected ExtendedExpression assertProtoExtendedExpressionRoundrip(
```
vibhatha, line 46: nit:
```suggestion
  protected ExtendedExpression assertProtoExtendedExpressionRoundtrip(
```

isthmus/src/test/java/io/substrait/isthmus/ExtendedExpressionTestBase.java
```diff
@@ -0,0 +1,72 @@
...
+      String query, SqlExpressionToSubstrait s, String schemaToLoad)
+      throws IOException, SqlParseException {
+    return assertProtoExtendedExpressionRoundrip(
+        query, s, tpchSchemaCreateStatements(schemaToLoad));
+  }
+
+  protected ExtendedExpression assertProtoExtendedExpressionRoundrip(
```
vibhatha, line 53: nit:
```suggestion
  protected ExtendedExpression assertProtoExtendedExpressionRoundtrip(
```

isthmus/src/test/resources/tpch/schema_error.sql
```diff
@@ -0,0 +1,36 @@
...
+  L_TAX DECIMAL,
+  L_RETURNFLAG CHAR(1),
+  L_LINESTATUS CHAR(1),
+  L_SHIPDATE DATE,
+  L_COMMITDATE DATE,
+  L_RECEIPTDATE DATE,
+  L_SHIPINSTRUCT CHAR(25),
+  L_SHIPMODE CHAR(10),
+  L_COMMENT VARCHAR(44)
+);
```
vibhatha, line 36: nit: add a new line

vibhatha (commented):
@davisusanibar I added a few comments. Overall the PR LGTM!

vibhatha:
@vbarua should we cover more ground on tests? Not necessarily for this PR itself but for the sustainability in the longer term?

isthmus/src/test/resources/tpch/schema_error.sql
davisusanibar: Changed

isthmus/src/main/java/io/substrait/isthmus/SqlExpressionToSubstrait.java
davisusanibar: I have moved to the inline method. Thank you.

vibhatha (approved):
Thanks @davisusanibar, PR LGTM!

vbarua (approved):
I went ahead a pushed a commit to remove the parquet related data and config changes as they weren't needed anymore. 

Thanks for adding the duplicate column check, and thanks for working on this. Changes look good!

PR #193, opened 2023-11-02
test: move common logic into TestBase

vbarua:


vbarua:
Reducing test code duplication by moving more common logic into the TestBase abstract class.

core/src/test/java/io/substrait/type/proto/AggregateRoundtripTest.java
```diff
@@ -21,8 +21,6 @@
 import org.junit.jupiter.api.Test;
 
 public class AggregateRoundtripTest extends TestBase {
-  static final org.slf4j.Logger logger =
-      org.slf4j.LoggerFactory.getLogger(AggregateRoundtripTest.class);
```
vbarua, line 25: These loggers aren't used or needed so removing them to reduce noise.

core/src/test/java/io/substrait/type/proto/LocalFilesRoundtripTest.java
```diff
@@ -25,15 +25,7 @@
 
 public class LocalFilesRoundtripTest extends TestBase {
 
-  SimpleExtension.ExtensionCollection extensions = defaultExtensionCollection;
-
-  public LocalFilesRoundtripTest() throws IOException {}
```
vbarua, line 30: This constructor is not needed.

EpsilonPrime (approved):
Nice cleanup!

vbarua:
CI is failing due to
```
/home/runner/work/substrait-java/substrait-java/core/src/test/java/io/substrait/type/proto/LocalFilesRoundtripTest.java:61: error: switch expressions are not supported in -source 8
    return switch (pathTypeCase) {
           ^
  (use -source 14 or higher to enable switch expressions)
      case URI_PATH -> builder.pathType(FileOrFiles.PathType.URI_PATH).path("path");
```
which are in code this PR did not touch. Locally, tests run and build fine.

I've seen a number of builds fail with issues like this. In theory Jabel should be converting this code, but it doesn't seem to fire sometimes.

vbarua:
Filed https://github.com/substrait-io/substrait-java/issues/195 for the build issue.

PR #199, opened 2023-11-07
feat: exhaustive copy on write visitors

vbarua:
BREAKING CHANGE: RelCopyOnWriteVisitor now extends RelVisitor and has generic type parameter

### DIFF
--- a/core/src/main/java/io/substrait/relation/CopyOnWriteVisitor.java
+++ b/core/src/main/java/io/substrait/relation/CopyOnWriteVisitor.java
@@ -29,9 +29,10 @@ public abstract class CopyOnWriteVisitor<T, E extends Exception> {
   /**
    * Applies the given transformation function to each item in the list. If any of the list items
    * are transformed, returns a new list in which each item is either
+   *
    * <ul>
-   *     <li>a transformed new item replacing an old item</li>
-   *     <li>the original item in the position it was in</li>
+   *   <li>a transformed new item replacing an old item
+   *   <li>the original item in the position it was in
    * </ul>
    *
    * @param items the list of items to transform
--- a/core/src/main/java/io/substrait/relation/CopyOnWriteVisitor.java
+++ b/core/src/main/java/io/substrait/relation/CopyOnWriteVisitor.java
@@ -1,0 +1,57 @@
+package io.substrait.relation;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+import java.util.Optional;
+import java.util.function.Supplier;
+
+/** Base class for Copy On Write visitors. Provides common utils. */
+public abstract class CopyOnWriteVisitor<T, E extends Exception> {
+
+  protected interface TransformFunction<T, E extends Exception> {
+    Optional<T> apply(T t) throws E;
+  }
+
+  protected boolean allEmpty(Optional<?>... optionals) {
+    return Arrays.stream(optionals).noneMatch(Optional::isPresent);
+  }
+
+  /** The `or` method on Optional instances is a Java 9+ feature */
+  protected static <T> Optional<T> or(Optional<T> left, Supplier<? extends Optional<T>> right) {
+    if (left.isPresent()) {
+      return left;
+    } else {
+      return right.get();
+    }
+  }
+
+  /**
+   * Applies the given transformation function to each item in the list. If any of the list items
+   * are transformed, returns a new list in which each item is either
+   * <ul>
+   *     <li>a transformed new item replacing an old item</li>
+   *     <li>the original item in the position it was in</li>
+   * </ul>
+   *
+   * @param items the list of items to transform
+   * @param transform the transformation function to apply to each item
+   * @return An empty optional if none of the items have changed. An optional containing a new list
+   *     otherwise.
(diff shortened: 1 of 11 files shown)


core/src/main/java/io/substrait/relation/ExprCopyOnWriteVisitor.java
```diff
@@ -0,0 +1,366 @@
...
+    extends CopyOnWriteVisitor<Expression, EXCEPTION>
+    implements ExpressionVisitor<Optional<Expression>, EXCEPTION> {
+
+  private final RelCopyOnWriteVisitor<EXCEPTION> relCopyOnWriteVisitor;
+
+  public ExprCopyOnWriteVisitor(RelCopyOnWriteVisitor<EXCEPTION> relCopyOnWriteVisitor) {
+    this.relCopyOnWriteVisitor = relCopyOnWriteVisitor;
+  }
+
+  protected final RelCopyOnWriteVisitor<EXCEPTION> getRelVisitor() {
```
EpsilonPrime, line 20: Is it worth abbreviating relation here (we do elsewhere so it seems justified).  How about Expression in the name of this class?  That seems less justified.

core/src/main/java/io/substrait/relation/RelCopyOnWriteVisitor.java
```diff
@@ -22,90 +16,129 @@
...
   }
 
-  private static boolean allEmpty(Optional<?>... optionals) {
-    for (Optional<?> optional : optionals) {
-      if (optional.isPresent()) {
-        return false;
-      }
-    }
-    return true;
+  protected ExprCopyOnWriteVisitor<EXCEPTION> getExprVisitor() {
```
EpsilonPrime, line 38: Also consider here as well.

EpsilonPrime (approved):

core/src/main/java/io/substrait/relation/ExprCopyOnWriteVisitor.java
vbarua: We use the `Rel` prefix for a number of classes already so I'm inclined to keep `RelCopyOnWriteVisitor`. We use the full `Expression` prefix elsewhere though, so I'll switch this class.

vbarua:
These changes introduce a new public ExpressionCopyOnWriteVisitor class and switch the RelCopyOnWriteVisitor over to use the RelVisitor (instead of AbstractRelVisitor) to force the implementation of all visit methods.

The purpose of this is to ensure a complete visitation over relations and expressions in order to enable relation rewrites THROUGH subqueries in expressions.

core/src/main/java/io/substrait/relation/ExpressionCopyOnWriteVisitor.java
```diff
@@ -0,0 +1,365 @@
...
+            .match(match.orElse(expr.match()))
+            .switchClauses(switchClauses.orElse(expr.switchClauses()))
+            .defaultClause(defaultClause.orElse(expr.defaultClause()))
+            .build());
+  }
+
+  protected Optional<Expression.SwitchClause> visitSwitchClause(
+      Expression.SwitchClause switchClause) throws EXCEPTION {
+    return switchClause
+        .then()
```
patientstreetlight, line 170: Does this not need to also visit over the `switchClause.condition()`?

patientstreetlight (commented):
Mostly minor comments and a question

core/src/main/java/io/substrait/relation/ExpressionCopyOnWriteVisitor.java
vbarua: I thought about this and wasn't sure of what the correct behaviour should be actually.

The condition clause is a `Literal` and not an `Expression`, so it's guaranteed not to contain an `Expression`.

While `Literal` does extend `Expression`, meaning we can call `.accept` on it, the visitor can return _any_ kind of `Expression` and not just a `Literal` which feels like it could break in subtle ways.

Part of my intent with having an explicit `visitSwitchClause` method was to allow users to override it if they wanted to muck with the condition themselves.

patientstreetlight: > the visitor can return any kind of Expression and not just a Literal which feels like it could break in subtle ways.

ah!  This is a great point.  I wonder if it would be worth a comment to that effect for future readers.

core/src/main/java/io/substrait/relation/CopyOnWriteVisitor.java
```diff
@@ -0,0 +1,58 @@
...
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+import java.util.Optional;
+import java.util.function.Supplier;
+
+/** Base class for Copy On Write visitors. Provides common utils. */
+public abstract class CopyOnWriteVisitor<T, E extends Exception> {
+
+  protected boolean allEmpty(Optional<?>... optionals) {
```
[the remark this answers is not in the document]
vbarua: Good idea. Done.

patientstreetlight (approved):
:shipit: thanks Victor!

Issue #198, opened 2023-11-03
Implement MergeJoinRel


danepitkin:
Let's add the final physical join operator to Substrait Java for completeness. NLJ (https://github.com/substrait-io/substrait-java/pull/188) and HashJoin (https://github.com/substrait-io/substrait-java/pull/187) have already been added.

PR #201, opened 2023-11-08
feat: add MergeJoinRel

danepitkin:
Closes https://github.com/substrait-io/substrait-java/issues/198

### DIFF
--- a/core/src/main/java/io/substrait/relation/ProtoRelConverter.java
+++ b/core/src/main/java/io/substrait/relation/ProtoRelConverter.java
@@ -542,33 +542,6 @@ public class ProtoRelConverter {
     return builder.build();
   }
 
-  private NestedLoopJoin newNestedLoopJoin(NestedLoopJoinRel rel) {
-    Rel left = from(rel.getLeft());
-    Rel right = from(rel.getRight());
-    Type.Struct leftStruct = left.getRecordType();
-    Type.Struct rightStruct = right.getRecordType();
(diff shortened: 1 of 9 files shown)


danepitkin:
The `MergeJoin` implementation is identical to the `HashJoin` implementation. Any changes unrelated to `MergeJoin` that you see are just code snippets being moved around (I grouped the three physical operator implementations together for better readability).

core/src/main/java/io/substrait/relation/ProtoRelConverter.java
```diff
@@ -103,6 +102,12 @@ public Rel from(io.substrait.proto.Rel rel) {
       case HASH_JOIN -> {
         return newHashJoin(rel.getHashJoin());
       }
+      case NESTED_LOOP_JOIN -> {
```
EpsilonPrime, line 105: Why is the order Hash, Nested, Merge over alphabetical sort?

danepitkin: This is the order in the Substrait spec, but I agree alphabetical would be better. Will update!

EpsilonPrime: And here.

EpsilonPrime (approved):

core/src/main/java/io/substrait/relation/ProtoRelConverter.java
danepitkin: oof, thank you

vbarua (approved):
Change looks good to me! Thanks for the cleanup.

Issue #204, opened 2023-11-21
feat(calcite): support in-predicate in Ishtmus


rkondakov:
`ExpressionRexConverter` does not support `Expression.InPredicate`

vbarua:
Implemented in https://github.com/substrait-io/substrait-java/pull/205

Will be available in the next release.

PR #205, opened 2023-11-21
feat(isthmus): support in-predicate (#204)

rkondakov:
Added support for `Expression.InPredicate` into `ExpressionRexConverter`


vbarua (approved):
Changes look good to me. Thanks for adding this!

Issue #127, opened 2023-02-01
Implement ExtendedExpression


ianmcook:
https://github.com/substrait-io/substrait/pull/405 added an `ExtendedExpression` message to represent standalone expressions (expressions that are not part of plans). It would be nice to have this implemented in the Substrait Java library.

PR #206, opened 2023-11-24
feat: enable support from/to pojo/protobuf for extended expressions

davisusanibar:
To closes https://github.com/substrait-io/substrait-java/issues/127


core/src/main/java/io/substrait/extended/expression/ExtendedExpressionProtoConverter.java
```diff
@@ -0,0 +1,50 @@
...
+
+/**
+ * Converts from {@link io.substrait.extended.expression.ExtendedExpression} to {@link
+ * ExtendedExpression}
+ */
+public class ExtendedExpressionProtoConverter {
+  public ExtendedExpression toProto(
+      io.substrait.extended.expression.ExtendedExpression extendedExpression) {
+
+    ExtendedExpression.Builder extendedExpressionBuilder = ExtendedExpression.newBuilder();
```
danepitkin, line 18:
```suggestion
    ExtendedExpression.Builder builder = ExtendedExpression.newBuilder();
```
nit: this is just for readability

core/src/main/java/io/substrait/extended/expression/ProtoExtendedExpressionConverter.java
```diff
@@ -0,0 +1,84 @@
+package io.substrait.extended.expression;
+
+import io.substrait.expression.Expression;
+import io.substrait.expression.proto.ProtoExpressionConverter;
+import io.substrait.extension.*;
```
danepitkin, line 5: Can we explicitly list imports instead of using `*`?

core/src/main/java/io/substrait/extended/expression/ProtoExtendedExpressionConverter.java
```diff
@@ -0,0 +1,84 @@
...
+            .advancedExtension(
+                Optional.ofNullable(
+                    extendedExpression.hasAdvancedExtensions()
+                        ? extendedExpression.getAdvancedExtensions()
+                        : null))
+            .baseSchema(namedStruct);
+    return builder.build();
+  }
+
+  private io.substrait.type.NamedStruct convertNamedStrutProtoToPojo(NamedStruct namedStruct) {
```
danepitkin, line 70:
```suggestion
  private io.substrait.type.NamedStruct convertNamedStructProtoToPojo(NamedStruct namedStruct) {
```

core/src/main/java/io/substrait/extension/ImmutableExtensionLookup.java
```diff
@@ -30,14 +31,16 @@ public static class Builder {
     private final Map<Integer, SimpleExtension.FunctionAnchor> functionMap = new HashMap<>();
     private final Map<Integer, SimpleExtension.TypeAnchor> typeMap = new HashMap<>();
 
-    public Builder from(Plan p) {
+    public Builder from(
```
danepitkin, line 34: Nice refactor! we could also overload the API and add `from(Plan p)` and `from(ExtendedExpression e)` that call `from(simpleExtensionURIs, simpleExtensionDeclarations)`.

core/src/main/java/io/substrait/extended/expression/ProtoExtendedExpressionConverter.java
danepitkin: Would it make sense to implement this function as `fromProto` in this file? https://github.com/substrait-io/substrait-java/blob/main/core/src/main/java/io/substrait/type/NamedStruct.java

danepitkin (commented):
This looks awesome! I took a first pass at the feature, but haven't reviewed the tests yet.

core/src/test/java/io/substrait/extended/expression/ProtoExtendedExpressionConverterTest.java
```diff
@@ -0,0 +1,80 @@
...
+            .map(
+                declaration ->
+                    ExpressionCreator.scalarFunction(
+                        declaration,
+                        TypeCreator.REQUIRED.BOOLEAN,
+                        ImmutableFieldReference.builder()
+                            .addSegments(FieldReference.StructField.of(0))
+                            .type(TypeCreator.REQUIRED.decimal(10, 2))
+                            .build(),
+                        ExpressionCreator.i32(false, 183)));
```
vbarua, line 36: Could you create the expression here explicitly rather than pulling it out from the extensions like this. I'm not a fan of this pattern because I think:
1. It makes the test harder to read. There's a lot of extra code to pick out one declaration rather than creating one.
2. It makes the test depend on the _ordering_ of declarations in the default collection. There are 13 add variants and the first one is the decimal one. If the ordering changes though (and it's not guaranteed to be stable) the test will start to fail randomly.

There's is a 
```java
protected SubstraitBuilder b
```
in the `TestBase` class with a `scalarFunction` method that can be used for this.

core/src/main/java/io/substrait/extended/expression/ExtendedExpression.java
```diff
@@ -0,0 +1,28 @@
+package io.substrait.extended.expression;
```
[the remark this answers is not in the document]
vbarua: I like this idea as well. Having an `.../extended` directory makes me think there could be more than one item in it. `extendedexpression` maps to a singular `ExtensionExpression` better, at least to me.

core/src/main/java/io/substrait/extension/ImmutableExtensionLookup.java
vbarua: +1 to keeping the existing `from(Plan p)` and having it call your newly refactor method.

core/src/main/java/io/substrait/extended/expression/ExtendedExpression.java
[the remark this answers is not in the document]
davisusanibar: Changed

core/src/main/java/io/substrait/extended/expression/ExtendedExpression.java
```diff
@@ -0,0 +1,28 @@
...
+
+  public abstract List<String> getExpectedTypeUrls();
+
+  // creating simple extensions, such as extensionURIs and extensions, is performed on the fly
+
+  public abstract Optional<AdvancedExtension> getAdvancedExtension();
+
+  @Value.Immutable
+  public abstract static class ExpressionReference {
+    public abstract Expression getReferredExpr();
```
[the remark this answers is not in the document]
davisusanibar: I have added AggregateFunction and marked it as an UnsupportedOperationException for now.

core/src/main/java/io/substrait/extended/expression/ProtoExtendedExpressionConverter.java
davisusanibar: This is the result of applying `./gradlew spotlessApply` let me check if there are changes needed

core/src/main/java/io/substrait/extended/expression/ProtoExtendedExpressionConverter.java
davisusanibar: Changed

core/src/main/java/io/substrait/extension/ImmutableExtensionLookup.java
davisusanibar: Added

core/src/test/java/io/substrait/extended/expression/ProtoExtendedExpressionConverterTest.java
davisusanibar: Added

core/src/test/java/io/substrait/extendedexpression/ProtoExtendedExpressionConverterTest.java
```diff
@@ -0,0 +1,78 @@
...
+import java.util.ArrayList;
+import java.util.List;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
+
+public class ProtoExtendedExpressionConverterTest extends TestBase {
+  static final String NAMESPACE = "/functions_arithmetic_decimal.yaml";
+
+  @Test
+  public void fromTest() throws IOException {
```
danepitkin, line 22: It might be better to have a single `roundtripExtendedExpression` test so that you can test both to/from proto and compare the full objects afterwards (instead of `toProtoTest` and `fromProtoTest`).

danepitkin (commented):
Thanks David! I left a few more comments

core/src/main/java/io/substrait/type/NamedStruct.java
```diff
@@ -21,4 +22,20 @@ default io.substrait.proto.NamedStruct toProto(TypeProtoConverter typeProtoConve
         .addAllNames(names())
         .build();
   }
+
+  static io.substrait.type.NamedStruct convertNamedStructProtoToPojo(
```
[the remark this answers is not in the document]
davisusanibar: Changed

core/src/main/java/io/substrait/extendedexpression/ExtendedExpressionProtoConverter.java
```diff
@@ -0,0 +1,58 @@
...
+                .setExpression(expressionProto)
+                .addAllOutputNames(expressionReference.getOutputNames());
+        builder.addReferredExpr(expressionReferenceBuilder);
+      } else if (expressionType
+          instanceof io.substrait.extendedexpression.ExtendedExpression.AggregateFunctionType) {
+        throw new UnsupportedOperationException(
+            "Aggregate function types are not supported in conversion to proto Extended Expressions for now");
+      } else {
+        throw new UnsupportedOperationException(
+            "Only Expression or Aggregate Function type are supported in conversion to proto Extended Expressions for now");
```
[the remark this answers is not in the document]
davisusanibar: Changed

core/src/main/java/io/substrait/extendedexpression/ExtendedExpressionProtoConverter.java
```diff
@@ -0,0 +1,58 @@
...
+                        .getExpression());
+        ExpressionReference.Builder expressionReferenceBuilder =
+            ExpressionReference.newBuilder()
+                .setExpression(expressionProto)
+                .addAllOutputNames(expressionReference.getOutputNames());
+        builder.addReferredExpr(expressionReferenceBuilder);
+      } else if (expressionType
+          instanceof io.substrait.extendedexpression.ExtendedExpression.AggregateFunctionType) {
+        throw new UnsupportedOperationException(
+            "Aggregate function types are not supported in conversion to proto Extended Expressions for now");
```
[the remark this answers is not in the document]
davisusanibar: Just added.

core/src/main/java/io/substrait/extendedexpression/ProtoExtendedExpressionConverter.java
```diff
@@ -0,0 +1,77 @@
+package io.substrait.extendedexpression;
+
+import io.substrait.expression.Expression;
+import io.substrait.expression.proto.ProtoExpressionConverter;
+import io.substrait.extension.*;
```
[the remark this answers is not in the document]
davisusanibar: Changed.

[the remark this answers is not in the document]
davisusanibar: Solved by applying https://github.com/substrait-io/substrait-java/pull/191#discussion_r1408378391

core/src/test/java/io/substrait/extendedexpression/ExtendedExpressionProtoConverterTest.java
```diff
@@ -0,0 +1,67 @@
...
+import io.substrait.expression.*;
+import io.substrait.type.ImmutableNamedStruct;
+import io.substrait.type.Type;
+import io.substrait.type.TypeCreator;
+import java.util.ArrayList;
+import java.util.List;
+import org.junit.jupiter.api.Test;
+
+public class ExtendedExpressionProtoConverterTest extends TestBase {
+  static final String NAMESPACE = "/functions_arithmetic_decimal.yaml";
```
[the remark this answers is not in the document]
davisusanibar: These variables are private in SubstraitBuilder, so just maintain them here for testing purposes.

core/src/test/java/io/substrait/extendedexpression/ProtoExtendedExpressionConverterTest.java
davisusanibar: Just maintain only roundtrip class for Expression, Aggregation and Expression + Aggregation

core/src/main/java/io/substrait/extendedexpression/ProtoExtendedExpressionConverter.java
```diff
@@ -0,0 +1,85 @@
...
+    io.substrait.type.NamedStruct namedStruct =
+        io.substrait.type.NamedStruct.fromProto(baseSchemaProto, protoTypeConverter);
+
+    ProtoExpressionConverter protoExpressionConverter =
+        new ProtoExpressionConverter(
+            functionLookup, this.extensionCollection, namedStruct.struct(), null);
+
+    List<ExtendedExpression.ExpressionReference> expressionReferences = new ArrayList<>();
+    for (ExpressionReference expressionReference : extendedExpression.getReferredExprList()) {
+      if (expressionReference.getExprTypeCase().getNumber() == 1) { // Expression
```
danepitkin, line 52: optional: is there another provided option that's better than comparing with `getNumber()`? If not, then I think this is okay.

core/src/test/java/io/substrait/extendedexpression/ExtendedExpressionRoundTripTest.java
```diff
@@ -0,0 +1,229 @@
...
+import io.substrait.type.TypeCreator;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
+
+public class ExtendedExpressionRoundTripTest extends TestBase {
+  static final String NAMESPACE = "/functions_arithmetic_decimal.yaml";
```
danepitkin, line 19: nit: could we put this here instead? https://github.com/substrait-io/substrait-java/blob/e73c6cac7a5b62d026ccd13eca97bfae02395aab/core/src/main/java/io/substrait/dsl/SubstraitBuilder.java#L48

core/src/test/java/io/substrait/extendedexpression/ExtendedExpressionRoundTripTest.java
```diff
@@ -0,0 +1,229 @@
+package io.substrait.extendedexpression;
```
danepitkin, line 1: optional nit: the tests might be more readable with helper functions for `createTestExpressionReference()`, `createTestAggregateReference()` and `verifyRoundTrip()`. It could help reduce code duplication, too.

danepitkin (commented):
Nice work, @davisusanibar ! Left a few minor comments, but overall LGTM.

core/src/main/java/io/substrait/extendedexpression/ExtendedExpressionProtoConverter.java
```diff
@@ -0,0 +1,66 @@
...
+
+import io.substrait.expression.Expression;
+import io.substrait.expression.proto.ExpressionProtoConverter;
+import io.substrait.extension.ExtensionCollector;
+import io.substrait.proto.AggregateFunction;
+import io.substrait.proto.ExpressionReference;
+import io.substrait.proto.ExtendedExpression;
+import io.substrait.type.proto.TypeProtoConverter;
+
+/** Converts from {@link ExtendedExpression} to {@link ExtendedExpression} */
```
vbarua, line 11: minor: As is these docs links are the same class. If we fully qualify them it's clearer that they are distinct ExtendedExpression classes.
java
```
/** Converts from {@link io.substrait.extendedexpression.ExtendedExpression} to {@link io.substrait.proto.ExtendedExpression} */
```

core/src/main/java/io/substrait/extendedexpression/ExtendedExpressionProtoConverter.java
```diff
@@ -0,0 +1,60 @@
...
+
+    final ExpressionProtoConverter expressionProtoConverter =
+        new ExpressionProtoConverter(functionCollector, null);
+
+    for (io.substrait.extendedexpression.ExtendedExpression.ExpressionReference
+        expressionReference : extendedExpression.getReferredExpressions()) {
+      io.substrait.extendedexpression.ExtendedExpression.ExpressionTypeReference
+          expressionTypeReference = expressionReference.getExpressionType();
+      if (expressionTypeReference
+          instanceof io.substrait.extendedexpression.ExtendedExpression.ExpressionType et) {
```
vbarua, line 27: I made a small change here (https://github.com/substrait-io/substrait-java/pull/206/commits/92d2cc57e46e4b4972e5251477abf6d9baf4300b) to bind the variables your instanceof checking to make the PR easier to read. Using this, this way you can avoid the need to recast them.

core/src/main/java/io/substrait/extendedexpression/ExtendedExpression.java
```diff
@@ -0,0 +1,41 @@
...
+
+  public abstract List<String> getExpectedTypeUrls();
+
+  // creating simple extensions, such as extensionURIs and extensions, is performed on the fly
+
+  public abstract Optional<AdvancedExtension> getAdvancedExtension();
+
+  @Value.Immutable
+  public abstract static class ExpressionReference {
+    public abstract ExpressionTypeReference getExpressionType();
```
vbarua, line 25: I think that having the ExpressionReference contain the type of the reference within itself is a bit clunky as a design. Looking at the `ExtendedExpressionProtoConverter`, we have to do instanceof checks to figure out the type and then make decision based on the type.

I think a cleaner way to approach this is with inheritance with something like:
```java
  public interface ExpressionReferenceBase {
    public List<String> getOutputNames();

    public io.substrait.proto.ExpressionReference toProto(...);
  }

  @Value.Immutable
  public abstract static class ExpressionReference implements ExpressionReferenceBase {
    public abstract Expression getExpression();

    @Override
    public io.substrait.proto.ExpressionReference toProto(...) {
      return ...
    }
  }

  @Value.Immutable
  public abstract static class AggregateFunctionReference implements ExpressionReferenceBase {
    public abstract Aggregate.Measure getMeasure();

    @Override
    public io.substrait.proto.ExpressionReference toProto(ExpressionProtoConverter converter) {
      return null;
    }
  }
```
effectively, each class knows how to convert itself.

@davisusanibar what do you think?

danepitkin:
Thanks, Victor, for pointing out there is still some work to do to implement AggregateFunctions. It seems like it may be best to push AggregateFunction support to a follow up issue to keep this PR small.

core/src/main/java/io/substrait/extendedexpression/ExtendedExpressionProtoConverter.java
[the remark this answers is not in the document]
vbarua: I think this actually may be a fair bit of work to do properly. In the interest of keeping this PR small and moving this work along, I think we could include AggregateFunction support as a future change.

What do you think @danepitkin

core/src/test/java/io/substrait/extendedexpression/ExtendedExpressionRoundTripTest.java
danepitkin: Nevermind, these are private so lets ignore this comment for now.

core/src/main/java/io/substrait/extendedexpression/ExtendedExpression.java
```diff
@@ -0,0 +1,41 @@
...
+  public abstract static class ExpressionTypeReference {}
+
+  @Value.Immutable
+  public abstract static class ExpressionType extends ExpressionTypeReference {
+    public abstract Expression getExpression();
+  }
+
+  @Value.Immutable
+  public abstract static class AggregateFunctionType extends ExpressionTypeReference {
+    public abstract AggregateFunction getMeasure();
```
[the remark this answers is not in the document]
davisusanibar: I'm sorry, that was my mistake. Just update the code to use POJOs instead of PROTOs.

davisusanibar:
> Thanks, Victor, for pointing out there is still some work to do to implement AggregateFunctions. It seems like it may be best to push AggregateFunction support to a follow up issue to keep this PR small.

Hi Dane, just doing my last attempt to try to cover AggregateFunction ... I could suggest that it is needed to have a uniform API that won't change that dramatically later.

core/src/main/java/io/substrait/extendedexpression/ExtendedExpressionProtoConverter.java
```diff
@@ -0,0 +1,64 @@
...
+    final ExpressionProtoConverter expressionProtoConverter =
+        new ExpressionProtoConverter(functionCollector, null);
+
+    for (io.substrait.extendedexpression.ExtendedExpression.ExpressionReferenceBase
+        expressionReference : extendedExpression.getReferredExpressions()) {
+      if (expressionReference
+          instanceof io.substrait.extendedexpression.ExtendedExpression.ExpressionReference et) {
+        io.substrait.proto.Expression expressionProto =
+            expressionProtoConverter.visit(
+                (Expression.ScalarFunctionInvocation) et.getExpression());
```
vbarua, line 31: Why are we restricting this to scalar functions invocations? Are the other expression types not valid here?

core/src/main/java/io/substrait/extendedexpression/ProtoExtendedExpressionConverter.java
vbarua: IMO the nicer way of doing this is with a switch statement like
```java
      switch (expressionReference.getExprTypeCase()) {
        case EXPRESSION -> {
         ...
        }
        case MEASURE -> {
        ...
        }
        default -> throw new UnsupportedOperationException(
            "Only Expression or Aggregate Function type are supported in conversion from proto Extended Expressions");
      }
```

which avoids the magic number problem and is easier to read than if-else-if (at least for me).

core/src/main/java/io/substrait/extendedexpression/ExtendedExpression.java
```diff
@@ -0,0 +1,36 @@
...
+public abstract class ExtendedExpression {
+  public abstract List<ExpressionReferenceBase> getReferredExpressions();
+
+  public abstract NamedStruct getBaseSchema();
+
+  public abstract List<String> getExpectedTypeUrls();
+
+  // creating simple extensions, such as extensionURIs and extensions, is performed on the fly
+
+  public abstract Optional<AdvancedExtension> getAdvancedExtension();
```
vbarua, line 21: This needs to be the POJO AdvancedExtension not the proto one.

vbarua: Wait, the Plan POJO also use the protobuf AdvancedExtension.

https://github.com/substrait-io/substrait-java/blob/e73c6cac7a5b62d026ccd13eca97bfae02395aab/core/src/main/java/io/substrait/plan/Plan.java#L3C2-L3C45

I'll file an issue to fix this in the future.

vbarua: Filed: https://github.com/substrait-io/substrait-java/issues/210

core/src/main/java/io/substrait/extendedexpression/ExtendedExpressionProtoConverter.java
davisusanibar: Ah, I see, I'm only considering Filter and Projection cases. Let me review changes needed to be more generic.

davisusanibar: Changed

core/src/main/java/io/substrait/extendedexpression/ProtoExtendedExpressionConverter.java
danepitkin: +1

danepitkin:
I'll let @vbarua comment on if there is more work needed for AggregateFunctions, as I'm not yet that familiar with them. If there is more work required, we should do it in a follow up PR to keep this one small. The Expression support LGTM and I wouldn't want to hold up merging it, especially if the SQL -> ExtendedExpression PR would benefit from it!

core/src/main/java/io/substrait/relation/AggregateFunctionProtoConverter.java
```diff
@@ -0,0 +1,43 @@
...
+        .setInvocation(measure.getFunction().invocation().toProto())
+        .setOutputType(measure.getFunction().getType().accept(typeProtoConverter))
+        .addAllArguments(
+            IntStream.range(0, args.size())
+                .mapToObj(i -> args.get(i).accept(aggFuncDef, i, argVisitor))
+                .collect(java.util.stream.Collectors.toList()))
+        .setFunctionReference(
+            functionCollector.getFunctionReference(measure.getFunction().declaration()))
+        .build();
+  }
```
vbarua, line 42: This duplicates code `RelProtoConverter#toProto(Aggregate.Measure measure)`.

It's fine for now, but we should consider unifying it in the future.

core/src/test/java/io/substrait/extendedexpression/ExtendedExpressionRoundTripTest.java
```diff
@@ -0,0 +1,134 @@
...
+  private static Stream<Arguments> expressionReferenceProvider() {
+    return Stream.of(
+        Arguments.of(getI32LiteralExpression()),
+        Arguments.of(getFieldReferenceExpression()),
+        Arguments.of(getScalarFunctionExpression()),
+        Arguments.of(getImmutableAggregateFunctionReference()));
+  }
+
+  @ParameterizedTest
+  @MethodSource("expressionReferenceProvider")
```
vbarua, line 32: TIL about `@MethodSource`. That's pretty nifty.

vbarua (commented):
Thanks for adding this. I left some comments about things that we can improve in the future, but overall looks great.

I would like two thing before approving and merging:
1. Could you update the `ProtoExtendedExpressionConverter` to not use comparisons to `getNumber` and instead compare the protobuf enums.
2. Could you answer my question about the `@Desugar` annotation.

core/src/main/java/io/substrait/relation/AggregateFunctionProtoConverter.java
```diff
@@ -0,0 +1,43 @@
...
+        .setPhase(measure.getFunction().aggregationPhase().toProto())
+        .setInvocation(measure.getFunction().invocation().toProto())
+        .setOutputType(measure.getFunction().getType().accept(typeProtoConverter))
+        .addAllArguments(
+            IntStream.range(0, args.size())
+                .mapToObj(i -> args.get(i).accept(aggFuncDef, i, argVisitor))
+                .collect(java.util.stream.Collectors.toList()))
+        .setFunctionReference(
+            functionCollector.getFunctionReference(measure.getFunction().declaration()))
+        .build();
```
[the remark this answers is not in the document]
davisusanibar: I will create a ticket to address this issue.

core/src/main/java/io/substrait/relation/ProtoAggregateFunctionConverter.java
```diff
@@ -0,0 +1,61 @@
...
+    return Aggregate.Measure.builder()
+        .function(
+            AggregateFunctionInvocation.builder()
+                .arguments(functionArgs)
+                .declaration(aggregateFunction)
+                .outputType(protoTypeConverter.from(measure.getOutputType()))
+                .aggregationPhase(Expression.AggregationPhase.fromProto(measure.getPhase()))
+                .invocation(Expression.AggregationInvocation.fromProto(measure.getInvocation()))
+                .build())
+        .build();
```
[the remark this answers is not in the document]
davisusanibar: I will create a ticket to address this issue.

core/src/main/java/io/substrait/extension/ExtensionCollector.java
```diff
@@ -93,11 +109,14 @@ public void addExtensionsToPlan(Plan.Builder builder) {
               .build();
       extensionList.add(decl);
     }
-
-    builder.addAllExtensionUris(uris.values());
-    builder.addAllExtensions(extensionList);
+    return new SimpleExtensions(uris, extensionList);
   }
 
+  @Desugar
```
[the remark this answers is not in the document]
davisusanibar: It might be possible to pass this Jabel validation https://github.com/bsideup/jabel/blob/master/jabel-javac-plugin/src/main/java/com/github/bsideup/jabel/RecordsRetrofittingTaskListener.java#L484

core/src/main/java/io/substrait/extendedexpression/ExtendedExpressionProtoConverter.java
```diff
@@ -0,0 +1,63 @@
...
+    ExtendedExpression.Builder builder = ExtendedExpression.newBuilder();
+    ExtensionCollector functionCollector = new ExtensionCollector();
+
+    final ExpressionProtoConverter expressionProtoConverter =
+        new ExpressionProtoConverter(functionCollector, null);
+
+    for (io.substrait.extendedexpression.ExtendedExpression.ExpressionReferenceBase
+        expressionReference : extendedExpression.getReferredExpressions()) {
+      if (expressionReference
+          instanceof io.substrait.extendedexpression.ExtendedExpression.ExpressionReference et) {
```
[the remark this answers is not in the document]
davisusanibar: I will create a ticket to address this issue.

vbarua (approved):
Changes look good. Thanks for working with us on this @davisusanibar!

danepitkin:
Thanks, @davisusanibar ! Nice work!!

PR #212, opened 2023-12-15
Added support for char_length

YannisBres:
This PR adds a string-related scalar function mapping in `FunctionMappings` to support generating plans for queries that use the [`char_length`](https://www.postgresql.org/docs/current/functions-string.html#id-1.5.8.10.5.2.2.6.1.1.1) function.
It also introduces very basic unit tests for string-related functions, including the newly supported `char_length` one.

### DIFF
--- a/isthmus/src/test/java/io/substrait/isthmus/StringFunctionTest.java
+++ b/isthmus/src/test/java/io/substrait/isthmus/StringFunctionTest.java
@@ -10,7 +10,7 @@ public class StringFunctionTest extends PlanTestBase {
 
   @ParameterizedTest
   @ValueSource(strings = {"c16", "vc32"})
-  void char_length(String column) throws Exception {
+  void charLength(String column) throws Exception {
     String query = String.format("SELECT char_length(%s) FROM strings", column);
     assertSqlSubstraitRelRoundTrip(query, CREATES);
   }
--- a/isthmus/src/test/java/io/substrait/isthmus/StringFunctionTest.java
+++ b/isthmus/src/test/java/io/substrait/isthmus/StringFunctionTest.java
@@ -1,0 +1,66 @@
+package io.substrait.isthmus;
+
+import java.util.List;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.ValueSource;
+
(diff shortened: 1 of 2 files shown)


isthmus/src/test/java/io/substrait/isthmus/StringFunctionTest.java
```diff
@@ -0,0 +1,66 @@
...
+
+  @ParameterizedTest
+  @ValueSource(strings = {"c16", "vc32"})
+  void char_length(String column) throws Exception {
+    String query = String.format("SELECT char_length(%s) FROM strings", column);
+    assertSqlSubstraitRelRoundTrip(query, CREATES);
+  }
+
+  @ParameterizedTest
+  @ValueSource(strings = {"vc32"})
```
YannisBres, line 19: Using the `||` concatenation operator fails on CHAR columns with "`IllegalArgumentException: Unable to convert call ||(char<16>?, char<16>?) at RexExpressionConverter.visitCall(RexExpressionConverter.java:83)`".
This should be investigated.

isthmus/src/test/java/io/substrait/isthmus/StringFunctionTest.java
```diff
@@ -0,0 +1,66 @@
...
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.ValueSource;
+
+public class StringFunctionTest extends PlanTestBase {
+
+  static List<String> CREATES = List.of("CREATE TABLE strings (c16 CHAR(16), vc32 VARCHAR(32))");
+
+  @ParameterizedTest
+  @ValueSource(strings = {"c16", "vc32"})
+  void charLength(String column) throws Exception {
```
vbarua, line 13: I made a very small change here: `char_length` -> `charLength`

vbarua (approved):
Thanks for adding this mapping, and also for the additional tests for other string functions!

isthmus/src/test/java/io/substrait/isthmus/StringFunctionTest.java
YannisBres: 6 months of *&#@$%! Go and I forgot good ol' Java conventions.  What a shame!  😉

PR #213, opened 2023-12-18
ci: update github actions

vbarua:


.github/workflows/pr.yml
```diff
@@ -78,18 +78,18 @@ jobs:
       run: |
         value=`mv isthmus/build/graal/isthmus isthmus/build/graal/isthmus-${{ matrix.os }}`
     - name: Publish artifact
-      uses: actions/upload-artifact@master
+      uses: actions/upload-artifact@v4
```
vbarua, line 81: I suspect that this is what was causing the failure. All of the download actions are `@v3`, but the upload action is `@master`. The [upload-artifact](https://github.com/actions/upload-artifact) action had a 4.0 release recently, and I suspect the `@v3` version of download-artifact can't handle the v4 uploads.

danepitkin:
LGTM!

Issue #155, opened 2023-06-26
Calcite Aggregates Built w/ Out Of Order Grouping Keys Have Incorrect Layout


vbarua:
# Code Example
```java
public class SubstraitRelNodeConverterTest extends PlanTestBase {
      @Test
      void groupingKeyOrdering() {
          Plan.Root root = b.root(
          b.aggregate(input -> b.grouping(input, 2, 0), input -> List.of(),
                  b.namedScan(List.of("foo"), List.of("a", "b", "c"), List.of(R.I64, R.I64, R.STRING))));
          var relNode = converter.convert(root.getInput());
          // Fails because the output record has shape (a, c) because Calcite "forgets" the grouping key order
          assertRowMatch(relNode.getRowType(), R.STRING, R.I64);
      }
}
```

# Explanation
When build an aggregate using the RelBuilder, Calcite converts the grouping keys into an immutable bit set and then uses the order of the keys in the bit set to generate the output record shape. That means that if you hand it a grouping key like in the example of (2, 0), Calcite will output the row as (0, 2) (which the rest of the Substrait plan is not necessarily expecting).

This behaviour in Calcite appears to be somewhat expected, as the Calcite Sql Parser appears to apply projections before aggregations to account for this behaviour.

vbarua:
Fixed by https://github.com/substrait-io/substrait-java/pull/214

PR #214, opened 2023-12-21
feat(isthmus): improved Calcite support for Substrait Aggregate rels 

vbarua:
Substrait Aggregates that contain expressions that are not field
references and/or grouping keys that are not in input order require
extra processing to be converted to Calcite Aggregates successfully AND
correctly

### DIFF
--- a/core/src/main/java/io/substrait/dsl/SubstraitBuilder.java
+++ b/core/src/main/java/io/substrait/dsl/SubstraitBuilder.java
@@ -389,6 +389,11 @@ public class SubstraitBuilder {
         .collect(java.util.stream.Collectors.toList());
   }
 
+  public Expression.SortField sortField(
+      Expression expression, Expression.SortDirection sortDirection) {
+    return Expression.SortField.builder().expr(expression).direction(sortDirection).build();
+  }
+
   public SwitchClause switchClause(Expression.Literal condition, Expression then) {
     return SwitchClause.builder().condition(condition).then(then).build();
   }
--- a/core/src/main/java/io/substrait/dsl/SubstraitBuilder.java
+++ b/core/src/main/java/io/substrait/dsl/SubstraitBuilder.java
@@ -530,6 +530,48 @@ public class SubstraitBuilder {
 
   // Scalar Functions
 
+  public Expression.ScalarFunctionInvocation negate(Expression expr) {
+    // output type of negate is the same as the input type
+    var outputType = expr.getType();
+    return scalarFn(
+        FUNCTIONS_ARITHMETIC,
+        String.format("negate:%s", ToTypeString.apply(outputType)),
+        outputType,
+        expr);
+  }
+
+  public Expression.ScalarFunctionInvocation add(Expression left, Expression right) {
+    return arithmeticFunction("add", left, right);
+  }
+
+  public Expression.ScalarFunctionInvocation subtract(Expression left, Expression right) {
+    return arithmeticFunction("substract", left, right);
+  }
+
+  public Expression.ScalarFunctionInvocation multiply(Expression left, Expression right) {
+    return arithmeticFunction("multiply", left, right);
+  }
+
+  public Expression.ScalarFunctionInvocation divide(Expression left, Expression right) {
+    return arithmeticFunction("divide", left, right);
+  }
+
+  private Expression.ScalarFunctionInvocation arithmeticFunction(
+      String fname, Expression left, Expression right) {
+    var leftTypeStr = ToTypeString.apply(left.getType());
+    var rightTypeStr = ToTypeString.apply(right.getType());
+    var key = String.format("%s:%s_%s", fname, leftTypeStr, rightTypeStr);
+
+    var isOutputNullable = left.getType().nullable() || right.getType().nullable();
+    var outputType = left.getType();
+    outputType =
+        isOutputNullable
+            ? TypeCreator.asNullable(outputType)
+            : TypeCreator.asNotNullable(outputType);
+
+    return scalarFn(FUNCTIONS_ARITHMETIC, key, outputType, left, right);
+  }
+
   public Expression.ScalarFunctionInvocation equal(Expression left, Expression right) {
     return scalarFn(
         DefaultExtensionCatalog.FUNCTIONS_COMPARISON, "equal:any_any", R.BOOLEAN, left, right);
--- a/core/src/main/java/io/substrait/dsl/SubstraitBuilder.java
+++ b/core/src/main/java/io/substrait/dsl/SubstraitBuilder.java
@@ -437,49 +437,76 @@ public class SubstraitBuilder {
   }
 
   public AggregateFunctionInvocation min(Rel input, int field) {
-    Type inputType = input.getRecordType().fields().get(field);
-    // min output is always nullable
+    return min(fieldReference(input, field));
+  }
+
+  public AggregateFunctionInvocation min(Expression expr) {
     return singleArgumentArithmeticAggregate(
-        input, field, "min", TypeCreator.asNullable(inputType));
+        expr,
+        "min",
+        // min output is always nullable
+        TypeCreator.asNullable(expr.getType()));
   }
 
   public AggregateFunctionInvocation max(Rel input, int field) {
-    Type inputType = input.getRecordType().fields().get(field);
-    // max output is always nullable
+    return max(fieldReference(input, field));
+  }
+
+  public AggregateFunctionInvocation max(Expression expr) {
+
     return singleArgumentArithmeticAggregate(
-        input, field, "max", TypeCreator.asNullable(inputType));
+        expr,
+        "max",
+        // max output is always nullable
+        TypeCreator.asNullable(expr.getType()));
   }
 
   public AggregateFunctionInvocation avg(Rel input, int field) {
-    Type inputType = input.getRecordType().fields().get(field);
-    // avg output is always nullable
+    return avg(fieldReference(input, field));
+  }
+
+  public AggregateFunctionInvocation avg(Expression expr) {
     return singleArgumentArithmeticAggregate(
-        input, field, "avg", TypeCreator.asNullable(inputType));
+        expr,
+        "avg",
+        // avg output is always nullable
+        TypeCreator.asNullable(expr.getType()));
   }
 
   public AggregateFunctionInvocation sum(Rel input, int field) {
-    Type inputType = input.getRecordType().fields().get(field);
-    // sum output is always nullable
+    return sum(fieldReference(input, field));
+  }
+
+  public AggregateFunctionInvocation sum(Expression expr) {
     return singleArgumentArithmeticAggregate(
-        input, field, "sum", TypeCreator.asNullable(inputType));
+        expr,
+        "sum",
+        // sum output is always nullable
+        TypeCreator.asNullable(expr.getType()));
   }
 
   public AggregateFunctionInvocation sum0(Rel input, int field) {
-    // sum0 output is always NOT NULL I64
-    return singleArgumentArithmeticAggregate(input, field, "sum0", R.I64);
+    return sum(fieldReference(input, field));
+  }
+
+  public AggregateFunctionInvocation sum0(Expression expr) {
+    return singleArgumentArithmeticAggregate(
+        expr,
+        "sum0",
+        // sum0 output is always NOT NULL I64
+        R.I64);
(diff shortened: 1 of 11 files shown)


vbarua:
Fixes https://github.com/substrait-io/substrait-java/issues/155

core/src/main/java/io/substrait/dsl/SubstraitBuilder.java
```diff
@@ -440,48 +479,75 @@ public AggregateFunctionInvocation count(Rel input, int field) {
...
+    return min(fieldReference(input, field));
+  }
+
+  public AggregateFunctionInvocation min(Expression expr) {
     return singleArgumentArithmeticAggregate(
-        input, field, "min", TypeCreator.asNullable(inputType));
+        expr,
+        "min",
+        // min output is always nullable
+        TypeCreator.asNullable(expr.getType()));
```
vbarua, line 490: Added arithmetic aggregate functions that take expression directly.

core/src/main/java/io/substrait/dsl/SubstraitBuilder.java
```diff
@@ -57,35 +57,63 @@ public SubstraitBuilder(SimpleExtension.ExtensionCollection extensions) {
...
+        aggsFn.andThen(
+            functionInvocations ->
+                functionInvocations.stream().map(this::measure).collect(Collectors.toList()));
+    return aggregateM(groupingFn, measuresFn, input);
+  }
+
+  public Aggregate aggregateM(
+      Function<Rel, Aggregate.Grouping> groupingFn,
+      Function<Rel, List<Aggregate.Measure>> measuresFn,
       Rel input) {
```
vbarua, line 82: I tried to add this as an overload of `aggregate` like:
```java
  public Aggregate aggregate(
      Function<Rel, Aggregate.Grouping> groupingFn,
      Function<Rel, List<Aggregate.Measure>> measuresFn,
      Rel input) {
```
but it results in clashes as both methods have the same erasure.

core/src/main/java/io/substrait/dsl/SubstraitBuilder.java
```diff
@@ -494,6 +560,48 @@ private AggregateFunctionInvocation singleArgumentArithmeticAggregate(
...
+    return arithmeticFunction("substract", left, right);
+  }
+
+  public Expression.ScalarFunctionInvocation multiply(Expression left, Expression right) {
+    return arithmeticFunction("multiply", left, right);
+  }
+
+  public Expression.ScalarFunctionInvocation divide(Expression left, Expression right) {
+    return arithmeticFunction("divide", left, right);
+  }
```
vbarua, line 587: I needed to nest some functions within aggregate function calls for testing purposes, so I added some easy ones.

isthmus/src/main/java/io/substrait/isthmus/AggregateValidator.java
```diff
@@ -0,0 +1,223 @@
...
+ * org.apache.calcite.rel.core.Aggregate} rels
+ *
+ * <p>The code in this class can:
+ *
+ * <ul>
+ *   <li>Check for these cases
+ *   <li>Rewrite the Substrait {@link Aggregate} such that it can be converted to Calcite
+ * </ul>
+ */
+public class AggregateValidator {
```
vbarua, line 27: I'm open to suggestions for a better name for this.

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelVisitor.java
```diff
@@ -271,6 +271,7 @@ Aggregate.Measure fromAggCall(RelNode input, Type.Struct inputType, AggregateCal
     if (call.filterArg != -1) {
       builder.preMeasureFilter(FieldReference.newRootStructReference(call.filterArg, inputType));
     }
+    // TODO: handle the collation on the AggregateCall
```
vbarua, line 274: Filed an issues for this: https://github.com/substrait-io/substrait-java/issues/215

patientstreetlight (approved):
Mostly had minor style questions/comments, but overall I think this looks good, and very welcome.

core/src/main/java/io/substrait/dsl/SubstraitBuilder.java
```diff
@@ -57,35 +57,63 @@ public SubstraitBuilder(SimpleExtension.ExtensionCollection extensions) {
...
+
+  public Aggregate.Measure measure(AggregateFunctionInvocation aggFn, Expression preMeasureFilter) {
+    return Aggregate.Measure.builder().function(aggFn).preMeasureFilter(preMeasureFilter).build();
+  }
+
   public Aggregate aggregate(
       Function<Rel, Aggregate.Grouping> groupingFn,
-      Function<Rel, List<AggregateFunctionInvocation>> measuresFn,
+      Function<Rel, List<AggregateFunctionInvocation>> aggsFn,
```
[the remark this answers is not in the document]
vbarua: The SubstraitBuilder is intended to help construct plan trees in code. It's more akin to a DSQL than a builder (I probably should have called it something else in hindsight).

 So for example when building:
```java
              b.aggregate(
                  input -> b.grouping(input, 0, 2),
                  input -> List.of(b.count(input, 0)),
                  <inputRel>));
```
`input` is set to `inputRel`, and is eventually passed to a FieldReference constructor which can use the input to set it's type.

Without functions, with a signature like:
```java
  public Aggregate aggregate(
      Aggregate.Grouping groupingFn,
      List<AggregateFunctionInvocation> measuresFn
      List<AggregateFunctionInvocation>
)
```

You could either do something like
```java
var inputRel = <inputRel> // subtree
var agg = 
              b.aggregate(
                  b.grouping(inputRel, 0, 2),
                  List.of(b.count(inputRel, 0)),
                  <inputRel>));
```
and build your relation tree piecemeal and still get the nice field reference utilities or

```java
var inputRel = <inputRel>
var agg = 
              b.aggregate(
                  b.grouping(
                     b.fieldReference(0, <Type Of Input Field 0>), 
                     b.fieldReference(2, <Type Of Input Field 2>)),
                  List.of(b.count(b.fieldReference(0, <Type Of Input Field 0>))),
                  <inputRel>));
```
and create the field references yourself, type and all.

TLDR: Using functions like this let's us construct field references more ergonomically. Building trees piecemeal would also work as a design.

core/src/main/java/io/substrait/dsl/SubstraitBuilder.java
```diff
@@ -57,35 +57,63 @@ public SubstraitBuilder(SimpleExtension.ExtensionCollection extensions) {
...
+      Rel input) {
+    Function<Rel, List<Aggregate.Measure>> measuresFn =
+        aggsFn.andThen(
+            functionInvocations ->
+                functionInvocations.stream().map(this::measure).collect(Collectors.toList()));
+    return aggregateM(groupingFn, measuresFn, input);
+  }
+
+  public Aggregate aggregateM(
```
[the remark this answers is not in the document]
vbarua: Thinking about this some more, I removed the the `aggregateM` methods and updated the `aggregate` methods to consume `Aggregate.Measure` instead of `AggregateFunctionInvocation`. I then updated the aggregate functions methods to return `Aggregate.Measure` as well.

Because the `aggregate` functions now consume `Aggregate.Measure`, we can now layer in additional utils to tweak the measure as needed like:
```java
            input ->
                List.of(
                    withPreMeasureFilter(
                        b.sum(input, 0), b.equal(b.fieldReference(input, 1), b.i32(42)))),
```

core/src/main/java/io/substrait/dsl/SubstraitBuilder.java
```diff
@@ -426,6 +459,12 @@ public Aggregate.Grouping grouping(Rel input, int... indexes) {
     return Aggregate.Grouping.builder().addAllExpressions(columns).build();
   }
 
+  public Aggregate.Grouping grouping(Expression... expressions) {
+    List<Expression> exprs =
+        Arrays.stream(expressions).collect(java.util.stream.Collectors.toList());
+    return Aggregate.Grouping.builder().addAllExpressions(exprs).build();
```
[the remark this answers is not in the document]
vbarua: I don't think this works because `addAllExpressions` requires an `Iterable<? extends Expression> elements`, not an   `Iterator<Expression>`.

That being said, there is an `addExpressions(Expression... elements)` methods that can be used directly 😅 .

isthmus/src/main/java/io/substrait/isthmus/AggregateValidator.java
vbarua: This inspired me to use `PreCalciteAggregateValidator` and `PreCalciteAggregateTransformer`.

It's very similar to `AggregatePreCalciteTransformer` but with extra emphasis on the `PreCalcite` part of it.

vbarua:
To briefly summarise what this PR is about. Calcite has a more restrictive definition of aggregate relations than Substrait does, which means that not all Substrait aggregates can be converted to Calcite aggregates. The primary restriction is that Calcite aggregates can only take field references in various places that Substrait allows full expressions.

Luckily, this is easy to address by simply moving those expressions to a project before the aggregate, and then updating the aggregate to use references.

The various comments in the `PreCalciteAggregateValdidator` try and document these various restrictions in more detail.

vbarua:
Addresses https://github.com/substrait-io/substrait-java/issues/155

core/src/main/java/io/substrait/dsl/SubstraitBuilder.java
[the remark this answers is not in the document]
patientstreetlight: fwiw `() -> Arrays.stream(expressions).iterator()` _is_ an `Iterable`, since it is a lambda which, when called, produces an `Iterator`.

isthmus/src/main/java/io/substrait/isthmus/PreCalciteAggregateValidator.java
vibhatha, line 1: @vbarua I am not very fluent with the Calcite component, so I will leave someone else to verify this part.

isthmus/src/test/java/io/substrait/isthmus/ComplexAggregateTest.java
vibhatha, line 1: General comment: I see the expected output is hand-calculated (meaning determined by ourselves), is there a way to use Calcite to get the optimal and validate with it? This is an out of scope area for me, but just curious if we can do things differently.

vibhatha:
@vbarua seems like there is a conflict.

vibhatha (approved):
@vbarua Thanks for working on this. The overall change LGTM! 
Though I would appreciate someone with better Calcite knowledge take a look.

isthmus/src/test/java/io/substrait/isthmus/ComplexAggregateTest.java
vbarua: The transformation code that I'm testing here is purely going from Substrait -> Substrait. My intent with these tests was to demonstrate that the `transformToValidCalciteAggregate(aggregate)` produces equivalent _Substrait_ plans, no Calcite required.

Calcite only becomes relevant in the validation code when we call
```java
    // Substrait POJO -> Calcite
    new SubstraitToCalcite(EXTENSION_COLLECTION, typeFactory).convert(pojo);
```
which tries to convert the given input to a Calcite plan. One thing that's hard to show here is that if you comment out
```java
    if (!PreCalciteAggregateValidator.isValidCalciteAggregate(aggregate)) {
      aggregate =
          PreCalciteAggregateValidator.PreCalciteAggregateTransformer
              .transformToValidCalciteAggregate(aggregate);
    }
```
in the `SubstraitRelNodeConverter#visit(Aggregate aggregate)` method all of these fail.

> is there a way to use Calcite to get the optimal and validate with it?

Something that's kind of cursed about this is that Calcite already does something like this, but _only_ when you give it SQL input. When it converts from its SqlNode representation to the RelNode representation it applies similar transforms to this.

I don't think copying exactly what Calcite does is a good idea here, because I'd rather perform straightforward (if verbose) transformation here that may not necessarily be the _best_ but is easier to reason about. This also happens _before_ any actual optimizations a user would perform with Calcite, as this transformation is just to allow us to convert the Substrait plan into a Calcite representation.

Does this answer what you were asking?

vbarua:
I'm going to merge these changes. They will only take effect for substrait plans that would previously have resulted in exceptions, so they shouldn't be surprising anyone.

Issue #211, opened 2023-12-12
[Isthmus][CLI][GraalVM] Failed to build Isthmus native image locally using GraalVM 22.0.0.2


davisusanibar:
# Issue

Currently Isthmus is using GraalVM 22.0.0.2 https://github.com/substrait-io/substrait-java/blob/main/isthmus/build.gradle.kts#L104

It is necessary to run `./gradlew nativeImage` in order to build Isthmus native image, and the process will finish with the following message:

```
./gradlew nativeImage

> Task :isthmus:extractGraalTooling
Downloading: Release index file from oca.opensource.oracle.com
Error: Error reading component list: https://oca.opensource.oracle.com/gds/meta-data.json

> Task :isthmus:extractGraalTooling FAILED

Build substrait FAILURE reason:                                
    Execution failed for task ':isthmus:extractGraalTooling':
        org.gradle.process.internal.ExecException: Process 'command '/Users/dsusanibar/.gradle/caches/com.palantir.graal/22.0.0.2/17/graalvm-ce-java17-22.0.0.2/Contents/Home/bin/gu'' finished with non-zero exit value 3
```

# Reason of the error

Isthmus is using [Palantir Gradle Graal](https://github.com/substrait-io/substrait-java/blob/main/isthmus/build.gradle.kts#L7) to download and locally cache a GraalVM installation and make available select parts of the GraalVM compiler for use in Gradle builds.

GraalVM 22.0.0.2 is available to download, but not to install a new image, as you can see at:

````
$ pwd
/Users/dsusanibar/.gradle/caches/com.palantir.graal/22.0.0.2/17/graalvm-ce-java17-22.0.0.2/Contents/Home/bin

$ ./gu --version
GraalVM Updater 22.0.0.2

$ ./gu available
Downloading: Release index file from oca.opensource.oracle.com
Error: Error reading component list: https://oca.opensource.oracle.com/gds/meta-data.json

$ ./gu install native-image
Downloading: Release index file from oca.opensource.oracle.com
Error: Error reading component list: https://oca.opensource.oracle.com/gds/meta-data.json
````

# Workaround
- Option 1: After moving from GraalVM 22.0.0.2 to [GraalVM 22.1.0](https://www.graalvm.org/release-notes/22_1/), everything is working as expected.

````
$ pwd
/Users/dsusanibar/.gradle/caches/com.palantir.graal/22.1.0/17/graalvm-ce-java17-22.1.0/Contents/Home/bin
[...]

PR #216, opened 2023-12-26
feat(isthmus): cli can now output JSON, TEXT and BINARY proto format

vbarua:


vbarua:
Partial fix for: https://github.com/substrait-io/substrait-java/issues/211

EpsilonPrime:
Technically text is protobuf text.  We are still looking to add the Substrait text format.

vbarua:
@EpsilonPrime good point.

What do you think about the names `PROTOJSON`, `PROTOTEXT` instead of `JSON` and `TEXT` to make it more explicity where they come from?

vibhatha:
@vbarua thank you for working on this.

EpsilonPrime:
That works for me.  The existing implementation can take an explicit format
type and it can also auto detect the type so having some way to eventually
provide auto detect as an option would be nice as well.

vbarua:
@EpsilonPrime @vibhatha 

If these changes look reasonable now, I would appreciate an approval when y'all have a chance 🙇

vibhatha (approved):
@vbarua LGTM. Thanks for working on this.

EpsilonPrime (approved):

PR #217, opened 2023-12-26
feat: introduce DefaultExtensionCatalog

vbarua:
contains static strings for default extensions

### DIFF
--- a/core/src/main/java/io/substrait/type/YamlRead.java
+++ b/core/src/main/java/io/substrait/type/YamlRead.java
@@ -3,6 +3,7 @@ package io.substrait.type;
 import com.fasterxml.jackson.databind.DeserializationFeature;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.dataformat.yaml.YAMLFactory;
+import io.substrait.extension.DefaultExtensionCatalog;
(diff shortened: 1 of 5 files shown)


vbarua:
It's come up a number of times that we're copying the names of extensions in various places. This change adds a `DefaultExtensionCatalog` class whose sole purpose is to hold static strings of extension names for re-use.

core/src/main/java/io/substrait/type/YamlRead.java
```diff
@@ -25,7 +26,8 @@ public class YamlRead {
 
   public static void main(String[] args) throws Exception {
     try {
-      System.out.println("Read: " + YamlRead.class.getResource("/functions_boolean.yaml"));
+      System.out.println(
```
vibhatha, line 29: nit: @vbarua just curious do we need a println here? Or should we have used logs?

vibhatha (approved):
This PR LGTM!

core/src/main/java/io/substrait/type/YamlRead.java
vbarua: This looks to be some old debugging code. Trying to run the main function throws a `FileNotFoundException` for me, so I'm not sure if it works anymore. I could be worth removing, but I think that would be better as it's own change.

Issue #219, opened 2024-01-17
[Release]: Semantic release requires Node version 20.8.1 or higher


davisusanibar:
Semantic release requires Node version 20.8.1 or higher

Detailed error: https://github.com/substrait-io/substrait-java/actions/runs/7516403502/job/20461523677

**Run semantic release**:
7s
Run ./ci/release/run.sh
npm WARN EBADENGINE Unsupported engine {
npm WARN EBADENGINE   package: 'semantic-release@23.0.0',
npm WARN EBADENGINE   required: { node: '>=20.8.[1](https://github.com/substrait-io/substrait-java/actions/runs/7516403502/job/20461523677#step:8:1)' },
npm WARN EBADENGINE   current: { node: 'v[18](https://github.com/substrait-io/substrait-java/actions/runs/7516403502/job/20461523677#step:8:19).[19](https://github.com/substrait-io/substrait-java/actions/runs/7516403502/job/20461523677#step:8:20).0', npm: '10.2.3' }
npm WARN EBADENGINE }
npm WARN deprecated read-pkg-up@11.0.0: Renamed to read-package-up
[semantic-release]: node version >=[20](https://github.com/substrait-io/substrait-java/actions/runs/7516403502/job/20461523677#step:8:21).8.1 is required. Found v18.19.0.

See https://github.com/semantic-release/semantic-release/blob/master/docs/support/node-version.md for more details and solutions.

PR #220, opened 2024-01-17
fix: semantic release requires Node version 20.8.1 or higher

davisusanibar:
To closes: https://github.com/substrait-io/substrait-java/issues/219


vbarua (approved):
Thanks for fixing this!

Issue #221, opened 2024-01-18
Some exceptions are not very informative for the client code


rkondakov:
Some exceptions in the code do not answer the question "what was wrong" and sometimes it requires debugging to figure out the problem with substrait plan. I propose to add some clarification to several exceptions in the code

davisusanibar:
Hello @rkondakov, could this issue be closed as part of https://github.com/substrait-io/substrait-java/pull/222?

rkondakov:
@davisusanibar yes, it's done. Thanks!

PR #222, opened 2024-01-18
fix: some exceptions are not very informative for the client code (#221)

rkondakov:
Some clarification was added to the non-informative exceptions.


EpsilonPrime (approved):
Thanks, these improved exceptions should make developers' lives better!

vbarua (approved):
LGTM! Thanks for the clarifications.

PR #224, opened 2024-02-06
feat(isthmus): support for replace

bvolpato:
Function is available in [Calcite](https://github.com/apache/calcite/blob/4f5a08f5429ff905a2ebd96600b0966ba9e228ea/core/src/main/java/org/apache/calcite/sql/fun/SqlStdOperatorTable.java#L1581-L1583) and [Substrait's functions_string.yaml](https://github.com/substrait-io/substrait/blob/3ded94d69e7ae329f2730b7634ba34900250e84a/extensions/functions_string.yaml#L771-L790), so just registering the mapping here.


bvolpato:
`Build and Test Java` seems broken? Complaining about things unrelated to my change:

```
```

vbarua (approved):
Thanks for adding the mapping!

PR #225, opened 2024-02-07
fix: update bad URLs in the release readme

bvolpato:
Minor doc fix: I was reading up about the release process and noticed the artifacts link was resolving to 404.


EpsilonPrime (approved):
Thanks for the fix!

PR #226, opened 2024-02-08
feat(isthmus): add up-converting signature matchers, and coerce types to match 

bvolpato:
My attempt to take a stab at the `TODO` in the code:

```
  // TODO: define up-converting matchers.
```

We've got here since we noticed that functions that have `list<any1>, any1` arguments wouldn't match `list<string>, string` without implementing a proper `CallConverter` as adapter:

- The `FunctionConverter` was not matching using `directMatchKey` since `list_any` != `list_str`.

- When falling back to the `leastRestrictive` logic, `list<string>` + `string` would also not agree, so no functions were returned.

I've kept the `leastRestrictive` logic intact, but added another fallback matcher that checks if all arguments will match (using the existent `isMatch` logic, which relies on [IgnoreNullableAndParameters](https://github.com/substrait-io/substrait-java/blob/main/isthmus/src/main/java/io/substrait/isthmus/expression/IgnoreNullableAndParameters.java) internally).

--

Let me know if there are any concerns or thoughts on how to make this better.

### DIFF
--- a/isthmus/src/main/java/io/substrait/isthmus/expression/FunctionConverter.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/expression/FunctionConverter.java
@@ -1,6 +1,12 @@
 package io.substrait.isthmus.expression;
 
-import com.google.common.collect.*;
+import com.google.common.collect.ArrayListMultimap;
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.ListMultimap;
+import com.google.common.collect.Multimap;
+import com.google.common.collect.Multimaps;
+import com.google.common.collect.Streams;
 import io.substrait.expression.Expression;
 import io.substrait.expression.ExpressionCreator;
 import io.substrait.expression.FunctionArg;
@@ -179,7 +185,7 @@ public abstract class FunctionConverter<
           // Make sure that arguments & return are within bounds and match the types
           if (function.returnType() instanceof ParameterizedType
               && isMatch(outputType, (ParameterizedType) function.returnType())
-              && argumentsMatchType(inputTypes, args)) {
+              && inputTypesSatisfyDefinedArguments(inputTypes, args)) {
             return Optional.of(function);
           }
         }
@@ -188,15 +194,30 @@ public abstract class FunctionConverter<
       };
     }
 
-    private static boolean argumentsMatchType(
+    /**
+     * Checks to see if the given input types satisfy the function arguments given. Checks that
+     *
+     * <ul>
+     *   <li>Variadic arguments all have the same input type
+     *   <li>Matched wildcard arguments (i.e.`any`, `any1`, `any2`, etc) all have the same input
+     *       type
+     * </ul>
+     *
+     * @param inputTypes input types to check against arguments
+     * @param args expected arguments as defined in a {@link SimpleExtension.Function}
+     * @return true if the {@code inputTypes} satisfy the {@code args}, false otherwise
+     */
+    private static boolean inputTypesSatisfyDefinedArguments(
         List<Type> inputTypes, List<SimpleExtension.Argument> args) {
 
       Map<String, Set<Type>> wildcardToType = new HashMap<>();
       for (int i = 0; i < inputTypes.size(); i++) {
         Type givenType = inputTypes.get(i);
-        // Variadic arguments should match the last argument's type
         SimpleExtension.ValueArgument wantType =
-            (SimpleExtension.ValueArgument) args.get(Integer.min(i, args.size() - 1));
+            (SimpleExtension.ValueArgument)
+                args.get(
+                    // Variadic arguments should match the last argument's type
+                    Integer.min(i, args.size() - 1));
 
         if (!isMatch(givenType, wantType.value())) {
           return false;
@@ -213,7 +234,8 @@ public abstract class FunctionConverter<
       }
 
       // If all the types match, check if the wildcard types are compatible.
-      // Note: We could ignore the "any" key here if we decide to not match non-enumerated types.
+      // TODO: Determine if non-enumerated wildcard types (i.e. `any` as opposed to `any1`) need to
+      //   have the same type.
       return wildcardToType.values().stream().allMatch(s -> s.size() == 1);
     }
 
--- a/isthmus/src/main/java/io/substrait/isthmus/expression/FunctionConverter.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/expression/FunctionConverter.java
@@ -1,6 +1,5 @@
 package io.substrait.isthmus.expression;
 
-import com.github.bsideup.jabel.Desugar;
 import com.google.common.collect.*;
 import io.substrait.expression.Expression;
 import io.substrait.expression.ExpressionCreator;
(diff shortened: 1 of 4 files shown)


isthmus/src/main/java/io/substrait/isthmus/expression/FunctionConverter.java
```diff
@@ -327,33 +348,76 @@ public Optional<T> attemptMatch(C call, Function<RexNode, Expression> topLevelCo
...
+
+        Optional<T> coerced = matchCoerced(call, outputType, operands);
+        if (coerced.isPresent()) {
+          return coerced;
         }
       }
       return Optional.empty();
     }
+
+    private Optional<T> matchByLeastRestrictive(
```
vbarua, line 364: Good call extracting this into a method, it's much easier to read ✨

vbarua (commented):
Started taking a look and left some minor feedback. I can look at this some more tomorrow. I need to be fresher to really grok the type matching code 🧠

isthmus/src/main/java/io/substrait/isthmus/expression/FunctionConverter.java
```diff
@@ -367,17 +431,15 @@ public interface GenericCall {
    * not for nullability or parameter mismatches.
    */
   private List<Expression> coerceArguments(List<Expression> arguments, Type type) {
```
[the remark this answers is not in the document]
bvolpato: Done, thanks! https://github.com/substrait-io/substrait-java/pull/226/commits/68b3be13719ec14ad15376876196c70bdec62dfa

isthmus/src/test/resources/extensions/functions_custom.yaml
```diff
@@ -6,19 +6,58 @@ types:
...
+          value: any1
+      return: string
   - name: "to_b_type"
     description: "converts a nullable a_type to a b_type"
     impls:
       - args:
           - name: arg1
             value: u!a_type?
         return: u!b_type
+  - name: "custom_scalar_arrayany"
```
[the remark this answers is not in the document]
bvolpato: Done, thanks! https://github.com/substrait-io/substrait-java/pull/226/commits/68b3be13719ec14ad15376876196c70bdec62dfa

isthmus/src/test/java/io/substrait/isthmus/CustomFunctionTest.java
```diff
@@ -99,6 +113,51 @@ public RelDataType toCalcite(Type.UserDefined type) {
...
+          SqlFunctionCategory.USER_DEFINED_FUNCTION);
+
+  static final SqlFunction customScalarArrayAnyFn =
+      new SqlFunction(
+          "custom_scalar_arrayany",
+          SqlKind.OTHER_FUNCTION,
+          ReturnTypes.explicit(varcharArrayType),
+          null,
+          null,
+          SqlFunctionCategory.USER_DEFINED_FUNCTION);
```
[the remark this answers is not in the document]
bvolpato: Yeah, Calcite type is called Array, but I agree that we should keep the function names consistent with the Substrait model / protos. https://github.com/substrait-io/substrait-java/pull/226/commits/68b3be13719ec14ad15376876196c70bdec62dfa

isthmus/src/test/java/io/substrait/isthmus/CustomFunctionTest.java
```diff
@@ -99,6 +113,51 @@ public RelDataType toCalcite(Type.UserDefined type) {
...
+          ReturnTypes.explicit(SqlTypeName.VARCHAR),
+          null,
+          null,
+          SqlFunctionCategory.USER_DEFINED_FUNCTION);
+
+  static final SqlFunction customScalarListAnyFn =
+      new SqlFunction(
+          "custom_scalar_listany",
+          SqlKind.OTHER_FUNCTION,
+          ReturnTypes.explicit(varcharArrayType),
```
vbarua, line 129: `custom_scalar_listany` is defined as returning a `LIST<any1>`. The return type here doesn't match that.

If I update the `R.STRING` types to be `R.I64` types in the `customScalarListAnyRoundtrip` the test fails.

isthmus/src/test/java/io/substrait/isthmus/CustomFunctionTest.java
```diff
@@ -99,6 +113,51 @@ public RelDataType toCalcite(Type.UserDefined type) {
...
+          null,
+          null,
+          SqlFunctionCategory.USER_DEFINED_FUNCTION);
+
+  static final SqlFunction customScalarListAnyAndAnyFn =
+      new SqlFunction(
+          "custom_scalar_listany_any",
+          SqlKind.OTHER_FUNCTION,
+          ReturnTypes.explicit(varcharArrayType),
```
vbarua, line 138: `custom_scalar_listany_any` is defined as returning a `LIST<any1>`. The return type here doesn't match that.

isthmus/src/test/resources/extensions/functions_custom.yaml
```diff
@@ -6,19 +6,58 @@ types:
...
+            value: any1
+        return: LIST<any1>
+  - name: "custom_scalar_liststring"
+    description: "custom function that takes list of string"
+    impls:
+      - args:
+          - name: list
+            value: LIST<string>
+        return: LIST<string>
+  - name: "custom_scalar_liststring_any"
```
vbarua, line 52: Trying to make sure the Calcite function definitions matched the signatures was tricky because I had to reference back and forth between the code and the extension. What do you think about encoding the return type with a pattern like `custom_scalar_liststring_any_to_liststring`.

vbarua (commented):
Left some more comments.

isthmus/src/test/java/io/substrait/isthmus/CustomFunctionTest.java
bvolpato: We have a problem if I try to make the type based on `SqlTypeName.ANY`:

```
java.lang.UnsupportedOperationException: Unable to convert the type ANY
	at io.substrait.isthmus.TypeConverter.toSubstrait(TypeConverter.java:141)
```

To add ANY in the TypeConverter we'll need to create an Any `io.substrait.type.Type`. Any concerns? Should I do within this PR?

isthmus/src/test/resources/extensions/functions_custom.yaml
bvolpato: +1, much less cognitive load

isthmus/src/test/java/io/substrait/isthmus/CustomFunctionTest.java
vbarua: Ah. To clarify, I think we need a better return type inference here. Something like [ReturnTypes.ARG0](https://github.com/apache/calcite/blob/31d66e797d3ca66d154262b15ee35876804df963/core/src/main/java/org/apache/calcite/sql/type/ReturnTypes.java#L155-L160) but that inspects the array element type (which I think may exists within Calcite somewhere).

isthmus/src/main/java/io/substrait/isthmus/expression/FunctionConverter.java
```diff
@@ -170,8 +171,30 @@ public boolean allowedArgCount(int count) {
...
     private static <F extends SimpleExtension.Function> SignatureMatcher<F> getSignatureMatcher(
         SqlOperator operator, List<F> functions) {
-      // TODO: define up-converting matchers.
-      return (a, b) -> Optional.empty();
+      return (inputTypes, outputType) -> {
+        for (F function : functions) {
+          List<SimpleExtension.Argument> args = function.requiredArguments();
+
+          // Skip if the number of arguments doesn't match or the return type doesn't match
+          if (args.size() != inputTypes.size()
```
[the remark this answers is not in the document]
bvolpato: Good point -- I've added support to variadic arguments.

vbarua (approved):
Thanks for adding this in!

PR #227, opened 2024-02-09
feat: add support for empty list literals

patientstreetlight:
The substrait proto definition provides a representation of empty list literals which is separate from the
representation of non-empty list literals:

```
message Literal {
  oneof literal_type {
    ...
    List list = 30;
    Type.List empty_list = 31;
    ...
  }
}
```

This change gives substrait-java a representation for empty list literals, along with conversions to/from
protobuf and to/from Calcite in Isthmus.

BREAKING CHANGE:
* ExpressionVisitor has a new method to visit Expression.EmptyListLiteral. Implementors which don’t extend AbstractExpressionVisitor will have to add their own implementation.
* Public class LiteralConstructorConverter now requires a TypeConverter to be passed to its constructor.
* ExpressionCopyOnWriteVisitor#visitLiteral() may now be called with Expression.EmptyListLiteral.

### DIFF
--- a/core/src/main/java/io/substrait/expression/proto/ExpressionProtoConverter.java
+++ b/core/src/main/java/io/substrait/expression/proto/ExpressionProtoConverter.java
@@ -205,6 +205,23 @@ public class ExpressionProtoConverter implements ExpressionVisitor<Expression, R
   }
 
   @Override
+  public Expression visit(io.substrait.expression.Expression.EmptyListLiteral expr)
+      throws RuntimeException {
+    return lit(
+        builder -> {
+          var protoListType = expr.getType().accept(typeProtoConverter);
+          builder
+              .setEmptyList(protoListType.getList())
+              // For empty lists, the Literal message's own nullable field should be ignored
+              // in favor of the nullability of the Type.List in the literal's
+              // empty_list field. But for safety we set the literal's nullable field
+              // to match in case any readers either look in the wrong location
+              // or want to verify that they are consistent.
+              .setNullable(expr.nullable());
+        });
+  }
+
+  @Override
   public Expression visit(io.substrait.expression.Expression.StructLiteral expr) {
     return lit(
         bldr -> {
--- a/core/src/main/java/io/substrait/expression/proto/ProtoExpressionConverter.java
+++ b/core/src/main/java/io/substrait/expression/proto/ProtoExpressionConverter.java
@@ -314,6 +314,12 @@ public class ProtoExpressionConverter {
           literal.getList().getValuesList().stream()
               .map(this::from)
               .collect(java.util.stream.Collectors.toList()));
+      case EMPTY_LIST -> {
+        // literal.getNullable() is intentionally ignored in favor of the nullability
+        // specified in the literal.getEmptyList() type.
+        var listType = protoTypeConverter.fromList(literal.getEmptyList());
+        yield ExpressionCreator.emptyList(listType.nullable(), listType.elementType());
+      }
       default -> throw new IllegalStateException(
           "Unexpected value: " + literal.getLiteralTypeCase());
     };
--- a/core/src/main/java/io/substrait/type/proto/ProtoTypeConverter.java
+++ b/core/src/main/java/io/substrait/type/proto/ProtoTypeConverter.java
@@ -47,7 +47,7 @@ public class ProtoTypeConverter {
               type.getStruct().getTypesList().stream()
                   .map(this::from)
                   .collect(java.util.stream.Collectors.toList()));
-      case LIST -> n(type.getList().getNullability()).list(from(type.getList().getType()));
+      case LIST -> fromList(type.getList());
       case MAP -> n(type.getMap().getNullability())
           .map(from(type.getMap().getKey()), from(type.getMap().getValue()));
       case USER_DEFINED -> {
@@ -59,6 +59,10 @@ public class ProtoTypeConverter {
           "Unsupported user defined reference: " + type);
       case KIND_NOT_SET -> throw new UnsupportedOperationException("Type is not set: " + type);
     };
+  }
+
+  public Type.ListType fromList(io.substrait.proto.Type.List list) {
+    return n(list.getNullability()).list(from(list.getType()));
   }
 
   public static boolean isNullable(io.substrait.proto.Type.Nullability nullability) {
--- a/isthmus/src/main/java/io/substrait/isthmus/expression/ExpressionRexConverter.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/expression/ExpressionRexConverter.java
@@ -16,6 +16,7 @@ import io.substrait.type.StringTypeVisitor;
 import io.substrait.type.Type;
 import io.substrait.util.DecimalUtil;
 import java.math.BigDecimal;
+import java.util.Collections;
 import java.util.List;
 import java.util.Set;
 import java.util.concurrent.TimeUnit;
@@ -246,6 +247,13 @@ public class ExpressionRexConverter extends AbstractExpressionVisitor<RexNode, R
     List<RexNode> args =
         expr.values().stream().map(l -> l.accept(this)).collect(Collectors.toList());
     return rexBuilder.makeCall(SqlStdOperatorTable.ARRAY_VALUE_CONSTRUCTOR, args);
+  }
+
+  @Override
+  public RexNode visit(Expression.EmptyListLiteral expr) throws RuntimeException {
+    var calciteType = typeConverter.toCalcite(typeFactory, expr.getType());
+    return rexBuilder.makeCall(
+        calciteType, SqlStdOperatorTable.ARRAY_VALUE_CONSTRUCTOR, Collections.emptyList());
   }
 
   @Override
--- a/isthmus/src/main/java/io/substrait/isthmus/expression/LiteralConstructorConverter.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/expression/LiteralConstructorConverter.java
@@ -3,6 +3,7 @@ package io.substrait.isthmus.expression;
 import io.substrait.expression.Expression;
 import io.substrait.expression.ExpressionCreator;
 import io.substrait.isthmus.CallConverter;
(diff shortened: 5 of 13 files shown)


core/src/main/java/io/substrait/expression/proto/ProtoExpressionConverter.java
```diff
@@ -314,6 +314,8 @@ public Expression.Literal from(io.substrait.proto.Expression.Literal literal) {
           literal.getList().getValuesList().stream()
               .map(this::from)
               .collect(java.util.stream.Collectors.toList()));
+      case EMPTY_LIST -> ExpressionCreator.emptyList(
+          literal.getNullable(), protoTypeConverter.from(literal.getEmptyList().getType()));
```
vbarua, line 318: Looking at this helped me identify some potential ambiguity in the spec. There is a nullability field in the `Literal` message AND a nullability field in the `Type.List` message.

Based on how null literals are written about in the spec, I think we should be using the nullability information from the `Type.List`.

I've created a different PR to clear up the spec here: https://github.com/substrait-io/substrait/pull/602



When declaring the

vbarua: > I think you're saying we should be using (2) and ignoring (1) for the nullability of the list literal itself?

Yes, the `Type.List` messages encodes the nullability of both the wrapper and its elements. Using the nullability therein seems consistent with how the nullability for null literals, another type that encodes it's own nullability) is handled.

core/src/main/java/io/substrait/expression/proto/ExpressionProtoConverter.java
```diff
@@ -204,6 +204,19 @@ public Expression visit(io.substrait.expression.Expression.ListLiteral expr) {
...
   }
 
+  @Override
+  public Expression visit(io.substrait.expression.Expression.EmptyListLiteral expr)
+      throws RuntimeException {
+    return lit(
+        builder -> {
+          var protoListType = expr.getType().accept(typeProtoConverter);
+          // For empty lists, the literal's own nullable field is ignored in favor
+          // of the nullability of the enclosed Type.List.
```
EpsilonPrime, line 214: For safety I'd argue that all writers should update both locations and that the readers should read from the documented location.  A really paranoid reader could check that the two match but I doubt anyone would want to do that.

patientstreetlight: Thanks David, that makes sense.  I've updated this to set the same nullability in both places.

isthmus/src/main/java/io/substrait/isthmus/expression/ExpressionRexConverter.java
```diff
@@ -248,6 +251,34 @@ public RexNode visit(Expression.ListLiteral expr) throws RuntimeException {
...
+    public RelDataType inferReturnType(SqlOperatorBinding opBinding) {
+      return type;
+    }
+  }
+
+  @Override
+  public RexNode visit(Expression.EmptyListLiteral expr) throws RuntimeException {
+    var calciteType = typeConverter.toCalcite(typeFactory, expr.getType());
+    return rexBuilder.makeCall(
+        new EmptySqlArrayValueConstructor(calciteType), Collections.emptyList());
```
vbarua, line 279: The docs for `makeCall` mention
```
If you already know the return type of the call, then {@link #makeCall(org.apache.calcite.rel.type.RelDataType, org.apache.calcite.sql.SqlOperator, java.util.List)} is preferred.
 ```

I tried out the following here
```java
    return rexBuilder.makeCall(
        calciteType, new SqlArrayValueConstructor(), Collections.emptyList());
```
here and your test passed. This code bypasses return type inference so it doesn't fail.

Would it make sense to you to construct it directly like this and avoid the need for the `EmptySqlArrayValueConstructor` abstraction?

isthmus/src/main/java/io/substrait/isthmus/expression/LiteralConstructorConverter.java
```diff
@@ -18,29 +19,53 @@ public class LiteralConstructorConverter implements CallConverter {
...
+            false,
+            call.operands.stream()
+                .map(t -> ((Expression.Literal) topLevelConverter.apply(t)))
+                .collect(java.util.stream.Collectors.toList())));
+  }
+
+  private Optional<Expression> toEmptyListLiteral(RexCall call) {
+    var calciteElementType = call.getType().getComponentType();
+    var substraitElementType = typeConverter.toSubstrait(calciteElementType);
+    return Optional.of(ExpressionCreator.emptyList(false, substraitElementType));
```
vbarua, line 69: Instead of setting then nullability to `false` here, should we use the nullability we get from Calcite in `call.getType.isNullable`? 

I _suspect_ it should always be `false`, but I'm not sure.

isthmus/src/main/java/io/substrait/isthmus/expression/LiteralConstructorConverter.java
```diff
@@ -18,29 +19,53 @@ public class LiteralConstructorConverter implements CallConverter {
...
-              .map(t -> ((Expression.Literal) topLevelConverter.apply(t)))
-              .collect(java.util.stream.Collectors.toList());
-      Map<Expression.Literal, Expression.Literal> items = new HashMap<>();
-      assert literals.size() % 2 == 0;
-      for (int i = 0; i < literals.size(); i += 2) {
-        items.put(literals.get(i), literals.get(i + 1));
-      }
-      return Optional.of(ExpressionCreator.map(false, items));
+      return toMapLiteral(call, topLevelConverter);
     }
```
vbarua, line 38: This is much nicer to read with the new helpers you extracted ✨

vbarua (commented):
Overall this looks good. I do have some questions and suggestion in various places.

core/src/main/java/io/substrait/expression/proto/ExpressionProtoConverter.java
```diff
@@ -204,6 +204,22 @@ public Expression visit(io.substrait.expression.Expression.ListLiteral expr) {
...
+          var protoListType = expr.getType().accept(typeProtoConverter);
+          // For empty lists, the literal's own nullable field should be ignored
+          // in favor of the nullability of the Type.List in the literal's
+          // empty_list field. But for safety we set the literal's nullable field
+          // to match in case any readers either look in the wrong location
+          // or want to verify that they are consistent.
+          var literalNullable = expr.nullable();
+          builder.setEmptyList(protoListType.getList()).setNullable(literalNullable);
+        });
+  }
```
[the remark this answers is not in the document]
patientstreetlight: I like the suggestion!  Both moving the comment and the tweak to the language 👍

core/src/main/java/io/substrait/type/proto/ProtoTypeConverter.java
```diff
@@ -61,6 +61,10 @@ public Type from(io.substrait.proto.Type type) {
     };
   }
 
+  public Type.ListType fromList(io.substrait.proto.Type.List list) {
```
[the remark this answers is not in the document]
patientstreetlight: This is used from `ProtoExpressionCreator` as:
```
      case EMPTY_LIST -> {
        // literal.getNullable() is intentionally ignored in favor of the nullability
        // specified in the literal.getEmptyList() type.
        var listType = protoTypeConverter.fromList(literal.getEmptyList());
        yield ExpressionCreator.emptyList(listType.nullable(), listType.elementType());
```

which:
* has a ` io.substrait.proto.Type.List` from `literal.getEmptyList()`
* wants a pojo `Type.ListType` to be able to grab its element type.

so I like having `ProtoTypeConverter#fromList()` be public for this use-case, which does exactly what `ProtoExpressionCreator` needs.

The alternate would be to use the original `ProtoTypeConverter#from(io.substrait.proto.Type type)` method, which is possible, but feels clunkier to me:
```
      case EMPTY_LIST -> {
        // literal.getNullable() is intentionally ignored in favor of the nullability
        // specified in the literal.getEmptyList() type.
        var typeWrapper =
            io.substrait.proto.Type.newBuilder().setList(literal.getEmptyList()).build();
        var listType = (Type.ListType) protoTypeConverter.from(typeWrapper);
        yield ExpressionCreator.emptyList(listType.nullable(), listType.elementType());
```

which I don't love for 2 reasons:
* it requires wrapping the `io.substrait.proto.Type.List` in an outer `io.substrait.proto.Type`.
* it requires down-casting the result to `Type.ListType` - this requires this client code to know that `io.substrait.proto.Type.List` maps to `Type.ListType`, and doesn't get any compile-time checking that this is actually true.

isthmus/src/test/java/io/substrait/isthmus/EmptyArrayLiteralTest.java
```diff
@@ -0,0 +1,26 @@
...
+  void emptyArrayLiteral() {
+    var colType = N.I8;
+    var emptyListLiteral = ExpressionCreator.emptyList(false, N.I8);
+    var rel =
+        b.project(
+            input -> List.of(emptyListLiteral),
+            Rel.Remap.offset(1, 1),
+            b.namedScan(List.of("t"), List.of("col"), List.of(colType)));
+    assertFullRoundTrip(rel);
+  }
```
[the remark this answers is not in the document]
patientstreetlight: I added the test, which also exposed a problem to which you had already proposed a solution in https://github.com/substrait-io/substrait-java/pull/227#discussion_r1490349870

While I was here, I also took the liberty of fixing a related bug in old code in `LiteralConstructorConverter#toNonEmptyListLiteral()`, which likewise should also look at `call.getType().isNullable()`.

isthmus/src/main/java/io/substrait/isthmus/expression/ExpressionRexConverter.java
patientstreetlight: This is a great idea - seems to work, and no need for the `EmptySqlArrayValueConstructor` ✨

vbarua (approved):
Thanks for adding this in! Changes look good 🎉

PR #230, opened 2024-02-13
fix(isthmus): allow for conversion of plans containing Calcite aggregate functions

bvolpato:
Some of the Calcite CoreRules might bring their own functions at a late stage, so I've decided to make the converter flexible enough and handle the superclass types correctly (since we already do a similar conversion to `APPROX_COUNT_DISTINCT` in the same spot).

Another alternative would be to improve the matchers and check if names match or if it's an `instanceof` the passing call -- today `signatures` is an `IdentityHashMap`, so it's really strict.

### DIFF
--- a/isthmus/src/main/java/io/substrait/isthmus/AggregateFunctions.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/AggregateFunctions.java
@@ -24,19 +24,29 @@ public class AggregateFunctions {
   public static SqlAggFunction SUM0 = new SubstraitSumEmptyIsZeroAggFunction();
 
   /**
-   * Utility class to possibly convert the SqlAggFunction from the native Calcite implementation to
-   * the Substrait subclasses present here, in case they have definitions.
+   * Some Calcite rules, like {@link
+   * org.apache.calcite.rel.rules.AggregateExpandDistinctAggregatesRule}, introduce the default
+   * Calcite aggregate functions into plans.
+   *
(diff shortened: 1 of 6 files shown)


bvolpato:
FYI Check failures are unrelated to the change:

```
/home/runner/work/substrait-java/substrait-java/core/src/test/java/io/substrait/type/proto/LocalFilesRoundtripTest.java:61: error: switch expressions are not supported in -source 8
    return switch (pathTypeCase) {
           ^
  (use -source 14 or higher to enable switch expressions)
      case URI_PATH -> builder.pathType(FileOrFiles.PathType.URI_PATH).path("path");
```

vbarua (approved):
Thanks for these changes!

I was poking around to add the `conversionHandlesBuiltInSum0CallAddedByRule` test and made some small changes along the way as well.

If the code in https://github.com/substrait-io/substrait-java/pull/230/commits/6c0c6fdf20c78ef6adec91a7981992ca8ae6da3d looks good to you, we can merge this in.

bvolpato:
> Thanks for these changes!
> 
> I was poking around to add the `conversionHandlesBuiltInSum0CallAddedByRule` test and made some small changes along the way as well.
> 
> If the code in [6c0c6fd](https://github.com/substrait-io/substrait-java/commit/6c0c6fdf20c78ef6adec91a7981992ca8ae6da3d) looks good to you, we can merge this in.

It looks good to me! Thanks for all the improvements / suggestions!

PR #231, opened 2024-02-14
feat: add POJOs for ConsistentPartitionWindowRel and WindowRelFunction

bvolpato:
Adding the POJO for ConsistentPartitionWindowRel and WindowRelFunction: https://github.com/substrait-io/substrait/blob/main/proto/substrait/algebra.proto#L233-L267

I wasn't sure about the organization/place of some parts since it's not physically under `Expression`, but there's **a lot** in common, even highlighted in the proto:

>  // This message mirrors the `WindowFunction` message but removes the fields defining the partition,
>  // sorts, and bounds, since those must be consistent across the various functions in this rel.  Refer
>  // to the `WindowFunction` message for a description of these fields.

So I tried to keep it as a new base class, but tried to keep the converters as close as they were to make it more concise + simplifying reuse.

>
>
--

I'll have an Isthmus change following (this is essentially a cleanup / split of https://github.com/substrait-io/substrait-java/pull/228), and @vbarua rightfully suggested breaking it down.


core/src/main/java/io/substrait/relation/RelCopyOnWriteVisitor.java
```diff
@@ -380,6 +400,24 @@ protected Optional<Expression.SortField> visitSortField(Expression.SortField sor
         .map(expr -> Expression.SortField.builder().from(sortField).expr(expr).build());
   }
 
+  protected Optional<ConsistentPartitionWindow.WindowRelFunctionInvocation> visitWindowRelFunction(
```
vbarua, line 403: minor: I would put this directly under the `visit(ConsistentPartitionWindow consistentPartitionWindow)` definition as it's not really a generic utility in the same way that `visitExprList`, in that it operates only on classes found within the `ConsistentPartitionWindow` unlike the the other methods which operate on classes that can exist in multiple relations.

core/src/main/java/io/substrait/relation/RelCopyOnWriteVisitor.java
```diff
@@ -342,6 +342,26 @@ public Optional<Rel> visit(NestedLoopJoin nestedLoopJoin) throws EXCEPTION {
...
+  public Optional<Rel> visit(ConsistentPartitionWindow consistentPartitionWindow) throws EXCEPTION {
+    var windowFunctions =
+        transformList(consistentPartitionWindow.getWindowFunctions(), this::visitWindowRelFunction);
+    var partitionExpressions =
+        transformList(
+            consistentPartitionWindow.getPartitionExpressions(),
+            t -> t.accept(getExpressionCopyOnWriteVisitor()));
+    var sorts = transformList(consistentPartitionWindow.getSorts(), this::visitSortField);
+
+    return Optional.of(
```
vbarua, line 355: To follow the pattern of the other copy on write visitors, this should return empty _unless_ any of the Optionals `windowFunctions`, `partitionExpressions` and `sorts` is non-empty.

core/src/main/java/io/substrait/expression/proto/ProtoExpressionConverter.java
```diff
@@ -250,6 +219,80 @@ public Expression from(io.substrait.proto.Expression expr) {
...
+    var partitionExprs =
+        windowFunction.getPartitionsList().stream().map(this::from).collect(Collectors.toList());
+    var sortFields =
+        windowFunction.getSortsList().stream()
+            .map(this::fromSortField)
+            .collect(Collectors.toList());
+    var options =
+        windowFunction.getOptionsList().stream()
+            .map(this::fromFunctionOption)
+            .collect(Collectors.toMap(FunctionOption::getName, Function.identity()));
```
vbarua, line 243: Oooh, nice addition. I hadn't noticed we weren't extraction and converting the functions options ✨

core/src/main/java/io/substrait/expression/proto/ProtoExpressionConverter.java
```diff
@@ -318,4 +361,28 @@ public Expression.Literal from(io.substrait.proto.Expression.Literal literal) {
...
+        .expr(from(s.getExpr()))
+        .build();
+  }
+
+  public FunctionOption fromFunctionOption(io.substrait.proto.FunctionOption o) {
+    return ImmutableFunctionOption.builder()
+        .name(o.getName())
+        .addAllValues(o.getPreferenceList())
+        .build();
+  }
```
vbarua, line 387: Good call extracting these utils ✨ 

Having these public makes sense to me as well to let library users apply them to their own custom protos if they want.

core/src/main/java/io/substrait/expression/proto/ProtoExpressionConverter.java
```diff
@@ -318,4 +361,28 @@ public Expression.Literal from(io.substrait.proto.Expression.Literal literal) {
...
+      int argumentsCount,
+      FunctionArg.ProtoFrom argVisitor,
+      SimpleExtension.Function declaration,
+      Function<Integer, FunctionArgument> argFunction) {
+    return IntStream.range(0, argumentsCount)
+        .mapToObj(i -> argVisitor.convert(declaration, i, argFunction.apply(i)))
+        .collect(Collectors.toList());
+  }
+
+  public ImmutableExpression.SortField fromSortField(SortField s) {
```
vbarua, line 375: minor: This should return `Expression.SortField`

core/src/main/java/io/substrait/relation/RelProtoConverter.java
```diff
@@ -75,6 +77,43 @@ private List<SortField> toProtoS(Collection<Expression.SortField> sorts) {
         .collect(java.util.stream.Collectors.toList());
   }
 
+  private List<ConsistentPartitionWindowRel.WindowRelFunction> toProtoWindowRelFunctions(
```
vbarua, line 80: minor: Can you move this to directly after the definition of `public Rel visit(ConsistentPartitionWindow` as it's the only caller of this.

core/src/main/java/io/substrait/relation/RelProtoConverter.java
```diff
@@ -75,6 +77,43 @@ private List<SortField> toProtoS(Collection<Expression.SortField> sorts) {
...
+                                  .build())
+                      .collect(java.util.stream.Collectors.toList());
+
+              return ConsistentPartitionWindowRel.WindowRelFunction.newBuilder()
+                  .setInvocation(f.invocation().toProto())
+                  .setPhase(f.aggregationPhase().toProto())
+                  .setOutputType(toProto(f.outputType()))
+                  .addAllArguments(arguments)
+                  .addAllOptions(options)
+                  .setFunctionReference(functionCollector.getFunctionReference(f.declaration()))
```
vbarua, line 111: I think we need to set the `lowerBound`, `upperBound` and `boundsType` as well.

core/src/main/java/io/substrait/relation/ConsistentPartitionWindow.java
```diff
@@ -0,0 +1,70 @@
...
+  public <O, E extends Exception> O accept(RelVisitor<O, E> visitor) throws E {
+    return visitor.visit(this);
+  }
+
+  public static ImmutableConsistentPartitionWindow.Builder builder() {
+    return ImmutableConsistentPartitionWindow.builder();
+  }
+
+  @Value.Immutable
+  public abstract static class WindowRelFunctionInvocation {
```
vbarua, line 46: Not for this PR. It might make sense in the future to define an interface of common types that we can share between `WindowRelFunctionInvocation` and `WindowFunctionInvocation`.

Constructing a `WindowRelFunctionInvocation` from that interface would be straightforward (as it would have all the fields), and then construction a `WindowRelFunctionInvocation` would just required providing the partitions and sorts.

We might also be able to have `WindowFunctionInvocation` extend from `WindowRelFunctionInvocation`, but I vaguely recall trying something like that in the past and the Immutable library not liking it.

core/src/test/java/io/substrait/type/proto/WindowFunctionRoundtripTest.java
```diff
@@ -0,0 +1,73 @@
...
+                    Expression.SortField.builder()
+                        .expr(expression)
+                        .from(b.sortField(expression, Expression.SortDirection.ASC_NULLS_FIRST))
+                        .build()))
+            .build();
+
+    Rel protoWindowRel = to.toProto(aggRel);
+    assertEquals(protoWindowRel, to.toProto(from.from(protoWindowRel)));
+    assertEquals(
+        protoWindowRel.getWindow().getWindowFunctions(0).getInvocation(), invocation.toProto());
```
vbarua, line 64: As written, this test doesn't catch the missing bounds conversion. I would recommend doing something like:

```java
    # note: whenever we're mixing proto and pojo code I find it useful to use fully qualified paths for clarity
    io.substrait.proto.Rel protoRel = to.toProto(aggRel);
    io.substrait.relation.Rel aggRel2 = from.from(protoRel);
    assertEquals(aggRel, aggRel2);
```

We should compare the example we build directly, in this case `aggRel` with the result of converting to and from proto. In this case the `toProto` conversion is lossy, but because we aren't checking against our built example we miss this.

vbarua (changes requested):
Missing a couple field conversions, but overall this looks good.

> I wasn't sure about the organization/place of some parts since it's not physically under Expression, but there's a lot in common, even highlighted in the proto:

I can't a think of a alternate organization at this point either. Grouping the `WindowRelFunctionInvocation ` conversion code alongside the equivalent `WindowFunctionInvocation` conversion code makes sense to me, and at least keeps it together. There might be some nicer abstraction we could use to unify these somehow, but I think that would be a bigger refactor and should be left for another time.

core/src/main/java/io/substrait/expression/proto/ProtoExpressionConverter.java
```diff
@@ -250,6 +219,80 @@ public Expression from(io.substrait.proto.Expression expr) {
     };
   }
 
+  public ImmutableExpression.WindowFunctionInvocation fromWindowFunction(
```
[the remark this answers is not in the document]
bvolpato: Thanks! Overlook when using extract method on IntelliJ

core/src/main/java/io/substrait/expression/proto/ProtoExpressionConverter.java
bvolpato: ditto

core/src/main/java/io/substrait/relation/ConsistentPartitionWindow.java
bvolpato: Yeah I agree -- it's subtly different and can be annoying to have concise builders and everything.

core/src/main/java/io/substrait/relation/RelCopyOnWriteVisitor.java
bvolpato: Good catch

core/src/main/java/io/substrait/relation/RelCopyOnWriteVisitor.java
```diff
@@ -380,6 +400,24 @@ protected Optional<Expression.SortField> visitSortField(Expression.SortField sor
...
   }
 
+  protected Optional<ConsistentPartitionWindow.WindowRelFunctionInvocation> visitWindowRelFunction(
+      ConsistentPartitionWindow.WindowRelFunctionInvocation windowRelFunctionInvocation)
+      throws EXCEPTION {
+    return Optional.of(
+        ConsistentPartitionWindow.WindowRelFunctionInvocation.builder()
+            .arguments(
+                visitFunctionArguments(windowRelFunctionInvocation.arguments())
+                    .orElse(windowRelFunctionInvocation.arguments()))
```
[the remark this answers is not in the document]
bvolpato: Done, thanks!

core/src/main/java/io/substrait/relation/RelProtoConverter.java
bvolpato: Ack

core/src/main/java/io/substrait/relation/RelProtoConverter.java
bvolpato: Very good catch! 💯

core/src/test/java/io/substrait/type/proto/WindowFunctionRoundtripTest.java
bvolpato: Thanks! It does catch the error and used it to verify the fix.

bvolpato:
Thanks for the great review @vbarua 💯.

All good points! Pushed https://github.com/substrait-io/substrait-java/pull/231/commits/1bb98364ccaf93271f506a06d959eb73a8d3b1d1 addressing them.

core/src/main/java/io/substrait/relation/RelCopyOnWriteVisitor.java
```diff
@@ -377,14 +377,8 @@ protected Optional<ConsistentPartitionWindow.WindowRelFunctionInvocation> visitW
 
     return Optional.of(
         ConsistentPartitionWindow.WindowRelFunctionInvocation.builder()
+            .from(windowRelFunctionInvocation)
```
vbarua, line 380: Using from here both:
* saves use some lines of code
* prevents issues if we ever add fields to the `ConsistentPartitionWindow` (as we would then need to make sure we updated this code to manually copy over those fields)

core/src/test/java/io/substrait/type/proto/WindowFunctionRoundtripTest.java
```diff
@@ -1,71 +0,0 @@
...
-import io.substrait.relation.RelProtoConverter;
-import io.substrait.relation.VirtualTableScan;
-import io.substrait.type.TypeCreator;
-import java.io.IOException;
-import java.math.BigDecimal;
-import java.util.Arrays;
-import java.util.Collections;
-import org.junit.jupiter.api.Test;
-
-public class WindowFunctionRoundtripTest extends TestBase {
```
vbarua: Renamed to `ConsistentPartitionWindowRelRoundtripTest` as we're testing the roundtripability of the rel as a whole.

core/src/test/java/io/substrait/type/proto/ConsistentPartitionWindowRelRoundtripTest.java
```diff
@@ -0,0 +1,62 @@
...
+  void consistentPartitionWindowRoundtrip() {
+    var windowFunctionDeclaration =
+        defaultExtensionCollection.getWindowFunction(
+            SimpleExtension.FunctionAnchor.of(
+                DefaultExtensionCatalog.FUNCTIONS_ARITHMETIC, "lead:any"));
+    Rel input =
+        b.namedScan(
+            Arrays.asList("test"),
+            Arrays.asList("a", "b", "c"),
+            Arrays.asList(R.I64, R.I16, R.I32));
```
vbarua, line 29: Using the builder to setup the input as it's more readable (IMO)

core/src/test/java/io/substrait/type/proto/ConsistentPartitionWindowRelRoundtripTest.java
```diff
@@ -0,0 +1,62 @@
...
+    Rel rel1 =
+        ImmutableConsistentPartitionWindow.builder()
+            .input(input)
+            .windowFunctions(
+                Arrays.asList(
+                    ConsistentPartitionWindow.WindowRelFunctionInvocation.builder()
+                        .declaration(windowFunctionDeclaration)
+                        // lead(a)
+                        .arguments(Arrays.asList(b.fieldReference(input, 0)))
```
vbarua, line 38: Specify at least one argument to make sure argument conversions are wired in.

core/src/test/java/io/substrait/type/proto/ConsistentPartitionWindowRelRoundtripTest.java
```diff
@@ -0,0 +1,62 @@
...
+                    Expression.SortField.builder()
+                        // SORT BY c
+                        .expr(b.fieldReference(input, 2))
+                        .direction(Expression.SortDirection.ASC_NULLS_FIRST)
+                        .build()))
+            .build();
+
+    io.substrait.proto.Rel protoRel = relProtoConverter.toProto(rel1);
+    io.substrait.relation.Rel rel2 = protoRelConverter.from(protoRel);
+    assertEquals(rel1, rel2);
```
vbarua, line 60: Use the `relProtoConverter` and `protoRelConverter` defined in `TestBase` to avoid having to set them up ourselves.

vbarua (approved):
Changes look good.

Made a small refactor in https://github.com/substrait-io/substrait-java/pull/231/commits/05476ed53c9b053fefc4541fb080dc8d1684692c and simplified the test in https://github.com/substrait-io/substrait-java/pull/231/commits/e595bf96f6ec45009e45df6a4201db7d37007b79, see my PR comments for my reasoning.

PR #232, opened 2024-02-14
feat: support UDT literals using bytestring/binary representation in calcite

cheikhachraf:
calcite does not accept RelDatatypes with sqltypename.OTHER and has no way of representing it as of now.
It will reject it at different steps such as [RexLiteral Precondition type check](https://github.com/apache/calcite/blob/c774c313a81d01c4e3e77cf296d04839c5ab04c0/core/src/main/java/org/apache/calcite/rex/RexLiteral.java#L231) or when building using the RexBuilder.

There's a [ticket for this on Calcite's board](https://issues.apache.org/jira/browse/IGNITE-18798) but nothing as of of now.

one hacky solution would be:
1. convert the substrait Expression.UDTLiteral into a VARBINARY/ByteString type. (the issue with this is that the Userdatatype.getsqlName would have to be set to sqlname.varbinary, otherwise the type check in calcite will fail)
2. and then when converting back, just check if the RelDatatype inherits from our UserDataType and use that to do the conversion correctly.

### DIFF
--- a/core/src/main/java/io/substrait/expression/proto/ExpressionProtoConverter.java
+++ b/core/src/main/java/io/substrait/expression/proto/ExpressionProtoConverter.java
@@ -239,7 +239,6 @@ public class ExpressionProtoConverter implements ExpressionVisitor<Expression, R
 
   @Override
   public Expression visit(io.substrait.expression.Expression.UserDefinedLiteral expr) {
-
     var typeReference =
         extensionCollector.getTypeReference(SimpleExtension.TypeAnchor.of(expr.uri(), expr.name()));
     return lit(
--- a/core/src/main/java/io/substrait/expression/proto/ExpressionProtoConverter.java
+++ b/core/src/main/java/io/substrait/expression/proto/ExpressionProtoConverter.java
@@ -1,5 +1,7 @@
 package io.substrait.expression.proto;
 
+import com.google.protobuf.Any;
+import com.google.protobuf.InvalidProtocolBufferException;
 import io.substrait.expression.ExpressionVisitor;
 import io.substrait.expression.FieldReference;
 import io.substrait.expression.FunctionArg;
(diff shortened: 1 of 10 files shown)


vbarua:
Thanks for working on this!

One thing I think would be helpful would be to add a full roundtrip test to verify that Substrait inputs with user-defined type literals can go from `PROTO -> POJO -> CALCITE` and back fully.

There's an example for testing user-defined types in [CustomFunctionTest](https://github.com/substrait-io/substrait-java/blob/main/isthmus/src/test/java/io/substrait/isthmus/CustomFunctionTest.java), it could make sense to add a test to this class (maybe with a rename) or make a new class.

There is code for something like this in [assertFullRoundTrip(Rel pojo1)](https://github.com/substrait-io/substrait-java/blob/63dd3051958eea650229ed0926161ae0edd40700/isthmus/src/test/java/io/substrait/isthmus/PlanTestBase.java#L185C18-L185C48), however that assumes that only standard extensions and functions are loaded.

core/src/main/java/io/substrait/expression/proto/ExpressionProtoConverter.java
```diff
@@ -217,6 +219,28 @@ public Expression visit(io.substrait.expression.Expression.StructLiteral expr) {
...
+        bldr -> {
+          try {
+            bldr.setNullable(expr.nullable())
+                .setUserDefined(
+                    Expression.Literal.UserDefined.newBuilder()
+                        .setTypeReference(
+                            expr.getType()
+                                .accept(typeProtoConverter)
+                                .getUserDefined()
+                                .getTypeReference())
```
vbarua, line 235: I think it would be a bit clearer here if we retrieve the type reference we need directly using
```java
      var typeReference = extensionCollector.getTypeReference(SimpleExtension.TypeAnchor.of(expr.uri(), expr.name()));
```

I had to trace through the type conversion here to verify it did this.

cheikhachraf: done

vbarua (approved):
Nice work finding the REINTERPRET code in Calcite, that works really well for this ✨ 

Overall the changes look good. I did spruce things up a bit by:
* Updating the test to include a check going from POJO -> PROTO -> POJO
* Turned your comment into a docstring on the REINTERPRET converter to be able to link to the code in question.

vbarua:
We've had enough activity this week for there to have been merge conflicts 😅 
I rebased this PR off of the current `main` branch to deal with this.

Issue #195, opened 2023-11-02
deterministic build failures due to Jabel


vbarua:
Periodically, Jabel does not appear to trigger correctly during builds and results in builds failing, locally and in CI, due to errors related to features it should be transforming.

See https://github.com/substrait-io/substrait-java/pull/193#issuecomment-1791675960 for an example

vbarua:
The pull request I made on-top of the prior change builds fine, even though the code that was causing the error has been updated.
https://github.com/substrait-io/substrait-java/pull/194

PR #233, opened 2024-02-22
ci: disable task caches in gradle build

bvolpato:
It seems that we have a lot of flakiness on the `Build and Test Java > Build with Gradle` step, and there's some suspicion that it might come from caching intermediate results and Jabel.

Adding `--rerun-tasks` seems like a good balance -- we don't disable cache completely (so we keep Maven artifacts) but we rebuild any unchanged dependencies to prevent such problems.


vbarua (approved):
This is an attempt to workaround the issue described in https://github.com/substrait-io/substrait-java/issues/195, which also affects CI.

PR #234, opened 2024-02-23
feat(isthmus): add WindowRelFunctionConverter

bvolpato:
We recently added POJOs for ConsistentPartitionWindowRel and WindowRelFunction. https://github.com/substrait-io/substrait-java/pull/231

The next step is starting to get isthmus support for it, so I went ahead and created WindowRelFunctionConverter so we can start using it to properly bind our window function calls.

It is heavily based on WindowFunctionConverter (although both the input and output changes - input from `RexOver` to `RexWinAggCall`, and output from `WindowFunctionInvocation` to `WindowRelFunctionInvocation`), so I extracted some converters to top-level classes to improve reusability.

### DIFF
--- a/isthmus/src/main/java/io/substrait/isthmus/expression/SortFieldConverter.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/expression/SortFieldConverter.java
@@ -1,0 +1,32 @@
+package io.substrait.isthmus.expression;
+
+import io.substrait.expression.Expression;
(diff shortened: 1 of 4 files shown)


vbarua (approved):
Thanks for adding this in!

Extracting the SortFieldConverter and WindowBoundConverter should be useful for anyone trying to implement or extend the function converters, and the WindowRelFunctionConverter should help anyone experimenting with ConsistentPartitionWindowRel.

Issue #192, opened 2023-10-31
[ISTHMUS] ClassCastException when converting ORDER BY clause which isn't a column reference


patientstreetlight:
The substrait for a SQL query like:

```
SELECT c FROM t ORDER BY c + 1
```

is currently failing to convert to calcite with a `ClassCastException`, since it doesn't expect the `ORDER BY` clause to be arbitrarily complex expressions, only `RexSlot`s.

```
java.lang.ClassCastException: class org.apache.calcite.rex.RexLiteral cannot be cast to class org.apache.calcite.rex.RexSlot (org.apache.calcite.rex.RexLiteral and org.apache.calcite.rex.RexSlot are in unnamed module of loader 'app')
	at io.substrait.isthmus.SubstraitRelNodeConverter.toRelFieldCollation(SubstraitRelNodeConverter.java:347)
	at io.substrait.isthmus.SubstraitRelNodeConverter.lambda$visit$7(SubstraitRelNodeConverter.java:318)
```

Full substrait file:
[class-cast-exception.json](https://github.com/substrait-io/substrait-java/files/13218405/class-cast-exception.json)

nielspardon:
This appears to be resolved. Full expressions in `ORDER BY` are now supported by #322 (`feat(isthmus): support full expressions in SortRel`).

The `Sort` visitor previously cast each sort-field `RexNode` to `RexSlot`, which failed for anything other than a plain field reference. It now routes each field through `relBuilder.sort(...)` with the full `RexNode`, so arbitrary expressions like `ORDER BY c + 1` convert without a `ClassCastException`.

Coverage exists in `ComplexSortTest` — e.g. `handleComplex2ExpressionSort` exercises `ORDER BY a::INT DESC, -b + 42 ASC NULLS LAST`, which includes arithmetic in the sort key.

Closing as fixed.

PR #235, opened 2024-02-27
[WIP] fix(isthmus): fix rel converter for sort when slot is wrapped

bvolpato:
Attempt to fix https://github.com/substrait-io/substrait-java/issues/192

It works for cases in which RexSlot is wrapped (as we visit + recurse over all the expressions), but it doesn't work yet for literals. Unclear what we should do, I'll do some more digging.


vbarua:
The [RelCollation](https://calcite.apache.org/javadocAggregate/org/apache/calcite/rel/RelCollation.html) class, which is used to capture the ordering in Calcite Sort relations, only appears to support column inputs (and not arbitrary expressions).

This is actually fairly similar to the issue of complex expressions in Aggregate group keys. Calcite only allows field references to be used as grouping keys, but Substrait is more permissive. The approach that was taken for that was to add a Project before the Aggregate to execute the expressions and provide field references for use in the Aggregate. That could potentially work here.

See https://github.com/substrait-io/substrait-java/pull/214

PR #237, opened 2024-03-07
fix: match functions using coercion first instead of least restrictive

bvolpato:
When trying to bind a function such as `LAG(FP64, 1)`, due to LEAD/LAG having function variants that take a single input (`any`), the current matcher is using least restrictive matcher using types (FP64, I32), and ends up matching the wrong function.

The `matchCoerced` function added by https://github.com/substrait-io/substrait-java/pull/226 handles this case correctly, but because the least restrictive takes precedence, it doesn't use proper types.

Inverting the matchers didn't break any of the existing tests, and I added tests to cover the LEAD/LAG scenarios, so I think this is safe.

### DIFF
--- a/isthmus/src/test/java/io/substrait/isthmus/WindowFunctionTest.java
+++ b/isthmus/src/test/java/io/substrait/isthmus/WindowFunctionTest.java
@@ -2,7 +2,11 @@ package io.substrait.isthmus;
 
 import static org.junit.jupiter.api.Assertions.assertThrows;
 
+import io.substrait.expression.Expression;
+import io.substrait.expression.WindowBound;
+import io.substrait.relation.Rel;
 import java.io.IOException;
+import java.util.List;
 import org.apache.calcite.sql.parser.SqlParseException;
 import org.junit.jupiter.api.Nested;
 import org.junit.jupiter.api.Test;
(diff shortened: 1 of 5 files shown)


isthmus/src/test/java/io/substrait/isthmus/WindowFunctionTest.java
```diff
@@ -170,4 +184,76 @@ void rejectQueriesWithIgnoreNulls() {
...
+
+  @ParameterizedTest
+  @ValueSource(strings = {"lag", "lead"})
+  void lagLeadFunctions(String function) {
+    Rel rel =
+        substraitBuilder.project(
+            input ->
+                List.of(
+                    substraitBuilder.windowFn(
+                        "/functions_arithmetic.yaml",
```
vbarua, line 196: minor: we have `DefaultExtensionCatalog.FUNCTIONS_ARITHMETIC` for this

isthmus/src/test/java/io/substrait/isthmus/WindowFunctionTest.java
```diff
@@ -170,4 +184,76 @@ void rejectQueriesWithIgnoreNulls() {
...
+
+  @ParameterizedTest
+  @ValueSource(strings = {"lag", "lead"})
+  void lagLeadWithOffset(String function) {
+    Rel rel =
+        substraitBuilder.project(
+            input ->
+                List.of(
+                    substraitBuilder.windowFn(
+                        "/functions_arithmetic.yaml",
```
vbarua, line 219: ditto above

isthmus/src/test/java/io/substrait/isthmus/WindowFunctionTest.java
```diff
@@ -170,4 +184,76 @@ void rejectQueriesWithIgnoreNulls() {
...
+
+  @ParameterizedTest
+  @ValueSource(strings = {"lag", "lead"})
+  void lagLeadWithOffsetAndDefault(String function) {
+    Rel rel =
+        substraitBuilder.project(
+            input ->
+                List.of(
+                    substraitBuilder.windowFn(
+                        "/functions_arithmetic.yaml",
```
vbarua, line 243: ditto above

vbarua (approved):
Changes look good ✨

PR #238, opened 2024-03-07
build: update dependencies

vbarua:


isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
```diff
@@ -316,6 +316,7 @@ private AggregateCall fromMeasure(Aggregate.Measure measure) {
         distinct,
         false,
         false,
+        Collections.emptyList(),
```
vbarua, line 319: The API for `AggregateCall.create` changed between Calcite 0.34 and 0.36.

bvolpato (approved):
Looks much cleaner, thank you!

bvolpato:
Oopsie Graal checks don't like a Calcite bump

> com.google.common.collect.CollectCollectors was unintentionally initialized at build time. To see why com.google.common.collect.CollectCollectors got initialized use --trace-class-initialization=com.google.common.collect.CollectCollectors
>
> org.apache.calcite.util.ReflectUtil was unintentionally initialized at build time. To see why org.apache.calcite.util.ReflectUtil got initialized use --trace-class-initialization=org.apache.calcite.util.ReflectUtil

bvolpato:
Ah ok, we already have exceptions [here](https://github.com/substrait-io/substrait-java/blob/main/isthmus/build.gradle.kts#L109), we could perhaps add those 2 now.

gradle.properties
```diff
@@ -13,11 +13,15 @@ org.jetbrains.gradle.plugin.idea-ext.version=0.5
...
+# library version
+antlr.version=4.13.1
+calcite.version=1.36.0
+guava.version=32.1.3-jre
+immutables.version=2.10.1
+jackson.version=2.16.1
+junit.version=5.8.1
 protobuf.version=3.17.1
 slf4j.version=1.7.25
-jackson.version=2.12.4
```
vbarua, line 20: Extracted the versions of some common dependencies to make it easier to keep them in sync. Also made sure these were being used in the build.

isthmus/build.gradle.kts
```diff
@@ -71,43 +71,126 @@ java {
...
-  )
```
vbarua, line 110: Extracted this out into a list which I then turn into a string for readability purposes. Also sorted the values.

bvolpato: Looks much better 👏

bvolpato (approved):

EpsilonPrime (approved):

PR #239, opened 2024-03-21
feat: add some more BaseProtoTypes

xiedeyantu:


core/src/main/java/io/substrait/type/proto/BaseProtoTypes.java
```diff
@@ -41,6 +48,13 @@ public BaseProtoTypes(Type.Nullability nullability) {
     INTERVAL_DAY = wrap(Type.IntervalDay.newBuilder().setNullability(nullability).build());
     INTERVAL_YEAR = wrap(Type.IntervalYear.newBuilder().setNullability(nullability).build());
     UUID = wrap(Type.UUID.newBuilder().setNullability(nullability).build());
+    VARCHAR = wrap(Type.VarChar.newBuilder().setNullability(nullability).build());
+    FIXED_CHAR = wrap(Type.FixedChar.newBuilder().setNullability(nullability).build());
+    FIXED_BINARY = wrap(Type.FixedBinary.newBuilder().setNullability(nullability).build());
+    DECIMAL = wrap(Type.Decimal.newBuilder().setNullability(nullability).build());
+    MAP = wrap(Type.Map.newBuilder().setNullability(nullability).build());
```
EpsilonPrime, line 55: Without a type you're going to need to get a builder, add the type, and then build it again.  Is it worth having a base type for this case?

core/src/main/java/io/substrait/type/proto/BaseProtoTypes.java
```diff
@@ -22,6 +22,13 @@ abstract class BaseProtoTypes<T, I> {
   public final T INTERVAL_DAY;
   public final T INTERVAL_YEAR;
   public final T UUID;
+  public final T VARCHAR;
+  public final T FIXED_CHAR;
+  public final T FIXED_BINARY;
+  public final T DECIMAL;
+  public final T MAP;
+  public final T LIST;
+  public final T STRUCT;
```
EpsilonPrime, line 31: Timestamp was recently deprecated in favor of PrecisionTimestamp.  Should we add it here as well (although it additionally has the precision which might be a similar issue to decimal.

core/src/main/java/io/substrait/type/proto/BaseProtoTypes.java
```diff
@@ -41,6 +48,13 @@ public BaseProtoTypes(Type.Nullability nullability) {
     INTERVAL_DAY = wrap(Type.IntervalDay.newBuilder().setNullability(nullability).build());
     INTERVAL_YEAR = wrap(Type.IntervalYear.newBuilder().setNullability(nullability).build());
     UUID = wrap(Type.UUID.newBuilder().setNullability(nullability).build());
+    VARCHAR = wrap(Type.VarChar.newBuilder().setNullability(nullability).build());
+    FIXED_CHAR = wrap(Type.FixedChar.newBuilder().setNullability(nullability).build());
+    FIXED_BINARY = wrap(Type.FixedBinary.newBuilder().setNullability(nullability).build());
+    DECIMAL = wrap(Type.Decimal.newBuilder().setNullability(nullability).build());
```
EpsilonPrime, line 54: Decimal also has precision (and scale). Should we have a method for creating a decimal instead?

core/src/main/java/io/substrait/type/proto/BaseProtoTypes.java
xiedeyantu: I think this pr can be close, this is not a good adding.

EpsilonPrime:
A number of these are fine such as interval_day.  I'll come up with a list.

core/src/main/java/io/substrait/type/proto/BaseProtoTypes.java
EpsilonPrime: VARCHAR, FIXED_CHAR, FIXED_BINARY all should be good to go.

DECIMAL, MAP, LIST, and STRUCT all have some sort of parameter and thus need to be functions.

EpsilonPrime: Hmm, except that those three types all expect to use the literal size as their parameter so without even a literal they aren't fully specified.  I agree, there aren't any more types that work with a simple type.

If we do want something for these they'd have to be functions (perhaps returning cached values) but there's not much perceivable gain there.

xiedeyantu: > Hmm, except that those three types all expect to use the literal size as their parameter so without even a literal they aren't fully specified. I agree, there aren't any more types that work with a simple type.
> 
> If we do want something for these they'd have to be functions (perhaps returning cached values) but there's not much perceivable gain there.

Maybe VARCHAR, FIXED_CHAR, FIXED_BINARY also need a size parameter？

EpsilonPrime (approved):

core/src/main/java/io/substrait/type/proto/BaseProtoTypes.java
EpsilonPrime: I do wonder if these are going to be directly usable without a size parameter.  But I haven't written any code using these.  Let's see if someone else has an opinion.

Issue #248, opened 2024-04-23
Split isthmus into CLI and library components


bestbeforetoday:
The isthmus component currently contains both library code that is useful to other components, and a CLI command implementation. This means that code that depends on the isthmus component as a library is forced to pull in CLI-specific dependencies, which includes a concrete logging provider (as described in issue #243).

I propose splitting the current content of the isthmus component into two: a library and CLI component. The library component could remain as _isthmus_ since projects likely already depend on that component, or could be renamed to something like _substrait-calcite_. The CLI portion could be named _isthmus-cli_, or could remain as _isthmus_ if the library component used a different name.

bestbeforetoday:
If you agree with this approach, I would be happy to take on this change.

vbarua:
I brought this up in the substrait sync yesterday and the response was generally positive.

In terms of your proposal, I think splitting out the CLI component into `isthmus-cli` and leaving the core package as `isthmus` makes the most sense, as this will minimize the amount of works our downstreams will need to do to migrate.

bestbeforetoday:
Resolved by PR #249 and published in [v0.31.0](https://github.com/substrait-io/substrait-java/releases/tag/v0.31.0).

PR #244, opened 2024-04-17
feat: allow deployment time selection of logging framework #243

bestbeforetoday:
Depend on slf4j-api instead of the slf4j-jdk14 logging framework. This allows consumers to select their preferred logging framework at deployment time by adding an appropriate SLF4J provider to their runtime classpath.
    
If no SLF4J provider is available at runtime, no logging output will be produced. The code still functions as expected.
    
The isthmus CLI command continues to log using java.util.logging by including the slf4j-jdk14 provider in its runtime classpath.

PR #249, opened 2024-04-25
feat(isthumus): extract isthmus CLI into isthmus-cli project #248

bestbeforetoday:
This avoid transitive dependencies on graal-sdk and picocli components for projects that consume isthmus as a library.


bestbeforetoday:
Reverted change to .editorconfig (despite helping me to get correct formatting locally) as the change applied to all files and so fails indentation checks for some file types.

bestbeforetoday:
Once #244 is merged, I will rebase on that change and mark this PR for review.

PR #252, opened 2024-04-26
fix: handle LIMIT ALL queries using sentinel value in FetchRel::count

vbarua:
FetchRels with no count set could not be distinguished from FetchRels
with count set to 0.

To account for this, the spec now uses -1 as a sentinel value for this.

feat: new limit and offset methods on SubstraitBuilder

### DIFF
--- a/isthmus/src/main/java/io/substrait/isthmus/SubstraitRelVisitor.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/SubstraitRelVisitor.java
@@ -29,6 +29,7 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
+import java.util.OptionalLong;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 import org.apache.calcite.rel.RelFieldCollation;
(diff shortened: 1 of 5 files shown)


vbarua:
This work is based on changes introduced in https://github.com/substrait-io/substrait/pull/627

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelVisitor.java
```diff
@@ -283,22 +284,30 @@ public Rel visit(org.apache.calcite.rel.core.Match match) {
...
-        sort.getCollation().getFieldCollations().stream()
-            .map(t -> toSortField(t, input.getRecordType()))
-            .collect(java.util.stream.Collectors.toList());
-    var convertedSort = Sort.builder().addAllSortFields(fields).input(input).build();
-    if (sort.fetch == null && sort.offset == null) {
-      return convertedSort;
+    Rel input = apply(sort.getInput());
+    Rel output = input;
+
+    // SortRel is added BEFORE FetchRel, if present
```
EpsilonPrime, line 290: Before can be a bit confusing -- perhaps make it clearer that one is operating on the other?

EpsilonPrime (approved):

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelVisitor.java
vbarua: This is a bit ambiguous for sure. Added a bit more language around this.

EpsilonPrime (approved):

PR #255, opened 2024-05-03
fix: account for struct fields in VirtualTableScan check

### LINKED ISSUE #254: StructTypes in a VirtualTableScan
Hey! I've been working on a Spark to Substrait (and back) converter, forked from https://github.com/apache/incubator-gluten/tree/v1.1.0/substrait/substrait-spark  / https://github.com/substrait-io/substrait-java/pull/90/files (currently the fork is private but I hope to open source it too)

While aiming to support struct types in LocalRelations/VirtualTables, I ran into the check here:
https://github.com/substrait-io/substrait-java/blob/72bab63edf6c4ffb12c3c4b0e4f49d066e0c5524/core/src/main/java/io/substrait/relation/VirtualTableScan.java#L33

If I've understood right, the "names" here should be a flattened list of field names, including column names but also recursively all names from struct types. This also aligns with the [code in Isthmus](https://github.com/substrait-io/substrait-java/blob/72bab63edf6c4ffb12c3c4b0e4f49d066e0c5524/isthmus/src/main/java/io/substrait/isthmus/TypeConverter.java#L135).  Overall, that means the check `r.fields().size() != names.size()` will trigger since there will be more names than top-level fields.

I'm pretty new to Substrait still so I may also be mistaken, but if my understanding is right, would it make sense to either:
a) remove the check,
b) change the check to confirm that names.size() >= r.fields().size(), or
c) iterate through fields to count the sub-fields as well before comparing the sizes?

Blizzara:
Fixes #254 and adds a test

### DIFF
--- a/core/src/test/java/io/substrait/relation/VirtualTableScanTest.java
+++ b/core/src/test/java/io/substrait/relation/VirtualTableScanTest.java
@@ -1,0 +1,52 @@
+package io.substrait.relation;
+
+import static io.substrait.expression.ExpressionCreator.*;
+import static org.junit.jupiter.api.Assertions.*;
+
+import io.substrait.expression.Expression;
+import java.util.HashMap;
+import java.util.Map;
+import org.junit.jupiter.api.Test;
+
+class VirtualTableScanTest {
+
+  @Test
+  void check() {
+    VirtualTableScan virtualTableScan =
+        ImmutableVirtualTableScan.builder()
+            .addDfsNames(
+                "string",
+                "struct",
+                "struct_field1",
+                "struct_field2",
+                "list",
+                "list_struct_field1",
+                "map",
+                "map_key_struct_field1",
+                "map_value_struct_field1")
+            .addRows(
+                struct(
+                    false,
+                    string(false, "string_val"),
+                    struct(
+                        false,
+                        string(false, "struct_field1_val"),
+                        string(false, "struct_field2_val")),
+                    list(false, struct(false, string(false, "list_struct_field1_val"))),
+                    map(
+                        false,
+                        mapOf(
+                            struct(false, string(false, "map_key_struct_field1_val")),
+                            struct(false, string(false, "map_value_struct_field1_val"))))))
+            .build();
+    assertDoesNotThrow(virtualTableScan::check);
+  }
+
+  private Map<Expression.Literal, Expression.Literal> mapOf(
+      Expression.Literal key, Expression.Literal value) {
+    // Map.of() comes only in Java 9 and the "core" module is on Java 8
+    HashMap<Expression.Literal, Expression.Literal> map = new HashMap<>();
+    map.put(key, value);
+    return map;
+  }
+}
--- a/core/src/test/java/io/substrait/relation/VirtualTableScanTest.java
+++ b/core/src/test/java/io/substrait/relation/VirtualTableScanTest.java
@@ -1,7 +1,10 @@
 package io.substrait.relation;
 
-import static io.substrait.expression.ExpressionCreator.*;
-import static org.junit.jupiter.api.Assertions.*;
+import static io.substrait.expression.ExpressionCreator.list;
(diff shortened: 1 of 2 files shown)


core/src/test/java/io/substrait/relation/VirtualTableScanTest.java
```diff
@@ -0,0 +1,52 @@
...
+                "struct_field1",
+                "struct_field2",
+                "list",
+                "list_struct_field1",
+                "map",
+                "map_key_struct_field1",
+                "map_value_struct_field1")
+            .addRows(
+                struct(
+                    false,
```
vbarua, line 29: If I'm understanding the naming correctly, doesn't this boolean correspond to the name "string", and then the string below it corresponds to the name "struct"?

core/src/test/java/io/substrait/relation/VirtualTableScanTest.java
```diff
@@ -0,0 +1,52 @@
...
+                    struct(
+                        false,
+                        string(false, "struct_field1_val"),
+                        string(false, "struct_field2_val")),
+                    list(false, struct(false, string(false, "list_struct_field1_val"))),
+                    map(
+                        false,
+                        mapOf(
+                            struct(false, string(false, "map_key_struct_field1_val")),
+                            struct(false, string(false, "map_value_struct_field1_val"))))))
```
vbarua, line 40: If I understand how the naming stuff is supposed to work here, I believe only nested _structs_ contribute to names in the schema, so its not valid to have names for structs accessed _through_ lists or maps.

This is based on the example in https://substrait.io/tutorial/sql_to_substrait/#types-and-schemas

I'm not super familiar with how this is supposed to work though.

cc: @jacques-n

Blizzara: Hmm, the example only has a list containing non-structs - so then there are no nested names. I don't see anything there indicating that a struct nested inside a list shouldn't contribute to names, and while it doesn't explicitly mention lists, I believe the part here would mean them as well: `Here, names is the names of all fields. In nested schemas, this includes the names of subfields in depth-first order.` It seems to me to make sense to include them, given SQL does allow names within structs inside lists as well :)

Isthmus does include nested fields from inside arrays:
```
./isthmus-cli/build/graal/isthmus "select col from t" -c "create table t (col row(sf1 int null, sf2 varchar) array)"
..
             "baseSchema": {
                "names": ["COL", "SF1", "SF2"],
                "struct": {
                  "types": [{
                    "list": {
                      "type": {
                        "struct": {
                          "types": [{
                            "i32": {
                              "typeVariationReference": 0,
                              "nullability": "NULLABILITY_NULLABLE"
                            }
                          }, {
..
```

Blizzara:
@vbarua @jacques-n I've signed the CLA and think this would be ready for review/to be merged!

core/src/test/java/io/substrait/relation/VirtualTableScanTest.java
jacques-n: Yeah, structs insides lists can have names when output. Yes, the names should be included in the depth-first naming list.

core/src/test/java/io/substrait/relation/VirtualTableScanTest.java
```diff
@@ -0,0 +1,52 @@
+package io.substrait.relation;
+
+import static io.substrait.expression.ExpressionCreator.*;
+import static org.junit.jupiter.api.Assertions.*;
```
vbarua, line 4: Could you import these explicitly instead of with `*`.

Blizzara: Done! [86a19aa](https://github.com/substrait-io/substrait-java/pull/255/commits/86a19aae9697c89f16f335ffd5994d0328b6fae8)

vbarua (approved):

PR #256, opened 2024-05-07
feat: support MATCH_RECOGNIZE

vbarua:
The code in this PR is me thinking out loud about modelling MATCH_RECOGNIZE in Substrait.
It builds off of the work in https://github.com/substrait-io/substrait/pull/636.

This should not be considered final. The primary reason I've wired this into substrait-java is to be able to produce example plans quickly.


vbarua:
Halting this work for now due to changes in priorities.

PR #209, opened 2023-12-06
feat(isthmus): support for SQL expressions in CLI

davisusanibar:
To close https://github.com/substrait-io/substrait-java/issues/207

PR #258, opened 2024-05-18
chore: shadow ANTLR dependency

Blizzara:
ANTLR had a breaking change in its serialization format between 4.9 and 4.10+, meaning libraries that depend on one are incompatible with libraries that depend on another.

Spark (until Spark 4.x) depends on ANTRL 4.9, so this project bumping ANTLR 4.10+ made it impossible to use both Spark and substrait-java in the same project.

Shadowing and relocating the ANTLR dependency removes the issues with incompatibility since substrait-spark's ANTLR is now independent of any other ANTLRs.

### DIFF
--- a/core/build.gradle.kts
+++ b/core/build.gradle.kts
@@ -9,6 +9,7 @@ plugins {
   id("antlr")
   id("com.google.protobuf") version "0.8.17"
   id("com.diffplug.spotless") version "6.11.0"
+  id("com.github.johnrengelman.shadow") version "7.1.2"
   signing
 }
 
@@ -73,6 +74,13 @@ var JACKSON_VERSION = properties.get("jackson.version")
 var JUNIT_VERSION = properties.get("junit.version")
 var SLF4J_VERSION = properties.get("slf4j.version")
 
+// This allows specifying deps to be shadowed so that they don't get included in the POM file
+val shadowImplementation by configurations.creating
+
+configurations[JavaPlugin.COMPILE_ONLY_CONFIGURATION_NAME].extendsFrom(shadowImplementation)
+
+configurations[JavaPlugin.TEST_IMPLEMENTATION_CONFIGURATION_NAME].extendsFrom(shadowImplementation)
+
 dependencies {
   testImplementation("org.junit.jupiter:junit-jupiter-api:${JUNIT_VERSION}")
   testImplementation("org.junit.jupiter:junit-jupiter-params:${JUNIT_VERSION}")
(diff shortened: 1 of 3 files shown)


isthmus/build.gradle.kts
```diff
@@ -108,6 +108,8 @@ tasks {
     archiveBaseName.set("isthmus")
     manifest { attributes(mapOf("Main-Class" to "io.substrait.isthmus.PlanEntryPoint")) }
   }
+
+  classes { dependsOn(":core:shadowJar") }
```
Blizzara, line 112: This is a bit ugly, but seems to work 🤷

core/build.gradle.kts
```diff
@@ -9,6 +9,7 @@ plugins {
   id("antlr")
   id("com.google.protobuf") version "0.8.17"
   id("com.diffplug.spotless") version "6.11.0"
+  id("com.github.johnrengelman.shadow") version "7.1.2"
```
vbarua, line 12: Just poking around the docs for this, is there any reason not to use the latest version which is 8.1.1?

bestbeforetoday: That would require Gradle v8: https://github.com/johnrengelman/shadow#latest-test-compatibility

vbarua: Ah, that seems like work for a different day then.

core/build.gradle.kts
```diff
@@ -85,14 +93,30 @@ dependencies {
...
   implementation("org.slf4j:slf4j-api:${SLF4J_VERSION}")
   annotationProcessor("org.immutables:value:${IMMUTABLES_VERSION}")
   compileOnly("org.immutables:value-annotations:${IMMUTABLES_VERSION}")
   annotationProcessor("com.github.bsideup.jabel:jabel-javac-plugin:0.4.2")
   compileOnly("com.github.bsideup.jabel:jabel-javac-plugin:0.4.2")
 }
 
+configurations[JavaPlugin.API_CONFIGURATION_NAME].let { apiConfiguration ->
+  // Workaround for https://github.com/gradle/gradle/issues/820
+  apiConfiguration.setExtendsFrom(apiConfiguration.extendsFrom.filter { it.name != "antlr" })
```
vbarua, line 106: That was a rabbit-hole. Thanks for documenting this 🙇

isthmus/build.gradle.kts
```diff
@@ -108,6 +108,8 @@ tasks {
     archiveBaseName.set("isthmus")
     manifest { attributes(mapOf("Main-Class" to "io.substrait.isthmus.PlanEntryPoint")) }
```
vbarua, line 109: I just noticed that we reference PlanEntryPoint here even though that was renamed to `IsthmusEntryPoint` as part of https://github.com/substrait-io/substrait-java/pull/209/files and now that lives in `isthmus-cli` after https://github.com/substrait-io/substrait-java/pull/249/files.

I'm not a Gradle expert so I'm not sure what this is for, but I suspect it's not load-bearing 😅 

In any case, not something we need to fix as part of this change, just noting it down.

vbarua (approved):
Changes look reasonable to me. Thanks for documenting all the Gradle build with comments, it made it much easier to understand what was going on. I will merge this in before the end of the week after CI passes.

PR #261, opened 2024-05-28
fix: map FP32 to REAL

ViggoC:
According to the [type definition of calcite ](https://calcite.apache.org/docs/reference.html#scalar-types), `REAL` is 4 byte floating point and `FLOAT` is synonym for DOUBLE which is 8 byte floating point.

### DIFF
--- a/isthmus/src/main/java/io/substrait/isthmus/TypeConverter.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/TypeConverter.java
@@ -71,7 +71,7 @@ public class TypeConverter {
       case SMALLINT -> creator.I16;
       case INTEGER -> creator.I32;
       case BIGINT -> creator.I64;
-      case FLOAT -> creator.FP32;
+      case REAL -> creator.FP32;
       case DOUBLE -> creator.FP64;
       case DECIMAL -> {
         if (type.getPrecision() > 38) {
@@ -203,7 +203,7 @@ public class TypeConverter {
 
     @Override
     public RelDataType visit(Type.FP32 expr) {
-      return t(n(expr), SqlTypeName.FLOAT);
+      return t(n(expr), SqlTypeName.REAL);
     }
 
     @Override
--- a/isthmus/src/main/java/io/substrait/isthmus/TypeConverter.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/TypeConverter.java
@@ -72,7 +72,7 @@ public class TypeConverter {
       case INTEGER -> creator.I32;
       case BIGINT -> creator.I64;
       case REAL -> creator.FP32;
-      case DOUBLE -> creator.FP64;
+      case FLOAT, DOUBLE -> creator.FP64;
       case DECIMAL -> {
         if (type.getPrecision() > 38) {
           throw new UnsupportedOperationException(
--- a/isthmus/src/main/java/io/substrait/isthmus/expression/LiteralConverter.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/expression/LiteralConverter.java
@@ -104,7 +104,7 @@ public class LiteralConverter {
         }
         throw new UnsupportedOperationException("Unable to handle char type: " + val);
       }
-      case DOUBLE -> fp64(n, bd(literal).doubleValue());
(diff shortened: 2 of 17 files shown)


EpsilonPrime (approved):

isthmus/src/main/java/io/substrait/isthmus/TypeConverter.java
```diff
@@ -71,7 +71,7 @@ private Type toSubstrait(RelDataType type, List<String> names) {
       case SMALLINT -> creator.I16;
       case INTEGER -> creator.I32;
       case BIGINT -> creator.I64;
-      case FLOAT -> creator.FP32;
+      case REAL -> creator.FP32;
       case DOUBLE -> creator.FP64;
```
vbarua, line 75: Somewhat cursedly, even though FLOAT is a synonym for DOUBLE, within Calcite they are still 2 different SqlTypeNames. If we create tables using `FLOAT` now, the type conversion fails because of the now missing conversion.

Could you add a mapping for FLOAT to `FP64` along with a test to confirm that we can handle mapping `SqlTypeName.FLOAT` to it. `CalciteTypeTest.java` would be a good file for the test.

ViggoC: I see, but the problem comes to how do we map FP64 back to the calcite type, it can be FLOAT or DOUBLE. `testType` in CalciteTypeTest requires a 1 to 1 map.

vbarua: > how do we map FP64 back to the calcite type

That's a good question. Given that FLOAT is defined as the synonym for DOUBLE, and that we were already using DOUBLE as the returned type, I think keeping that as is makes the most sense.

> it can be FLOAT or DOUBLE. testType in CalciteTypeTest requires a 1 to 1 map.

I think having a test that checks that DOUBLE roundtrips to FP64 and back (which we already have) and then a separate check that FLOAT gets mapped to FP64 (which we don't have yet) is sufficient here.

isthmus/src/main/java/io/substrait/isthmus/expression/LiteralConverter.java
```diff
@@ -105,7 +105,7 @@ public Expression.Literal convert(RexLiteral literal) {
         throw new UnsupportedOperationException("Unable to handle char type: " + val);
       }
       case DOUBLE -> fp64(n, bd(literal).doubleValue());
-      case FLOAT -> fp32(n, bd(literal).floatValue());
```
vbarua, line 108: We should map FLOAT to `fp64` here.

isthmus/src/main/java/io/substrait/isthmus/TypeConverter.java
vbarua: Oh, you already added this in
https://github.com/substrait-io/substrait-java/pull/261/files#diff-ed9b94983bf865c9d1fd5cb424c02e9b3ab3f04501a853d95a21ab28fbcdcd57R63-R70 ✨

isthmus/src/main/java/io/substrait/isthmus/expression/LiteralConverter.java
ViggoC: I add a test cast in CalciteLiteralTest. But I found out there is a paradox on precision definition of FLOAT in calcite. In  `org.apache.calcite.rex.RexBuilder#clean` calcite a pass a MathContext.DECIMAL32 to create a FLOAT literal.

ViggoC: @vbarua The problem has been fixed by https://github.com/apache/calcite/pull/3492 and released in calcite 1.37. Do you think we should upgrade the version of Calcite?

vbarua: > Do you think we should upgrade the version of Calcite?

I think that's a good idea, but it should probably be it's own PR. Does it make sense to you to upgrade the Calcite version first and then base this branch on top of that work?

vbarua (approved):
Changes look good! Thanks for adding this.

PR #262, opened 2024-05-28
feat(isthmus): support PrecisionTimestamp in isthmus

### LINKED ISSUE #246: integrate PrecisionTimestamp and PrecisionTimestampTZ into Isthmus
The changes in https://github.com/substrait-io/substrait-java/pull/247 added POJOs for PrecisionTimestamp and PrecisionTimestampTZ but did not integrate them into Isthmus.

ViggoC:
Fix #246

### DIFF
--- a/isthmus/src/main/java/io/substrait/isthmus/TypeConverter.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/TypeConverter.java
@@ -225,11 +225,21 @@ public class TypeConverter {
 
     @Override
     public RelDataType visit(Type.PrecisionTimestamp expr) {
+      int maxPrecision = typeFactory.getTypeSystem().getMaxPrecision(SqlTypeName.TIMESTAMP);
+      if (expr.precision() > maxPrecision) {
+        throw new UnsupportedOperationException(
+            "unsupported timestamp precision " + expr.precision() + ", max precision in calcite type system is " + maxPrecision);
+      }
       return t(n(expr), SqlTypeName.TIMESTAMP, expr.precision());
     }
 
     @Override
     public RelDataType visit(Type.PrecisionTimestampTZ expr) throws RuntimeException {
+      int maxPrecision = typeFactory.getTypeSystem().getMaxPrecision(SqlTypeName.TIMESTAMP);
+      if (expr.precision() > maxPrecision) {
+        throw new UnsupportedOperationException(
+            "unsupported timestamp_tz precision " + expr.precision() + ", max precision in calcite type system is " + maxPrecision);
+      }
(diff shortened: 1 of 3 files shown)


ViggoC:
@vbarua The Dry-run release chek fails, how can I fixed it.

ViggoC:
@vbarua All checks have passed, could you review this PR.

isthmus/src/main/java/io/substrait/isthmus/TypeConverter.java
```diff
@@ -241,6 +223,16 @@ public RelDataType visit(Type.Timestamp expr) {
       return t(n(expr), SqlTypeName.TIMESTAMP, 6);
     }
 
+    @Override
+    public RelDataType visit(Type.PrecisionTimestamp expr) {
+      return t(n(expr), SqlTypeName.TIMESTAMP, expr.precision());
+    }
```
vbarua, line 229: One subtle thing about this is that by default this won't actually convert the type correctly if the precision is above 6. For example, adding 9 as one of the precision values to your tests below results in the following error:

```
Expected :PrecisionTimestamp{nullable=true, precision=9}
Actual   :PrecisionTimestamp{nullable=true, precision=6}
```

There is a max precision associated with the Calcite RelDataTypeSystem used when creating an instance of the Calcite type. It can be retrieved as follows:
```java
typeFactory.getTypeSystem().getMaxPrecision(SqlTypeName.TIMESTAMP)
```
The default value of which is 6. With this RelDataTypeSystem, any Substrait Timestamp type with precision greater than 6 gets truncated down to 6. This isn't immediately obvious to users, as I discovered as a user myself a while back.

What do you think about checking that the precision being requested by the user in the Substrait type is within the given RelDataTypeSystem for the conversion? It could be good to throw a RuntimeException here which pushes the user to override the RelDataTypeSystem when this happens.

ViggoC: Yes, I know what you mean, that why I remove the hard code precision limitation, so that the behavior is totally depends on calcite type factory. As for the implicit precision truncation, it's a implement detail in type factory, and it's better to send out the warning by calcite type factory. This feature may confuse all users that not familiar with calcite type system, not only the users of substrait. But checking in the TypeConverter also make sense.

vbarua (approved):
Changes looks good to me.

@ViggoC I made a pair of changes:
* Switch to `String.format` and applied spotless
* Used `TIMESTAMP_WITH_LOCAL_TIME_ZONE` to get the max precision for PrecisionTimestampTZ

If those changes look good to you I can merge this in.

ViggoC:
@vbarua thanks, it looks good to me.

PR #264, opened 2024-05-30
ci: update deprecated GitHub actions

bestbeforetoday:
Actions based on Node versions lower than Node 20 are deprecated. See:

https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20/


vbarua (approved):
Changes look good. Thanks for the updated 🙇

PR #265, opened 2024-05-30
ci: cancel in-progress build on PR update

bestbeforetoday:
This change modifies the pull request build definition so that previous builds of a pull request are cancelled when a new change is pushed. This avoids redundant pull request builds continuing to run to completion and wasting build resources.

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -2,6 +2,11 @@ name: PR Build Check
 
 on:
   pull_request:
+
+concurrency:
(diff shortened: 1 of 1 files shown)


bestbeforetoday:
You can see that a force-push to the PR branch cancelled an in-progress build of the previous PR state:

https://github.com/substrait-io/substrait-java/actions/runs/9305341696

vbarua (approved):
This is super handy ✨ 
Thanks for adding this.

PR #266, opened 2024-05-31
build: bump into calcite 1.37

ViggoC:


ViggoC:
@vbarua Bump calcite version for #261, please take a look, all tests passed locally.

PR #268, opened 2024-06-06
fix(isthmus): correct SLF4J dependency

bestbeforetoday:
- Remove concrete slf4j-jdk14 provider dependency from isthmus.
- Add concrete slf4j-jdk14 provider dependency to isthmus-cli.

Library components should only depend on slf4j-api, allowing consumers to select an appropriate concrete logging provider implementation. Being a self-contained (CLI) application, isthmus-cli can depend on a specific SLF4J provider.


bestbeforetoday:
It looks to me that I introduced this problem when splitting isthmus into isthumus (library) and istmus-cli (application) in PR #249.

PR #269, opened 2024-06-11
build: add vulnerability scan to PR build

bestbeforetoday:
Continue build on vulnerability detection.

Update protobuf dependency to address vulnerabilities:

- CVE-2022-25647
- CVE-2022-3509
- CVE-2022-3510
- CVE-2022-3171

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -34,6 +34,42 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - uses: gradle/actions/wrapper-validation@v3
+  cyclonedx-sbom:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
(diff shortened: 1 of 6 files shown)


bestbeforetoday:
I had to back off from using the OSV-Scanner GitHub Action since multiple scan runs all appear to upload security analysis results with the same name. The current change just uses the osv-scanner CLI command directly and so will not populate the repository's Security tab.

vbarua:
> The current change just uses the osv-scanner CLI command directly and so will not populate the repository's Security tab.

How will we see the results of the scan? Are they added as comments to PRs?

bestbeforetoday:
The scan jobs will fail if a vulnerability is detected and will show up as failed build checks against the PR. Unless they are required checks, it should not prevent changes being merged though.

vbarua (approved):
Changes look good to me. Thanks for adding this capability to our build ✨

PR #272, opened 2024-06-19
fix: set VirtualTableScan schema explicitly

Blizzara:
VirtualTableScan in Substrait is expected to contain both a NamedSchema (field names in dfs form and types) like any other read rel, plus the actual data rows. However substrait-java was ignoring the types of the NamedSchema, instead grabbing type info from the first row of the data. This however is not always sufficient since the data can contain varying nullability row-by-row.

This also just simplifies VirtualTableScan and makes it behave more like one would expect by looking at the Substrait spec as well as other ReadRels

### DIFF
--- a/core/src/main/java/io/substrait/relation/ProtoRelConverter.java
+++ b/core/src/main/java/io/substrait/relation/ProtoRelConverter.java
@@ -334,12 +334,11 @@ public class ProtoRelConverter {
                       .collect(java.util.stream.Collectors.toList()))
               .build());
     }
-    var fieldNames =
-        rel.getBaseSchema().getNamesList().stream().collect(java.util.stream.Collectors.toList());
+
     var builder =
         VirtualTableScan.builder()
             .filter(Optional.ofNullable(rel.hasFilter() ? converter.from(rel.getFilter()) : null))
-            .addAllDfsNames(fieldNames)
+            .initialSchema(NamedStruct.fromProto(rel.getBaseSchema(), protoTypeConverter))
             .rows(structLiterals);
 
     builder
--- a/core/src/main/java/io/substrait/relation/VirtualTableScan.java
+++ b/core/src/main/java/io/substrait/relation/VirtualTableScan.java
@@ -1,7 +1,6 @@
 package io.substrait.relation;
 
 import io.substrait.expression.Expression;
-import io.substrait.type.NamedStruct;
 import io.substrait.type.Type;
 import io.substrait.type.TypeVisitor;
 import java.util.List;
@@ -9,8 +8,6 @@ import org.immutables.value.Value;
 
 @Value.Immutable
 public abstract class VirtualTableScan extends AbstractReadRel {
-
-  public abstract List<String> getDfsNames();
 
   public abstract List<Expression.StructLiteral> getRows();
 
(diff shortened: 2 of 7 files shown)


core/src/main/java/io/substrait/relation/VirtualTableScan.java
```diff
@@ -36,11 +33,6 @@ protected void check() {
             .allMatch(r -> r.getType().accept(new NamedFieldCountingTypeVisitor()) == names.size());
   }
 
-  @Override
-  public final NamedStruct getInitialSchema() {
-    return NamedStruct.of(getDfsNames(), (Type.Struct) getRows().get(0).getType());
```
Blizzara, line 41: The only benefit of this I can think of is to kind of ensure the schema matches between the rows and the initialSchema. However, given substrait allows those to be specified separately, I think it makes sense to keep that separation here?

Also, given substrait literals specify nullability on each literal, the schema might be different for row 0 than for row 1. For the producers, it's often easier to ensure the nullability is correct for the initialSchema than for each literal, since the literals are usually converted independently.

core/src/main/java/io/substrait/relation/ProtoRelConverter.java
```diff
@@ -334,12 +334,11 @@ private VirtualTableScan newVirtualTable(ReadRel rel) {
...
               .build());
     }
-    var fieldNames =
-        rel.getBaseSchema().getNamesList().stream().collect(java.util.stream.Collectors.toList());
+
     var builder =
         VirtualTableScan.builder()
             .filter(Optional.ofNullable(rel.hasFilter() ? converter.from(rel.getFilter()) : null))
-            .addAllDfsNames(fieldNames)
+            .initialSchema(NamedStruct.fromProto(rel.getBaseSchema(), protoTypeConverter))
```
vbarua, line 341: Setting and using the schema directly here is a better idea 👍

core/src/main/java/io/substrait/relation/VirtualTableScan.java
```diff
@@ -10,8 +9,6 @@
 @Value.Immutable
 public abstract class VirtualTableScan extends AbstractReadRel {
 
-  public abstract List<String> getDfsNames();
```
vbarua, line 13: Removing this is a breaking change as it removes the `addAllDfsNames` method on the `ImmutableVirtualTableScanBuilder`, but that's reasonable I think.

It is causing a failure in `SubstraitRelVisitor` which uses the `addAllDfsNames`, but we should be able to replace that with a call to `initialSchema`.

core/src/main/java/io/substrait/relation/VirtualTableScan.java
vbarua: > given substrait allows those to be specified separately, I think it makes sense to keep that separation here?

That makes sense to me and it also avoids the issue you highligh

> the schema might be different for row 0 than for row 1.

vbarua (changes requested):
I like this change. Thanks for proposing it! The build is failing, but if you fix that I'm happy to approve and merge this.

core/src/main/java/io/substrait/relation/VirtualTableScan.java
Blizzara: Ah right, turns out there were two tests that had similar issues as well, should be all fixed by https://github.com/substrait-io/substrait-java/pull/272/commits/14b7a1b1a8efa7b9706d6e1fa1f5a5930078b05f

vbarua (approved):

PR #273, opened 2024-06-20
feat: deprecate Timestamp and TimestampTZ visit functions

mbwhite:
By marking the function as @Deprecated and providing a default implementation makes users code a lot cleaner.

Directly implementations of the interface previously would need to implement these methods and cope with the Deprecation warning.

With this change if they are using it, no problems; but they are not obliged to implement it.

### DIFF
--- a/core/src/main/java/io/substrait/type/TypeVisitor.java
+++ b/core/src/main/java/io/substrait/type/TypeVisitor.java
@@ -23,8 +23,10 @@ public interface TypeVisitor<R, E extends Throwable> {
 
   R visit(Type.Time type) throws E;
 
+  @Deprecated
   R visit(Type.TimestampTZ type) throws E;
 
+  @Deprecated
   R visit(Type.Timestamp type) throws E;
 
   R visit(Type.PrecisionTimestamp type) throws E;


core/src/main/java/io/substrait/type/TypeVisitor.java
```diff
@@ -23,9 +23,17 @@ public interface TypeVisitor<R, E extends Throwable> {
 
   R visit(Type.Time type) throws E;
 
-  R visit(Type.TimestampTZ type) throws E;
+  @Deprecated
+  default R visit(Type.TimestampTZ type) throws E {
```
vbarua, line 27: Marking the visitor methods as deprecated makes sense to me, but I'm not sure if providing a default implementation is what we want here. There is `TypeThrowsVisitor` to allow users to implement partial visitors, but ideally the core visitor should force users to be exhaustive.

mbwhite: thanks @vbarua 

I would agree that the `TypeThrowsVisitor` is useful for partial visitor implementations; and that the core visitor should be exhaustive. My thinking is that it's not appropriate or helpful to ask implementations to implement a deprecated method.  (it's a relatively new codebase as well not at v1 as well).. A default implementation could even defer to the new API if there was a 1:1 mapping between old and new APIs (which in this case I think there is).

Marking as `@deprecated` certainly;  will go with your thoughts on the `default` implementation.

vbarua: > My thinking is that it's not appropriate or helpful to ask implementations to implement a deprecated method.

I suspect that those 2 fields will be around for a while (order of months to years) as various other systems will need updates to switch over. I think it's safest to have users of the library keep that in mind by forcing them to handle it. That being said

> default implementation could even defer to the new API if there was a 1:1 mapping between old and new APIs

if we can manage this that would solve this issue for me.

mbwhite: Had a look at being going from the `Timestamp` to the `PrecisionTimestamp` variants.  The PrecisionTimestamps don't have a default value for precision - so unless a value is available to create that when a default implementation won't work. 

@vbarua  would you rather modifying this pr to just do @depecated additions, or close this one and open a fresh one?

vbarua: Let's start with just deprecating for now.

mbwhite: @vbarua I've updated the PR to just include the deprecation annotation

vbarua (approved):
Thanks for the changes!

PR #278, opened 2024-07-02
fix: properly support Options for Scalar/Aggregate/Window functions

Blizzara:
[Options](https://substrait.io/expressions/scalar_functions/#options) were supported in the Java immutables but in most cases dropped when converting into protobuf

This
- fixes the protobuf conversion,
- adds a builder for FunctionOption
- adds a ExpressionCreator functions with options
- does some refactoring

### DIFF
--- a/core/src/main/java/io/substrait/expression/AggregateFunctionInvocation.java
+++ b/core/src/main/java/io/substrait/expression/AggregateFunctionInvocation.java
@@ -3,7 +3,6 @@ package io.substrait.expression;
 import io.substrait.extension.SimpleExtension;
 import io.substrait.type.Type;
 import java.util.List;
-import java.util.Map;
 import org.immutables.value.Value;
 
 @Value.Immutable
@@ -12,7 +11,7 @@ public abstract class AggregateFunctionInvocation {
 
   public abstract List<FunctionArg> arguments();
 
-  public abstract Map<String, FunctionOption> options();
+  public abstract List<FunctionOption> options();
 
   public abstract Expression.AggregationPhase aggregationPhase();
 
--- a/core/src/main/java/io/substrait/expression/ExpressionCreator.java
+++ b/core/src/main/java/io/substrait/expression/ExpressionCreator.java
@@ -285,7 +285,7 @@ public class ExpressionCreator {
       SimpleExtension.ScalarFunctionVariant declaration,
       Type outputType,
       FunctionArg... arguments) {
-      return scalarFunction(declaration, outputType, Arrays.asList(), Arrays.asList(arguments));
+    return scalarFunction(declaration, outputType, Arrays.asList(), Arrays.asList(arguments));
   }
 
   public static Expression.ScalarFunctionInvocation scalarFunction(
--- a/core/src/main/java/io/substrait/expression/ExpressionCreator.java
+++ b/core/src/main/java/io/substrait/expression/ExpressionCreator.java
@@ -10,6 +10,7 @@ import java.math.BigDecimal;
 import java.time.Instant;
 import java.time.LocalDateTime;
 import java.time.ZoneOffset;
+import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
 import java.util.UUID;
@@ -284,20 +285,25 @@ public class ExpressionCreator {
       SimpleExtension.ScalarFunctionVariant declaration,
       Type outputType,
       FunctionArg... arguments) {
-    return Expression.ScalarFunctionInvocation.builder()
-        .declaration(declaration)
-        .outputType(outputType)
-        .addArguments(arguments)
-        .build();
+      return scalarFunction(declaration, outputType, Arrays.asList(), Arrays.asList(arguments));
   }
 
   public static Expression.ScalarFunctionInvocation scalarFunction(
       SimpleExtension.ScalarFunctionVariant declaration,
       Type outputType,
       Iterable<? extends FunctionArg> arguments) {
+    return scalarFunction(declaration, outputType, Arrays.asList(), arguments);
+  }
+
+  public static Expression.ScalarFunctionInvocation scalarFunction(
+      SimpleExtension.ScalarFunctionVariant declaration,
+      Type outputType,
+      List<? extends FunctionOption> options,
+      Iterable<? extends FunctionArg> arguments) {
     return Expression.ScalarFunctionInvocation.builder()
         .declaration(declaration)
         .outputType(outputType)
+        .options(options)
         .addAllArguments(arguments)
         .build();
   }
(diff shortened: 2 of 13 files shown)


core/src/main/java/io/substrait/expression/AggregateFunctionInvocation.java
```diff
@@ -12,7 +11,7 @@ public abstract class AggregateFunctionInvocation {
 
   public abstract List<FunctionArg> arguments();
 
-  public abstract Map<String, FunctionOption> options();
```
Blizzara, line 15: This seemed weird to me, since the name is already kept inside the FunctionOption, so this just duplicates the source of truth.

core/src/main/java/io/substrait/expression/ExpressionCreator.java
```diff
@@ -284,20 +285,25 @@ public static Expression.ScalarFunctionInvocation scalarFunction(
...
+      SimpleExtension.ScalarFunctionVariant declaration,
+      Type outputType,
+      Iterable<? extends FunctionArg> arguments) {
+    return scalarFunction(declaration, outputType, Arrays.asList(), arguments);
   }
 
   public static Expression.ScalarFunctionInvocation scalarFunction(
       SimpleExtension.ScalarFunctionVariant declaration,
       Type outputType,
+      List<? extends FunctionOption> options,
```
Blizzara, line 301: I wasn't sure if it's better to change the existing functions or add new ones given this is an optional parameter, so I added new ones. But happy to remove the old ones too if that'd be better?

core/src/main/java/io/substrait/extendedexpression/ProtoExtendedExpressionConverter.java
```diff
@@ -64,9 +64,12 @@ public ExtendedExpression from(io.substrait.proto.ExtendedExpression extendedExp
           break;
         case MEASURE:
           io.substrait.relation.Aggregate.Measure measure =
-              new ProtoAggregateFunctionConverter(
-                      functionLookup, extensionCollection, protoExpressionConverter)
-                  .from(expressionReference.getMeasure());
+              io.substrait.relation.Aggregate.Measure.builder()
```
Blizzara, line 67: this change made it easier to reuse ProtoAggregateFunctionConverter

core/src/main/java/io/substrait/expression/proto/ProtoExpressionConverter.java
```diff
@@ -393,10 +397,7 @@ public Expression.SortField fromSortField(SortField s) {
         .build();
   }
 
-  public FunctionOption fromFunctionOption(io.substrait.proto.FunctionOption o) {
-    return ImmutableFunctionOption.builder()
-        .name(o.getName())
-        .addAllValues(o.getPreferenceList())
-        .build();
+  public static FunctionOption fromFunctionOption(io.substrait.proto.FunctionOption o) {
```
vbarua, line 400: It is somewhat awkward to make this static so it can be re-used in the `ProtoAggregateFunctionConverter`, but I also can't think of a better place to put this.

core/src/main/java/io/substrait/expression/FunctionOption.java
```diff
@@ -9,4 +9,8 @@ public abstract class FunctionOption {
   public abstract String getName();
 
   public abstract List<String> values();
+
+  public static ImmutableFunctionOption.Builder builder() {
+    return ImmutableFunctionOption.builder();
+  }
```
vbarua, line 15: Good addition ✨

core/src/main/java/io/substrait/expression/AggregateFunctionInvocation.java
vbarua: I imagine this is/was intended to allow for easy lookup of options given a name, so that consumers could look up the relevant options for a given function. However given that:
```
  // Name of the option to set. If the consumer does not recognize the
  // option, it must reject the plan. The name is matched case-insensitively
  // with option names defined for the function.
```
https://github.com/substrait-io/substrait/blob/7dbbf0468083d932a61b9c720700bd6083558fa9/proto/substrait/algebra.proto#L741-L744

It probably is safer to have this be a list and force consumers to process _all_ of them.

core/src/test/java/io/substrait/type/proto/ConsistentPartitionWindowRelRoundtripTest.java
```diff
@@ -36,7 +36,12 @@ void consistentPartitionWindowRoundtrip() {
                         .declaration(windowFunctionDeclaration)
                         // lead(a)
                         .arguments(Arrays.asList(b.fieldReference(input, 0)))
-                        .options(Collections.emptyMap())
+                        .options(
+                            Arrays.asList(
+                                FunctionOption.builder()
+                                    .name("option")
+                                    .addValues("VALUE1", "VALUE2")
+                                    .build()))
```
vbarua, line 44: meta: The [lead function](https://github.com/substrait-io/substrait/blob/7dbbf0468083d932a61b9c720700bd6083558fa9/extensions/functions_arithmetic.yaml#L1750) doesn't actually define any options. Potentially, we should reject both:
* Building plans w/ invalid options.
* Reading protobuf plans in w/ invalid options.

I think this is outside the scope of this PR.

vbarua (commented):
Started taking a look. This seems fairly reasonable so far. I can spend some more time on this tomorrow.

core/src/test/java/io/substrait/type/proto/ConsistentPartitionWindowRelRoundtripTest.java
Blizzara: Fair 😅 but yep agreed it'd be a separate PR

vbarua:
> any ideas why the editorconfig check is complaining? 

Not only where they unrelated to your PR, they also haven't been modified in a long time. I've fixed the issues in:
https://github.com/substrait-io/substrait-java/pull/280

core/src/main/java/io/substrait/relation/RelProtoConverter.java
```diff
@@ -345,13 +348,8 @@ private List<ConsistentPartitionWindowRel.WindowRelFunction> toProtoWindowRelFun
...
               var options =
-                  f.options().entrySet().stream()
-                      .map(
-                          o ->
-                              FunctionOption.newBuilder()
-                                  .setName(o.getKey())
-                                  .addAllPreference(o.getValue().values())
-                                  .build())
+                  f.options().stream()
+                      .map(ExpressionProtoConverter::from)
```
vbarua, line 352: meta: I think re-using this is good, but also think it's a bit weird that this lives in `ExpressionProtoConverter` and is used in `RelProtoConverter`. I'm not sure what a better place for this is yet, so I think it's fine to keep this here for now.

core/src/main/java/io/substrait/relation/ProtoAggregateFunctionConverter.java
```diff
@@ -47,15 +49,17 @@ public io.substrait.relation.Aggregate.Measure from(
...
+            .map(ProtoExpressionConverter::fromFunctionOption)
+            .collect(Collectors.toList());
+    return AggregateFunctionInvocation.builder()
+        .arguments(functionArgs)
+        .declaration(aggregateFunction)
+        .outputType(protoTypeConverter.from(measure.getOutputType()))
+        .aggregationPhase(Expression.AggregationPhase.fromProto(measure.getPhase()))
+        .invocation(Expression.AggregationInvocation.fromProto(measure.getInvocation()))
+        .options(options)
         .build();
```
vbarua, line 63: The other advantage of this is that by forcing callers to wrap this in a Measure themselves, it also makes it easier for them to include the PreMeasure filter when it's available.

core/src/main/java/io/substrait/expression/ExpressionCreator.java
vbarua: I think were getting close to the point where this pattern (individual constructor variants with different field combinations) may not be good as we start to explode the number of function combinations. These are fine for now, but if were to add more beyond these it might make sense to revisit this utility function approach.

vbarua (approved):
Overall this looks reasonable to me.

core/src/main/java/io/substrait/expression/ExpressionCreator.java
Blizzara: That's fair - I decided to revert the additions and add some docstrings to point to using the builder directly in https://github.com/substrait-io/substrait-java/pull/278/commits/9aae3ec7461e9363997d5f7e856bd290b91c9668, does that seem alright to you?

PR #279, opened 2024-07-03
Take `sorts` into account inside aggregate functions

gabotechs:
Aggregate functions can take inner ORDER BY statements that will sort the underlaying data before the aggregation, for example:
```sql
SELECT string_agg("foo", ', ' ORDER BY "foo" DESC) FROM "tbl"
```
In substrait this is reflected as a `sorts` field inside an aggregate function's measure. This PR adds support for loading that field in `ProtoTypeConverter`


gabotechs:
I'm unsure about how to contribute tests for this, any guidance there is welcomed

vbarua:
For testing this, I would suggest adding a sort to the aggregate roundtrip test in 
https://github.com/substrait-io/substrait-java/blob/3e553eee981feb11a64b6c2fef6daf1fe377945a/core/src/test/java/io/substrait/type/proto/AggregateRoundtripTest.java

to make sure that sorts can be read from protos and output to protos.

gabotechs:
No longer working on this

PR #280, opened 2024-07-03
ci: editorconfig updates

vbarua:


vbarua:
Weird. CI still fails for the editorconfig check https://github.com/substrait-io/substrait-java/actions/runs/9781058298/job/27004211456?pr=280 even though those files haven't been modified recently.

core/src/main/antlr/SubstraitType.g4
```diff
@@ -99,83 +99,83 @@ SingleQuote: '\'';
...
 
 Identifier
   : ('a'..'z' | 'A'..'Z' | '_' | '$') ('a'..'z' | 'A'..'Z' | '_' | '$' | Digit)*
   ;
 
 LineComment
-	:	'//' ~[\r\n]* -> channel(HIDDEN)
-	;
+  : '//' ~[\r\n]* -> channel(HIDDEN)
+  ;
```
vbarua, line 111: I'm choosing spaces over tabs here because that it what is used in examples in the ANTLR repo
https://github.com/antlr/grammars-v4/tree/master

.editorconfig
```diff
@@ -19,3 +19,6 @@ trim_trailing_whitespace = unset
 
 [*.gradle.kts]
 indent_size = 2
+
+[.gitmodules]
+indent_style = tab
```
vbarua, line 24: .gitmodule has not been updated in 2 years, but the check is firing (and failing) now for some reason.
Adding an explicit configuration for .gitmodule.

vbarua:
Still not sure why editorconfig started checking these files only now. In any case:
* Added an explicit config to use tabs for .gitmodule
* Updated [SubstraitType.g4](https://github.com/substrait-io/substrait-java/pull/280/files#diff-e399bff231da922efc6f882ffbf20eb852a70b9d9df86b8fee666158d72bb862) to use only spaces, instead of a mix of tabs and spaces.

EpsilonPrime (approved):

PR #283, opened 2024-07-17
feat: support PrecisionTimestamp and PrecisionTimestampTZ literals

Blizzara:


Blizzara:
cc @vbarua

vbarua:
Will look at this today.

core/src/main/java/io/substrait/expression/ExpressionCreator.java
```diff
@@ -102,17 +114,59 @@ public static Expression.TimestampLiteral timestamp(
...
+        .precision(precision)
+        .build();
+  }
+
+  public static Expression.PrecisionTimestampLiteral precisionTimestamp(
+      boolean nullable, LocalDateTime value) {
+    var epochMicro =
+        TimeUnit.SECONDS.toMicros(value.toEpochSecond(ZoneOffset.UTC))
+            + TimeUnit.NANOSECONDS.toMicros(value.toLocalTime().getNano());
+    return precisionTimestamp(nullable, epochMicro, 6);
```
vbarua, line 150: The docs for [LocalDateTime](https://docs.oracle.com/javase/8/docs/api/java/time/LocalDateTime.html) indicate
> Time is represented to nanosecond precision.

Should we be using nanoseconds (9) here instead of microseconds (6)?

core/src/main/java/io/substrait/expression/ExpressionCreator.java
```diff
@@ -102,17 +114,59 @@ public static Expression.TimestampLiteral timestamp(
...
+        .precision(precision)
+        .build();
+  }
+
+  public static Expression.PrecisionTimestampTZLiteral precisionTimestampTZ(
+      boolean nullable, Instant value) {
+    var epochMicro =
+        TimeUnit.SECONDS.toMicros(value.getEpochSecond())
+            + TimeUnit.NANOSECONDS.toMicros(value.getNano());
+    return precisionTimestampTZ(nullable, epochMicro, 6);
```
vbarua, line 167: The docs for [Instant](https://docs.oracle.com/javase/8/docs/api/java/time/Instant.html) indicate that
> the class stores a long representing epoch-seconds and an int representing nanosecond-of-second

Should we be using nanoseconds (9) here instead of microseconds (6)?

Relatedly, are all possible Instant values representable as PrecistionTimestampTz literals?

vbarua (commented):
Overall, this looks good. I have two questions around the ExpressionCreator constructors.

core/src/main/java/io/substrait/expression/ExpressionCreator.java
Blizzara: I have no real preference here - just kept it like this so that it aligns as closely as possible to the existing timestamp creators.

Given the LocalDate (and also Instant) contain more data than just a int64, (they have seconds and then the subseconds separately), I guess there's some dates that cannot be represented as only nanonseconds but can in microseconds. Quick search suggests that's about +- 292 years. So while we can switch this to nanoseconds, it could overflow for some inputs (if those inputs are like year 3000).

Frankly, I think I'd rather keep this as is, to make switching from the earlier timestamp behavior as easy as possible. If people want more control, they can always use the other creator func that lets them decide the precision. Alternatively, we could even consider not adding this second constructor. Thoughts?

core/src/main/java/io/substrait/expression/ExpressionCreator.java
Blizzara: See the other answer. Not all instants are representable as PTTz, I believe, since instant stores in total 96 bits of data, while PTTz is only 64 bits.

core/src/main/java/io/substrait/expression/ExpressionCreator.java
vbarua: > Alternatively, we could even consider not adding this second constructor. 

Let's leave the new constructors out for now. It sounds like we may need to consider their behaviours more carefully now that we're not eliding to microseconds by default, but we don't need to wait to figure that out before merging the rest of this in.

Blizzara: Done! https://github.com/substrait-io/substrait-java/pull/283/commits/f255e6c1293488740b8b5100e4bb0a10d572f554

PR #284, opened 2024-07-18
feat: check that VirtualTableScan field names correspond to schema

vbarua:


core/src/test/java/io/substrait/relation/VirtualTableScanTest.java
```diff
@@ -39,8 +39,8 @@ void check() {
                     R.struct(
                         R.STRING,
                         R.struct(R.STRING, R.STRING),
-                        R.list(R.STRING),
-                        R.map(R.STRING, R.STRING))))
+                        R.list(R.struct(R.STRING)),
+                        R.map(R.struct(R.STRING), R.struct(R.STRING)))))
```
vbarua, line 43: The types here did not match the actual row types, which led to some confusion.

core/src/main/java/io/substrait/relation/VirtualTableScan.java
```diff
@@ -24,13 +24,15 @@ public abstract class VirtualTableScan extends AbstractReadRel {
   @Value.Check
   protected void check() {
     var names = getInitialSchema().names();
+
+    assert names.size() == NamedFieldCountingTypeVisitor.countNames(this.getInitialSchema().struct());
```
vbarua, line 28: It may make more sense to move this check into the NamedStruct itself in the long-term. Avoiding this for now because there's some stuff I'd like to verify around names going to and from Calcite and Substrait.

core/src/test/java/io/substrait/relation/VirtualTableScanTest.java
vbarua, line 43: The types here did not match the actual row types, which led to some confusion.

core/src/main/java/io/substrait/relation/VirtualTableScan.java
```diff
@@ -24,13 +24,16 @@ public abstract class VirtualTableScan extends AbstractReadRel {
   @Value.Check
   protected void check() {
     var names = getInitialSchema().names();
+
+    assert names.size()
+        == NamedFieldCountingTypeVisitor.countNames(this.getInitialSchema().struct());
```
vbarua, line 29: It may make more sense to move this check into the NamedStruct itself in the long-term. Avoiding this for now because there's some stuff I'd like to verify around names going to and from Calcite and Substrait.

EpsilonPrime (approved):

PR #285, opened 2024-08-09
feat: opening ProtoRelConverter for extension

bvolpato:
I have a need to override some of the behavior in [ProtoRelConverter.java](https://github.com/substrait-io/substrait-java/compare/main...bvolpato:open-proto-rel-converter?expand=1#diff-dd48c750d05998223daf061d1d409a5691a8e3c22e93cf68467f393a062b785c)'s `newExtensionTable`, but today I can't because the class was designed to be all private.

I'd like to make that open if possible, to allow easily extending it instead of having to duplicate any code. I'm aware that there are some extension points / customization entries such as `detailFromExtensionTable`, but that signature isn't enough to have roundtrip-able leafs, because I can only derive the schema from the `Rel` itself, not from the data packed in the `Any` instance.

### DIFF
--- a/core/src/main/java/io/substrait/relation/ProtoRelConverter.java
+++ b/core/src/main/java/io/substrait/relation/ProtoRelConverter.java
@@ -117,7 +117,7 @@ public class ProtoRelConverter {
     }
   }
 
-  private Rel newRead(ReadRel rel) {
+  protected Rel newRead(ReadRel rel) {
     if (rel.hasVirtualTable()) {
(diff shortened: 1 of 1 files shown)


vbarua (approved):
This looks reasonable to me. It should allow consumers to extend the ProtoRelConverter behaviour more easily.

jacques-n:
I generally suggest using composition instead of subclassing. Exposing these methods as protected creates some weird testing patterns as things progress. I'd be much more inclined to say let's convert to interface and final this class.

PR #286, opened 2024-08-16
fix(core): row type should output the window function aggregation results

bvolpato:
We've caught a problem where the `deriveRecordType` was very inconsistent with Calcite's `RowType`, and debugging those closely, Substrait was inconsistently adding the partition expression as part of the record type, which doesn't make sense (I mean in general, there might be cases that you are outputting the partition field).

I've fixed the `deriveRecordType` method to output based on the `windowFunctions` instead, and added a test that validates whether the types are still the expected after the round-trip, for the original case and added one new plan that had 2 distinct aggregations.

### DIFF
--- a/core/src/main/java/io/substrait/relation/ConsistentPartitionWindow.java
+++ b/core/src/main/java/io/substrait/relation/ConsistentPartitionWindow.java
@@ -29,7 +29,7 @@ public abstract class ConsistentPartitionWindow extends SingleInputRel implement
         .struct(
             Stream.concat(
                 initial.fields().stream(),
-                getPartitionExpressions().stream().map(Expression::getType)));
+                getWindowFunctions().stream().map(WindowRelFunctionInvocation::outputType)));
   }
 
   @Override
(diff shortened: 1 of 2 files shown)


core/src/main/java/io/substrait/relation/ConsistentPartitionWindow.java
```diff
@@ -29,7 +29,7 @@ protected Type.Struct deriveRecordType() {
         .struct(
             Stream.concat(
                 initial.fields().stream(),
-                getPartitionExpressions().stream().map(Expression::getType)));
+                getWindowFunctions().stream().map(WindowRelFunctionInvocation::outputType)));
```
vbarua, line 32: This is consistent with the language in the [spec](https://substrait.io/relations/physical_relations/#consistent-partition-window-operation)
> Same as Project operator (input followed by each window expression).

vbarua (approved):
Thanks for fixing and adding a test as well.

Issue #290, opened 2024-08-28
[ISTHMUS] WindowFunctionConverter should be able to use a non-default TypeConverter


patientstreetlight:
`FunctionConverter` has 2 constructors, one of which uses a default `TypeConverter` which doesn't know about user-defined types:

```
  public FunctionConverter(List<F> functions, RelDataTypeFactory typeFactory) {
    this(functions, Collections.EMPTY_LIST, typeFactory, TypeConverter.DEFAULT);
  }
```

and the other which allows setting a custom `TypeConverter`:

```
  public FunctionConverter(
      List<F> functions,
      List<FunctionMappings.Sig> additionalSignatures,
      RelDataTypeFactory typeFactory,
      TypeConverter typeConverter) {
```

Some of `FunctionConverter`'s subclasses, like `ScalarFunctionConverter` and `AggregateFunctionConverter`, likewise provide 2 constructors matching those from `FunctionConverter`, which allows setting a custom `TyperConverter` if need be.

But `WindowFunctionConverter` only offers a single constructor which doesn't allow setting `TypeConverter`:
```
  public WindowFunctionConverter(
      List<SimpleExtension.WindowFunctionVariant> functions, RelDataTypeFactory typeFactory) {
    super(functions, typeFactory);
  }
```

PR #291, opened 2024-08-29
feat(isthmus): allow WindowFunctionConverter to use a non-default TypeConverter

patientstreetlight:
This to fix https://github.com/substrait-io/substrait-java/issues/290


bvolpato (approved):
This makes sense 💯 Thanks Chris!

vbarua (approved):
Good idea. This brings these converters in line with the ScalarFunctionConverter and AggregateFunctionConverter.

PR #295, opened 2024-09-20
feat: add ExpandRel support to core and spark

andrew-coleman:
This PR implements the `Expand` relation in the `core` and `spark` modules.
It is required in the spark module to support optimised queries that contain `distinct` aggregations, and increases the number of successful test cases in the TPCDS suite.


spark/src/main/scala/io/substrait/spark/logical/ToSubstraitRel.scala
```diff
@@ -72,7 +71,8 @@ class ToSubstraitRel extends AbstractLogicalPlanVisitor with Logging {
     val substraitExps = expression.aggregateFunction.children.map(toExpression(output))
     val invocation =
       SparkExtension.toAggregateFunction.apply(expression, substraitExps)
-    relation.Aggregate.Measure.builder.function(invocation).build()
+    val filter = expression.filter map toExpression(output)
+    relation.Aggregate.Measure.builder.function(invocation).preMeasureFilter(Optional.ofNullable(filter.orNull)).build()
```
vbarua, line 75: I assume including the filter is an opportunistic bug fix?

I don't think it needs to be it's own PR, but in the future if you could make it a separate commit with a `fix(spark)` style message it makes it easier for me to notice and call out such changes in the release log.

core/src/main/java/io/substrait/dsl/SubstraitBuilder.java
```diff
@@ -313,6 +315,37 @@ private Project project(
...
+      Function<Rel, Iterable<? extends Expand.ExpandField>> fieldsFn,
+      Optional<Rel.Remap> remap,
+      Optional<List<String>> outputNames,
+      Rel input) {
+    var fields = fieldsFn.apply(input);
+    var expand = Expand.builder().input(input).fields(fields).remap(remap);
+    outputNames.ifPresent(
+        names -> expand.hint(RelCommon.Hint.newBuilder().addAllOutputNames(names).build()));
+    return expand.build();
+  }
```
vbarua, line 347: In my opinion I don't think it makes sense to include an explicit `outputNames` parameter like this in these builders.  We should either make variants that pass a full `hints` object, or not include a `hints` param at all and instead have users utilize the pattern of:

```java
Expand.builder.from(
  expand(...)
).hint(...)
.build()
```

I lean towards not including `hints` at all, as I think setting them is very engine specific and if we have `expand` utilize methods with `hints` params it would make sense for all the other methods to also have variants with `hints` and that blows up the method count.

spark/src/main/scala/io/substrait/spark/expression/FunctionMappings.scala
```diff
@@ -67,6 +67,7 @@ class FunctionMappings {
     s[Count]("count"),
     s[Min]("min"),
     s[Max]("max"),
+    s[First]("any_value"),
```
vbarua, line 70: While you can treat `any_value` with `first`, I don't think you can treat `first` as `any_value`. Is this mapping bi-directional?

core/src/main/java/io/substrait/relation/Expand.java
```diff
@@ -0,0 +1,64 @@
...
+  @Value.Immutable
+  public abstract static class ExpandField {
+    public abstract Optional<SwitchingField> getSwitchingField();
+
+    public abstract Optional<Expression> getConsistentField();
+
+    public static ImmutableExpand.ExpandField.Builder builder() {
+      return ImmutableExpand.ExpandField.builder();
+    }
+  }
```
vbarua, line 54: A class with two optional fields, only 1 of which is set at any point in time, feels like a code smell to me. I think this should be an interface `ExpandField` with two implementing classes.

vbarua (changes requested):
Left some questions and some comments.

spark/src/main/scala/io/substrait/spark/logical/ToSubstraitRel.scala
andrew-coleman: Sure, I've split this into a separate commit.

spark/src/main/scala/io/substrait/spark/expression/FunctionMappings.scala
andrew-coleman: Sure, my first proposal was to add the `first` function to `functions_aggregate_general.yaml`, but this suggested was given as an alternative.
https://github.com/substrait-io/substrait/pull/697#issuecomment-2315684133

vbarua: Ah, if Spark itself handles `First` as `any_value` this makes sense. Though truth be told if I'm being pedantic this sounds like Spark doesn't actually support `First` correctly. Not a Substrait problem though.

core/src/main/java/io/substrait/relation/ProtoRelConverter.java
```diff
@@ -389,6 +397,37 @@ protected Project newProject(ProjectRel rel) {
...
+                              case CONSISTENT_FIELD -> Expand.ConsistentField.builder()
+                                  .expression(converter.from(expandField.getConsistentField()))
+                                  .build();
+                              case SWITCHING_FIELD -> Expand.SwitchingField.builder()
+                                  .duplicates(
+                                      expandField.getSwitchingField().getDuplicatesList().stream()
+                                          .map(converter::from)
+                                          .collect(java.util.stream.Collectors.toList()))
+                                  .build();
+                              case FIELDTYPE_NOT_SET -> Expand.ConsistentField.builder().build();
```
vbarua, line 420: IMO this should be an UnsupportedOperationException. If this isn't set correctly the plan is bad.

core/src/main/java/io/substrait/relation/RelProtoConverter.java
```diff
@@ -374,14 +371,44 @@ public Rel visit(Project project) throws RuntimeException {
...
+        .getFields()
+        .forEach(
+            expandField -> {
+              if (expandField instanceof Expand.ConsistentField) {
+                builder.addFields(
+                    ExpandRel.ExpandField.newBuilder()
+                        .setConsistentField(
+                            toProto(((Expand.ConsistentField) expandField).getExpression()))
+                        .build());
+
```
vbarua, line 395: Can we use the
```java
              if (expandField instanceof Expand.ConsistentField cf) {
                builder.addFields(
                    ExpandRel.ExpandField.newBuilder()
                        .setConsistentField(
                            toProto(cf.getExpression()))
                        .build());
```
version of instanceof, or is that too modern?

core/src/main/java/io/substrait/relation/RelProtoConverter.java
```diff
@@ -374,14 +371,44 @@ public Rel visit(Project project) throws RuntimeException {
...
+                builder.addFields(
+                    ExpandRel.ExpandField.newBuilder()
+                        .setSwitchingField(
+                            ExpandRel.SwitchingField.newBuilder()
+                                .addAllDuplicates(
+                                    ((Expand.SwitchingField) expandField)
+                                        .getDuplicates().stream()
+                                            .map(this::toProto)
+                                            .collect(Collectors.toList())))
+                        .build());
```
vbarua, line 406: Can you make an explicit instanceof check for the SwitchingField type here, and then throw an exception on else.

We should have a visitor for ExpandField eventually, but I won't block this PR on it.

core/src/main/java/io/substrait/hint/Hint.java
```diff
@@ -0,0 +1,23 @@
...
+import io.substrait.proto.RelCommon;
+import java.util.List;
+import java.util.Optional;
+import org.immutables.value.Value;
+
+@Value.Immutable
+public abstract class Hint {
+  public abstract Optional<String> getAlias();
+
+  public abstract List<String> getOutputNames();
```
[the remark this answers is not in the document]
vbarua: The protos don't mark these fields as optional, so we can't distinguish between empty and unset can we 😢

vbarua (commented):
Left some minor comments, can do a full pass tomorrow. I want to get this in this week.

core/src/main/java/io/substrait/relation/ProtoRelConverter.java
```diff
@@ -389,6 +397,38 @@ protected Project newProject(ProjectRel rel) {
...
+                                  .build();
+                              case FIELDTYPE_NOT_SET -> throw new UnsupportedOperationException(
+                                  "Expand fields not set");
+                            })
+                    .collect(java.util.stream.Collectors.toList()));
+
+    builder
+        .commonExtension(optionalAdvancedExtension(rel.getCommon()))
+        .remap(optionalRelmap(rel.getCommon()))
+        .hint(optionalHint(rel.getCommon()));
```
vbarua, line 428: Made an issue to track handling of Hint information for _all_ relations: https://github.com/substrait-io/substrait-java/issues/297

spark/src/main/scala/io/substrait/spark/logical/ToLogicalPlan.scala
```diff
@@ -82,11 +81,15 @@ class ToLogicalPlan(spark: SparkSession) extends DefaultRelVisitor[LogicalPlan]
         )
         throw new IllegalArgumentException(msg)
       })
+
+    val filter = Option(measure.getPreMeasureFilter.orElse(null))
+      .map(_.accept(expressionConverter))
```
vbarua, line 86: Note to myself to include this fix in the release notes.

spark/src/main/scala/io/substrait/spark/logical/ToLogicalPlan.scala
```diff
@@ -193,6 +196,27 @@ class ToLogicalPlan(spark: SparkSession) extends DefaultRelVisitor[LogicalPlan]
     }
   }
 
+  override def visit(expand: relation.Expand): LogicalPlan = {
+    val child = expand.getInput.accept(this)
+    val names = expand.getHint.get().getOutputNames.asScala
```
vbarua, line 201: What happens if this is not set? Can you generate names for this on the fly?

I'm okay with leaving this as is for now, but generally when consuming a plan hints should be entirely optional for producers to set.

vbarua (approved):
Left some minor comments, but nothing blocking. I will merge this before Friday at latest (I've set a reminder for myself).

vbarua (approved):
Left some minor comments, but nothing blocking. I will merge this before Friday at latest (I've set a reminder for myself).

PR #296, opened 2024-09-24
feat(spark): support for FetchRel offset field

andrew-coleman:
Add missing support for the ‘offset’ clause in the spark module.

### DIFF
--- a/spark/src/main/scala/io/substrait/spark/logical/ToLogicalPlan.scala
+++ b/spark/src/main/scala/io/substrait/spark/logical/ToLogicalPlan.scala
@@ -160,11 +160,20 @@ class ToLogicalPlan(spark: SparkSession) extends DefaultRelVisitor[LogicalPlan] 
   }
   override def visit(fetch: relation.Fetch): LogicalPlan = {
     val child = fetch.getInput.accept(this)
-    val limit = Literal(fetch.getCount.getAsLong.intValue(), IntegerType)
-    fetch.getOffset match {
-      case 1L => GlobalLimit(limitExpr = limit, child = child)
-      case -1L => LocalLimit(limitExpr = limit, child = child)
-      case _ => visitFallback(fetch)
+    val limit = fetch.getCount.getAsLong.intValue()
+    val offset = fetch.getOffset.intValue()
+    val toLiteral = (i: Int) => Literal(i, IntegerType)
+    if (limit >= 0) {
+      val limitExpr = toLiteral(limit)
+      if (offset > 0) {
+        GlobalLimit(limitExpr,
+          Offset(toLiteral(offset),
+            LocalLimit(toLiteral(offset + limit), child)))
+      } else {
+        GlobalLimit(limitExpr, LocalLimit(limitExpr, child))
+      }
+    } else {
+        Offset(toLiteral(offset), child)
     }
   }
   override def visit(sort: relation.Sort): LogicalPlan = {
--- a/spark/src/main/scala/io/substrait/spark/logical/ToSubstraitRel.scala
+++ b/spark/src/main/scala/io/substrait/spark/logical/ToSubstraitRel.scala
@@ -170,23 +170,37 @@ class ToSubstraitRel extends AbstractLogicalPlanVisitor with Logging {
     case other => throw new UnsupportedOperationException(s"Unknown type: $other")
   }
 
-  private def fetchBuilder(limit: Long, global: Boolean): relation.ImmutableFetch.Builder = {
-    val offset = if (global) 1L else -1L
-    relation.Fetch
-      .builder()
-      .count(limit)
+  private def fetch(child: LogicalPlan, offset: Long, limit: Long = -1): relation.Fetch = {
+    relation.Fetch.builder()
+      .input(visit(child))
       .offset(offset)
-  }
-  override def visitGlobalLimit(p: GlobalLimit): relation.Rel = {
-    fetchBuilder(asLong(p.limitExpr), global = true)
-      .input(visit(p.child))
+      .count(limit)
       .build()
   }
 
+  override def visitGlobalLimit(p: GlobalLimit): relation.Rel = {
+    p match {
+      case OffsetAndLimit((offset, limit, child)) => fetch(child, offset, limit)
+      case GlobalLimit(IntegerLiteral(globalLimit), LocalLimit(IntegerLiteral(localLimit), child))
+        if globalLimit == localLimit => fetch(child, 0, localLimit)
+      case _ =>
+        throw new UnsupportedOperationException(s"Unable to convert the limit expression: $p")
+    }
+  }
+
   override def visitLocalLimit(p: LocalLimit): relation.Rel = {
-    fetchBuilder(asLong(p.limitExpr), global = false)
-      .input(visit(p.child))
-      .build()
+    val localLimit = asLong(p.limitExpr)
+    p.child match {
+      case OffsetAndLimit((offset, limit, child)) if localLimit >= limit =>
+        fetch(child, offset, limit)
+      case GlobalLimit(IntegerLiteral(globalLimit), child) if localLimit >= globalLimit =>
+        fetch(child, 0, globalLimit)
+      case _ => fetch(p.child, 0, localLimit)
+    }
+  }
+
+  override def visitOffset(p: Offset): relation.Rel = {
+    fetch(p.child, asLong(p.offsetExpr))
   }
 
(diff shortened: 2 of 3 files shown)


spark/src/main/scala/io/substrait/spark/logical/ToLogicalPlan.scala
```diff
@@ -157,11 +156,16 @@ class ToLogicalPlan(spark: SparkSession) extends DefaultRelVisitor[LogicalPlan]
...
     val child = fetch.getInput.accept(this)
-    val limit = Literal(fetch.getCount.getAsLong.intValue(), IntegerType)
-    fetch.getOffset match {
-      case 1L => GlobalLimit(limitExpr = limit, child = child)
-      case -1L => LocalLimit(limitExpr = limit, child = child)
-      case _ => visitFallback(fetch)
+    if (fetch.getCount.isPresent) {
+      val limit = Literal(fetch.getCount.getAsLong.intValue(), IntegerType)
+      fetch.getOffset match {
+        case 1L => GlobalLimit(limitExpr = limit, child = child)
```
Blizzara, line 162: fwiw this is just misusing the field, but I know it was there already so not a problem for this PR 😅

spark/src/main/scala/io/substrait/spark/logical/ToLogicalPlan.scala
```diff
@@ -157,11 +156,16 @@ class ToLogicalPlan(spark: SparkSession) extends DefaultRelVisitor[LogicalPlan]
...
+        case -1L => LocalLimit(limitExpr = limit, child = child)
+        case _ => visitFallback(fetch)
```
vbarua, line 164: Can Spark not handle the case where both the `offset` and `count` are set?

spark/src/main/scala/io/substrait/spark/logical/ToLogicalPlan.scala
vbarua: If I understand this, you're using offset `1` or `-1` as a switching field? What's needed to not do this? At the very least, could we used something like `-2` to signal this to avoid using a valid `offset` value as a magic number?

spark/src/main/scala/io/substrait/spark/logical/ToLogicalPlan.scala
acruise: `OFFSET` is supported as of [Spark 3.4](https://spark.apache.org/releases/spark-release-3-4-0.html#features).

andrew-coleman: > Can Spark not handle the case where both the offset and count are set?

Yes, the unit test in this PR sets `limit` and `offset`.   Spark handles `limit` by breaking it into two logical relations `LocalLimit` and `GlobalLimit` (there's an explanation [here](https://towardsdatascience.com/stop-using-the-limit-clause-wrong-with-spark-646e328774f5)).  

In the case of the test query in this PR:

```sql
select l_partkey from lineitem where l_shipdate < date '1998-01-01' " +
        "order by l_shipdate asc, l_discount desc limit 100 offset 1000
```

the spark logical plan is:

```
    GlobalLimit 100
    +- Offset 1000
       +- LocalLimit 1100
          +- Project [l_partkey#7997L]
             +- Sort [l_shipdate#8006 ASC NULLS FIRST, l_discount#8002 DESC NULLS LAST], true
                +- Project [l_partkey#7997L, l_shipdate#8006, l_discount#8002]
                   +- Filter (isnotnull(l_shipdate#8006) AND (l_shipdate#8006 < 1998-01-01))
                      +- Relation ...
```

I'm not sure how this would best translate into a Substrait plan.

The handling of the `limit` clause was already in the codebase as inherited from Gluten.  As @Blizzara noted, this PR adds the `offset`.  But I can understand that this whole area needs rethinking.

spark/src/main/scala/io/substrait/spark/logical/ToLogicalPlan.scala
vbarua: @Blizzara Do you know if there's anything outside of substrait-spark that's relying on this magic number, or is it something that we can change.

spark/src/main/scala/io/substrait/spark/logical/ToLogicalPlan.scala
vbarua: Based on what you've indicated, is converting `FetchRel(count=X, offset = Y, <input>)` into:
```
    GlobalLimit X
    +- Offset Y
       +- LocalLimit X + Y
          + <converted input>
```
viable?

Generally speaking, Substrait consumers should be written in such a way that they accept all valid Substrait plans, even those not built specifically by or for Spark in this case. It's perfectly valid to set both fields of the FetchRel, even if Spark doesn't.

andrew-coleman: > viable?

Could be... I'll do some more testing to get a better understanding of this.

> Generally speaking, ...

Completely agree.  One of the weaknesses of the current test suite it that is relies entirely on round-tripping the query plans.  I need to improve on this, somehow.

westonpace: From reading the Spark PRs here is my understanding.

When converting from Substrait -> Spark:

```
Fetch(limit>=0, offset>0) => LocalLimit -> GlobalLimitAndOffset
Fetch(limit<0, offset>0) => Offset
Fetch(limit>=0, offset==0) => LocalLimit -> GlobalLimit
```

When converting from Spark -> Substrait then the following logic should work:

```
if current_node == LocalLimit:
  if next_node == GlobalLimitAndOffset:
    if current_node.limit < next_node.limit:
      raise Exception("this plan cannot be converted into Substrait")
    else:
      yield Fetch(limit=next_node.limit, offset=next_node.offset)
      skip(next_node)
  elif next_node == GlobalLimit:
    if current_node.limit < next_node.limit:
      raise Exception("this plan cannot be converted into Substrait")
    else:
      yield Fetch(limit=next_node.limit, offset=0)
      skip(next_node)
  else:
    raise Exception("this plan cannot be converted into Substrait")

# This cannot have been preceded by LocalLimit because we would have
# skipped it above so it is just a global limit
if current_node == GlobalLimit:
  yield Fetch(limit=current_node.limit, offset=0)

if current_node == Offset:
  yield Fetch(limit=-1, offset=current_node.offset)

if current_node == GlobalLimitAndOffset:
  yield Fetch(limit=current_node.limit, offset=current_node.offset)
```

The above logic should always round trip successfully (unless I made a mistake in my understanding)

The case where there is either a `LocalLimit` alone or a `LocalLimit` followed by `GlobalLimit` or `GlobalLimitAndOffset` but the `LocalLimit`'s limit is smaller cannot be supported today.

This plan implements a per-group limit and the closest equivalent Substrait plan would be something like...

```
┌──────┐  ┌───────────────────────────┐  ┌───────┐          ┌───────────┐
│ SCAN │  │ FILTER (partition_id = 1) │  │ LIMIT │ ────┬────│ UNION ALL │
└──────┘  └───────────────────────────┘  └───────┘     │    └───────────┘
                                                       │                 
                                                       │                 
┌──────┐  ┌───────────────────────────┐  ┌───────┐     │                 
│ SCAN │  │ FILTER (partition_id = 2) │  │ LIMIT │ ────┼                 
└──────┘  └───────────────────────────┘  └───────┘     │                 
                                                       │                 
                                                       │                 
                                                       │                 
                  ...                                  │                 
                                                       │                 
                                                       │                 
                                                       │                 
┌──────┐  ┌───────────────────────────┐  ┌───────┐     │                 
│ SCAN │  │ FILTER (partition_id = N) │  │ LIMIT │ ────┘                 
└──────┘  └───────────────────────────┘  └───────┘                       
```

andrew-coleman: Thank you @westonpace, that's really helpful.  I think I've captured that in the amended commit, and added a couple of extra tests

spark/src/main/scala/io/substrait/spark/logical/ToLogicalPlan.scala
Blizzara: I'm not aware of anything else, I'd be happy to see it changed! (I had already changed it in our internal fork of substrait-spark, but the code of this PR is nicer than mine 🎉 )

vbarua (approved):
This looks reasonable to me. Thanks fork pushing this forward!

PR #298, opened 2024-09-30
feat(spark): add MakeDecimal support

andrew-coleman:
The Spark query optimiser injects an internal function (MakeDecimal) when numeric literals appear in a query. This commit adds support for this, which drastically improves the pass rate for the TPC-DS test suite.

### DIFF
--- a/spark/src/main/scala/io/substrait/spark/expression/ToSparkExpression.scala
+++ b/spark/src/main/scala/io/substrait/spark/expression/ToSparkExpression.scala
@@ -16,10 +16,10 @@
  */
 package io.substrait.spark.expression
 
-import io.substrait.spark.{DefaultExpressionVisitor, HasOutputStack, ToSubstraitType}
+import io.substrait.spark.{DefaultExpressionVisitor, HasOutputStack, SparkExtension, ToSubstraitType}
 import io.substrait.spark.logical.ToLogicalPlan
-import org.apache.spark.sql.catalyst.expressions.{CaseWhen, Cast, Expression, In, Literal, NamedExpression, ScalarSubquery}
-import org.apache.spark.sql.types.{Decimal, NullType}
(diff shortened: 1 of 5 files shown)


spark/src/main/scala/io/substrait/spark/expression/ToSparkExpression.scala
```diff
@@ -131,23 +131,32 @@ class ToSparkExpression(
...
-            }
-            .mkString(", ")
-        )
-        throw new IllegalArgumentException(msg)
-      })
+    expr.declaration.name match {
+      case "make_decimal" => expr.outputType match {
+        // Need special case handing of this internal function (not nice, I know).
+        // Because the precision and scale arguments are extracted from the output type,
+        // we can't use the generic scalar function conversion mechanism here.
```
vbarua, line 138: > (not nice, I know)

This seems perfectly reasonable tbh.

spark/src/main/scala/io/substrait/spark/expression/ToSparkExpression.scala
```diff
@@ -131,23 +131,32 @@ class ToSparkExpression(
...
-              case e: SExpression => e.getType.accept(new StringTypeVisitor)
-              case t: Type => t.accept(new StringTypeVisitor)
-              case a => throw new IllegalStateException("Unexpected value: " + a)
-            }
-            .mkString(", ")
-        )
-        throw new IllegalArgumentException(msg)
-      })
+    expr.declaration.name match {
+      case "make_decimal" => expr.outputType match {
```
vbarua, line 135: Being pedantic, this should also check the `uri` of the declaration if it's feasible to avoid applying special handling to an externally defined `make_decimal` function.

vbarua (commented):
This seems pretty reasonable overall. There's likely always going to be functions that require some form of special handling. Once we accumulate more we can think about patterns to make it easier to work with them.

I did leave one comment about the matching logic for it though.

vbarua (approved):

PR #301, opened 2024-10-03
feat(spark): support Union All

andrew-coleman:
First part of the implementation of the `Set` relation in the spark module.

This PR only adds `UNION_ALL` capability, but that enables 
support for the “UNION ALL” clause for Spark SQL, increasing the
number of successful TPC-DS tests.


vbarua (approved):
Thanks for adding this!

PR #306, opened 2024-10-15
feat(spark): map from EqualNullSafe fn to is_not_distinct_from

andrew-coleman:
Add support for the <=> operator (NULL safe equals) which gets generated by the spark optimiser for certain queries.

Enables three more successful TPC-DS tests.

### DIFF
--- a/spark/src/main/scala/io/substrait/spark/expression/FunctionMappings.scala
+++ b/spark/src/main/scala/io/substrait/spark/expression/FunctionMappings.scala
@@ -47,6 +47,7 @@ class FunctionMappings {
     s[GreaterThan]("gt"),
     s[GreaterThanOrEqual]("gte"),
     s[EqualTo]("equal"),
+    s[EqualNullSafe]("is_not_distinct_from"),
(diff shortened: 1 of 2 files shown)


spark/src/main/scala/io/substrait/spark/expression/FunctionMappings.scala
```diff
@@ -47,6 +47,7 @@ class FunctionMappings {
     s[GreaterThan]("gt"),
     s[GreaterThanOrEqual]("gte"),
     s[EqualTo]("equal"),
+    s[EqualNullSafe]("is_not_distinct_from"),
```
vbarua, line 50: EqualNullSafe matches the behaviour of `is_not_distinct_from`

https://github.com/apache/spark/blob/31a411773a3e97adb833289f8c695b37802cfedb/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/predicates.scala#L1070-L1094

vbarua (approved):
LGTM

PR #307, opened 2024-10-16
feat(spark): add Window support

andrew-coleman:
To support the OVER clause in SQL

This fixes some of the TPC-DS tests.

### DIFF
--- a/spark/src/main/scala/io/substrait/spark/expression/FunctionMappings.scala
+++ b/spark/src/main/scala/io/substrait/spark/expression/FunctionMappings.scala
@@ -80,9 +80,22 @@ class FunctionMappings {
     s[HyperLogLogPlusPlus]("approx_count_distinct")
   )
 
+  val WINDOW_SIGS: Seq[Sig] = Seq(
+    s[RowNumber]("row_number"),
+    s[Rank]("rank"),
+    s[DenseRank]("dense_rank"),
+    s[PercentRank]("percent_rank"),
+    s[CumeDist]("cume_dist"),
+    s[NTile]("ntile"),
+    s[Lead]("lead"),
+    s[Lag]("lag"),
+    s[NthValue]("nth_value")
+  )
+
   lazy val scalar_functions_map: Map[Class[_], Sig] = SCALAR_SIGS.map(s => (s.expClass, s)).toMap
   lazy val aggregate_functions_map: Map[Class[_], Sig] =
     AGGREGATE_SIGS.map(s => (s.expClass, s)).toMap
+  lazy val window_functions_map: Map[Class[_], Sig] = WINDOW_SIGS.map(s => (s.expClass, s)).toMap
 }
 
 object FunctionMappings extends FunctionMappings
--- a/spark/src/main/scala/io/substrait/spark/expression/ToWindowFunction.scala
+++ b/spark/src/main/scala/io/substrait/spark/expression/ToWindowFunction.scala
@@ -1,0 +1,151 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.substrait.spark.expression
+
+import io.substrait.spark.expression.ToWindowFunction.fromSpark
+
+import org.apache.spark.sql.catalyst.expressions.{CurrentRow, Expression, FrameType, Literal, OffsetWindowFunction, RangeFrame, RowFrame, SpecifiedWindowFrame, UnboundedFollowing, UnboundedPreceding, UnspecifiedFrame, WindowExpression, WindowFrame, WindowSpecDefinition}
+import org.apache.spark.sql.catalyst.expressions.aggregate.AggregateExpression
+import org.apache.spark.sql.types.{IntegerType, LongType}
+
+import io.substrait.`type`.Type
+import io.substrait.expression.{Expression => SExpression, ExpressionCreator, FunctionArg, WindowBound}
+import io.substrait.expression.Expression.WindowBoundsType
+import io.substrait.expression.WindowBound.{CURRENT_ROW, UNBOUNDED, WindowBoundVisitor}
+import io.substrait.extension.SimpleExtension
+import io.substrait.relation.ConsistentPartitionWindow.WindowRelFunctionInvocation
+
+import scala.collection.JavaConverters
+
+abstract class ToWindowFunction(functions: Seq[SimpleExtension.WindowFunctionVariant])
+  extends FunctionConverter[SimpleExtension.WindowFunctionVariant, WindowRelFunctionInvocation](
+    functions) {
+
+  override def generateBinding(
+      sparkExp: Expression,
+      function: SimpleExtension.WindowFunctionVariant,
+      arguments: Seq[FunctionArg],
+      outputType: Type): WindowRelFunctionInvocation = {
+
+    val (frameType, lower, upper) = sparkExp match {
+      case WindowExpression(
+            _,
+            WindowSpecDefinition(_, _, SpecifiedWindowFrame(frameType, lower, upper))) =>
+        (fromSpark(frameType), fromSpark(lower), fromSpark(upper))
+      case WindowExpression(_, WindowSpecDefinition(_, orderSpec, UnspecifiedFrame)) =>
+        if (orderSpec.isEmpty) {
+          (WindowBoundsType.ROWS, UNBOUNDED, UNBOUNDED)
+        } else {
+          (WindowBoundsType.RANGE, UNBOUNDED, CURRENT_ROW)
+        }
+
+      case _ => throw new UnsupportedOperationException(s"Unsupported window expression: $sparkExp")
+    }
+
+    ExpressionCreator.windowRelFunction(
+      function,
+      outputType,
+      SExpression.AggregationPhase.INITIAL_TO_RESULT, // use defaults...
+      SExpression.AggregationInvocation.ALL, // Spark doesn't define these
+      frameType,
+      lower,
+      upper,
+      JavaConverters.asJavaIterable(arguments)
+    )
+  }
+
+  def convert(
+      expression: WindowExpression,
+      operands: Seq[SExpression]): Option[WindowRelFunctionInvocation] = {
+    val cls = expression.windowFunction match {
+      case agg: AggregateExpression => agg.aggregateFunction.getClass
+      case other => other.getClass
+    }
+
+    Option(signatures.get(cls))
+      .flatMap(m => m.attemptMatch(expression, operands))
+  }
+
+  def apply(
+      expression: WindowExpression,
+      operands: Seq[SExpression]): WindowRelFunctionInvocation = {
+    convert(expression, operands).getOrElse(throw new UnsupportedOperationException(
+      s"Unable to find binding for call ${expression.windowFunction} -- $operands -- $expression"))
+  }
+}
+
+object ToWindowFunction {
+  def fromSpark(frameType: FrameType): WindowBoundsType = frameType match {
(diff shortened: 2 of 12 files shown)


vbarua:
>This will also fix several of the TPC-DS tests, but I will enable those in a future PR to try and avoid merge conflicts with other PRs that are currently open.

I've gone ahead and reviewed and merged the other PRs so we shouldn't have any issues with conflict there, beyond the one that exists now. I can review this PR tomorrow.

spark/src/main/scala/io/substrait/spark/expression/ToWindowFunction.scala
```diff
@@ -0,0 +1,141 @@
...
+  def convert(
+      expression: WindowExpression,
+      operands: Seq[SExpression]): Option[WindowRelFunctionInvocation] = {
+    val cls = expression.windowFunction match {
+      case agg: AggregateExpression => agg.aggregateFunction.getClass
+      case other => other.getClass
+    }
+
+    Option(signatures.get(cls))
+      .filter(m => m.allowedArgCount(2))
```
vbarua, line 77:
```java
m.allowedArgCount(2)
```
Is the intent to only filter out functions that take 2 arguments?

That line is also a no-op based on the implementation. Do we need this code?
```java
  def allowedArgCount(count: Int): Boolean = true
```

spark/src/main/scala/io/substrait/spark/expression/ToWindowFunction.scala
```diff
@@ -0,0 +1,141 @@
...
+
+  override def generateBinding(
+      sparkExp: Expression,
+      function: SimpleExtension.WindowFunctionVariant,
+      arguments: Seq[FunctionArg],
+      outputType: Type): WindowRelFunctionInvocation = {
+
+    val (frameType, lower, upper) = sparkExp match {
+      case WindowExpression(_: OffsetWindowFunction, _) =>
+        (WindowBoundsType.ROWS, UNBOUNDED, CURRENT_ROW)
```
vbarua, line 46: What is an `OffsetWindowFunction`. Does it somehow override the frame bounds?

vbarua (commented):
Left some questions.

spark/src/test/scala/io/substrait/spark/WindowPlan.scala
```diff
@@ -0,0 +1,67 @@
...
+    val query =
+      """
+        |SELECT name, salary,
+        |    LAG(salary) OVER (PARTITION BY dept ORDER BY salary) AS lag,
+        |    LEAD(salary, 1, 0) OVER (PARTITION BY dept ORDER BY salary) AS lead
+        |    FROM employees;
+        |
+        |""".stripMargin
+    assertSqlSubstraitRelRoundTrip(query)
+  }
```
Blizzara, line 66: maybe add a test with two different partitions in same select? I think it should pass fine but just in case

Blizzara (commented):
Thanks! Overall looks good, but there's some things in the bounds that I think aren't always correct

spark/src/main/scala/io/substrait/spark/expression/ToWindowFunction.scala
```diff
@@ -0,0 +1,141 @@
...
+      case WindowExpression(
+            _,
+            WindowSpecDefinition(_, _, SpecifiedWindowFrame(frameType, lower, upper))) =>
+        (fromSpark(frameType), fromSparkPreceding(lower), fromSparkFollowing(upper))
+      case WindowExpression(_, WindowSpecDefinition(_, _, UnspecifiedFrame)) =>
+        (WindowBoundsType.ROWS, UNBOUNDED, CURRENT_ROW)
```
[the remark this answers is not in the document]
andrew-coleman: Good spot, I hadn't noticed that comment.

spark/src/main/scala/io/substrait/spark/expression/ToWindowFunction.scala
andrew-coleman: You're right, I don't think it's needed either.  I'm guilty of cutting and pasting code that was originally part of the gluten project.  I'm sure it's there for historical reasons, but I'll remove it.

spark/src/main/scala/io/substrait/spark/expression/ToWindowFunction.scala
andrew-coleman: That's my understanding. TBH I'm not sure exactly how it should translate to substrait, but it's ignored by spark (it matches the lag/lead functions)
https://github.com/apache/spark/blob/master/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/windowExpressions.scala#L404

spark/src/main/scala/io/substrait/spark/expression/FunctionMappings.scala
```diff
@@ -73,9 +73,22 @@ class FunctionMappings {
...
 
+  val WINDOW_SIGS: Seq[Sig] = Seq(
+    s[RowNumber]("row_number"),
+    s[Rank]("rank"),
+    s[DenseRank]("dense_rank"),
+    s[PercentRank]("percent_rank"),
+    s[CumeDist]("cume_dist"),
+    s[NTile]("ntile"),
+    s[Lead]("lead"),
+    s[Lag]("lag"),
```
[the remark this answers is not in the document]
andrew-coleman: I haven't seen this - do you have a test case?

Blizzara:
> Thanks @Blizzara, there’s some really useful feedback here. I hadn’t realised you had already been implementing this, it would be awesome if you want to collaborate :)

Yea, shame on me - it's been on my todo-list ever since you added this into substrait-java to pull up our changes. Some of those changes are non-trivial refactorings which might be annoying, but I'll try to get to it!

> I’d only implemented as much as necessary to enable the windowing tests in the TPC-DS suite to pass. If you’ve got other tests that require additional logic in the converter, then it would be great to add these. But perhaps in a future PR in the interest of moving things forward incrementally?

I'd like to see the bounds things fixed, since currently this implementation doesn't adhere to the bounds (preceeding, following) [being strictly positive](https://github.com/substrait-io/substrait/blob/683f4179a058c2c99c04501b920a48ff372356ff/proto/substrait/algebra.proto#L1196). That's not a problem for the roundtrip tests, but it means a plan generated here may not be valid Substrait that can be consumed by other

Blizzara:
> Thanks @Blizzara, there’s some really useful feedback here. I hadn’t realised you had already been implementing this, it would be awesome if you want to collaborate :)

Starting with https://github.com/substrait-io/substrait-java/pull/311! After that I'll add structs (+ fix the name handling), and then I have some more complicated changes in how FunctionMappings works to allow for more complicated mappings.

spark/src/main/scala/io/substrait/spark/expression/ToWindowFunction.scala
```diff
@@ -0,0 +1,141 @@
...
+    case other => throw new UnsupportedOperationException(s"Unsupported bounds type: $other.")
+  }
+
+  def toSpark(boundsType: WindowBoundsType): FrameType = boundsType match {
+    case WindowBoundsType.ROWS => RowFrame
+    case WindowBoundsType.RANGE => RangeFrame
+    case other => throw new UnsupportedOperationException(s"Unsupported bounds type: $other.")
+  }
+
+  def fromSparkPreceding(bound: Expression): WindowBound = bound match {
```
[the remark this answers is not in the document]
andrew-coleman: Thanks, I've changed it this.

spark/src/main/scala/io/substrait/spark/expression/ToWindowFunction.scala
```diff
@@ -0,0 +1,141 @@
...
+    case Literal(l: Long, LongType) => Following.of(l)
+    case _ =>
+      throw new UnsupportedOperationException(s"Unsupported bounds expression ${bound.getClass}")
+  }
+
+  def toSparkPreceding(bound: WindowBound): Expression = bound match {
+    case UNBOUNDED => UnboundedPreceding
+    case CURRENT_ROW => CurrentRow
+    case p: Preceding => Literal(p.offset())
+    case _ => throw new UnsupportedOperationException(s"Unsupported bounds expression $bound")
```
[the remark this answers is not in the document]
andrew-coleman: Yours is nicer :).

spark/src/main/scala/io/substrait/spark/expression/FunctionMappings.scala
```diff
@@ -73,9 +73,22 @@ class FunctionMappings {
     s[HyperLogLogPlusPlus]("approx_count_distinct")
   )
 
+  val WINDOW_SIGS: Seq[Sig] = Seq(
+    s[RowNumber]("row_number"),
+    s[Rank]("rank"),
```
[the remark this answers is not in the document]
Blizzara: Ah you handle it here: https://github.com/substrait-io/substrait-java/pull/307/files#diff-78e1a9d4e9f4c7d8968b3ba83d3cc5222ade95d28e16b0328277c3f8c8a9d313R180

spark/src/main/scala/io/substrait/spark/expression/FunctionMappings.scala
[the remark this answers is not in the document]
Blizzara: Looks like you had, you just handle it here: https://github.com/substrait-io/substrait-java/pull/307/files#diff-78e1a9d4e9f4c7d8968b3ba83d3cc5222ade95d28e16b0328277c3f8c8a9d313R183 :D

spark/src/main/scala/io/substrait/spark/expression/ToWindowFunction.scala
Blizzara: You actually should get the frame here as well properly. Otherwise the bound will be incorrect for other consumers (unbounded preceeding and current row will not work for Lead, for example) and Spark will throw. You'll see that if you add `logicalPlan2.show()` into `SubstraitPlanTestBase::assertSqlToSubstraitRoundTrip` and re-run the lag/lead test. (We should also make the tests always evaluate the converted plans to ensure they are actually valid, I'll try to get to that)

Can be fixed by just removing this `case`.

spark/src/main/scala/io/substrait/spark/expression/FunctionMappings.scala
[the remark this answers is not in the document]
andrew-coleman: Oh yes - not known for my memory! 😂

spark/src/main/scala/io/substrait/spark/expression/ToWindowFunction.scala
andrew-coleman: Ok, I've removed it for now

spark/src/main/scala/io/substrait/spark/expression/ToWindowFunction.scala
```diff
@@ -0,0 +1,151 @@
...
+      arguments: Seq[FunctionArg],
+      outputType: Type): WindowRelFunctionInvocation = {
+
+    val (frameType, lower, upper) = sparkExp match {
+      case WindowExpression(
+            _,
+            WindowSpecDefinition(_, _, SpecifiedWindowFrame(frameType, lower, upper))) =>
+        (fromSpark(frameType), fromSpark(lower), fromSpark(upper))
+      case WindowExpression(_, WindowSpecDefinition(_, orderSpec, UnspecifiedFrame)) =>
+        if (orderSpec.isEmpty) {
```
Blizzara, line 50: just for posterity, this comes from Spark notes: https://github.com/apache/spark/blob/250f8affd04e4be14446dd02a1c52716e54a226d/sql/api/src/main/scala/org/apache/spark/sql/expressions/Window.scala#L36

spark/src/test/scala/io/substrait/spark/TPCDSPlan.scala
```diff
@@ -32,16 +32,16 @@ class TPCDSPlan extends TPCDSBase with SubstraitPlanTestBase {
   }
 
   // spotless:off
-  val successfulSQL: Set[String] = Set("q1", "q3", "q4", "q7", "q8",
-    "q11", "q13", "q14a", "q14b", "q15", "q16", "q18", "q19",
-    "q21", "q22", "q23a", "q23b", "q24a", "q24b", "q25", "q26", "q27", "q28", "q29",
-    "q30", "q31", "q32", "q33", "q37", "q38",
-    "q40", "q41", "q42", "q43", "q46", "q48",
+  val successfulSQL: Set[String] = Set("q1", "q3", "q4", "q5", "q7", "q8",
```
Blizzara, line 35: given how this is growing, might be worth starting to list the ones that _don't_ work instead 😄

Blizzara (approved):
LGTM, thanks!

spark/src/test/scala/io/substrait/spark/TPCDSPlan.scala
andrew-coleman: Exactly, I was going to do this after my next PR which will add support for more numeric functions.  The number of failing tests will be relatively small then.

PR #308, opened 2024-10-23
feat(spark): enable some missing string functions

andrew-coleman:
Add mappings for upper/lower/concat/coalesce


Blizzara (approved):
LGTM (don't think my approval counts tho, need @vbarua)

Blizzara:
Now it does count :D Thanks @andrew-coleman!

PR #309, opened 2024-10-23
feat(spark): bitwise functions

andrew-coleman:
Adds support in the spark module for 8-bit and 16-bit integer types and for some bitwise functions. The catalyst optimizer generates expressions using these for certain query types.

Note that `shift_right` (and other bit shifting functions) might want to be considered for the core substrait function catalog, but it has been added here (temporarily?) as spark extension pending a longer term discussion/decision on their wider utility.

### DIFF
--- a/spark/src/main/resources/spark.yml
+++ b/spark/src/main/resources/spark.yml
@@ -41,3 +41,22 @@ scalar_functions:
       - args:
           - value: i64
         return: DECIMAL<P,S>
+  - name: shift_right
+    description: >-
+      Bitwise (signed) shift right.
+      Params:
+        base – the base number to shift.
+        shift – number of bits to right shift.
+    impls:
+      - args:
+          - name: base
+            value: i64
+          - name: shift
+            value: i32
+        return: i64
+      - args:
+          - name: base
+            value: i32
+          - name: shift
+            value: i32
+        return: i32
--- a/spark/src/main/scala/io/substrait/spark/ToSubstraitType.scala
+++ b/spark/src/main/scala/io/substrait/spark/ToSubstraitType.scala
@@ -29,6 +29,8 @@ import scala.collection.JavaConverters.asScalaBufferConverter
 private class ToSparkType
   extends TypeVisitor.TypeThrowsVisitor[DataType, RuntimeException]("Unknown expression type.") {
 
+  override def visit(expr: Type.I8): DataType = ByteType
+  override def visit(expr: Type.I16): DataType = ShortType
   override def visit(expr: Type.I32): DataType = IntegerType
   override def visit(expr: Type.I64): DataType = LongType
 
--- a/spark/src/main/scala/io/substrait/spark/expression/ToSparkExpression.scala
+++ b/spark/src/main/scala/io/substrait/spark/expression/ToSparkExpression.scala
@@ -44,6 +44,15 @@ class ToSparkExpression(
       Literal.FalseLiteral
     }
   }
+
+  override def visit(expr: SExpression.I8Literal): Expression = {
+    Literal(expr.value().asInstanceOf[Byte], ToSubstraitType.convert(expr.getType))
+  }
+
+  override def visit(expr: SExpression.I16Literal): Expression = {
+    Literal(expr.value().asInstanceOf[Short], ToSubstraitType.convert(expr.getType))
+  }
+
   override def visit(expr: SExpression.I32Literal): Expression = {
     Literal(expr.value(), ToSubstraitType.convert(expr.getType))
   }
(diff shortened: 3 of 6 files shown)


spark/src/main/scala/io/substrait/spark/ToSubstraitType.scala
```diff
@@ -29,6 +29,8 @@ import scala.collection.JavaConverters.asScalaBufferConverter
 private class ToSparkType
   extends TypeVisitor.TypeThrowsVisitor[DataType, RuntimeException]("Unknown expression type.") {
 
+  override def visit(expr: Type.I8): DataType = ByteType
+  override def visit(expr: Type.I16): DataType = ShortType
```
Blizzara, line 33: while you're at it, mind adding these also to ToSparkType?

spark/src/main/scala/io/substrait/spark/expression/ToSparkExpression.scala
```diff
@@ -44,6 +44,15 @@ class ToSparkExpression(
       Literal.FalseLiteral
     }
   }
+
+  override def visit(expr: SExpression.I8Literal): Expression = {
+    Literal(expr.value().asInstanceOf[Byte], ToSubstraitType.convert(expr.getType))
+  }
+
+  override def visit(expr: SExpression.I16Literal): Expression = {
+    Literal(expr.value().asInstanceOf[Short], ToSubstraitType.convert(expr.getType))
```
Blizzara, line 53: and here also the other direction for the conversion?

spark/src/main/scala/io/substrait/spark/ToSubstraitType.scala
andrew-coleman: This is in `ToSparkType`.  The conversion in `ToSubstraitType` is already there further down this file. It converts both ways.

spark/src/main/scala/io/substrait/spark/expression/ToSparkExpression.scala
andrew-coleman: It's already covered in the other direction in `ToSubstraitExpression` on the following line:
`case SubstraitLiteral(substraitLiteral) => Some(substraitLiteral)`
The `unapply` method in this object invokes the conversion.

spark/src/main/scala/io/substrait/spark/ToSubstraitType.scala
Blizzara: Ah right, all good then!

spark/src/main/scala/io/substrait/spark/expression/ToSubstraitExpression.scala
```diff
@@ -144,6 +144,7 @@ abstract class ToSubstraitExpression extends HasOutputStack[Seq[Attribute]] {
             "org.apache.spark.sql.catalyst.expressions.PromotePrecision") =>
         translateUp(p.children.head)
       case CaseWhen(branches, elseValue) => translateCaseWhen(branches, elseValue)
+      case InSet(value, set) => translateIn(value, set.toSeq.map(v => Literal(v)))
```
Blizzara, line 147: nit: mind moving this next to the `case In(..)`?

spark/src/main/resources/spark.yml
```diff
@@ -41,3 +41,16 @@ scalar_functions:
       - args:
           - value: i64
         return: DECIMAL<P,S>
+  - name: shift_right
+    description: >-
+      Bitwise (signed) shift right.
+      Params:
+        base – the base number to shift.
+        shift – number of bits to right shift.
+    impls:
```
Blizzara, line 50: It looks like in Spark the base can be either int or long, and return type is set accordingly. I  we should add both options?

Quickly testing this, it works for `select shiftright(col, 2) from (values (bigint(1234)) as table(col))` but not for `select shiftright(col, 2) from (values (1234) as table(col))`, so yep I think we need to list both versions here.

Blizzara: Overall I think it's fine to add the function here initially, but it'd be good to also file the PR/Issue on the core functions since this seems general enough, I think :)

spark/src/main/scala/io/substrait/spark/expression/ToSubstraitExpression.scala
andrew-coleman: Sure, although `InSet` needs to come before `ScalarFunction` otherwise it matches the latter (since it's a `UnaryExpression`).  I've moved `In` up the list so they are together.

spark/src/main/resources/spark.yml
andrew-coleman: I agree, although that might take a bit longer ;)

andrew-coleman:
I'll fixup the conflict :)

Blizzara (approved):
Thanks!

PR #310, opened 2024-10-23
fix(spark): output columns of expand relation

andrew-coleman:
Fix the logic for deriving whether each output column of the expand relation is nullable or not.
If one or more of the projections can assign null to the column, then it must be defined as nullable.


Blizzara (approved):
Makes sense, thanks!

Blizzara:
@andrew-coleman ah this has a conflict in the tests it enables, mind merging/rebasing main in? :)

PR #311, opened 2024-10-23
feat(spark): add support for more types and literals (binary, list, map, intervals, timestamps)

Blizzara:
45d938700c8f9208e378e20be096a88a78289783 adds support for EmptyMap into `core` - if preferred, I can split that off into its own PR
ea7bdb16bc726e71c7eff531f2b3841e09f0106b adds support for bunch of types and literals into substrait-spark, as well as testing for all of the currently supported types
c43f3991e7d51927aeb305e2836ffaaa70ed0a12 adds Byte and Short (taken from #309, thanks @andrew-coleman - this just so that should be fine to merge these in either order and we'll have tests setup fully)

### DIFF
--- a/spark/src/main/scala/io/substrait/spark/ToSubstraitType.scala
+++ b/spark/src/main/scala/io/substrait/spark/ToSubstraitType.scala
@@ -53,25 +53,16 @@ private class ToSparkType
   override def visit(expr: Type.Bool): DataType = BooleanType
 
   override def visit(expr: Type.PrecisionTimestamp): DataType = {
-    if (expr.precision() != Util.MICROSECOND_PRECISION) {
-      throw new UnsupportedOperationException(
-        s"Unsupported precision for timestamp: ${expr.precision()}")
-    }
+    Util.assertMicroseconds(expr.precision())
     TimestampNTZType
   }
   override def visit(expr: Type.PrecisionTimestampTZ): DataType = {
-    if (expr.precision() != Util.MICROSECOND_PRECISION) {
-      throw new UnsupportedOperationException(
-        s"Unsupported precision for timestamp: ${expr.precision()}")
-    }
+    Util.assertMicroseconds(expr.precision())
     TimestampType
   }
 
   override def visit(expr: Type.IntervalDay): DataType = {
-    if (expr.precision() != Util.MICROSECOND_PRECISION) {
-      throw new UnsupportedOperationException(
-        s"Unsupported precision for intervalDay: ${expr.precision()}")
-    }
+    Util.assertMicroseconds(expr.precision())
     DayTimeIntervalType.DEFAULT
   }
 
--- a/spark/src/main/scala/io/substrait/spark/ToSubstraitType.scala
+++ b/spark/src/main/scala/io/substrait/spark/ToSubstraitType.scala
@@ -44,11 +44,47 @@ private class ToSparkType
 
   override def visit(expr: Type.Str): DataType = StringType
 
+  override def visit(expr: Type.Binary): DataType = BinaryType
+
   override def visit(expr: Type.FixedChar): DataType = StringType
 
   override def visit(expr: Type.VarChar): DataType = StringType
 
   override def visit(expr: Type.Bool): DataType = BooleanType
+
+  override def visit(expr: Type.PrecisionTimestamp): DataType = {
+    if (expr.precision() != Util.MICROSECOND_PRECISION) {
+      throw new UnsupportedOperationException(
+        s"Unsupported precision for timestamp: ${expr.precision()}")
+    }
+    TimestampNTZType
+  }
+  override def visit(expr: Type.PrecisionTimestampTZ): DataType = {
+    if (expr.precision() != Util.MICROSECOND_PRECISION) {
+      throw new UnsupportedOperationException(
+        s"Unsupported precision for timestamp: ${expr.precision()}")
+    }
+    TimestampType
+  }
+
+  override def visit(expr: Type.IntervalDay): DataType = {
+    if (expr.precision() != Util.MICROSECOND_PRECISION) {
+      throw new UnsupportedOperationException(
+        s"Unsupported precision for intervalDay: ${expr.precision()}")
+    }
+    DayTimeIntervalType.DEFAULT
+  }
+
+  override def visit(expr: Type.IntervalYear): DataType = YearMonthIntervalType.DEFAULT
+
+  override def visit(expr: Type.ListType): DataType =
+    ArrayType(expr.elementType().accept(this), containsNull = expr.elementType().nullable())
+
+  override def visit(expr: Type.Map): DataType =
+    MapType(
+      expr.key().accept(this),
+      expr.value().accept(this),
+      valueContainsNull = expr.value().nullable())
 }
 class ToSubstraitType {
 
@@ -81,10 +117,12 @@ class ToSubstraitType {
       case charType: CharType => Some(creator.fixedChar(charType.length))
       case varcharType: VarcharType => Some(creator.varChar(varcharType.length))
       case StringType => Some(creator.STRING)
-      case DateType => Some(creator.DATE)
-      case TimestampType => Some(creator.TIMESTAMP)
-      case TimestampNTZType => Some(creator.TIMESTAMP_TZ)
       case BinaryType => Some(creator.BINARY)
+      case DateType => Some(creator.DATE)
+      case TimestampNTZType => Some(creator.precisionTimestamp(Util.MICROSECOND_PRECISION))
+      case TimestampType => Some(creator.precisionTimestampTZ(Util.MICROSECOND_PRECISION))
+      case DayTimeIntervalType.DEFAULT => Some(creator.intervalDay(Util.MICROSECOND_PRECISION))
+      case YearMonthIntervalType.DEFAULT => Some(creator.INTERVAL_YEAR)
       case ArrayType(elementType, containsNull) =>
         convert(elementType, Seq.empty, containsNull).map(creator.list)
       case MapType(keyType, valueType, valueContainsNull) =>
--- a/spark/src/main/scala/io/substrait/spark/expression/ToSparkExpression.scala
+++ b/spark/src/main/scala/io/substrait/spark/expression/ToSparkExpression.scala
@@ -28,8 +28,9 @@ import io.substrait.`type`.{StringTypeVisitor, Type}
 import io.substrait.{expression => exp}
 import io.substrait.expression.{Expression => SExpression}
 import io.substrait.util.DecimalUtil
+import io.substrait.utils.Util
 
-import scala.collection.JavaConverters.asScalaBufferConverter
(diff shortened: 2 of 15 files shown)


spark/src/main/scala/io/substrait/spark/expression/ToSparkExpression.scala
```diff
@@ -68,15 +71,38 @@ class ToSparkExpression(
...
+  override def visit(expr: SExpression.ListLiteral): ArrayData = {
+    val array = expr.values().asScala.map(value => value.accept(this))
+    CatalystTypeConverters.convertToCatalyst(array).asInstanceOf[ArrayData]
+  }
+
+  override def visit(expr: SExpression.MapLiteral): MapData = {
+    val map = expr.values().asScala.map { case (key, value) => (key.accept(this), value.accept(this)) }
+    CatalystTypeConverters.convertToCatalyst(map).asInstanceOf[MapData]
+  }
+
```
Blizzara, line 105: looks like I'm missing intervals here, let me see if I can add those too

spark/src/main/scala/io/substrait/spark/expression/ToSubstraitLiteral.scala
```diff
@@ -34,6 +38,34 @@ class ToSubstraitLiteral {
...
 
+    private def sparkArray2Substrait(
+        arrayData: ArrayData,
+        elementType: DataType,
+        containsNull: Boolean): SExpression.Literal = {
+      val elements = arrayData.array.map(any => apply(Literal(any, elementType)))
+      if (elements.isEmpty) {
+        return emptyList(false, ToSubstraitType.convert(elementType, nullable = containsNull).get)
+      }
+      list(false, JavaConverters.asJavaIterable(elements)) // TODO: handle containsNull
```
Blizzara, line 49: Handling `containsNull` here is non-trivial, since Substrait doesn't have such a concept. The nullability of the non-empty list is taken from the nullability of the element type, which is gotten from the first element. So we'll need to encode `containsNull` into the `elementType`. It can be done, but requires some more work (as there's no "withNullable" method or anything) and this PR is already big enough, so I'll do it as a followup.

spark/src/main/scala/io/substrait/spark/ToSubstraitType.scala
```diff
@@ -128,7 +166,7 @@ class ToSubstraitType {
     )
   }
 
-  def toAttribute(namedStruct: NamedStruct): Seq[AttributeReference] = {
+  def toAttributeSeq(namedStruct: NamedStruct): Seq[AttributeReference] = {
```
Blizzara, line 169: just a rename

spark/src/main/scala/io/substrait/spark/ToSubstraitType.scala
```diff
@@ -42,11 +44,47 @@ private class ToSparkType
...
+  override def visit(expr: Type.PrecisionTimestampTZ): DataType = {
+    if (expr.precision() != Util.MICROSECOND_PRECISION) {
+      throw new UnsupportedOperationException(
+        s"Unsupported precision for timestamp: ${expr.precision()}")
+    }
+    TimestampType
+  }
+
+  override def visit(expr: Type.IntervalDay): DataType = {
+    if (expr.precision() != Util.MICROSECOND_PRECISION) {
```
vbarua, line 71: To confirm, the Spark `DayTimeIntervalType` encodes time down to seconds, but can store fractional seconds.

From the Spark code
> Internally, values of day-time intervals are stored in Long values as amount of time in terms of microseconds that are calculated by the formula: -/+ (24*60*60 * DAY + 60*60 * HOUR + 60 * MINUTE + SECOND) * 1000000

So the default / supported precision is microseconds?

spark/src/main/scala/io/substrait/spark/expression/ToSparkExpression.scala
```diff
@@ -68,15 +82,69 @@ class ToSparkExpression(
...
     Literal(Decimal(decimal), ToSubstraitType.convert(expr.getType))
   }
+
   override def visit(expr: SExpression.DateLiteral): Expression = {
     Literal(expr.value(), ToSubstraitType.convert(expr.getType))
   }
 
+  override def visit(expr: SExpression.PrecisionTimestampLiteral): Literal = {
+    Literal(
+      Util.toMicroseconds(expr.value(), expr.precision()),
```
vbarua, line 101: Is it generally safe to convert timestamp values like this? If we're going from nanoseconds to microseconds, this will be lossy. Would it better for reject plans with non-microsecond precision and force the callers to make their plans compatible by using microseconds explicitly?

spark/src/main/scala/io/substrait/spark/expression/ToSparkExpression.scala
```diff
@@ -68,15 +82,69 @@ class ToSparkExpression(
...
+
+  override def visit(expr: SExpression.PrecisionTimestampTZLiteral): Literal = {
+    Literal(
+      Util.toMicroseconds(expr.value(), expr.precision()),
+      ToSubstraitType.convert(expr.getType))
+  }
+
+  override def visit(expr: SExpression.IntervalDayLiteral): Literal = {
+    val micros =
+      (expr.days() * Util.SECONDS_PER_DAY + expr.seconds()) * Util.MICROSECOND_PRECISION +
```
vbarua, line 114: I think this is wrong? It's equivalent to
```java
(expr.days() * Util.SECONDS_PER_DAY + expr.seconds()) * 6 +
```

Do we need/want something like 
```java
(expr.days() * Util.SECONDS_PER_DAY + expr.seconds()) * Util.MICROSECONDS_PER_SECOND +
```

spark/src/main/scala/io/substrait/utils/Util.scala
```diff
@@ -21,6 +21,22 @@ import scala.collection.mutable.ArrayBuffer
...
+  def toMicroseconds(value: Long, precision: Int): Long = {
+    // Spark uses microseconds as a Long value for most time things
+    val factor = MICROSECOND_PRECISION - precision
+    // Doing this in a way that avoids floating point math
+    if (factor == 0) {
+      value
+    } else if (factor > 0) {
+      value * math.pow(10, factor).toLong
+    } else {
+      value / math.pow(10, -factor).toLong
```
vbarua, line 36: Should we fail in this case instead to avoid truncation / precision loss? Related to my comments in `ToSparkExpression`

spark/src/main/scala/io/substrait/spark/expression/ToSubstraitLiteral.scala
vbarua: I'm not entirely sure what you mean here.

> The nullability of the non-empty list is taken from the nullability of the element type, which is gotten from the first element.

Is this when going from Substrait to Spark or from Spark to Substrait?

vbarua (commented):
Overall this is looking pretty good. The tests help a lot. Left a couple of questions.

spark/src/main/scala/io/substrait/spark/ToSubstraitType.scala
Blizzara: Yea, Spark always stores the value as microseconds. 

It's possible this is unnecessarily strict, I checked that [on DataFusion we don't care](https://github.com/apache/datafusion/blob/6a3c0b0bce67553a4431b941d13fc995f310bee8/datafusion/substrait/src/logical_plan/consumer.rs#L1833) about the precision at all for the type conversion.. but I guess it doesn't hurt to be strict to begin with, we can relax it later if there's need.

spark/src/main/scala/io/substrait/spark/expression/ToSparkExpression.scala
Blizzara: Yeah, maybe not, I changed it in fd01c2f262534e10e71abf729eac306795df40be to be an assert instead. 

Though we may want to make it e.g. a user-choosable option later, or something, since e.g. Spark and DataFusion sometimes disagree on which precision to use (for IntervalDay, Spark uses micros and DF millis), and it'd be sad if the two would then be fundamentally incompatible. But for now we can just be strict.

spark/src/main/scala/io/substrait/spark/expression/ToSparkExpression.scala
Blizzara: Lol, you're right - and the test didn't fail this as we don't write any "days" or "seconds", just "subseconds". Good catch! Fixed in fd01c2f262534e10e71abf729eac306795df40be

spark/src/main/scala/io/substrait/utils/Util.scala
Blizzara: fd01c2f262534e10e71abf729eac306795df40be

spark/src/main/scala/io/substrait/spark/expression/ToSubstraitLiteral.scala
Blizzara: Substrait itself doesn't contain the "containsNull" arg, so the Substrait -> Spark conversion has to infer it, and the Spark -> Substrait conversion doesn't provide the correct information to infer it based on.

Easiest to see from the commented-out test:
Starting with Spark: `value: [1,null,3] type: ArrayType(IntegerType,true)`
Gets converted into Substrait literal: `ListLiteral{nullable=false, values=[I32Literal{nullable=false, value=1}, NullLiteral{nullable=false, type=I32{nullable=true}}, I32Literal{nullable=false, value=3}]}`

Note that there is no "containsNull" anywhere. So the way we infer it when converting back into Substrait is to look at the first element and it's `nullable` arg. Which in today's world is always `nullabe=false`, leading to:

Spark: `value: [1,null,3] type: ArrayType(IntegerType,false)`

(Funnily enough that's the case even if the first element _is_ a null.. but fixing that in itself wouldn't be enough, since the null might be elsewhere in the list or even in a different row, but we should have the "type" of the literal match across rows.)

I think fixing this requires changing the `ToSubstraitLiteral.convert/convertWithValue` so that the nullability they set can be overridden in this case.

spark/src/main/scala/io/substrait/spark/ToSubstraitType.scala
vbarua: Let's start strict and see how we go.

> I checked that [on DataFusion we don't care](https://github.com/apache/datafusion/blob/6a3c0b0bce67553a4431b941d13fc995f310bee8/datafusion/substrait/src/logical_plan/consumer.rs#L1833) about the precision at all for the type conversion.

That's the case for now, but that may also change if we want to enforce semantics with Substrait strictly.

vbarua (approved):
LGTM. I will merge this today.

spark/src/main/scala/io/substrait/spark/ToSubstraitType.scala
Blizzara: Yea, I was thinking about that today - problem is I do need Spark -> DataFusion to play nicely together and that's tough if Spark insists strictly on one precision and DF on another 😅 but then what is the risk of things going wrong if the precision don't match, that I'm not sure about.. on literals it all makes sense, but for types themselves?

PR #312, opened 2024-10-24
feat(spark): convert VirtualTableScan to LocalRelation 

Blizzara:
The other direction (Spark -> Substrait) was already supported, but this adds Substrait -> Spark and enables round-trip testing.

Also fixes LocalRelation -> VirtualTableScan for rows containing null values.

### DIFF
--- a/spark/src/main/scala/io/substrait/spark/ToSubstraitType.scala
+++ b/spark/src/main/scala/io/substrait/spark/ToSubstraitType.scala
@@ -128,7 +128,7 @@ class ToSubstraitType {
     )
   }
 
-  def toAttribute(namedStruct: NamedStruct): Seq[AttributeReference] = {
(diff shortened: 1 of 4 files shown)


spark/src/main/scala/io/substrait/spark/ToSubstraitType.scala
```diff
@@ -128,7 +128,7 @@ class ToSubstraitType {
     )
   }
 
-  def toAttribute(namedStruct: NamedStruct): Seq[AttributeReference] = {
+  def toAttributeSeq(namedStruct: NamedStruct): Seq[AttributeReference] = {
```
Blizzara, line 131: same rename as in #311

vbarua (approved):
LGTM. Thanks for adding this!

PR #313, opened 2024-10-25
fix: propagate sorts in aggregate function invocation proto->rel

bvolpato:
When consuming a Substrait proto plan that contains sort within aggregate calls, the SortFields were being dropped in the process.

This is what I mean by sort within aggregates:
> SELECT ARRAY_AGG(genre **ORDER BY genre**) FROM book

So this is a simple addition + test case change that closes that gap.


Blizzara (approved):
LGTM, thanks @bvolpato!

PR #315, opened 2024-10-28
fix: add proto roundtrips for Spark tests and fix issues it surfaces

Blizzara:
Adds testing for substrait-spark that going from POJO (ie. substrait-java plan) -> Proto -> POJO results in the same POJO. 

The test showed a bunch of cases where that assertion fails, mainly due to the java pojos containing a derived outputType which was in many cases incorrect when created from the proto.


core/src/main/java/io/substrait/relation/ProtoRelConverter.java
```diff
@@ -361,12 +361,7 @@ protected VirtualTableScan newVirtualTable(ReadRel rel) {
 
   protected Fetch newFetch(FetchRel rel) {
     var input = from(rel.getInput());
-    var builder = Fetch.builder().input(input).offset(rel.getOffset());
-    if (rel.getCount() != -1) {
-      // -1 is used as a sentinel value to signal LIMIT ALL
-      // count only needs to be set when it is not -1
-      builder.count(rel.getCount());
-    }
+    var builder = Fetch.builder().input(input).offset(rel.getOffset()).count(rel.getCount());
```
Blizzara, line 364: while the idea of not setting count if it's -1 is fine, this makes roundtrip tests fail if count is set in the pojo. Alternative fix is to ensure in the pojo it's never set if -1.

spark/src/main/scala/io/substrait/spark/logical/ToSubstraitRel.scala
```diff
@@ -131,7 +131,7 @@ class ToSubstraitRel extends AbstractLogicalPlanVisitor with Logging {
     val aggregates = collectAggregates(actualResultExprs, aggExprToOutputOrdinal)
     val aggOutputMap = aggregates.zipWithIndex.map {
       case (e, i) =>
-        AttributeReference(s"agg_func_$i", e.dataType)() -> e
+        AttributeReference(s"agg_func_$i", e.dataType, nullable = e.nullable)() -> e
```
Blizzara, line 134: these were causing wrong nullability for the type in the created pojos

spark/src/main/scala/io/substrait/spark/logical/ToSubstraitRel.scala
```diff
@@ -185,13 +185,10 @@ class ToSubstraitRel extends AbstractLogicalPlanVisitor with Logging {
...
-      case w: WindowExpression => fromWindowCall(w, window.child.output)
-      case a: Alias if a.child.isInstanceOf[WindowExpression] =>
-        fromWindowCall(a.child.asInstanceOf[WindowExpression], window.child.output)
-      case other =>
-        throw new UnsupportedOperationException(s"Unsupported window expression: $other")
-    }.asJava
+    val windowExpressions = window.windowExpressions
+        .flatMap(expr => expr.collect { case w: WindowExpression => w })
+        .map(fromWindowCall(_, window.child.output))
+      .asJava
```
Blizzara, line 191: the previous code would fail if there was a column like Alias(Alias(WindowExpression(..))), this catches those. It doesn't explicitly fail if there's some other wrappers for WindowExpressions than Alias, but I hope that's not the case in valid Spark plans

core/src/test/java/io/substrait/type/proto/ExtensionRoundtripTest.java
```diff
@@ -312,7 +312,7 @@ void scalarSubquery() {
                   Stream.of(
                           Expression.ScalarSubquery.builder()
                               .input(relWithEnhancement)
-                              .type(TypeCreator.REQUIRED.struct(TypeCreator.REQUIRED.I64))
+                              .type(TypeCreator.REQUIRED.I64)
```
Blizzara, line 315: not 100% sure about this, is it actually meant to return a struct type? given it's scalar that seems a bit weird

core/src/main/java/io/substrait/relation/Set.java
```diff
@@ -42,7 +46,38 @@ public static Set.SetOp fromProto(SetRel.SetOp proto) {
 
   @Override
   protected Type.Struct deriveRecordType() {
-    return getInputs().get(0).getRecordType();
+    // The different inputs may have schemas that differ in nullability, but not in type.
+    // In that case we should return a schema that is nullable where any of the inputs is nullable.
```
andrew-coleman, line 50: Looking at the docs for this (https://substrait.io/relations/logical_relations/#set-operation-types), the output nullability depends on which set operation is being performed.

Blizzara: yep, I realized that as well but forgot to fix 😅 I'll try to tomorrow..

andrew-coleman (approved):

spark/src/test/scala/io/substrait/spark/SubstraitPlanTestBase.scala
```diff
@@ -95,10 +95,11 @@ trait SubstraitPlanTestBase { self: SharedSparkSession =>
     val extensionCollector = new ExtensionCollector
     val bytes = new RelProtoConverter(extensionCollector).toProto(substraitPlan).toByteArray
 
-    // Read it back
+    // Read it back and ensure the proto roundtrip doesn't modify the plan
     val protoPlan = io.substrait.proto.Rel.parseFrom(bytes)
     val substraitPlan2 =
       new ProtoRelConverter(extensionCollector, SparkExtension.COLLECTION).from(protoPlan)
+    substraitPlan2.shouldEqualPlainly(substraitPlan)
```
vbarua, line 102: This is a good check to have generally ✨

core/src/main/java/io/substrait/relation/Set.java
```diff
@@ -42,7 +46,56 @@ public static Set.SetOp fromProto(SetRel.SetOp proto) {
...
+      } else if (!prioritizeFirst && anyOtherIsNullable) {
+        // For other INTERSECTIONs and UNIONs: if any other field is nullable, type should be
+        // nullable
+        fields.add(TypeCreator.asNullable(typeA));
+      } else {
+        // Can keep nullability as-is
+        fields.add(typeA);
+      }
+    }
+    return Type.Struct.builder().fields(fields).nullable(first.nullable()).build();
```
vbarua, line 98: The spec has an associated example for this derivation that would be good to encode as a test to make sure we're doing the right thing: https://substrait.io/relations/logical_relations/#set-operation-types

This could also be handled as it's own change, unelss this is needed for one of the tests.

vbarua (commented):
Found some time to actually look at this. Have one comment about the nullability of Scalar Subqueries, and one requests for tests for the Set output type derivation logic.

core/src/main/java/io/substrait/relation/Set.java
Blizzara: Test added in https://github.com/substrait-io/substrait-java/pull/315/commits/a67d4a34d69f2b71bfbb71a1460392ad56a8c2e9, and that was a good call - turns out my logic wasn't correct either.

spark/src/main/scala/io/substrait/spark/logical/ToSubstraitRel.scala
```diff
@@ -550,10 +550,12 @@ private[logical] class WithLogicalSubQuery(toSubstraitRel: ToSubstraitRel)
     expr match {
       case s: ScalarSubquery if s.outerAttrs.isEmpty && s.joinCond.isEmpty =>
         val rel = toSubstraitRel.visit(s.plan)
+        val t =
+          s.plan.schema.fields.head // Using this instead of s.dataType/s.nullable to get correct nullability
```
Blizzara, line 563: There's two ScalarSubquery classes in Spark, the one we're using here is https://github.com/apache/spark/blob/9fe78e3d33499060467ecdc0c2631beae0b0316c/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/subquery.scala#L411 which always sets nullability to true. However the other one picks the nullability based on the plan's schema: https://github.com/apache/spark/blob/9fe78e3d33499060467ecdc0c2631beae0b0316c/sql/core/src/main/scala/org/apache/spark/sql/execution/subquery.scala#L68. I'm not very sure why there's a difference, but picking it from the plan seems to align better with what we expect here (and makes the tests pass)

core/src/main/java/io/substrait/expression/proto/ProtoExpressionConverter.java
```diff
@@ -196,7 +198,21 @@ public Expression from(io.substrait.proto.Expression expr) {
...
+                            new TypeVisitor.TypeThrowsVisitor<Type, RuntimeException>(
+                                "Expected struct field") {
+                              @Override
+                              public Type visit(Type.Struct type) throws RuntimeException {
+                                if (type.fields().size() != 1) {
+                                  throw new UnsupportedOperationException(
+                                      "Scalar subquery must have exactly one field");
+                                }
+                                // Result can be null if the query returns no rows
+                                return TypeCreator.asNullable(type.fields().get(0));
```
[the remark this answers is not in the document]
Blizzara: Issue was that Spark was always reporting nullable=true, which meant the from-proto didn't match the from-spark version. However this "fix" I had was wrong indeed, I made a more correct one here: https://github.com/substrait-io/substrait-java/pull/315/commits/4f7f2746d95d65120b1f466f84963c6ec4e4f1f2. Thanks!

spark/src/main/scala/io/substrait/spark/logical/ToSubstraitRel.scala
andrew-coleman: They both have `override def nullable: Boolean = true` (line 69 in your second link)?

core/src/main/java/io/substrait/relation/Set.java
```diff
@@ -42,7 +46,88 @@ public static Set.SetOp fromProto(SetRel.SetOp proto) {
...
+      Type typeA = first.fields().get(i);
+      if (!typeA.nullable()) {
+        // Just to make this case explicit and to short-circuit, logic below would work without too
+        fields.add(typeA);
+        continue;
+      }
+      int finalI = i;
+      boolean anyOtherIsRequired = rest.stream().anyMatch(t -> !t.fields().get(finalI).nullable());
+      fields.add(anyOtherIsRequired ? TypeCreator.asNotNullable(typeA) : typeA);
+    }
```
andrew-coleman, line 129: Looks good.  Tiny nit - both of these functions share a lot of common code (only 2 lines differ) - could they be refactored somehow?  Not important though :)

spark/src/main/scala/io/substrait/spark/logical/ToSubstraitRel.scala
Blizzara: Woops indeed, I can't read 🤦. Still, I think the change here is fine nonetheless.

core/src/main/java/io/substrait/relation/Set.java
Blizzara: I had it before as a single function, but it ends up making it harder to read since while the difference is small, it's non-trivial :/

vbarua (approved):
Changes look good. Will merge them shortly.

PR #316, opened 2024-11-05
fix(spark): incorrect deriveRecordType() for Expand

andrew-coleman:
In the Expand relation, the record type was being calculated incorrectly, leading to errors when round-tripping to protobuf and back.

### DIFF
--- a/core/src/main/java/io/substrait/relation/Expand.java
+++ b/core/src/main/java/io/substrait/relation/Expand.java
@@ -4,7 +4,6 @@ import io.substrait.expression.Expression;
 import io.substrait.type.Type;
 import io.substrait.type.TypeCreator;
 import java.util.List;
-import java.util.stream.Stream;
 import org.immutables.value.Value;
 
 @Value.Enclosing
@@ -18,7 +17,7 @@ public abstract class Expand extends SingleInputRel {
   public Type.Struct deriveRecordType() {
     Type.Struct initial = getInput().getRecordType();
     return TypeCreator.of(initial.nullable())
-        .struct(Stream.concat(initial.fields().stream(), Stream.of(TypeCreator.REQUIRED.I64)));
+        .struct(getFields().stream().map(ExpandField::getType));
   }
 
   @Override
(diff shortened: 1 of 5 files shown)


andrew-coleman:
Incorporates the test code changes in draft PR https://github.com/substrait-io/substrait-java/pull/314

core/src/main/java/io/substrait/relation/Expand.java
```diff
@@ -51,8 +54,8 @@ public static ImmutableExpand.ConsistentField.Builder builder() {
   public abstract static class SwitchingField implements ExpandField {
     public abstract List<Expression> getDuplicates();
 
-    public Type getType() {
-      return getDuplicates().get(0).getType();
+    public Stream<Type> getTypes() {
+      return getDuplicates().stream().map(Expression::getType);
```
Blizzara, line 58: maybe I'm misunderstanding something here - but are you sure this is right? The [Substrait spec comments](https://github.com/substrait-io/substrait/blob/2e13d0b7b4a2b5ac8221e7f02cc8f04456c8f11c/proto/substrait/algebra.proto#L500) say 
> All duplicates must return the same type class but may differ in nullability.

so it doesn't seem right to take a list of the duplicates' types.

core/src/main/java/io/substrait/relation/Expand.java
```diff
@@ -17,8 +17,11 @@ public abstract class Expand extends SingleInputRel {
   @Override
   public Type.Struct deriveRecordType() {
     Type.Struct initial = getInput().getRecordType();
-    return TypeCreator.of(initial.nullable())
-        .struct(Stream.concat(initial.fields().stream(), Stream.of(TypeCreator.REQUIRED.I64)));
+    var fields =
+        getFields().isEmpty()
+            ? initial.fields().stream()
+            : Stream.concat(initial.fields().stream(), getFields().get(0).getTypes());
```
Blizzara, line 23: this also doesn't seem right (but possible I'm not reading this right!) - [the spec](https://github.com/substrait-io/substrait/blob/2e13d0b7b4a2b5ac8221e7f02cc8f04456c8f11c/proto/substrait/algebra.proto#L475) says 
> In addition to a field being emitted per input field an extra int64 field is emitted which
> contains a zero-indexed ordinal corresponding to the duplicate definition.

so I'd expect the record type to be something like `initial.fields().map([for each field handle nullability based on duplicates]) + Type.Int64`

andrew-coleman:
Yes, I was on the wrong track entirely with this.  Turns out I needed to transpose the projections matrix that's defined in Spark before conversion to Substrait (and vice-versa).

Blizzara (approved):
I've been mulling over this given it's still not exactly as Substrait spec intends, but at least it's closer (and fixes the tests). So let's merge this. Still @andrew-coleman you understand the Expand better than I do, would you be able to figure out how to fully align Spark and Substrait here please?

Issue #319, opened 2024-11-29
AggregationPhase: AGGREGATION_PHASE_UNSPECIFIED not understood substrait-java


mbwhite:
The protobuf defines the `AGGREGATION_PHASE_UNSPECIFIED` but does say it implies INTERMEDIATE_TO_RESULT

https://github.com/substrait-io/substrait/blob/55683fb1078bd50b9fe711a8e0084910f34199e6/proto/substrait/algebra.proto#L1582

substrait-java treats this as unknown, when it isn't really. (in fact duckdb produces this)

testing fix for this locally

mbwhite:
Just to clairfy - that the substrait-java throws an exception if it encounters this field; this is part that is being fixed.

Up to consumers as you say to handle the actually meaning.

Blizzara:
Hmm, given the proto docs [say](https://github.com/substrait-io/substrait/blob/55683fb1078bd50b9fe711a8e0084910f34199e6/proto/substrait/algebra.proto#L1581) 
```
// Implies `INTERMEDIATE_TO_RESULT`.
```
does it make sense to let consumers handle it, or should we just treat it as INTERMEDIATE_TO_RESULT directly in substrait-java?

mbwhite:
@Blizzara the [pr](https://github.com/substrait-io/substrait-java/pull/320) just enables the the value to be correctly returned to the user to do as they wish. 

The issue was that an exception was thrown;

The PR has approval and is waiting for merge

vbarua:
PR has been merged

PR #320, opened 2024-11-29
feat(pojo): add UNSPECIFIED value to AggregationPhase enum

mbwhite:
Resolves issue #319

### DIFF
--- a/core/src/main/java/io/substrait/expression/Expression.java
+++ b/core/src/main/java/io/substrait/expression/Expression.java
@@ -940,6 +940,7 @@ public interface Expression extends FunctionArg {
   }
 
   enum AggregationPhase {
+    UNSPECIFIED(io.substrait.proto.AggregationPhase.AGGREGATION_PHASE_UNSPECIFIED),
     INITIAL_TO_INTERMEDIATE(
         io.substrait.proto.AggregationPhase.AGGREGATION_PHASE_INITIAL_TO_INTERMEDIATE),
     INTERMEDIATE_TO_INTERMEDIATE(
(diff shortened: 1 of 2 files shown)


core/src/main/java/io/substrait/expression/Expression.java
```diff
@@ -940,6 +940,7 @@ public static AggregationInvocation fromProto(AggregateFunction.AggregationInvoc
   }
 
   enum AggregationPhase {
+    UNSPECIFIED(io.substrait.proto.AggregationPhase.AGGREGATION_PHASE_UNSPECIFIED),
```
EpsilonPrime, line 943: There are probably a lot of switches that need to be updated appropriate.  A search for INITIAL_TO_RESULT found many potential instances across Java and Scala.

mbwhite: The enumerations here are mapping from the protobuf enumerations; in this case here one of the valid values in the protobuf was missing from the Java enumerations; 

If a plan contained that it would fail immediately; which in the case of a plan coming from duckdb it did.

EpsilonPrime (approved):

core/src/main/java/io/substrait/expression/Expression.java
vbarua: > There are probably a lot of switches that need to be updated as appropriate. 

There's actually a Java warning enabled for exhaustive switch statements, which should trigger CI failures. I didn't see any pattern matching blocks in Java which would be affected.

I did find one in Scala running `./gradlew check` locally. I filed https://github.com/substrait-io/substrait-java/issues/321 for the fact that it didn't fail in CI.

vbarua:
@mbwhite I found one block in the Scala code missing a pattern for the UNSPECIFIED, which I've added. If my changes look good to you I can merge this in.

vbarua (approved):
Checked in with @mbwhite on this. Changes looks good.

PR #322, opened 2025-01-06
fix(isthmus): support expressions in Sort

yassram:
## Motivation

This PR adds support for expressions in Sort (today, only input references are supported https://github.com/substrait-io/substrait-java/issues/192). 

### Details
This PR uses Calcite's `relbuilder.sort(rexNodes)` rather than computing `RelFieldCollation` (which required input references only). Calcite then makes sure to expand the Sort into a `LogicalSort` and a `LogicalProject` to evaluate expressions if needed (an additional `LogicalProject` is applied which acts like a remap to remove any introduced fields to preserve the same output).

##### e1:
```sql
SELECT foo FROM bar ORDER BY foo
```
```swift
LogicalSort(sort0=[$0], dir0=[ASC])
  TableScan(table=[bar])
```
##### e2:
```sql
SELECT foo FROM bar ORDER BY foo::int + 1 DESC NULLS LAST
```
```swift
LogicalProject(foo=[$0])
  LogicalSort(sort0=[$1], dir0=[DESC-nulls-last])
    LogicalProject(foo=[$0], $f1=[+(CAST($0):BIGINT, 1)])
      TableScan(table=[bar])
```

See `ComplexSortTest.java` for more examples

### DIFF
--- a/isthmus/src/test/java/io/substrait/isthmus/ComplexSortTest.java
+++ b/isthmus/src/test/java/io/substrait/isthmus/ComplexSortTest.java
@@ -171,7 +171,7 @@ public class ComplexSortTest extends PlanTestBase {
   @Test
   void handleComplex2ExpressionSort() {
     // CREATE TABLE example (a VARCHAR, b INT)
-    // SELECT b, a FROM example ORDER BY a::INT > 0 DESC, -b + 42 ASC NULLS LAST
+    // SELECT b, a FROM example ORDER BY a::INT DESC, -b + 42 ASC NULLS LAST
 
     Rel rel =
         b.project(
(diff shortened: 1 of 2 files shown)


isthmus/src/test/java/io/substrait/isthmus/ComplexSortTest.java
```diff
@@ -0,0 +1,205 @@
...
+           Collation: [1]
+           LogicalSort(sort0=[$1], dir0=[ASC])
+             LogicalProject(a=[$0], a0=[CAST($0):INTEGER NOT NULL])
+               LogicalTableScan(table=[[example]])
+         """;
+
+    RelNode relReturned = substraitToCalcite.convert(rel);
+    var sw = new StringWriter();
+    relReturned.explain(new CollationRelWriter(sw));
+    assertEquals(expected, sw.toString());
```
vbarua, line 104: In your test strings
```
        """
         LogicalProject(a0=[$0])
           Collation: [1]
           LogicalSort(sort0=[$1], dir0=[ASC])
             LogicalProject(a=[$0], a0=[CAST($0):INTEGER NOT NULL])
               LogicalTableScan(table=[[example]])
         """;
```
I'm not sure how much extra value the `Collation` line is adding because that's already visible from the `LogicalSort`, albeit more verbosely. I think it would be cleaner to just compare the expected result to `relReturned.explain()`

vbarua (commented):
Overall looks good to me. It's good to be able to rely on built-in Calcite functionality to expand complex expressions in the sort fields.

I did leave one comment on your tests. The `CollationRelWriter` seems somewhat redundant given that that the same information is visible in the `LogicalSort`s

isthmus/src/test/java/io/substrait/isthmus/ComplexSortTest.java
yassram: The value in this tests for now is the absence of `Collation` on the top `RelNode` rather than the information of the collation itself. This means that the collation is not always propagated correctly in the logical plan (due to the "remap" that removes the collation field that is introduced for the `LogicalSort` by calcite in a child `LogicalProject`). 
I'm hopping that a following PR would fix this behavior, and I thought that displaying the `Collation` in this first PR would help to see the impact of the change on the propagation.

vbarua: > The value in this tests for now is the absence of Collation on the top RelNode rather than the information of the collation itself. 

Gotcha. A comment to that effect on the `CollationRelWriter` would be useful to highlight that.

> I thought that displaying the Collation in this first PR would help to see the impact of the change on the propagation.

A, perfect.

yassram: Added docstring to `CollationRelWriter` 👍

vbarua (approved):
Thanks for working on this!

PR #323, opened 2025-01-29
feat(spark): add support for DelimiterSeparatedTextReadOptions

andrew-coleman:
The handling of CSV in the spark module works only in a very limited way with many hard-coded assumptions.
This commit adds full support for delimited text support as defined in the `FileOrFiles` proto message


Blizzara (approved):
Looks good to me!

andrew-coleman:
Not sure why `osv-scanner` is failing.  Is that related to this PR?  It works locally with no issues. 🤔

.editorconfig
```diff
@@ -10,7 +10,7 @@ trim_trailing_whitespace = true
 [*.{yaml,yml}]
 indent_size = 2
 
-[{**/*.sql,**/OuterReferenceResolver.md,**gradlew.bat}]
+[{**/*.sql,**/OuterReferenceResolver.md,**gradlew.bat,**/*.parquet,**/*.orc}]
```
Blizzara, line 13: is this needed, no-one should be editing `.parquet` or `.orc` files by hand afaik?

spark/src/main/scala/io/substrait/spark/logical/ToLogicalPlan.scala
```diff
@@ -331,6 +334,34 @@ class ToLogicalPlan(spark: SparkSession) extends DefaultRelVisitor[LogicalPlan]
...
+          "quote" -> csv.getQuote,
+          "header" -> (csv.getHeaderLinesToSkip match {
+            case 0 => "false"
+            case 1 => "true"
+            case _ =>
+              throw new UnsupportedOperationException(
+                s"Cannot configure CSV reader to skip ${csv.getHeaderLinesToSkip} rows")
+          }),
+          "escape" -> csv.getEscape,
+          "maxColumns" -> csv.getMaxLineSize.toString
```
Blizzara, line 356: this sounds wrong, maxColumns in spark is for _columns_ while maxLineSize in Substrait is for _bytes_

spark/src/main/scala/io/substrait/spark/logical/ToSubstraitRel.scala
```diff
@@ -378,8 +381,25 @@ class ToSubstraitRel extends AbstractLogicalPlanVisitor with Logging {
...
 
-    val ff = new FileFormat.ParquetReadOptions {
-      override def toString: String = "csv" // TODO this is hardcoded at the moment
+    val format = fsRelation.fileFormat match {
+      case _: CSVFileFormat =>
+        new FileFormat.DelimiterSeparatedTextReadOptions {
+          // default values for options specified here:
+          // https://spark.apache.org/docs/latest/sql-data-sources-csv.html#data-source-option
+          override def getFieldDelimiter: String = fsRelation.options.getOrElse("delimiter", ",")
+          override def getMaxLineSize: Long =
```
Blizzara, line 390: same here

Blizzara (changes requested):

.editorconfig
andrew-coleman: I'm not sure, but the `editorconfig-checker` stage failed prior to adding this (at least for `.orc`, didn't do a separate check for `.parquet`):
https://github.com/substrait-io/substrait-java/actions/runs/13032936614/job/36356268259

spark/src/main/scala/io/substrait/spark/logical/ToLogicalPlan.scala
andrew-coleman: Good catch.  There doesn't seem to be a Spark equivalent option for this, so I'll just remove the mapping and ignore.

spark/src/main/scala/io/substrait/spark/logical/ToSubstraitRel.scala
andrew-coleman: Same as above, but I need to assign something, so I'll use the proto default value of `0`.

.editorconfig
Blizzara: Weird but okay :D

spark/src/main/scala/io/substrait/spark/logical/ToSubstraitRel.scala
```diff
@@ -378,8 +381,24 @@ class ToSubstraitRel extends AbstractLogicalPlanVisitor with Logging {
...
 
-    val ff = new FileFormat.ParquetReadOptions {
-      override def toString: String = "csv" // TODO this is hardcoded at the moment
+    val format = fsRelation.fileFormat match {
+      case _: CSVFileFormat =>
+        new FileFormat.DelimiterSeparatedTextReadOptions {
+          // default values for options specified here:
+          // https://spark.apache.org/docs/latest/sql-data-sources-csv.html#data-source-option
+          override def getFieldDelimiter: String = fsRelation.options.getOrElse("delimiter", ",")
+          override def getMaxLineSize: Long = 0 // No Spark equivalent. Assign proto default of 0
```
Blizzara, line 390: this is a bit weird still, but I see that it's a part of the proto, so I don't have better ideas either. I guess only problem would appear if someone uses spark-produced plans in some other system which would support & implement this, but if that happens we can figure it out then, I think.

Blizzara (approved):

PR #324, opened 2025-01-31
chore(spark): pin osv-scanner to v.1.9.2

andrew-coleman:
The `ghcr.io/google/osv-scanner:latest` docker image was updated to `v2.0.0-beta1`.
This breaks the build with the error:
`Failed to parse SBOM "/bom.json" with error: could not determine extractor suitable to this file`

This commit pins the image tag to last working release `v1.9.2`

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -69,7 +69,7 @@ jobs:
         with:
           name: cyclonedx-sbom
       - name: Scan
-        run: docker run --rm -v "${PWD}/${{ matrix.project }}/build/reports/bom.json:/bom.json" ghcr.io/google/osv-scanner --sbom /bom.json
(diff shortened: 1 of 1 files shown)


.github/workflows/pr.yml
```diff
@@ -130,7 +130,9 @@ jobs:
     - name: Install GraalVM native image
       run: gu install native-image
     - name: Build with Gradle
-      run: gradle nativeImage
+      run: |
+        ln -s isthmus-cli/proxies.json
```
andrew-coleman, line 134: The macos builder is failing because it is looking in the wrong directory for this file.  Not sure why, but to work around the problem, I've added a symlink to where it's looking.  Just to get the build going again pending a 'proper' fix

EpsilonPrime (approved):

PR #326, opened 2025-02-05
fix(isthmus): more standard SQL for TPCDS query 72

bestbeforetoday:
According to a related [Calcite issue](https://issues.apache.org/jira/browse/CALCITE-304), TPCDS query 72 uses Oracle-specific syntax for date addition, which defaults to days as the time unit when a numeric value is added to a date. A more standards-compliant form is to explicitly specify the time unit as an interval.


Blizzara (approved):

Blizzara:
Thanks!

PR #328, opened 2025-02-12
build: update to Calcite 1.38.0

vbarua:


vbarua:
My intent with this was to fix the build issues around the json-smart dependency, which is handled more succintly in https://github.com/substrait-io/substrait-java/pull/329

PR #334, opened 2025-02-27
build: correctly declare api dependencies

nielspardon:
Currently, both the `isthmus` and `core` projects are declaring all dependencies as `implementation` dependencies. Gradle has two types of dependencies `api` and `implementation` dependencies. From the Gradle documentation:

> The `api` configuration should be used to declare dependencies which are exported by the library API, whereas the `implementation` configuration should be used to declare dependencies which are internal to the component.

https://docs.gradle.org/current/userguide/java_library_plugin.html#sec:java_library_separation

For the [`core`](https://github.com/substrait-io/substrait-java/tree/main/core) project some of the protobuf Java classes are part of the Substrait Java API, e.g. `com.google.protobuf.ByteString` is exposed as part of the [`io.substrait.expression.Expression`](https://github.com/substrait-io/substrait-java/blob/d8f22f71d4b22b2047f5c6c97fc360342088334b/core/src/main/java/io/substrait/expression/Expression.java#L190) interface. Which means we should declare `com.google.protobuf:protobuf-java` as an `api` dependency instead of an `implementation` dependency.

Similarly, in the [`isthmus`](https://github.com/substrait-io/substrait-java/tree/main/isthmus) project classes coming from the dependency on the `core` project are part of the isthmus API and should be reported as an `api` dependency. Additionally, the Apache Calcite classes are part of the isthmus API and so the `org.apache.calcite:calcite-core` should also be declared as an `api` dependency.

In order to correctly declare these `api` dependencies we need to switch from the `java` to the `java-library` Gradle plugin.

If such dependencies are not declared as `api` dependencies Gradle projects depending on the `core` or `isthmus` libraries are forced to repeat the dependency declaration on the respective dependencies.

### DIFF
--- a/core/build.gradle.kts
+++ b/core/build.gradle.kts
@@ -2,7 +2,7 @@ import org.gradle.plugins.ide.idea.model.IdeaModel
 
 plugins {
   `maven-publish`
-  id("java")
+  id("java-library")
   id("idea")
   id("antlr")
   id("com.google.protobuf") version "0.9.4"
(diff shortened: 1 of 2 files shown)


vbarua (approved):
Thanks for this change, along with the super-clear explanation ✨

PR #335, opened 2025-02-28
fix(isthmus): sql day/time interval conversion

### LINKED ISSUE #327: [isthmus] INTERVAL DAY is interpreted incorrectly
When converting an INTERVAL DAY to Substrait using Isthmus, the result interprets the interval incorrectly. For example, with this input:

```sql
select interval '1' day
```

The following Substrait is generated:

```json
{
  "extensionUris": [],
  "extensions": [],
  "relations": [{
    "root": {
      "input": {
        "read": {
          "common": {
            "direct": {
            }
          },
          "baseSchema": {
            "names": ["EXPR$0"],
            "struct": {
              "types": [{
                "intervalDay": {
                  "typeVariationReference": 0,
                  "nullability": "NULLABILITY_REQUIRED",
                  "precision": 6
                }
              }],
              "typeVariationReference": 0,
              "nullability": "NULLABILITY_REQUIRED"
            }
          },
          "virtualTable": {
            "values": [{
              "fields": [{
                "intervalDayToSecond": {
                  "days": 0,
                  "seconds": 86,
                  "precision": 6,
                  "subseconds": "400000"
                },
                "nullable": false,
                "typeVariationReference": 0
              }]
            }]
          }
        }
      },
      "names": ["EXPR$0"]
    }
  }],
  "expectedTypeUrls": []
}
```

Note that, instead of an intervalDayToSecond with a value of 1 day, it produces a value of 86.4 seconds.

What appears to happen is that parsing the SQL produces a RexLiteral to represent the interval with the value `86400000:INTERVAL DAY`. This is the number of **milliseconds** in a day. The conversion in the Isthmus [LiteralConverter](https://github.com/substrait-io/substrait-java/blob/b4afdb25364eefa2094c7d86c2c19581a8054e1b/isthmus/src/main/java/io/substrait/isthmus/expression/LiteralConverter.java#L164-L187) then treats the value as **microseconds** (using a scale of 6) and produces an incorrect result.

bestbeforetoday:
RexLiteral day/time intervals were converted to Substrait using their scale. Despite all day/time interval RexLiteral created from SQL by Calcite having a (default) scale of 6, their value is always in milliseconds. This resulted in intervals generated by Calcite from SQL being wrong by a factor of 1000 seconds.

This change treats day/time interval RexLiteral values as milliseconds regardless of their claimed scale.

Closes #327

### DIFF
--- a/isthmus/src/main/java/io/substrait/isthmus/expression/LiteralConverter.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/expression/LiteralConverter.java
@@ -15,8 +15,8 @@ import java.time.*;
 import java.time.format.DateTimeFormatter;
 import java.time.format.DateTimeFormatterBuilder;
 import java.util.List;
+import java.util.Objects;
 import java.util.Optional;
-import java.util.concurrent.TimeUnit;
 import java.util.stream.Collectors;
 import org.apache.calcite.rel.type.RelDataType;
 import org.apache.calcite.rex.RexLiteral;
@@ -31,8 +31,6 @@ public class LiteralConverter {
   public LiteralConverter(TypeConverter typeConverter) {
     this.typeConverter = typeConverter;
   }
-
-  private static final long MICROS_IN_DAY = TimeUnit.DAYS.toMicros(1);
 
   static final DateTimeFormatter CALCITE_LOCAL_DATE_FORMATTER = DateTimeFormatter.ISO_LOCAL_DATE;
   static final DateTimeFormatter CALCITE_LOCAL_TIME_FORMATTER =
@@ -96,7 +94,7 @@ public class LiteralConverter {
       case SMALLINT -> i16(n, i(literal).intValue());
       case INTEGER -> i32(n, i(literal).intValue());
       case BIGINT -> i64(n, i(literal).longValue());
-      case BOOLEAN -> bool(n, (Boolean) literal.getValue());
+      case BOOLEAN -> bool(n, literal.getValueAs(Boolean.class));
       case CHAR -> {
         var val = literal.getValue();
         if (val instanceof NlsString nls) {
@@ -104,8 +102,9 @@ public class LiteralConverter {
         }
         throw new UnsupportedOperationException("Unable to handle char type: " + val);
       }
-      case FLOAT, DOUBLE -> fp64(n, ((Number) literal.getValue()).doubleValue());
-      case REAL -> fp32(n, ((Number) literal.getValue()).floatValue());
+      case FLOAT, DOUBLE -> fp64(n, literal.getValueAs(Double.class));
+      case REAL -> fp32(n, literal.getValueAs(Float.class));
+
       case DECIMAL -> {
         BigDecimal bd = bd(literal);
         yield decimal(n, bd, literal.getType().getPrecision(), literal.getType().getScale());
@@ -155,7 +154,7 @@ public class LiteralConverter {
         yield timestamp(n, ldt);
       }
       case INTERVAL_YEAR, INTERVAL_YEAR_MONTH, INTERVAL_MONTH -> {
-        var intervalLength = literal.getValueAs(BigDecimal.class).longValue();
+        long intervalLength = Objects.requireNonNull(literal.getValueAs(Long.class));
         var years = intervalLength / 12;
         var months = intervalLength - years * 12;
         yield intervalYear(n, (int) years, (int) months);
@@ -170,19 +169,15 @@ public class LiteralConverter {
           INTERVAL_MINUTE,
           INTERVAL_MINUTE_SECOND,
           INTERVAL_SECOND -> {
-        // we need to convert to microseconds.
-        // TODO: don't need to anymore
-        int scale = literal.getType().getScale();
-        var intervalLength = literal.getValueAs(BigDecimal.class).longValue();
-        var adjustedLength =
-            scale > 6
-                ? intervalLength / ((int) Math.pow(10, scale - 6))
-                : intervalLength * ((int) Math.pow(10, 6 - scale));
-        var days = adjustedLength / MICROS_IN_DAY;
-        var totalMicroseconds = adjustedLength - days * MICROS_IN_DAY;
-        var seconds = totalMicroseconds / 1_000_000;
-        var microseconds = totalMicroseconds - 1_000_000 * seconds;
-        yield intervalDay(n, (int) days, (int) seconds, microseconds, 6 /* micros */);
+        // Calcite represents day/time intervals in milliseconds, despite a default scale of 6.
+        var totalMillis = Objects.requireNonNull(literal.getValueAs(Long.class));
+        var interval = Duration.ofMillis(totalMillis);
+
+        var days = interval.toDays();
+        var seconds = interval.minusDays(days).toSeconds();
+        var micros = interval.toMillisPart() * 1000;
+
+        yield intervalDay(n, (int) days, (int) seconds, micros, 6);
       }
 
       case ROW -> {
--- a/isthmus/src/test/java/io/substrait/isthmus/CalciteLiteralTest.java
+++ b/isthmus/src/test/java/io/substrait/isthmus/CalciteLiteralTest.java
@@ -216,6 +216,29 @@ public class CalciteLiteralTest extends CalciteObjs {
   }
 
   @Test
+  void tIntervalDay() {
+    // Calcite always uses milliseconds
+    BigDecimal bd = new BigDecimal(TimeUnit.DAYS.toMillis(5));
+    RexLiteral intervalDayLiteral =
+        rex.makeIntervalLiteral(
+            bd,
+            new SqlIntervalQualifier(
+                org.apache.calcite.avatica.util.TimeUnit.DAY, -1, null, -1, SqlParserPos.ZERO));
+    var intervalDayExpr = intervalDay(false, 5, 0, 0, 6);
+
+    // rex --> expression
+    var convertedExpr = intervalDayLiteral.accept(rexExpressionConverter);
+    assertEquals(intervalDayExpr, convertedExpr);
+
+    // expression -> rex
+    RexLiteral convertedRex = (RexLiteral) intervalDayExpr.accept(expressionRexConverter);
+
+    // Compare value only. Ignore the precision in SqlIntervalQualifier in comparison.
+    assertEquals(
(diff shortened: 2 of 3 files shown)


bestbeforetoday:
I am not entirely satisfied with this as a fix, although there look to be other places in the existing code that assumes that Calcite intervals are millisecond precision. I will continue to see if there is a non-breaking way to change the Calcite behaviour so that substrait-java can use the RexLiteral scale value. If that turns out to be possible then I can revisit this.

isthmus/src/main/java/io/substrait/isthmus/expression/LiteralConverter.java
```diff
@@ -96,16 +94,17 @@ public Expression.Literal convert(RexLiteral literal) {
...
         var val = literal.getValue();
         if (val instanceof NlsString nls) {
           yield fixedChar(n, nls.getValue());
         }
         throw new UnsupportedOperationException("Unable to handle char type: " + val);
       }
-      case FLOAT, DOUBLE -> fp64(n, ((Number) literal.getValue()).doubleValue());
-      case REAL -> fp32(n, ((Number) literal.getValue()).floatValue());
+      case FLOAT, DOUBLE -> fp64(n, literal.getValueAs(Double.class));
+      case REAL -> fp32(n, literal.getValueAs(Float.class));
```
vbarua, line 106: TIL about `getValueAs`

isthmus/src/main/java/io/substrait/isthmus/expression/LiteralConverter.java
```diff
@@ -170,19 +169,15 @@ public Expression.Literal convert(RexLiteral literal) {
...
+        // Calcite represents day/time intervals in milliseconds, despite a default scale of 6.
+        var totalMillis = Objects.requireNonNull(literal.getValueAs(Long.class));
+        var interval = Duration.ofMillis(totalMillis);
+
+        var days = interval.toDays();
+        var seconds = interval.minusDays(days).toSeconds();
+        var micros = interval.toMillisPart() * 1000;
+
+        yield intervalDay(n, (int) days, (int) seconds, micros, 6);
       }
```
vbarua, line 181: This is also consistent with the conversion from Substrait to Calcite in 
https://github.com/substrait-io/substrait-java/blob/fe08fd42dbe8d343ded89c60b2f325972ecce60d/isthmus/src/main/java/io/substrait/isthmus/expression/ExpressionRexConverter.java#L236-L246

vbarua (commented):
Overall, this looks reasonable to me.

@Blizzara could you take a look as well as you've worked on this in the past.

vbarua (approved):

isthmus/src/main/java/io/substrait/isthmus/expression/LiteralConverter.java
```diff
@@ -170,19 +169,15 @@ public Expression.Literal convert(RexLiteral literal) {
...
-            scale > 6
-                ? intervalLength / ((int) Math.pow(10, scale - 6))
-                : intervalLength * ((int) Math.pow(10, 6 - scale));
-        var days = adjustedLength / MICROS_IN_DAY;
-        var totalMicroseconds = adjustedLength - days * MICROS_IN_DAY;
-        var seconds = totalMicroseconds / 1_000_000;
-        var microseconds = totalMicroseconds - 1_000_000 * seconds;
-        yield intervalDay(n, (int) days, (int) seconds, microseconds, 6 /* micros */);
+        // Calcite represents day/time intervals in milliseconds, despite a default scale of 6.
+        var totalMillis = Objects.requireNonNull(literal.getValueAs(Long.class));
```
Blizzara, line 173: so is the case here that Calcite has _always_
- literal.getType().getScale() = 6
- literal.getValue() returns millis?

isthmus/src/main/java/io/substrait/isthmus/expression/LiteralConverter.java
```diff
@@ -170,19 +169,15 @@ public Expression.Literal convert(RexLiteral literal) {
...
+        var interval = Duration.ofMillis(totalMillis);
+
+        var days = interval.toDays();
+        var seconds = interval.minusDays(days).toSeconds();
+        var micros = interval.toMillisPart() * 1000;
+
+        yield intervalDay(n, (int) days, (int) seconds, micros, 6);
```
Blizzara, line 180: alternatively you could return a precision of 3 and the millis value 🤷

isthmus/src/test/java/io/substrait/isthmus/CalciteLiteralTest.java
```diff
@@ -215,6 +215,29 @@ void tIntervalMillisecond() {
     bitest(intervalDaySecondExpr, intervalDaySecond);
   }
 
+  @Test
+  void tIntervalDay() {
+    // Calcite always uses milliseconds
+    BigDecimal bd = new BigDecimal(TimeUnit.DAYS.toMillis(5));
```
Blizzara, line 221: maybe use some value that's days + seconds + millis rather than just days?

isthmus/src/main/java/io/substrait/isthmus/expression/LiteralConverter.java
```diff
@@ -155,7 +154,7 @@ public Expression.Literal convert(RexLiteral literal) {
         yield timestamp(n, ldt);
       }
       case INTERVAL_YEAR, INTERVAL_YEAR_MONTH, INTERVAL_MONTH -> {
-        var intervalLength = literal.getValueAs(BigDecimal.class).longValue();
+        long intervalLength = Objects.requireNonNull(literal.getValueAs(Long.class));
```
Blizzara, line 157: I was curious why this change is okay but getValueAs() returns the object casted if the type matches, otherwise it has some conversion logic for some types, so here it'll first cast the object to BigDecimal and then call `.longValue()` on it, so basically what the previous code was doing 👍

Blizzara:
I'm not familiar with Calcite, but seems okay to me! I left couple notes/questions just in case they're helpful

isthmus/src/main/java/io/substrait/isthmus/expression/LiteralConverter.java
Blizzara: Ah, you actually already answered that in the PR description:
> Despite all day/time interval RexLiteral created from SQL by Calcite having a (default) scale of 6, their value is always in milliseconds. 

👌

isthmus/src/test/java/io/substrait/isthmus/CalciteLiteralTest.java
bestbeforetoday: This test is specifically testing day intervals since that is the case I hit the issue. It wouldn't be a good test of day intervals if it actually tested other time units. I am happy to add additional test cases though. There is already a day/decond test immediately above this one in the file. What other cases do you think it would be valuable to cover? Just day + second + millis?

isthmus/src/main/java/io/substrait/isthmus/expression/LiteralConverter.java
bestbeforetoday: Actually the `literal.getValueAs(BigDecimal.class)` that was there previously calls `liternal.getValueAs(Long.class)` internally before then converting it into a BigDecimal... which we then convert back to a long. It seems more sensible to just ask for the type we want. Functionally it is the same.

isthmus/src/main/java/io/substrait/isthmus/expression/LiteralConverter.java
bestbeforetoday: I don't know which approach is better. Given the current Calcite behaviour, 3 is correct. However, Calcite has a default fractional second precision of 6. My view is that Calcite is not behaving correctly and really should be returning the precision is says it is. If that gets fixed at some point, the precision here will revert to 6. I figured it was less impact not to change the precision (for now) to minimise the impact of this change to isthmus users, but very happy to take your guidance on the best approach. Just let me know what you prefer.

isthmus/src/main/java/io/substrait/isthmus/expression/LiteralConverter.java
Blizzara: Hm, I think the getValueAs(Decimal) on a interval type would end up in this code at the beginning of `getValueAs`:
```
    if (value == null || clazz.isInstance(value)) {
      return clazz.cast(value);
    }
```
while now we're going to 
```
    case INTERVAL_YEAR:
...
} else if (clazz == Long.class) {
        return clazz.cast(((BigDecimal) value).longValue());
```

isthmus/src/main/java/io/substrait/isthmus/expression/LiteralConverter.java
Blizzara: I have slight preference for returning "currently true" values (ie. 3) rather than "should be true" values, especially given it's not certain if Calcite will ever be fixed - and downstream users should anyways be willing to deal with either. But I have no personal experience or investment with Calcite so that's a very weak opinion.

isthmus/src/main/java/io/substrait/isthmus/expression/LiteralConverter.java
bestbeforetoday: Do you need me to change this back to we explicitly ask for a type we don't want, or is it OK to leave this change so we ask for the type we actually want and Calcite provides it in whichever way is best fits its current implementation?

Blizzara: Oh sorry this is totally fine by me 👍

isthmus/src/main/java/io/substrait/isthmus/expression/LiteralConverter.java
bestbeforetoday: I tried changing the precision to 3 and it throws up some problems with the existing tests in round-trip comparisons. Converting from Rex to Substrait we now get a sub-second precision of 3 and a millisecond value, which is fine. Converting the other way from Substrait to Rex gives a RexLiteral with a sub-second precision of 6 and a millisecond value, despite the qualifier being passed specifying a precision of 3. Calcite (currently) always stores intervals in milliseconds, and always specifies a sub-second precision of 6, despite the supplied qualifier being a correct sub-second interval of 3. It is very frustrating!

I know Calcite is really the correct place to fix this, and then picking up that change here will potentially cause some breakages. The Calcite fix seems much more involved.

PR #337, opened 2025-03-05
feat(pojo): support best_effort_filter on ReadRel messages

freakyzoidberg:
ReadRel spec contains a BestEffortFilter expression that is ignored by the Java ProtoRel and RelProto converter.

Adding support to it, similarly to the existing Filter expression.

Thanks


Blizzara (approved):
Thanks, seems sensible to me!

Issue #105, opened 2022-11-22
[ISTHMUS] Issues in string concatenation


YBRua:
Isthmus may run into problems when converting a string concatenation call to substrait plan.

#### 1. It does not support concatenating two string literals

The input to Isthmus is

```
"SELECT 'foo' || 'bar' FROM test;" -c "CREATE TABLE test(col_1 INTEGER NOT NULL, col_2 VARCHAR(10));"
```

and it throws an error. I suspect that this is because substrait only have two signatures for `concat`: `concat:vchar_vchar` and `concat:str_str`. In this case, both inputs are fixed char `fchar` so the signature match fails. Perhaps adding a cast somewhere can fix this?

Full stack trace below:

```
java.lang.IllegalArgumentException: Unable to convert call ||(char<3>, char<3>).
        at io.substrait.isthmus.expression.RexExpressionConverter.visitCall(RexExpressionConverter.java:78)
        at io.substrait.isthmus.expression.RexExpressionConverter.visitCall(RexExpressionConverter.java:19)
        at org.apache.calcite.rex.RexCall.accept(RexCall.java:189)
        at io.substrait.isthmus.SubstraitRelVisitor.toExpression(SubstraitRelVisitor.java:99)
        at java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:197)
        at java.util.Collections$2.tryAdvance(Collections.java:4853)
        at java.util.Collections$2.forEachRemaining(Collections.java:4861)
        at java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:509)
        at java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:499)
        at java.util.stream.ReduceOps$ReduceOp.evaluateSequential(ReduceOps.java:921)
        at java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234)
        at java.util.stream.ReferencePipeline.collect(ReferencePipeline.java:682)
        at io.substrait.isthmus.SubstraitRelVisitor.visit(SubstraitRelVisitor.java:156)
        at io.substrait.isthmus.SubstraitRelVisitor.visit(SubstraitRelVisitor.java:61)
        at io.substrait.isthmus.RelNodeVisitor.reverseAccept(RelNodeVisitor.java:95)
        at io.substrait.isthmus.SubstraitRelVisitor.apply(SubstraitRelVisitor.java:351)
[...]

PR #338, opened 2025-03-05
fix(isthmus): concat of different string types

### LINKED ISSUE #69: [ISTHMUS] [TPCDS] Fix Concat scalar function support
For example:
- `q5` has the expression: `'web_site' || web_site_id as id`
- `q66` has the expression: `'DHL' || ',' || 'BARIAN' as ship_carriers`
- `q80` has expression using ||

Implicit casting is not 

Queries in tpcds having this issue: `[5, 66, 80, 84]`

How to reproduce:
 Add a test to `TpcdsQueryNoValidation` like this:

```java
@ParameterizedTest
  @ValueSource(ints = {5})
  public void tpcdsQ5(int query) throws Exception {
    testQuery(query);
  }
```

On fixing, please move the fixed queries from the `tpcdsFailure` list to the `tpcdsSuccess` list

bestbeforetoday:
The type matching used when mapping Calcite to Substrait expressions required parameters and return type of the concat function to be either all varchar or all string. Any mixing of types caused a failure.

This change relaxes the type matching to allow string types (varchar, char and string literal) to be used interchangeably.

Closes #69
Contributes to #105

### DIFF
--- a/isthmus/src/main/java/io/substrait/isthmus/expression/FunctionConverter.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/expression/FunctionConverter.java
@@ -25,6 +25,7 @@ import java.util.IdentityHashMap;
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
 import java.util.function.Function;
@@ -149,7 +150,6 @@ public abstract class FunctionConverter<
     private final SqlOperator operator;
     private final List<F> functions;
     private final Map<String, F> directMap;
-    private final SignatureMatcher<F> matcher;
     private final Optional<SingularArgumentMatcher<F>> singularInputType;
     private final Util.IntRange argRange;
 
@@ -161,7 +161,6 @@ public abstract class FunctionConverter<
           Util.IntRange.of(
               functions.stream().mapToInt(t -> t.getRange().getStartInclusive()).min().getAsInt(),
               functions.stream().mapToInt(t -> t.getRange().getEndExclusive()).max().getAsInt());
-      this.matcher = getSignatureMatcher(operator, functions);
       this.singularInputType = getSingularInputType(functions);
       var directMap = ImmutableMap.<String, F>builder();
       for (var func : functions) {
@@ -178,21 +177,18 @@ public abstract class FunctionConverter<
       return argRange.within(count);
     }
 
-    private static <F extends SimpleExtension.Function> SignatureMatcher<F> getSignatureMatcher(
-        SqlOperator operator, List<F> functions) {
-      return (inputTypes, outputType) -> {
-        for (F function : functions) {
-          List<SimpleExtension.Argument> args = function.requiredArguments();
-          // Make sure that arguments & return are within bounds and match the types
-          if (function.returnType() instanceof ParameterizedType
-              && isMatch(outputType, (ParameterizedType) function.returnType())
-              && inputTypesSatisfyDefinedArguments(inputTypes, args)) {
-            return Optional.of(function);
-          }
+    private Optional<F> signatureMatch(List<Type> inputTypes, Type outputType) {
+      for (F function : functions) {
+        List<SimpleExtension.Argument> args = function.requiredArguments();
+        // Make sure that arguments & return are within bounds and match the types
+        if (function.returnType() instanceof ParameterizedType
+            && isMatch(outputType, (ParameterizedType) function.returnType())
+            && inputTypesMatchDefinedArguments(inputTypes, args)) {
+          return Optional.of(function);
         }
+      }
 
-        return Optional.empty();
-      };
+      return Optional.empty();
     }
 
     /**
@@ -208,7 +204,7 @@ public abstract class FunctionConverter<
      * @param args expected arguments as defined in a {@link SimpleExtension.Function}
      * @return true if the {@code inputTypes} satisfy the {@code args}, false otherwise
      */
-    private static boolean inputTypesSatisfyDefinedArguments(
+    private static boolean inputTypesMatchDefinedArguments(
         List<Type> inputTypes, List<SimpleExtension.Argument> args) {
 
       Map<String, Set<Type>> wildcardToType = new HashMap<>();
@@ -318,7 +314,7 @@ public abstract class FunctionConverter<
 
       assert (rexOperands.size() == opTypes.size());
 
-      if (rexOperands.size() == 0) {
+      if (rexOperands.isEmpty()) {
         return Stream.of("");
       } else {
         List<List<String>> argTypeLists =
@@ -357,13 +353,12 @@ public abstract class FunctionConverter<
       // try to do a direct match
       List<String> typeStrings =
           opTypes.stream().map(t -> t.accept(ToTypeString.INSTANCE)).collect(Collectors.toList());
-      Stream<String> possibleKeys =
-          matchKeys(call.getOperands().collect(Collectors.toList()), typeStrings);
+      Stream<String> possibleKeys = matchKeys(operandsList, typeStrings);
 
       Optional<String> directMatchKey =
           possibleKeys
               .map(argList -> name + ":" + argList)
-              .filter(k -> directMap.containsKey(k))
+              .filter(directMap::containsKey)
               .findFirst();
 
       if (directMatchKey.isPresent()) {
@@ -376,14 +371,13 @@ public abstract class FunctionConverter<
                       RexNode r = operandsList.get(i);
                       Expression o = operands.get(i);
                       if (EnumConverter.isEnumValue(r)) {
-                        return EnumConverter.fromRex(variant, (RexLiteral) r, i)
-                            .orElseGet(() -> null);
+                        return EnumConverter.fromRex(variant, (RexLiteral) r, i).orElse(null);
                       } else {
                         return o;
                       }
                     })
                 .collect(Collectors.toList());
-        boolean allArgsMapped = funcArgs.stream().filter(e -> e == null).findFirst().isEmpty();
+        boolean allArgsMapped = funcArgs.stream().filter(Objects::isNull).findFirst().isEmpty();
         if (allArgsMapped) {
           return Optional.of(generateBinding(call, variant, funcArgs, outputType));
         } else {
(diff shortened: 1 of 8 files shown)


isthmus/src/test/java/io/substrait/isthmus/ExpressionConvertabilityTest.java
```diff
@@ -65,6 +65,21 @@ public void inPredicate() throws IOException, SqlParseException {
             + "(SELECT L_SUPPKEY from LINEITEM where L_SUPPKEY < L_ORDERKEY)");
   }
 
+  @Test
+  public void concatStringLiteralAndVarchar() throws IOException, SqlParseException {
+    assertFullRoundTrip("select 'part_'||P_NAME from PART");
+  }
```
vbarua, line 71: When I look at this test I'm actually somewhat suprised that this doesn't throw an exception of some kind in the Substrait layer.

The POJO this produces has the following scalar function invocation
```
ScalarFunctionInvocation{
    declaration=concat:vchar,
    arguments=[
        FixedCharLiteral{nullable=false, value=part_},
        FieldReference{
            segments=[StructField{offset=1}],
            type=VarChar{nullable=true, length=55}}
    ]
    , options=[],
    outputType=VarChar{nullable=true, length=60}
}
```
As I understand it, this is an invalid invocation of `concat:vchar` from [functions_string.yaml](https://github.com/substrait-io/substrait/blob/10d53ca376cde9342a74d3480fcd2e77109d65cd/extensions/functions_string.yaml#L12-L21) 
because all arguments _must_ be varchars. Substrait, intentionally, doesn't have any form of implicit casting to allow calls to this function that mix string types. 

While SQL may allow us to `concat` different string types together, if we want to encode this into valid Substrait we'll need to upcast the arguments to the same type.

bestbeforetoday: Thank you for the feedback. The code changed is used by function matching code for cases where the types don't exactly match. Since the change didn't break any tests I (wrongly!) assumed it was successfully fixing the problem. I will dig more into the matching to find out if any type coercion should be happening and, depending on the result, either fix it or implement it.

bestbeforetoday:
@vbarua Do you have any guidance on the best testing approach? The existing and added tests did not catch the not-quite-correct change I made previously, and it would be good for those tests to give better confidence that the conversion is working correctly.

bestbeforetoday:
@vbarua I need some help untangling casting of arguments to the types required by a Substrait function. Given there is no exact match for the function signature, I can find a function to which I think I can coerce the supplied call. To do the coercion for operands that don't match the argument type required by the function, I need to create a cast. This requires the target `Type`. The `Function` can only give me the `ParameterizedType` of the arguments. Confusingly, `ParameterizedType` is a _superclass_ of `Type`, and the argument "types" I get from the `Function` are not instances of `Type`.

How do I get the `Type` values for a given `Function` arguments?

vbarua:
> Do you have any guidance on the best testing approach? The existing and added tests did not catch the not-quite-correct change I made previously, and it would be good for those tests to give better confidence that the conversion is working correctly.

I don't have a specific approach in mind. It will likely requiere a combination of things. One thing that would be good to investigate would be adding further validation and/or validators for plans generally to catch issues like that.

> How do I get the Type values for a given Function arguments?

Unfortunately I'm not too familiar with that machinary. Someone in the Slack might know more.

Additionally, I wanted to add that adding cast is one way to solve this problem. There might be others, but it's what came to my mind first at least.

bestbeforetoday:
@vbarua In the current (draft) iteration of the code I did some horrible, horrible ParameterizedType to Type mapping using visitors. This allows casts to be constructed for different string types, which at first glance look plausible. For example, this SQL:

```sql
select 'store' || s_store_name from store
```

generates Substrait including this (redacted) mapping:

```json
{
  "extensionUris": [{
    "extensionUriAnchor": 1,
    "uri": "/functions_string.yaml"
  }],
  "extensions": [{
    "extensionFunction": {
      "extensionUriReference": 1,
      "name": "concat:vchar"
    }
  }],
  "relations": [{
    "root": {
      "input": {
        "project": {
          [...]
          "expressions": [{
            "scalarFunction": {
              "outputType": {
                "varchar": {
                  "length": 55,
                  "nullability": "NULLABILITY_REQUIRED"
                }
              },
              "arguments": [{
                "value": {
                  "cast": {
                    "type": {
                      "varchar": {
                        "length": 5,
                        "nullability": "NULLABILITY_REQUIRED"
                      }
                    },
                    "input": {
                      "literal": {
                        "fixedChar": "store"
                      }
                    },
                    "failureBehavior": "FAILURE_BEHAVIOR_THROW_EXCEPTION"
                  }
                }
              }, {
                "value": {
                  "selection": {
                    "directReference": {
                      "structField": {
                        "field": 5
                      }
                    },
                    "rootReference": {
                    }
                  }
                }
              }]
            }
          }]
        }
      },
      "names": ["EXPR$0"]
    }
  }]
}
```

Unfortunately not all cases round-trip, and it breaks a few other test cases. If this looks to you like it is heading the right direction, I can try to get the failing test cases working again. What do you think?

vbarua:
That does seem reasonable to me. One other thing to note is that `concat:vchar` is defined with a length param `L1`:
```
    impls:
      - args:
          - value: "varchar<L1>"
            name: "input"
        variadic:
          min: 1
        options:
          null_handling:
            values: [ IGNORE_NULLS, ACCEPT_NULLS ]
        return: "varchar<L1>"
```

As I understand the spec currently, if the output type of your `concat:vchar` declaration is `varchar<55>`, then all inputs must also be `varchar<55>`.

> I did some horrible, horrible ParameterizedType to Type mapping using visitors.

Another way to approach this might be to upcast the arguments in _Calcite_ before mapping into Substrait, though I don't actually know if there is anything for this within Calcite.

isthmus/src/test/java/io/substrait/isthmus/ExpressionConvertabilityTest.java
vbarua: The type system change you added definitely helped.
```
ScalarFunctionInvocation{
    declaration=concat:vchar,
    arguments=[
        Cast{type=VarChar{nullable=true, length=55}, input=FixedCharLiteral{nullable=false, value=part_}, failureBehavior=THROW_EXCEPTION},
        FieldReference{segments=[StructField{offset=1}], type=VarChar{nullable=true, length=55}}
    ],
    options=[],
    outputType=VarChar{nullable=true, length=60}
}
```

Both inputs are now varchars 🎉 

The output schema doesn't quite match Substrait still, which would expect `VarChar<55>` instead of `VarChar<60>`.

That's happening because the Calcite definition of [CONCAT](https://github.com/apache/calcite/blob/61e6dd11706aed0ee617a251643022b331c7d466/core/src/main/java/org/apache/calcite/sql/fun/SqlStdOperatorTable.java#L260-L281)
uses [ReturnTypes.DYADIC_STRING_SUM_PRECISION_NULLABLE](https://github.com/apache/calcite/blob/61e6dd11706aed0ee617a251643022b331c7d466/core/src/main/java/org/apache/calcite/sql/type/ReturnTypes.java#L1342-L1347) for its return type inference which ends up doing the following
> Type-inference strategy for concatenating two string arguments. The result type of a call is:
> * the same type as the input types but with the combined length of the two first types
> * if types are of char type the type with the highest coercibility will be used
> * result is varying if either input is; otherwise fixed
> 
> Pre-requisites:
> * input types must be of the same string type
> * types must be comparable without casting

I think we can tackle that as part of https://github.com/substrait-io/substrait-java/issues/379

vbarua (commented):
I just noticed that this was ready for re-review. Took a preliminary look at everything except the FunctionConverter, for which I need fresh eyes.

isthmus/src/main/java/io/substrait/isthmus/expression/FunctionConverter.java
```diff
@@ -479,56 +454,30 @@ public interface GenericCall {
...
-        for (SignatureMatcher m : matchers) {
-          var t = m.tryMatch(types, outputType);
-          if (t.isPresent()) {
-            return t;
-          }
-        }
-        return Optional.empty();
-      };
-    };
-  }
```
vbarua, line 519: REVIEW NOTE: this code is unused

isthmus/src/main/java/io/substrait/isthmus/expression/FunctionConverter.java
```diff
@@ -479,56 +454,30 @@ public interface GenericCall {
...
-
-  private static boolean isMatch(Type inputType, ParameterizedType type) {
-    if (type.isWildcard()) {
-      return true;
-    }
-    return inputType.accept(new IgnoreNullableAndParameters(type));
-  }
```
vbarua, line 526: REVIEW NOTE: this code is redundant with `isMatch(ParameterizedType actualType, ParameterizedType targetType) ` below

vbarua (commented):
Overall this looks reasonable to me.

@bestbeforetoday could you confirm for me that there are no functionality changes in `FunctionConverter.java`? From what I reviewed, it looks like there are only cleanups and code simplifications.

Additionally, could you move the tests in ExpressionConvertabilityTest.java as per https://github.com/substrait-io/substrait-java/pull/338/files#r2038494522 and fix the conflicts.

Once those things are done, I'm happy to merge this.

vbarua (changes requested):
Marking as waiting for changes.

isthmus/src/main/java/io/substrait/isthmus/expression/FunctionConverter.java
```diff
@@ -411,53 +404,35 @@ private Optional<T> matchByLeastRestrictive(
...
-        declaration.validateOutputType(coercedArgs, outputType);
-        return Optional.of(
-            generateBinding(
-                call,
-                out.get(),
-                coercedArgs.stream().map(FunctionArg.class::cast).collect(Collectors.toList()),
-                outputType));
-      }
-      return Optional.empty();
+      var out = singularInputType.orElseThrow().tryMatch(type, outputType);
```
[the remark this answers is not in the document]
bestbeforetoday: Correct. No functional change by using `orElseThrow()`. The Java API documentation for `get()` says that `orElseThrow()` is preferred. I guess just for consistency with the other `orElse...` methods.

isthmus/src/test/java/io/substrait/isthmus/ExpressionConvertabilityTest.java
```diff
@@ -65,6 +65,21 @@ public void inPredicate() throws IOException, SqlParseException {
...
+
+  @Test
+  public void concatCharAndVarchar() throws Exception {
+    assertProtoPlanRoundrip("select P_BRAND||P_NAME from PART");
+  }
+
+  @Test
+  public void concatStringLiteralAndChar() throws Exception {
+    assertProtoPlanRoundrip("select 'brand_'||P_BRAND from PART");
+  }
```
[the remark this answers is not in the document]
bestbeforetoday: Done.

vbarua (approved):
Changes look good, I will merge this later today.

vbarua:
@bestbeforetoday thanks for working on this.

In the future, it would be helpful to call out that changes are mostly refactors and/or split them into their own PRs. The FunctionConverter code is pretty hairy, even for me, so I kept trying to find a time when I was mentally fresh to review them, but it turned out they were easy to reviews and I could have done it anytime really.

bestbeforetoday:
@vbarua Are you able to merge this change so I can safely start additional changes to improve TPC-DS compatibility on top of it?

vbarua:
@bestbeforetoday thanks for the reminder. Merged.

PR #340, opened 2025-03-08
feat(isthmus): convert Substrait EmptyScan to Calcite Values relation

vbarua:


vbarua:
A Calcite Values relation is converted to a Substrait EmptyScan when it contains no tuples.
https://github.com/substrait-io/substrait-java/blob/780a0bb22a6cc625cd118eeb59a5f013bc468ba0/isthmus/src/main/java/io/substrait/isthmus/SubstraitRelVisitor.java#L120-L125

This PR adds the missing mapping from Substrait EmptyScan to Calcite Values relation.

PR #342, opened 2025-03-12
feat(spark): support Struct type and literal, and include names for struct fields

Blizzara:
BREAKING CHANGE: Plan root's "names" field now includes nested names in addition to the top field names. The new behavior is how Substrait spec defines the names, but this change breaks compatibility of existing plans containing struct fields since the names list will not match.


spark/src/main/scala/io/substrait/spark/logical/ToSubstraitRel.scala
```diff
@@ -460,7 +459,11 @@ class ToSubstraitRel extends AbstractLogicalPlanVisitor with Logging {
     ImmutablePlan.builder
       .roots(
         Collections.singletonList(
-          ImmutableRoot.builder().input(rel).addAllNames(p.output.map(_.name).asJava).build()
+          ImmutableRoot
+            .builder()
+            .input(rel)
+            .addAllNames(ToSubstraitType.toNamedStruct(p.schema).names())
```
Blizzara, line 465: This changes the produced plans to include DFS names in the root

spark/src/main/scala/io/substrait/spark/logical/ToSubstraitRel.scala
```diff
@@ -342,7 +341,7 @@ class ToSubstraitRel extends AbstractLogicalPlanVisitor with Logging {
   }
 
   private def buildNamedScan(schema: StructType, tableNames: List[String]): relation.NamedScan = {
-    val namedStruct = toNamedStruct(schema)
+    val namedStruct = ToSubstraitType.toNamedStruct(schema)
```
Blizzara, line 344: Just being more explicit here

spark/src/main/scala/io/substrait/spark/ToSubstraitType.scala
```diff
@@ -122,52 +163,35 @@ class ToSubstraitType {
...
-    schema.fields.foreach(
-      field => {
-        names.add(field.name)
-        children.add(apply(field.dataType, field.nullable))
-      })
-    val struct = creator.struct(children)
-    NamedStruct.of(names, struct)
-  }
 
-  def toStructType(namedStruct: NamedStruct): StructType = {
```
Blizzara, line 150: these were previously inside `ToSubstraitType`, which is confusing given they return Spark types. Now they're moved into the ToSparkType object above, but also rewritten to use the newly-available struct type code.

(We should also split this file, but I'd rather do that as FLUP to keep the diff clean)

spark/src/main/scala/io/substrait/spark/ToSubstraitType.scala
```diff
@@ -76,13 +83,47 @@ private class ToSparkType
...
 }
-class ToSubstraitType {
 
+object ToSparkType {
   def convert(typeExpression: TypeExpression): DataType = {
-    typeExpression.accept(new ToSparkType)
+    typeExpression.accept(new ToSparkType(Seq.empty))
+  }
+
+  def toStructType(namedStruct: NamedStruct): StructType = {
```
Blizzara, line 114: moved from below, see comment below

spark/src/main/scala/io/substrait/spark/ToSubstraitType.scala
```diff
@@ -122,52 +163,35 @@ class ToSubstraitType {
...
-      fields = namedStruct
-        .struct()
-        .fields()
-        .asScala
-        .map(t => (t, convert(t)))
-        .zip(namedStruct.names().asScala)
-        .map { case ((t, d), name) => StructField(name, d, t.nullable()) }
-    )
+  def toNamedStruct(schema: StructType): NamedStruct = {
+    val dfsNames = JavaConverters.seqAsJavaList(fieldNamesDfs(schema))
```
Blizzara, line 179: this changes anything using `toNamedStruct` to produce the full list of names rather than only top-level

spark/src/main/scala/io/substrait/utils/Util.scala
```diff
@@ -64,9 +64,9 @@ object Util {
     @tailrec
     def seqToOptionHelper(s: Seq[Option[T]], accum: Seq[T] = Seq[T]()): Option[Seq[T]] = {
       s match {
-        case Some(head) :: Nil =>
+        case Seq(Some(head)) =>
```
Blizzara, line 67: I'm not sure why but the earlier code just wasn't working

Blizzara:
@andrew-coleman @vbarua looking for your review for this :) Note that this is a breaking change since the currently produced plans aren't valid Substrait - they lack the inner names, this adds those, but that changes the meaning of the names-fields.

For Andrew: I saw you had added a bit of struct handling in your PR, this is compatible with that but adds the rest of the handling as well.

spark/src/main/scala/io/substrait/spark/ToSubstraitType.scala
```diff
@@ -76,13 +83,47 @@ private class ToSparkType
...
+  override def visit(expr: Type.Struct): DataType =
+    StructType(
+      expr.fields.asScala.zipWithIndex
+        .map {
+          case (f, i) =>
+            val name = if (dfsNames.nonEmpty) {
+              require(nameIdx < dfsNames.size)
+              val n = dfsNames(nameIdx)
+              nameIdx += 1
+              n
```
andrew-coleman, line 96:
```suggestion
              dfsNames(i)
```
Won't `nameIdx` always be the same `i`? If so, no need to increment a separate counter.
Incrementing an object-scoped counter assumes that the `.map` is behaving like a `foreach` - i.e. guaranteed order of execution, but I think (in theory, at least) the map could execute its functions concurrently (not sure it does, in practice), just assembling the results back into the original sequence.

andrew-coleman: Looking at this further, I guess it's doing this because it's doing a depth-first traversal of a potentially nested struct.

andrew-coleman (approved):

spark/src/main/scala/io/substrait/spark/ToSubstraitType.scala
Blizzara: Yup, exactly

spark/src/test/scala/io/substrait/spark/SubstraitPlanTestBase.scala
```diff
@@ -85,30 +87,39 @@ trait SubstraitPlanTestBase { self: SharedSparkSession =>
...
+    // conversion can be reversed and repeated leading to same plan:
+    val substraitPlan = new ToSubstraitRel().convert(sparkPlan)
 
-    // Serialize to protobuf byte array
-    val extensionCollector = new ExtensionCollector
-    val bytes = new RelProtoConverter(extensionCollector).toProto(substraitPlan).toByteArray
+    // Serialize to protobuf byte array and read it back to ensure the proto-roundtrip
+    val bytes = new PlanProtoConverter().toProto(substraitPlan).toByteArray
+    val substraitPlan2 = new ProtoPlanConverter(SparkExtension.COLLECTION).from(io.substrait.proto.Plan.parseFrom(bytes)))
+    // TODO: assert substraitPlan2 == substraitPlan
```
Blizzara, line 99: this will be added in https://github.com/substrait-io/substrait-java/pull/315

PR #344, opened 2025-03-14
feat(spark): support SELECT without FROM clause

andrew-coleman:
Add support for the Spark OneRowRelation, which is the leaf node that represents selects without a from clause.

A follow-on PR will use this to aid testing.


spark/src/main/scala/io/substrait/spark/logical/ToSubstraitRel.scala
```diff
@@ -501,6 +503,12 @@ class ToSubstraitRel extends AbstractLogicalPlanVisitor with Logging {
               s"******* Unable to convert the plan to a substrait relation: " +
                 s"${logicalRelation.relation.toString}")
         }
+      case _: OneRowRelation =>
+        relation.EmptyScan
```
Blizzara, line 507: EmptyScan doesn't produce any rows, so it's not an exact match for OneRowRelation, no? Would it make sense to convert it to VirtualTableScan with a single (empty) row in it instead?

PR #345, opened 2025-03-17
chore: update to substrait 0.69.0

andrew-coleman:
Required to pick up the decimal round() function definition


EpsilonPrime (approved):

spark/src/main/scala/io/substrait/spark/ToSubstraitType.scala
```diff
@@ -52,6 +52,10 @@ private class ToSparkType
 
   override def visit(expr: Type.Bool): DataType = BooleanType
 
+  override def visit(expr: Type.PrecisionTime): DataType = {
+    Util.assertMicroseconds(expr.precision())
+    TimestampType
+  }
```
vbarua, line 58: Two questions:
1.  From what I'm seeing, Spark doesn't have a direct representation for Time types. Do the semantics of TimestampType match the Substrait expecations for Time? There is no existing mapping for the old `Type.Time` either.
2. Assuming 1, does the PrecisionTime type have a timezone? Should this be `TimestampNTZType` instead.

vbarua (commented):
Overall looks good. I did have one question about mapping PrecisionTime into Spark.

spark/src/main/scala/io/substrait/spark/ToSubstraitType.scala
andrew-coleman: I don't know TBH... I've taken out the mapping entirely so that it will default to throwing an exception (like the old `Type.Time`).
This can be implemented in the future, once it's better understood.  I only intended to do a minimal implementation of this new type in order to fix the build error caused by pulling in the latest substrait library :)

vbarua: > I've taken out the mapping entirely

That works for me, especially as it wasn't already handled on the deprecated path.

vbarua (approved):

PR #346, opened 2025-03-19
feat(spark): support merged structures

andrew-coleman:
For a certain types of queries, the spark optimiser will rewrite the plan to create structs of merged sub-queries.
These plans don’t convert directly to substrait.  This commit provides a matcher for these structures and converts them back to their original form before translating them to Substrait.

### DIFF
--- a/spark/src/main/scala/io/substrait/spark/expression/ToSubstraitExpression.scala
+++ b/spark/src/main/scala/io/substrait/spark/expression/ToSubstraitExpression.scala
@@ -19,6 +19,7 @@ package io.substrait.spark.expression
 import io.substrait.spark.{HasOutputStack, ToSubstraitType}
 
 import org.apache.spark.sql.catalyst.expressions._
+import org.apache.spark.sql.catalyst.plans.logical.{Aggregate, Project}
 import org.apache.spark.substrait.SparkTypeUtil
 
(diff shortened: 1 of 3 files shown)


Blizzara:
> These plans don’t convert directly to substrait.

what is the part that doesn't convert? is it just that we haven't supported Structs yet, or is there something else too?

andrew-coleman:
The Spark functions `CreateNamedStruct` and `GetStructField` don't have equivalents in Substrait.
The `CreateNamedStruct` function represents the SQL function `named_struct`, but trying to model that in Substrait hits the problem that the return type parameters depends on the function arguments.  So I figured best avoid this for now :).

If we can get these PRs in, then I've just got one more to do (which I'm currently progressing) to get 100% pass rate on the TPC-DS suite :)

Blizzara:
> The Spark functions `CreateNamedStruct` and `GetStructField` don't have equivalents in Substrait. The `CreateNamedStruct` function represents the SQL function `named_struct`, but trying to model that in Substrait hits the problem that the return type parameters depends on the function arguments. So I figured best avoid this for now :).

I think it works by just dropping the names from it 😄 then it creates a struct, and the names get handled by the naming logic.

Anyways, this PR is an improvement over current state and doesn't prevent us from doing that in future, so 👍 

> If we can get these PRs in, then I've just got one more to do (which I'm currently progressing) to get 100% pass rate on the TPC-DS suite :)

Nice! 🚀

PR #347, opened 2025-03-19
feat(spark): support for round, floor and ceil functions

andrew-coleman:
Support the decimal versions of `round`, `ceil` and `floor`.


Blizzara (approved):
LGTM, thanks!

vbarua (approved):
Thanks for adding this ✨

PR #348, opened 2025-03-20
fix(spark): enable aliased expressions to round-trip

andrew-coleman:
A number of the TPC-DS tests were failing because the query contains multiple aliases to the same expression, causing a potential mismatch in the reference index.  Although the plans were equivalent, the substrait POJO comparison failed. This commit uses the `hint` field of the Rel message to store the alias names, and restore them back to the Spark plan to match the original.

### DIFF
--- a/spark/src/main/scala/io/substrait/spark/logical/ToLogicalPlan.scala
+++ b/spark/src/main/scala/io/substrait/spark/logical/ToLogicalPlan.scala
@@ -16,7 +16,7 @@
  */
 package io.substrait.spark.logical
 
-import io.substrait.spark.{DefaultRelVisitor, SparkExtension, ToSparkType}
+import io.substrait.spark.{DefaultRelVisitor, SparkExtension, ToSparkType, ToSubstraitType}
 import io.substrait.spark.expression._
 
 import org.apache.spark.sql.SparkSession
@@ -33,7 +33,7 @@ import org.apache.spark.sql.execution.datasources.csv.CSVFileFormat
 import org.apache.spark.sql.execution.datasources.orc.OrcFileFormat
 import org.apache.spark.sql.execution.datasources.parquet.ParquetFileFormat
 import org.apache.spark.sql.internal.SQLConf
-import org.apache.spark.sql.types.{ArrayType, DataType, DataTypes, IntegerType, MapType, StructField, StructType}
+import org.apache.spark.sql.types.{DataType, IntegerType, StructField, StructType}
 
 import io.substrait.`type`.{NamedStruct, StringTypeVisitor, Type}
 import io.substrait.{expression => exp}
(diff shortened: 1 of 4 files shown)


spark/src/main/scala/io/substrait/spark/logical/ToLogicalPlan.scala
```diff
@@ -249,12 +249,21 @@ class ToLogicalPlan(spark: SparkSession) extends DefaultRelVisitor[LogicalPlan]
       case a: Aggregate => (a.aggregateExpressions, false)
       case other => (other.output, true)
     }
+    val names = if (project.getHint.isPresent) {
+      project.getHint.get().getOutputNames.asScala
```
Blizzara, line 253: is the `hint.outputNames` supposed to be just the column names, or also the possible inner struct field names in DFS form (like the other Substrait "names" fields)?

andrew-coleman: It's just the column names, used to distinguish different aliases to the same underlying expression.  Otherwise the Spark optimiser de-duplicates them causing the round-trip equality check to fail (even though they are equivalent).

Blizzara: The [docstring](https://github.com/substrait-io/substrait/blob/4a472447392d9beba0401987b4d845079241c30d/proto/substrait/algebra.proto#L44) there says
> // Assigns alternative output field names for any relation.  Equivalent to the names field
> // in RelRoot but applies to the output of the relation this RelCommon is attached to.

There's multiple ways to infer that, but "equivalent to the names field in RelRoot" would indicate it should be the DFS listing, including the inner names, since that's what the "names" field in RelRoot is?

(Using it as the DFS would help with the named_struct issue too ;))

andrew-coleman: Hi @Blizzara, I'm not really sure what you want me to do here.  Do you have a test case in mind that demonstrates the problem?  Thanks :)

Blizzara: We should have the "names" hint contain all of the names, including the inner names in the schema, so that it matches the "names" field on the RelRoot. You can see an example of how I'm planning on doing it for RelRoot here: https://github.com/substrait-io/substrait-java/pull/342 (I should merge that but it fails for the the test added in #346  as it needs a fix from https://github.com/substrait-io/substrait-java/pull/315 - it's actually a good catch from that test.)

spark/src/main/scala/io/substrait/spark/logical/ToSubstraitRel.scala
```diff
@@ -340,9 +342,11 @@ class ToSubstraitRel extends AbstractLogicalPlanVisitor with Logging {
       p.child.output.count(o => !existenceJoins.contains(o.exprId.id)),
       expressions.size
     )
+
     relation.Project.builder
       .remap(remap)
       .expressions(expressions.asJava)
+      .hint(Hint.builder.addAllOutputNames(ToSubstraitType.toNamedStruct(p.schema).names()).build())
```
Blizzara, line 349: let's update visitExpand below to also use toNamedStruct().names() ?

spark/src/main/scala/io/substrait/spark/logical/ToLogicalPlan.scala
```diff
@@ -250,12 +250,21 @@ class ToLogicalPlan(spark: SparkSession) extends DefaultRelVisitor[LogicalPlan]
...
+      List.empty
+    }
 
     withOutput(output) {
-      val projectList =
+      val projectExprs =
         project.getExpressions.asScala
           .map(expr => expr.accept(expressionConverter))
-          .map(toNamedExpression)
+      val projectList = if (names.size == projectExprs.size) {
```
Blizzara, line 263: ah so this just skips using the names if there are inner structs? I guess that works, for a bit more complete solution I think you could use ToSparkType.toStructType to construct the expected schema (like [here](https://github.com/substrait-io/substrait-java/blob/d73723274f46ad6329fa97874788d3255ff3cd7a/spark/src/main/scala/io/substrait/spark/logical/ToLogicalPlan.scala#L424)) and then pick the column names from there, and for even more complete do that + reuse the `renameAndCastExprs`.

Blizzara (approved):
Thanks!

andrew-coleman:
Hi @Blizzara, I notice you've approved but not merged.  Is there anything more you need me to do here?
I've got more in the pipeline, but it needs this to go in first :). Many thanks!

EpsilonPrime (approved):

PR #349, opened 2025-03-20
chore: uncomment working TPC-DS tests

bestbeforetoday:
TPC-DS test cases 17, 24, 39 and 78 previously failed. Earlier commits mean that they now run successfully. This change includes these tests in the test suite to ensure that regressions are not introduced that cause them to start failing again.


Blizzara (approved):
Thanks!

PR #350, opened 2025-03-21
fix: use SubstraitTypeSystem when parsing SQL

vbarua:
BREAKING CHANGE: SQL parsing no longer uses the default Calcite type system

### DIFF
--- a/isthmus/src/main/java/io/substrait/isthmus/SubstraitTypeSystem.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/SubstraitTypeSystem.java
@@ -10,9 +10,6 @@ import org.apache.calcite.sql.parser.SqlParserPos;
 import org.apache.calcite.sql.type.SqlTypeName;
 
 public class SubstraitTypeSystem extends RelDataTypeSystemImpl {
-  static final org.slf4j.Logger logger =
-      org.slf4j.LoggerFactory.getLogger(SubstraitTypeSystem.class);
-
   public static final RelDataTypeSystem TYPE_SYSTEM = new SubstraitTypeSystem();
 
   private SubstraitTypeSystem() {}
--- a/isthmus/src/test/java/io/substrait/isthmus/PlanTestBase.java
+++ b/isthmus/src/test/java/io/substrait/isthmus/PlanTestBase.java
@@ -101,26 +101,29 @@ public class PlanTestBase {
   protected List<RelNode> assertSqlSubstraitRelRoundTrip(String query, List<String> creates)
       throws Exception {
     // sql <--> substrait round trip test.
-    // Assert (sql -> substrait) and (sql -> substrait -> calcite rel -> substrait) are same.
-    // Return list of sql -> substrait rel -> Calcite rel.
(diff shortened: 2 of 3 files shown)


isthmus/src/main/java/io/substrait/isthmus/SubstraitTypeSystem.java
```diff
@@ -10,9 +10,6 @@
 import org.apache.calcite.sql.type.SqlTypeName;
 
 public class SubstraitTypeSystem extends RelDataTypeSystemImpl {
-  static final org.slf4j.Logger logger =
-      org.slf4j.LoggerFactory.getLogger(SubstraitTypeSystem.class);
```
vbarua, line 14: cleanup: opportunistic removal of unused logger

isthmus/src/test/java/io/substrait/isthmus/PlanTestBase.java
```diff
@@ -101,26 +101,29 @@ protected List<RelNode> assertSqlSubstraitRelRoundTrip(String query) throws Exce
...
     SqlToSubstrait s = new SqlToSubstrait();
+
+    // 1. SQL -> Calcite RelRoot
     for (RelRoot relRoot : s.sqlToRelNode(query, creates)) {
-      Rel pojoRel = SubstraitRelVisitor.convert(relRoot, EXTENSION_COLLECTION);
+      // 2. Calcite RelRoot  -> Substrait Rel
+      Rel pojo1 = SubstraitRelVisitor.convert(relRoot, EXTENSION_COLLECTION);
 
-      // 2. substrait rel -> Calcite Rel
-      RelNode relnodeRoot = new SubstraitToSql().substraitRelToCalciteRel(pojoRel, creates);
```
vbarua, line 114: The SqlToSubstrait conversion uses the schema as encoded by the `creates`, which is a list of SQL CREATE statements. These are also used when parsing the original query above.

Using the SubstraitToCalcite convert method, whichs generates a Calcite schema from the given Substrait plan, works as an extra check to make sure that the schema was encoded correctly into the plan, instead of just assuming the same schema.

Switching to SubstraitToCalcite here also exposed the bug that I'm fixing, because SubstraitToCalcite uses the SubstraitTypeSystem, while SqlToCalcite uses the default Calcite type system which has a low maximum precision.

vbarua:
Was doing some work related to SQL parsing when I noticed some weirdness around decimals. Specifically, some tests failed  because the precision between expected and actual results didn't line up. The number 19 kept showing up, and once I figured out that was the default Calcite precision I realized that we weren't using the SubstraitTypeSystem when parsing SQL which has different limits.

isthmus/src/test/java/io/substrait/isthmus/PlanTestBase.java
vbarua: It would be, but at the moment the `SubraitRelVisitor#convert(RelRoot root, ...)` method currently generates a Substrait `Rel` from the Calcite `RelRoot`.

Updating that conversion to generate a `Plan.Root` would make sense, but I would consider it beyond the scope of this.

PR #351, opened 2025-03-21
feat: Add `unquotedcasing` option to Isthmus CLI

geoffreyclaude:
Adds a command line argument to Isthmus CLI to allow specifying Calcite's "unquoted casing" policy.

## Why?

I want to add Substrait based sqllogictests to [DataFusion](https://github.com/apache/datafusion), converting the SQL to Substrait through Isthmus.
The main blocker right now is that Isthmus generates Substrait plans with uppercase tables and columns, whereas DataFusion normalizes to lowercase by default.

### DIFF
--- a/isthmus-cli/src/main/java/io/substrait/isthmus/cli/IsthmusEntryPoint.java
+++ b/isthmus-cli/src/main/java/io/substrait/isthmus/cli/IsthmusEntryPoint.java
@@ -19,6 +19,7 @@ import io.substrait.proto.Plan;
 import java.io.IOException;
 import java.util.List;
 import java.util.concurrent.Callable;
+import org.apache.calcite.avatica.util.Casing;
 import org.apache.calcite.sql.validate.SqlConformanceEnum;
 import picocli.CommandLine;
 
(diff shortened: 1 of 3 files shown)


geoffreyclaude:
@vbarua @bvolpato would you mind reviewing?

bvolpato:
> @vbarua @bvolpato would you mind reviewing?

Will do, thanks for the nudge.

FYI it's missing spotless, this should do the trick

```
./gradlew spotlessApply
```

bvolpato:
Another one on the commit message

```
⧗   input: feat: Add `unquotedcasing` option to Isthmus CLI
✖   subject must not be sentence-case, start-case, pascal-case, upper-case [subject-case]
```

bvolpato (approved):

ingomueller-net:
Is it possible that this change breaks the Python bindings? The Isthmus tests in `consumer-testing` break from the previous to this commit with the following error:

```
cls = <class 'jpype._jclass.JClass'>, jc = 'io.substrait.isthmus.SqlToSubstrait', loader = None, initialize = True
...
            if len(params) > 0 and len(params) != len(ret.class_.getTypeParameters()):
                raise TypeError(
                    "Java generic class '%s' length mismatch" % (ret.__name__))
            return ret
    
        # Pass to class factory to create the type
>       return _jpype._getClass(jc)
E       TypeError: Class io.substrait.isthmus.SqlToSubstrait is not found

.../lib/python3.12/site-packages/jpype/_jclass.py:99: TypeError
```

PR #352, opened 2025-03-21
refactor: remove IOException from SimpleExtension.loadDefaults()

vbarua:


vbarua:
`SimpleExtension.loadDefaults()` doesn't actually throw an IOException. This PR removes the throw on it, and then propagates the removal through the rest of the codebase.

isthmus/src/test/java/io/substrait/isthmus/PlanTestBase.java
```diff
@@ -87,7 +78,7 @@ protected Plan assertProtoPlanRoundrip(String query, SqlToSubstrait s, List<Stri
     return plan;
   }
 
-  protected void assertPlanRoundrip(Plan plan) throws IOException, SqlParseException {
+  protected void assertPlanRoundtrip(Plan plan) {
```
vbarua, line 81: Opportunistic typo rename `Roundrip` -> `Roundtrip`. This our internal test methods so it should not affect users.

core/src/main/java/io/substrait/extension/SimpleExtension.java
```diff
@@ -692,7 +692,7 @@ public ExtensionCollection merge(ExtensionCollection extensionCollection) {
     }
   }
 
-  public static ExtensionCollection loadDefaults() throws IOException {
+  public static ExtensionCollection loadDefaults() {
```
vbarua, line 695: This doesn't actually throw an `IOException`. Removing this throws lets us cleanup a bunch of other code.

Blizzara: I was curious what it does instead given it does load some files, but it catches and throws a RuntimeException: https://github.com/substrait-io/substrait-java/pull/352/files#diff-d507a961a7a2ce9617ab3f8fdefcc4f647ef2207956133522b402ec6a79cc26cR728 👍

Blizzara (approved):

PR #353, opened 2025-03-24
feat(isthmus): add support for scalar subqueries

nielspardon:
Currently, the Substrait to Calcite mappings to not support scalar subqueries like the follow SQL example:

```sql
select 
  n_name 
from 
  tpch.tiny.nation 
where 
  n_regionkey = (
    select r_regionkey from tpch.tiny.region where r_name = 'EUROPE')
```

This PR adds support for mapping scalar subqueries from Substrait into Calcite and adds a test case showing that it works now.

Additionally:
- it changes `RelCreator` to be created with a list of CREATE SQL statements to initialize the schema of the `RelBuilder`
- the previous change requires `RelCreator` to extend `SqlConverterBase` so that we can reuse the `registerCreateTables()` method
- in `SubstraitExpressionConverterTest` it gets the `ExpressionRexConverter` from `SubstraitRelNodeConverter` in order to make sure the `ExpressionRexConverter` is properly set up to parse sub queries using `SubstraitRelNodeConverter`

### DIFF
--- a/isthmus/src/main/java/io/substrait/isthmus/SubstraitToCalcite.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/SubstraitToCalcite.java
@@ -2,14 +2,15 @@ package io.substrait.isthmus;
 
 import io.substrait.extension.SimpleExtension;
 import io.substrait.plan.Plan;
-import io.substrait.relation.AbstractRelVisitor;
 import io.substrait.relation.NamedScan;
 import io.substrait.relation.Rel;
+import io.substrait.relation.RelCopyOnWriteVisitor;
 import io.substrait.type.NamedStruct;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import java.util.function.Function;
 import org.apache.calcite.jdbc.CalciteSchema;
 import org.apache.calcite.jdbc.LookupCalciteSchema;
@@ -179,10 +180,11 @@ public class SubstraitToCalcite {
     }
   }
 
-  private static class NamedStructGatherer extends AbstractRelVisitor<Void, RuntimeException> {
+  private static class NamedStructGatherer extends RelCopyOnWriteVisitor<RuntimeException> {
     Map<List<String>, NamedStruct> tableMap;
 
     private NamedStructGatherer() {
+      super();
       this.tableMap = new HashMap<>();
     }
 
@@ -193,16 +195,13 @@ public class SubstraitToCalcite {
     }
 
     @Override
-    public Void visit(NamedScan namedScan) {
+    public Optional<Rel> visit(NamedScan namedScan) {
+      Optional<Rel> result = super.visit(namedScan);
+
       List<String> tableName = namedScan.getNames();
       tableMap.put(tableName, namedScan.getInitialSchema());
-      return null;
-    }
 
-    @Override
-    public Void visitFallback(Rel rel) {
-      for (Rel input : rel.getInputs()) input.accept(this);
-      return null;
+      return result;
     }
   }
 }
--- a/isthmus/src/test/java/io/substrait/isthmus/SubstraitExpressionConverterTest.java
+++ b/isthmus/src/test/java/io/substrait/isthmus/SubstraitExpressionConverterTest.java
@@ -1,16 +1,22 @@
 package io.substrait.isthmus;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertInstanceOf;
 
 import io.substrait.dsl.SubstraitBuilder;
+import io.substrait.expression.Expression;
 import io.substrait.isthmus.expression.ExpressionRexConverter;
-import io.substrait.isthmus.expression.ScalarFunctionConverter;
-import io.substrait.isthmus.expression.WindowFunctionConverter;
+import io.substrait.relation.Project;
 import io.substrait.relation.Rel;
+import io.substrait.relation.Rel.Remap;
 import io.substrait.type.Type;
 import io.substrait.type.TypeCreator;
 import java.util.List;
+import org.apache.calcite.rel.RelNode;
+import org.apache.calcite.rel.logical.LogicalProject;
 import org.apache.calcite.rel.type.RelDataType;
+import org.apache.calcite.rex.RexNode;
+import org.apache.calcite.sql.SqlKind;
 import org.junit.jupiter.api.Test;
 
 public class SubstraitExpressionConverterTest extends PlanTestBase {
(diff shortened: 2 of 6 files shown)


isthmus/src/test/java/io/substrait/isthmus/RelCreator.java
```diff
@@ -25,19 +26,37 @@
 import org.apache.calcite.sql2rel.SqlToRelConverter;
 import org.apache.calcite.sql2rel.StandardConvertletTable;
 import org.apache.calcite.tools.RelBuilder;
+import org.apache.calcite.util.Pair;
 
-public class RelCreator {
+public class RelCreator extends SqlConverterBase {
```
vbarua, line 31: I have a preference for not doing this if possible. I have some in progress work to (hopefully) simplify how this works overall, and also I think there's a better approach which I've mentioned in your test.

isthmus/src/test/java/io/substrait/isthmus/SubstraitExpressionConverterTest.java
```diff
@@ -43,6 +51,54 @@ public void switchExpression() {
...
+                        TypeCreator.NULLABLE.fixedChar(25),
+                        TypeCreator.NULLABLE.varChar(152)))));
+
+    Expression.ScalarSubquery expr =
+        Expression.ScalarSubquery.builder()
+            .type(TypeCreator.REQUIRED.I64)
+            .input(subQueryRel)
+            .build();
+
+    RexNode calciteExpr = expr.accept(converter);
```
vbarua, line 94: I think an easier way to test this would be to use a `SubstraitToCalcite` instance and invoking the `convert` method on it. You would need to wrap your `expr` itself in another Project relation, which could have as input a EmptyScan.

This does a couple of things:
1. It helps to verify that the code for collecting/generating the schema from a Substrait plan goes through ScalarSubqueries correctly.
2. It decouples this test from the any of the TPCH / SQL parsing code.

isthmus/src/test/java/io/substrait/isthmus/SubstraitExpressionConverterTest.java
```diff
@@ -43,6 +51,54 @@ public void switchExpression() {
...
+                      b.fieldReference(input, 1),
+                      Expression.StrLiteral.builder().nullable(false).value("EUROPE").build());
+                },
+                b.namedScan(
+                    List.of("REGION"),
+                    List.of("r_regionkey", "r_name", "r_comment"),
+                    List.of(
+                        TypeCreator.REQUIRED.I64,
+                        TypeCreator.NULLABLE.fixedChar(25),
+                        TypeCreator.NULLABLE.varChar(152)))));
```
vbarua, line 86: Could you use the `commonTable` in conjunction with the comment below.

isthmus/src/test/java/io/substrait/isthmus/SubstraitExpressionConverterTest.java
```diff
@@ -43,6 +51,54 @@ public void switchExpression() {
...
+     * from
+     *   region
+     * where
+     *   r_name = 'EUROPE'
+     */
+    Rel subQueryRel =
+        b.project(
+            input -> {
+              return List.of(b.fieldReference(input, 0));
+            },
```
vbarua, line 70: minor: you should be able to elide the `return` statement in a couple of places
```java
        b.project(
            input -> b.fieldReferences(input, 0),
            ...
```

vbarua (commented):
The conversion code looks good overall. I left some comments that will hopefully simplify the testing of this a little bit.

isthmus/src/test/java/io/substrait/isthmus/SubstraitExpressionConverterTest.java
nielspardon: I get that the TPC-H schema SQL parsing is a bit ugly right now and it would be good to find a better way. What I'm a bit worried about is that there are other places in the code which rely on the schema being fully described within the provided Substrait plan and there currently is no way to provide the schema from the outside in `SubstraitCalcite` which is another thing I would like to fix in a future PR. This causes problems when one tries to parse the Substrait plan generated by `SqlToSubstrait` in the `TpchQueryNoValidation` test with `SubstraitToCalcite`. It needs access to the same schema that `SqlToSubstrait` used in order to successfully parse the Substrait plan.

I will see how to incorporate your suggestion for the test cases though.

isthmus/src/test/java/io/substrait/isthmus/RelCreator.java
nielspardon: I reverted that change.

isthmus/src/test/java/io/substrait/isthmus/SubstraitExpressionConverterTest.java
nielspardon: I'm using `commonTable` now.

isthmus/src/test/java/io/substrait/isthmus/SubstraitExpressionConverterTest.java
nielspardon: this should be cleaned up

isthmus/src/test/java/io/substrait/isthmus/SubstraitExpressionConverterTest.java
nielspardon: I had to implement a `NamedStructExpressionGatherer` that also walks through the expressions of various relations to look for sub queries and apply the existing `NamedStructGatherer` to the sub queries in order to make it work without any external schema.

nielspardon: I saw a few places constructing a truly empty `EmptyScan` and added a `SubstraitBuilder.emptyScan()` method for these cases.

isthmus/src/main/java/io/substrait/isthmus/SubstraitToCalcite.java
```diff
@@ -181,9 +203,11 @@ private Pair<Integer, RelDataType> renameFields(
 
   private static class NamedStructGatherer extends AbstractRelVisitor<Void, RuntimeException> {
```
vbarua, line 204: I just realised this is using `AbstractRelVisitor` which means you had to add a bunch of boilerplate.

We don't yet have something like the Calcite `RelShuttle` interface, but if you switch to use the `RelCopyOnWriteVisitor` and `ExpressionCopyOnWriteVisitor` visitors (ignore the copying part of them) they have been implemented to be exhaustive in terms of visiting all relations and expressions so you can be guaranteed to see every NamedScan and you won't need to override any of the other visitors.

vbarua (approved):
Left one or comment with a simplifcation for the NamedStruct gatherer, but otherwise looks good to me.

isthmus/src/main/java/io/substrait/isthmus/SubstraitToCalcite.java
vbarua: Filed a ticket related to this for adding first-class relation and expression shuttle visitors
https://github.com/substrait-io/substrait-java/issues/356

nielspardon: done. I switched to the `RelCopyOnWriteVisitor`.

PR #354, opened 2025-03-24
feat(isthmus): support EXISTS and UNIQUE set predicates

nielspardon:
Should be merged after #353 since it shares the same test infrastructure.

Currently, the Substrait to Calcite mappings do not support exists and unique set predicates like the follow SQL example from TPC-H query no. 4:

```sql
select
  o.o_orderpriority,
  count(*) as order_count
from
  "orders" o

where
  o.o_orderdate >= date '1996-10-01'
  and o.o_orderdate < date '1996-10-01' + interval '3' month
  and
  exists (
    select
      *
    from
      "lineitem" l
    where
      l.l_orderkey = o.o_orderkey
      and l.l_commitdate < l.l_receiptdate
  )
group by
  o.o_orderpriority
order by
  o.o_orderpriority
```

This PR adds support for mapping exists and unique set predicates from Substrait into Calcite and adds a test case showing that it works now.

### DIFF
--- a/isthmus/src/main/java/io/substrait/isthmus/expression/ExpressionRexConverter.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/expression/ExpressionRexConverter.java
@@ -6,6 +6,7 @@ import io.substrait.expression.EnumArg;
 import io.substrait.expression.Expression;
 import io.substrait.expression.Expression.FailureBehavior;
 import io.substrait.expression.Expression.ScalarSubquery;
+import io.substrait.expression.Expression.SetPredicate;
 import io.substrait.expression.Expression.SingleOrList;
 import io.substrait.expression.Expression.Switch;
 import io.substrait.expression.FieldReference;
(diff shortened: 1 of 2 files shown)


isthmus/src/main/java/io/substrait/isthmus/expression/ExpressionRexConverter.java
```diff
@@ -545,4 +546,18 @@ public RexNode visit(ScalarSubquery expr) throws RuntimeException {
...
+  public RexNode visit(SetPredicate expr) throws RuntimeException {
+    RelNode inputRelnode = expr.tuples().accept(relNodeConverter);
+    switch (expr.predicateOp()) {
+      case PREDICATE_OP_EXISTS:
+        return RexSubQuery.exists(inputRelnode);
+      case PREDICATE_OP_UNIQUE:
+        return RexSubQuery.unique(inputRelnode);
+      case PREDICATE_OP_UNSPECIFIED:
+      default:
+        return super.visit(expr);
```
vbarua, line 560: minor: this is going to invoke `visitFallback` which will throw
```java
        String.format(
            "Expression %s of type %s not handled by visitor type %s.",
            expr, expr.getClass().getCanonicalName(), this.getClass().getCanonicalName()));
```
I think it would be helpful to include a more specific error here, along the lines of "Cannot handle SetPredicate with when PredicateOp is unspecified" to better guide users towards the issue.

vbarua (approved):
Looks good overall. Left one minor usability improvment.

isthmus/src/main/java/io/substrait/isthmus/expression/ExpressionRexConverter.java
nielspardon: I added a more specific exception that can also handle unimplemented new enum entries.
```
      case PREDICATE_OP_UNSPECIFIED:
      default:
            String.format(
                "Cannot handle SetPredicate when PredicateOp is %s.", expr.predicateOp().name()));
```

vbarua (approved):

PR #355, opened 2025-03-25
fix(isthmus): use explicit return type for scalar function expressions

nielspardon:
When converting a Substrait plan back to Apache Calcite which contains a datetime subtract such as the one generated by SQL-to-Substrait code for TPC-H query no. 1 the mapping fails with a `java.lang.ArrayIndexOutOfBoundsException` because the Substrait-to-Calcite mapping code relies on Calcite's type inference for determining the return type of the `subtract` function call. 

Since the Substrait plan contains an explicitly return type in this case for this function call we can simply provide the return type explicitly to Calcite and avoid using the return type inference.

This PRs provides the function return type from the Substrait plan to Calcite and adds a test case for verifying that the expression conversion no longer throws an Exception and sets the function return type correctly in Calcite.

### DIFF
--- a/isthmus/src/main/java/io/substrait/isthmus/expression/ExpressionRexConverter.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/expression/ExpressionRexConverter.java
@@ -321,7 +321,8 @@ public class ExpressionRexConverter extends AbstractExpressionVisitor<RexNode, R
             .mapToObj(i -> eArgs.get(i).accept(expr.declaration(), i, this))
             .collect(java.util.stream.Collectors.toList());
 
-    return rexBuilder.makeCall(operator, args);
+    RelDataType returnType = typeConverter.toCalcite(typeFactory, expr.outputType());
+    return rexBuilder.makeCall(returnType, operator, args);
   }
 
   private String callConversionFailureMessage(
--- a/isthmus/src/test/java/io/substrait/isthmus/SubstraitExpressionConverterTest.java
+++ b/isthmus/src/test/java/io/substrait/isthmus/SubstraitExpressionConverterTest.java
@@ -6,6 +6,9 @@ import static org.junit.jupiter.api.Assertions.assertThrows;
 
 import io.substrait.dsl.SubstraitBuilder;
 import io.substrait.expression.Expression;
+import io.substrait.expression.Expression.ScalarFunctionInvocation;
+import io.substrait.expression.ExpressionCreator;
+import io.substrait.extension.DefaultExtensionCatalog;
 import io.substrait.isthmus.expression.ExpressionRexConverter;
 import io.substrait.relation.Project;
 import io.substrait.relation.Rel;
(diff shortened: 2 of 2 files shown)


isthmus/src/main/java/io/substrait/isthmus/expression/ExpressionRexConverter.java
```diff
@@ -321,7 +321,8 @@ public RexNode visit(Expression.ScalarFunctionInvocation expr) throws RuntimeExc
             .mapToObj(i -> eArgs.get(i).accept(expr.declaration(), i, this))
             .collect(java.util.stream.Collectors.toList());
 
-    return rexBuilder.makeCall(operator, args);
+    RelDataType returnType = typeConverter.toCalcite(typeFactory, expr.outputType());
+    return rexBuilder.makeCall(returnType, operator, args);
```
vbarua, line 325: This change also brings us in line with the converison for window functions
https://github.com/substrait-io/substrait-java/blob/d73723274f46ad6329fa97874788d3255ff3cd7a/isthmus/src/main/java/io/substrait/isthmus/expression/ExpressionRexConverter.java#L345

and aggregate functions
https://github.com/substrait-io/substrait-java/blob/d73723274f46ad6329fa97874788d3255ff3cd7a/isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java#L303

isthmus/src/test/java/io/substrait/isthmus/SubstraitExpressionConverterTest.java
```diff
@@ -159,6 +162,32 @@ Rel createSubQueryRel() {
...
+    Project query = b.project(input -> List.of(expr), b.emptyScan());
+
+    SubstraitToCalcite substraitToCalcite = new SubstraitToCalcite(extensions, typeFactory);
+    RelNode calciteRel = substraitToCalcite.convert(query);
+    RexNode calciteExpr = ((LogicalProject) calciteRel).getProjects().get(0);
+
+    assertEquals(
+        TypeConverter.DEFAULT.toCalcite(typeFactory, TypeCreator.REQUIRED.DATE),
+        calciteExpr.getType());
+  }
```
vbarua, line 189: I like the change above, but I don't think this is the correct test for it.

The fact that the Calcite return type inference throws makes me think that we're mapping `"subtract:date_iday"` to the wrong Calcite function and/or that the return type inference as defined doesn't correspond with Substrait. Your (good) change sort of papers over the issue, but it's still there.

Ideally, the explicit return type in Substrait AND the inferred return type from Calcite should agree. In fact, when it doesn't it can problems during optimization. This is also an issue beyond the scope of your change here.

All that being said, I'm not sure what a nice way to check this. One idea I have is to construct an invalid scalar function call invoking `add:i64_i64` but with return type `string`, which is obviously wrong in both Calcite and Substrait, and check that the constructed call still has the wrong type. Effectively, we want to directly check a property like `functionConversionUsesGivenReturnType` even when the return type is wrong.

nielspardon: > The fact that the Calcite return type inference throws makes me think that we're mapping `"subtract:date_iday"` to the wrong Calcite function and/or that the return type inference as defined doesn't correspond with Substrait. Your (good) change sort of papers over the issue, but it's still there.

I did do a bit of digging while debugging the `ArrayIndexOutOfBoundsException`. What happens is that the default Apache Calcite datetime subtract (which we map to) is not a 2 argument function but a 3 argument function which expects the return type as the third argument:

```
A special operator for the subtraction of two DATETIMEs. The format of DATETIME subtraction is:

    "(" <datetime> "-" <datetime> ")" <interval qualifier>

This operator is special since it needs to hold the additional interval qualifier specification.
```

https://calcite.apache.org/javadocAggregate/org/apache/calcite/sql/fun/SqlStdOperatorTable.html#MINUS_DATE

https://github.com/apache/calcite/blob/b394abca407eea06ee8f8b50f30bd443f286bb42/core/src/main/java/org/apache/calcite/sql/fun/SqlStdOperatorTable.java#L1428-L1439

We're getting the `ArrayIndexOutOfBoundsException` because that third parameter is missing in this specific case.

Interestingly, there's an alternative implementation for Big Query which has 2 arguments:

https://calcite.apache.org/javadocAggregate/org/apache/calcite/sql/fun/SqlInternalOperators.html#MINUS_DATE2

Looks to me there's a bit of a special case going on here with the specific datetime subtract function that triggered my investigation into this and not necessarily a general type inference issue.

vbarua: Wrote https://github.com/substrait-io/substrait-java/issues/377 to capture the specific subtract issue

vbarua (approved):
LGTM

isthmus/src/test/java/io/substrait/isthmus/SubstraitExpressionConverterTest.java
vbarua: Further modifying these tests in https://github.com/substrait-io/substrait-java/pull/378

PR #358, opened 2025-03-26
build: make spotless run more reliably

nielspardon:
When I run `./gradlew clean build` on my local clone I constantly see Spotless tripping up on files in the `build` folder no longer existing which means I have to run the command multiple times. This suggests to me that the `targetExclude` configuration does not work as expected.

Instead this PRs configures the `target` to format every `*.java` file within any of the `src/*/java/**` directories which works more reliably and achieves the same goal as the current configuration.

This PR also adds `.metals` to the `.gitignore` for the VS Code users among us which uses metals for the Scala code.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,4 @@ out/**
 .pmdCache
 
 */bin
+.metals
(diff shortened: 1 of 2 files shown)


vbarua (approved):
Tested `./gradlew spotlessJavaApply` and spotless is still applied as expected. Changes look reasonable to me.

PR #360, opened 2025-03-28
fix(isthmus): support more datetime extract variants

nielspardon:
Currently, the Substrait to Calcite mappings only support mapping scalar functions with enum arguments for the `extract:req_ts` function which means it only supports the `extract` function with an enum argument and a timestamp but none of the other datetime extract functions defined in the default datetime extension.

This PR adds basic support for all datetime extract functions which are not using the indexing argument which indicates whether e.g. months should start to be counted from 0 instead of 1.

It is still a naive mapping from the enum values defined in the datetime extension YAML to Apache Calcite's `TimeUnitRange` enum which misses some of the Substrait enum values (e.g. `US_YEAR`, `PICOSECONDS`) or has spelling differences for some of the enum values (`ISOYEAR` vs. `ISO_YEAR`).

This PR at least allows some of the TPC-H queries to be mapped back from Substrait to Calcite which currently throws an `UnsupportedOperationException` due to missing support for date extract.

I would suggest to revisit the function mapping code in a follow-up PR to create a framework which supports more sophisticated mapping logic.

### DIFF
--- a/isthmus/src/main/java/io/substrait/isthmus/expression/EnumConverter.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/expression/EnumConverter.java
@@ -1,10 +1,13 @@
 package io.substrait.isthmus.expression;
 
-import com.google.common.collect.BiMap;
-import com.google.common.collect.HashBiMap;
 import io.substrait.expression.EnumArg;
 import io.substrait.extension.DefaultExtensionCatalog;
 import io.substrait.extension.SimpleExtension;
+import io.substrait.extension.SimpleExtension.Argument;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
 import java.util.Optional;
 import java.util.function.Supplier;
 import org.apache.calcite.avatica.util.TimeUnitRange;
@@ -25,16 +28,34 @@ import org.apache.calcite.sql.type.SqlTypeName;
  */
 public class EnumConverter {
 
-  private static final BiMap<Class<? extends Enum>, ArgAnchor> calciteEnumMap = HashBiMap.create();
+  private static final Map<ArgAnchor, Class<? extends Enum<?>>> calciteEnumMap = new HashMap<>();
 
   static {
+    // deprecated {@link io.substrait.type.Type.Timestamp}
     calciteEnumMap.put(
-        TimeUnitRange.class,
-        argAnchor(DefaultExtensionCatalog.FUNCTIONS_DATETIME, "extract:req_ts", 0));
+        argAnchor(DefaultExtensionCatalog.FUNCTIONS_DATETIME, "extract:req_ts", 0),
+        TimeUnitRange.class);
+    // deprecated {@link io.substrait.type.Type.TimestampTZ}
+    calciteEnumMap.put(
+        argAnchor(DefaultExtensionCatalog.FUNCTIONS_DATETIME, "extract:req_tstz_str", 0),
+        TimeUnitRange.class);
+
+    calciteEnumMap.put(
+        argAnchor(DefaultExtensionCatalog.FUNCTIONS_DATETIME, "extract:req_pts", 0),
+        TimeUnitRange.class);
+    calciteEnumMap.put(
+        argAnchor(DefaultExtensionCatalog.FUNCTIONS_DATETIME, "extract:req_ptstz_str", 0),
+        TimeUnitRange.class);
+    calciteEnumMap.put(
+        argAnchor(DefaultExtensionCatalog.FUNCTIONS_DATETIME, "extract:req_date", 0),
+        TimeUnitRange.class);
+    calciteEnumMap.put(
+        argAnchor(DefaultExtensionCatalog.FUNCTIONS_DATETIME, "extract:req_time", 0),
+        TimeUnitRange.class);
   }
 
-  private static Optional<Enum> constructValue(
-      Class<? extends Enum> cls, Supplier<Optional<String>> option) {
+  private static Optional<Enum<?>> constructValue(
+      Class<? extends Enum<?>> cls, Supplier<Optional<String>> option) {
     if (cls.isAssignableFrom(TimeUnitRange.class)) {
       return option.get().map(TimeUnitRange::valueOf);
     } else {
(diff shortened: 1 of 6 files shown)


mbwhite (approved):

isthmus/src/main/java/io/substrait/isthmus/expression/ExpressionRexConverter.java
```diff
@@ -209,20 +212,65 @@ public RexNode visit(Expression.DateLiteral expr) throws RuntimeException {
...
+    return rexBuilder.makeLiteral(
+        getTimestampString(expr.value(), expr.precision()),
+        typeConverter.toCalcite(typeFactory, expr.getType()));
+  }
+
+  private TimestampString getTimestampString(long microSec) {
+    return getTimestampString(microSec, 6);
+  }
+
+  private TimestampString getTimestampString(long value, int precision) {
```
vbarua, line 243: Good call extracting this. It's also easier to read, at least for me ✨

isthmus/src/main/java/io/substrait/isthmus/expression/ExpressionRexConverter.java
```diff
@@ -209,20 +212,65 @@ public RexNode visit(Expression.DateLiteral expr) throws RuntimeException {
...
+        {
+          long seconds = TimeUnit.NANOSECONDS.toSeconds(value);
+          int fracSecondsInNano = (int) (value - TimeUnit.SECONDS.toNanos(seconds));
+          return TimestampString.fromMillisSinceEpoch(TimeUnit.SECONDS.toMillis(seconds))
+              .withNanos(fracSecondsInNano);
+        }
+      default:
+        throw new UnsupportedOperationException(
+            String.format("Cannot handle PrecisionTimestamp with precision %d.", precision));
+    }
```
vbarua, line 273: Are there any tests that check the conversion of values with precision other than 6?

isthmus/src/test/java/io/substrait/isthmus/SubstraitExpressionConverterTest.java
```diff
@@ -202,4 +211,316 @@ void assertTypeMatch(RelDataType actual, Type expected) {
...
+            TypeCreator.REQUIRED.I64,
+            ImmutableEnumArg.builder().value("MONTH").build(),
+            ImmutableEnumArg.builder().value("ONE").build(),
+            Expression.DateLiteral.builder().value(0).build());
+
+    Project query = b.project(i -> List.of(reqReqDateFn), b.emptyScan());
+
+    SubstraitToCalcite substraitToCalcite = new SubstraitToCalcite(extensions, typeFactory);
+    assertThrows(UnsupportedOperationException.class, () -> substraitToCalcite.convert(query));
+  }
```
vbarua, line 525: Could you move these tests into `FunctionConversionTests` as they are verifying specific functions.

As in my changes in https://github.com/substrait-io/substrait-java/pull/378/files, you should also be able test these using an ExpressionRexConverter directly as well because these tests only use literals so there's no schema that needs to be collected.

isthmus/src/test/java/io/substrait/isthmus/SubstraitExpressionConverterTest.java
```diff
@@ -202,4 +211,316 @@ void assertTypeMatch(RelDataType actual, Type expected) {
...
+    assertEquals(SqlKind.LITERAL, operand2.getKind());
+    assertEquals(SqlTypeName.TIMESTAMP_WITH_LOCAL_TIME_ZONE, operand2.getType().getSqlTypeName());
+    assertEquals(
+        0, ((RexLiteral) operand2).getValueAs(TimestampString.class).getMillisSinceEpoch());
+
+    RexNode operand3 = extract.getOperands().get(2);
+    assertInstanceOf(RexLiteral.class, operand3);
+    assertEquals(SqlKind.LITERAL, operand3.getKind());
+    assertEquals(SqlTypeName.VARCHAR, operand3.getType().getSqlTypeName());
+    assertEquals("GMT", RexLiteral.stringValue(operand3));
```
vbarua, line 255: What do you think about doing something like
```java
    assertEquals(
        "EXTRACT(FLAG(MONTH), 1970-01-01 00:00:00:TIMESTAMP_WITH_LOCAL_TIME_ZONE(6), 'GMT':VARCHAR)",
        extract.toString());
```

for these tests. I _think_ it captures most if not all of what you're trying to verify, but it's more compact and readable.

vbarua (changes requested):
The functional changes look good overall. I did leave some simplification suggestions, and one request to move tests.

isthmus/src/main/java/io/substrait/isthmus/expression/EnumConverter.java
```diff
@@ -25,16 +28,34 @@
  */
 public class EnumConverter {
 
-  private static final BiMap<Class<? extends Enum>, ArgAnchor> calciteEnumMap = HashBiMap.create();
+  private static final Map<ArgAnchor, Class<? extends Enum<?>>> calciteEnumMap = new HashMap<>();
```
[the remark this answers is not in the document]
nielspardon: Yes, the `BiMap` design didn't really fit reality in that multiple Substrait `ArgAnchor` can map to the same enum class and then it turned out it was not really needed anyway. Fixes an issue and simplifies the code a little.

isthmus/src/main/java/io/substrait/isthmus/expression/EnumConverter.java
```diff
@@ -79,14 +101,14 @@ private static Optional<SimpleExtension.EnumArgument> findEnumArg(
...
         Object v = literal.getValue();
         if (!literal.isNull() && (v instanceof Enum)) {
-          Enum value = (Enum) v;
-          Optional<ArgAnchor> enumAnchor =
-              Optional.ofNullable(calciteEnumMap.getOrDefault(value.getClass(), null));
+          Enum<?> value = (Enum<?>) v;
+          Optional<ArgAnchor> enumAnchor = Optional.of(argAnchor(function, argIdx));
           yield enumAnchor
               .flatMap(en -> findEnumArg(function, en))
               .map(ea -> EnumArg.of(ea, value.name()));
```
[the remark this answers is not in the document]
nielspardon: good catch. changed it

isthmus/src/test/java/io/substrait/isthmus/SubstraitExpressionConverterTest.java
nielspardon: I changed the tests to this style

isthmus/src/test/java/io/substrait/isthmus/SubstraitExpressionConverterTest.java
nielspardon: I moved the tests

isthmus/src/main/java/io/substrait/isthmus/expression/ExpressionRexConverter.java
nielspardon: I added specific test cases in `ExpressionConvertabilityTest` for supported / unsupported precision values.

vbarua (approved):

PR #361, opened 2025-03-31
feat(isthmus) support calcite 1.39.0

mbwhite:
As a result of an (accidental) upgrade Calcite 1.39; conversions via Isthmus started to report that table names where unknown;

Isthmus has an external API that allows tables to be lookup via calls to a function - dynamic or 'lazy' lookup.  A special subclass of the CalciteSimpleSchema was created to support this.  This special subclass had to be in the same package as the CalciteSimpleSchema as it took advantage of being able to override a number of protected methods. 

Calcite 1.39.0 introduced similar support for 'lazy' table lookups; that implementation, though changed the protected methods sufficiently Isthmus now failed (indeed rebuilding the code fails to compile).

This PR removes the special subclass, and allows Ishtmus to work as before; One complexity is how namespace tables are handled with the original Isthmus API, and Calcite new lookup API.

This does make the code a lot more complex; for future releases, a suggestion would be to deprecate this function interface and replace it with accepting the same Calcite lookup API.

- `SubstraitCalciteSchema` is the replacement class; this is not an external but used internally; it's now using the Calcite lookup API
- Other changes are just to accommodate the new class - it's fulling the same role as the previous one, so no other logic has really changed. 
- Added a couple of end-end tests to add additional validation - as this was how the issue was first seen; though even without that the Calcite change resulted in > 40% test case failures.


isthmus/src/main/java/io/substrait/isthmus/SqlConverterBase.java
```diff
@@ -217,9 +210,10 @@ public DefinedTable(String name, RelDataTypeFactory factory, RelDataType type) {
 
     @Override
     public RelDataType getRowType(RelDataTypeFactory typeFactory) {
-      //      if (factory != typeFactory) {
-      //        throw new IllegalStateException("Different type factory than previously used.");
-      //      }
+      // if (factory != typeFactory) {
+      // throw new IllegalStateException("Different type factory than previously
+      // used.");
+      // }
```
nielspardon, line 216: not sure what policy is for `substrait-java` but if you are anyway touching those lines I guess they can also be removed

isthmus/src/main/java/io/substrait/isthmus/SubstraitCalciteSchema.java
```diff
@@ -0,0 +1,284 @@
...
+/**
+ * A subclass of the Calcite Schema for creation from a Substrait relation
+ *
+ * <p>Implementation note:
+ *
+ * <p>The external Isthmus API can take a function that will return the table schema when needed,
+ * rather than it being available up front.
+ *
+ * <p>This was implemented by a special subclass of the Calcite simple schena. Since this was
+ * changed in Caclite 1.39.0; it failed to work; the protected methods it extended from change.
```
nielspardon, line 36:
```suggestion
 * <p>This was implemented by a special subclass of the Calcite simple schema. Since this was
 * changed in Calcite 1.39.0; it failed to work; the protected methods it extended from changed.
```

isthmus/src/main/java/io/substrait/isthmus/SubstraitToCalcite.java
```diff
@@ -57,19 +48,13 @@ public SubstraitToCalcite(
...
-          if (table == null) {
-            return null;
-          }
-          return new SqlConverterBase.DefinedTable(
-              id.get(id.size() - 1),
-              typeFactory,
-              typeConverter.toCalcite(typeFactory, table.struct(), table.names()));
-        };
-    return LookupCalciteSchema.createRootSchema(lookup);
+    System.out.println(rel);
```
mbwhite, line 51: need to remove this

isthmus/src/main/java/io/substrait/isthmus/SubstraitCalciteSchema.java
```diff
@@ -0,0 +1,284 @@
...
+ *
+ * <p>The change, ironically, was support a laxy approach to looking up Calcite schemas. Good -
+ * *but* the external function in Isthmus is assuming it's going to get called with a fully
+ * namespaced table table. Which Calcite though sees as being subschemas.
```
nielspardon, line 40:
```suggestion
 * <p>The change, ironically, was to support a lazy approach to looking up Calcite schemas. Good -
 * *but* the external function in Isthmus is assuming it's going to get called with a fully
 * namespaced table name. Which Calcite though sees as being subschemas.
```

isthmus/src/main/java/io/substrait/isthmus/SubstraitCalciteSchema.java
```diff
@@ -0,0 +1,284 @@
...
+ *
+ * <p>This results is some 'complex' code below to try and map the laszy way calcite now works and
+ * maintain the existing Isthmus API
```
nielspardon, line 43:
```suggestion
 * <p>This results in some 'complex' code below to try and map the lazy way Calcite now works and
 * maintain the existing Isthmus API
```

isthmus/src/main/java/io/substrait/isthmus/SubstraitCalciteSchema.java
```diff
@@ -0,0 +1,284 @@
...
+
+  private Map<String, Table> tables;
+  private Function<List<String>, NamedStruct> tableLookup;
+
+  private RelDataTypeFactory typeFactory;
+  private TypeConverter typeConverter;
+
+  /**
+   * Maintain a track of the 'prefix' of this schema... i.e. allows recreation of the full
+   * quallified name of this subschema
```
nielspardon, line 58:
```suggestion
   * Maintain a track of the 'prefix' of this schema... i.e. allows recreation of the fully
   * qualified name of this subschema
```

isthmus/src/main/java/io/substrait/isthmus/SubstraitCalciteSchema.java
```diff
@@ -0,0 +1,284 @@
...
+  @Override
+  protected Map<String, Table> getTableMap() {
+    return tables;
+  }
+
+  @Override
+  public Lookup<? extends Schema> subSchemas() {
+    var defaultLookup = super.subSchemas();
+
+    // Note ono the lookups, calcite prefers calling getIgnoreCase() initially
```
nielspardon, line 84:
```suggestion
    // Note on the lookups, Calcite prefers calling getIgnoreCase() initially
```

isthmus/src/main/java/io/substrait/isthmus/SubstraitCalciteSchema.java
```diff
@@ -0,0 +1,284 @@
...
+
+      @Override
+      public Set<String> getNames(LikePattern pattern) {
+        return ImmutableSet.of();
+      }
+    };
+  }
+
+  /**
+   * Turn this into a root Calciteschema Choice of settings is based on current isthmus behaviour
```
nielspardon, line 175:
```suggestion
   * Turn this into a root CalciteSchema; choice of settings is based on current Isthmus behaviour
```

isthmus/src/main/java/io/substrait/isthmus/SubstraitToCalcite.java
nielspardon, line 51: I guess we don't need this leftover debug code anymore.

```suggestion
```

isthmus/src/main/java/io/substrait/isthmus/SubstraitCalciteSchema.java
```diff
@@ -0,0 +1,284 @@
...
+        throw new IllegalArgumentException("Specify either 'rel' or 'tableLookup' ");
+      }
+
+      if (rel != null) {
+        // If there are any named structs within the relation, gather these and convert
+        // them to a map of tables
+        // index by name; note that the name of the table is 'un-namespaced' here.
+        // This was the existing logic so it has not been altered.
+        Map<List<String>, NamedStruct> tableMap = NamedStructGatherer.gatherTables(rel);
+        System.out.println(tableMap);
```
nielspardon, line 236: I guess we don't need this leftover debug code anymore.

```suggestion
```

isthmus/src/test/java/io/substrait/isthmus/api/TestIsthmusEndToEnd.java
```diff
@@ -0,0 +1,97 @@
...
+import org.apache.calcite.adapter.java.ReflectiveSchema;
+import org.apache.calcite.jdbc.JavaTypeFactoryImpl;
+import org.apache.calcite.schema.Schema;
+import org.apache.calcite.sql.parser.SqlParseException;
+import org.junit.jupiter.api.Test;
+
+/** API level testing */
+public class TestIsthmusEndToEnd {
+
+  /** Conversion of a Substrait Protobuf to */
```
nielspardon, line 25: this comment looks incomplete

isthmus/src/test/java/io/substrait/isthmus/api/TestIsthmusEndToEnd.java
```diff
@@ -0,0 +1,97 @@
...
+  @Test
+  public void substraitToSqlViaCalcite() throws IOException {
+
+    // create the protobuf substarit plan
```
nielspardon, line 29:
```suggestion
    // create the protobuf Substrait plan
```

isthmus/src/test/java/io/substrait/isthmus/api/TestIsthmusEndToEnd.java
```diff
@@ -0,0 +1,97 @@
...
+    byte[] planProtobuf = Resources.toByteArray(Resources.getResource("substrait_sql_003.plan"));
+    io.substrait.proto.Plan protoPlan = io.substrait.proto.Plan.parseFrom(planProtobuf);
+    // convert this to the Substrait Plan POJO
+    ProtoPlanConverter converter = new io.substrait.plan.ProtoPlanConverter();
+    io.substrait.plan.Plan pojoPlan = converter.from(protoPlan);
+
+    System.out.println("POJO Substrait Plan::\n" + pojoPlan);
+
+    // To convert from Substrait first need to convert to calcite, and then calcite
```
nielspardon, line 38:
```suggestion
    // To convert from Substrait first need to convert to Calcite, and then Calcite
```

isthmus/src/test/java/io/substrait/isthmus/api/TestIsthmusEndToEnd.java
```diff
@@ -0,0 +1,97 @@
...
+    String sql = SubstraitToSql.toSql(calciteRel).replace("\n", " ");
+
+    assertEquals(
+        sql,
+        Resources.toString(
+            Resources.getResource("substrait_sql_003.sql"), Charset.defaultCharset()));
+    System.out.println(sql);
+  }
+
+  /** SQL to Substrait Using a set of `CREATE TABLE...` statements to define the schema */
```
nielspardon, line 58:
```suggestion
  /** SQL to Substrait using a set of `CREATE TABLE...` statements to define the schema */
```

isthmus/src/test/java/io/substrait/isthmus/api/TestIsthmusEndToEnd.java
```diff
@@ -0,0 +1,97 @@
...
+
+/** API level testing */
+public class TestIsthmusEndToEnd {
+
+  /** Conversion of a Substrait Protobuf to */
+  @Test
+  public void substraitToSqlViaCalcite() throws IOException {
+
+    // create the protobuf substarit plan
+    byte[] planProtobuf = Resources.toByteArray(Resources.getResource("substrait_sql_003.plan"));
```
nielspardon, line 30: other tests are using `SubstraitBuilder` to provide a Substrait plan for test cases which I think is more reliable than relying on a binary file in the tests resources which might have been encoded in an older version of the Substrait spec and which might break with future Substrait spec changes.

isthmus/src/main/java/io/substrait/isthmus/SqlConverterBase.java
mbwhite: Yes; though they were there before this PR, so not sure why they shown as changed. or why they were commented out at all. 

Might as well remove them - they will be in the git history if needed

isthmus/src/test/java/io/substrait/isthmus/api/TestIsthmusEndToEnd.java
mbwhite: fair point; will look at using that builder.

mbwhite:
> How urgent is the 1.39 upgrade for you? If deprecating / removing the TableLookupAPI simplifies this, I'd be open to doing that first. I can take a look at that this week as I have some other changes around SQL handling that might simplify this as well.

Not urgent - our 'version checker' wanted to bump the calcite version up, therefore this came up. No other reason to move to calcite version;  just pre-empting a problem!

No problem with dropping the lookup first;  in fact I was thinking of adding an API to accept the Calcite 1.39 loopup API instead.. The function of a dynamic schema is still present;

Interesting to see other SQL updates - there was one I was thinking of to make 

```
  public static List<String> toSql(Plan plan, SqlDialect dialect) {
    SimpleExtension.ExtensionCollection extensions = SimpleExtension.loadDefaults();

    SubstraitToCalcite substrait2Calcite = new SubstraitToCalcite(
        extensions, new JavaTypeFactoryImpl(SubstraitTypeSystem.TYPE_SYSTEM));

    return plan.getRoots().stream().map(root -> {
      var calciteRel = substrait2Calcite.convert(root).project();
      return SubstraitToSql.toSql(calciteRel);
    }).collect(Collectors.toList());

  }
  ```

vbarua:
@mbwhite my PR/proposal to remove the troublesome API is in https://github.com/substrait-io/substrait-java/pull/363

vbarua:
https://github.com/substrait-io/substrait-java/pull/363 has been merged in. Hopefully it can simplify this update.

Issue #362, opened 2025-03-31
[isthmus] simplify Substrait SQL API


vbarua:
## Goals
1. Isthmus provides a _reference_ implementation for converting from SQL to Calcite
    *  This includes any utilities or configurations needed to instantiate the Calcite [SqlParser](https://github.com/apache/calcite/blob/main/core/src/main/java/org/apache/calcite/sql/parser/SqlParser.java) and Calcite [SqlValidator](https://github.com/apache/calcite/blob/main/core/src/main/java/org/apache/calcite/sql/validate/SqlValidator.java) classes needed for parsing.
2. Isthmus provides a _reference_ implementation for converting to Calcite to SQL
    * This includes any utilities or configurations needs to instantiate [SqlDialect](https://github.com/apache/calcite/blob/main/core/src/main/java/org/apache/calcite/sql/SqlDialect.java) class.

The purpose of these reference implementation is to:
1. Better enable testing of Isthmus with TPCH queries and the like.
2. Provide users with a reference examples for configuring their own parsers and SQL generators.

## Non-Goals

Isthmus as a library is primarily intended for conversions between Calcite and Substrait.

### Flexible SQL Parser
Isthmus currently provides a [FeatureBoard](https://github.com/substrait-io/substrait-java/blob/main/isthmus/src/main/java/io/substrait/isthmus/FeatureBoard.java) for tweaking varoius settings in SQL Parsing.

This should be avoided. If end-users wish to configure their own SQL parsing behaviour, they should instantiate their own Calcite parser directly. This will help avoid coupling the Substrait parser to end-user systems. Parsing is a capability of Calcite, not Isthmus, so we should lean on Calcite and push users towards Calcite for this. Having a reference implementation in Substrait should help users configure their own parsers.

mbwhite:
To rephrase to make sure I've understood;  

- the main feature of Ishtmus is Substrait<->Calcite
- a secondary feature is the _reference_ implementation for SQL <-> Calcite;  but with a clear interface between the two so it could be removed and replaced if needed;

vbarua:
@mbwhite you're rephrase is accurate.

To rephrase/ expand on your second point, I want to modularize our SQL handling sufficiently so that we can pipeline the transformations. Effectively:
* `SqlToSubstrait` becomes `SqlToCalcite` followed by `CalciteToSubstrait`
* `SubstraitToSql` becomes `SubstraitToCalcite` followed by `CalciteToSql`

[...]

PR #363, opened 2025-04-01
feat(isthmus): new Prepare.CatalogReader based APIs for SQL to/from Substrait

vbarua:
BREAKING CHANGE: removed function-based table lookup conversion methods


vbarua:
This PR removes the following public methods:
* `SqlToSubstrait#execute(String sql, Function<List<String>, NamedStruct> tableLookup)`
* `SubstraitToSql#substraitRelToCalciteRel(Rel relRoot, Function<List<String>, NamedStruct> tableLookup)`

and adds 2 new methods that replace their functionality:
* `SqlToSubstrait#execute(String sql, Prepare.CatalogReader catalogReader)`
* `SubstraitToSql#substraitRelToCalciteRel(Rel relRoot, Prepare.CatalogReader catalog)`

The original API was introduced in https://github.com/substrait-io/substrait-java/pull/26 (along with https://github.com/substrait-io/substrait-java/pull/33) with the following intent:
> This uses a function to look up a table schema without a create table statement. I've intentionally kept the interface free from calcite objects so as to not export them as part of the isthmus public API.

At this point though, we expose a number of Calcite objects in our public API in order for users to refine conversion between Substrait and Calcite. 

Additionally, I would make the point that converting from SQL to Substrait is really 2 conversions:
1. SQL -> Calcite, which is a capability of Calcite that we can leverage
2. Calcite -> Substrait, which is the capability that Isthmus aims to provide

Given how tightly integrated 1 is with Calcite, I would argue that instead of hiding Calcite details from users it would be more productive to guide them towards using Calcite effectively to parse and unparse SQL. This is part of the aim of https://github.com/substrait-io/substrait-java/issues/362, which this PR works towards.

isthmus/src/main/java/io/substrait/isthmus/SqlConverterBase.java
```diff
@@ -197,8 +166,10 @@ private Validator(
     }
 
     public static Validator create(
-        RelDataTypeFactory factory, CalciteCatalogReader catalog, SqlValidator.Config config) {
-      return new Validator(SubstraitOperatorTable.INSTANCE, catalog, factory, config);
+        RelDataTypeFactory factory,
+        SqlValidatorCatalogReader validatorCatalog,
```
vbarua, line 170: Calcite only requires a `SqlValidatorCatalogReader`, not a full `CalciteCatalogReader` when constructing a validator.

isthmus/src/main/java/io/substrait/isthmus/SqlConverterBase.java
```diff
@@ -90,41 +84,16 @@ Pair<SqlValidator, CalciteCatalogReader> registerCreateTables(List<String> table
         }
       }
     }
-    return Pair.of(validator, catalogReader);
+    return catalogReader;
```
vbarua, line 87: I've decoupled the creation of the schema and the creation of the validator. The validator is not need as part of creating the schema. The validator can be created after with the schema we produce here. Splitting this apart makes this more explicit. It also simplifies some future changes I have in mind.

These methods are all internal.

isthmus/src/main/java/io/substrait/isthmus/SqlToSubstrait.java
```diff
@@ -32,40 +30,31 @@ public SqlToSubstrait(FeatureBoard features) {
...
-      throws SqlParseException {
-    var pair = registerCreateTables(tableLookup);
-    return executeInner(sql, factory, pair.left, pair.right);
-  }
-
   public Plan execute(String sql, List<String> tables) throws SqlParseException {
-    var pair = registerCreateTables(tables);
-    return executeInner(sql, factory, pair.left, pair.right);
+    CalciteCatalogReader catalogReader = registerCreateTables(tables);
+    SqlValidator validator = Validator.create(factory, catalogReader, SqlValidator.Config.DEFAULT);
```
vbarua, line 35: As indicated above, we now create the schema first and create validator from the schema. This works because the validator is not needed to validate the schema, but rather the query we are parsing.

isthmus/src/main/java/io/substrait/isthmus/SubstraitToSql.java
```diff
@@ -17,20 +16,16 @@ public class SubstraitToSql extends SqlConverterBase {
 
   public SubstraitToSql() {
     super(FEATURES_DEFAULT);
-    CalciteSchema rootSchema = CalciteSchema.createRootSchema(false);
```
vbarua, line 20: This was unused.

isthmus/src/test/java/io/substrait/isthmus/TableLookupTest.java
```diff
@@ -1,167 +0,0 @@
...
-import java.util.Objects;
-import java.util.function.Function;
-import org.apache.calcite.rel.RelNode;
-import org.apache.calcite.rel.RelRoot;
-import org.apache.calcite.sql.SqlKind;
-import org.apache.calcite.sql.parser.SqlParseException;
-import org.junit.jupiter.api.Assertions;
-import org.junit.jupiter.api.Test;
-
-public class TableLookupTest extends PlanTestBase {
```
vbarua, line 26: This test is no longer needed as this functionality is no longer supported.

vbarua:
Tagging @nielspardon @mbwhite as you've both been poking at the SQL conversion side of things.

Tagging @rymurr, @jinfengni as you added the original APIs that are being removed.

mbwhite:
Hi @vbarua agree with what you've proposed here; I understand the original intent, and the logic there but agree things have moved on as well. 

As found with Calcite 1.39, there's (another) lazy mechanism to get table schema information;  not sure how the `lookup` added in 1.39 is related to the `Prepare.CatalogReader` ?

nielspardon (approved):
LGTM

mbwhite (approved):

isthmus/src/main/java/io/substrait/isthmus/SqlToSubstrait.java
```diff
@@ -32,40 +30,31 @@ public SqlToSubstrait(FeatureBoard features) {
...
+    return executeInner(sql, validator, catalogReader);
   }
 
   public Plan execute(String sql, String name, Schema schema) throws SqlParseException {
```
vbarua, line 39: This API is somewhat redundant when we add `execute(String sql, Prepare.CatalogReader catalogReader)` as users can easily convert the `Schema` to a `Prepare.CatalogReader` satisfying instance themselves.

EpsilonPrime (approved):

Blizzara (approved):

PR #365, opened 2025-04-01
build: use OSV-Scanner v2 for vulnerability scan

bestbeforetoday:
- Use Gradle lockfile to assemble dependencies for vulnerability scanning, and avoid requiring the additional CycloneDX plugin.
- Update versions of Gradle actions.

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -33,27 +33,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
-      - uses: gradle/actions/wrapper-validation@v3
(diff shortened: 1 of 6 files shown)


bestbeforetoday:
Removed scanning of the **spark** subproject that I had initially added as part of this pull request since it contains many vulnerable dependencies that we likely cannot do much about since they are dictated by Spark.

vbarua (approved):
It's always nice when we can simplify the build files. Thanks for bumping the wrapper-version and setup-gradle tasks as well.

Verified that vulnerable dependencies are still detected by introducing a vulnerable dep in: https://github.com/substrait-io/substrait-java/pull/369

PR #367, opened 2025-04-03
feat(isthmus): always emit Cross relation when possible

vbarua:
BREAKING CHANGE: removed CrossJoinPolicy from FeatureBoard


vbarua:
The default policy is already to emit Cross relations
https://github.com/substrait-io/substrait-java/blob/385264096cd6f33997e53bf64d6b1bcbcff1fec3/isthmus/src/main/java/io/substrait/isthmus/FeatureBoard.java#L36-L42

If a downstream consumer can handle the Join relation, but not the Cross relation, it should be updated the Cross relation rather than having an optional processing override in our system.

EpsilonPrime (approved):

PR #372, opened 2025-04-04
feat(isthmus): always allow batch queries

vbarua:
BREAKING CHANGE: removed AllowsSqlBatch from FeatureBoard

### DIFF
--- a/isthmus-cli/src/test/java/io/substrait/isthmus/cli/IsthmusEntryPointTest.java
+++ b/isthmus-cli/src/test/java/io/substrait/isthmus/cli/IsthmusEntryPointTest.java
@@ -1,29 +1,17 @@
 package io.substrait.isthmus.cli;
 
-import static org.junit.jupiter.api.Assertions.assertFalse;
-import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.junit.jupiter.api.Assertions.assertEquals;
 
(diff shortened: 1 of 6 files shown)


vbarua:
There's not much point in controlling whether multiple statements can be processed. We might as well allow it by default. If users have issues with it, they should avoid giving the CLI multiple statements at once, which right now would fail anyways with an exception.

isthmus-cli/src/test/java/io/substrait/isthmus/cli/IsthmusEntryPointTest.java
```diff
@@ -1,29 +1,17 @@
...
+  void canProcessQuery() {
     IsthmusEntryPoint isthmusEntryPoint = new IsthmusEntryPoint();
-    new CommandLine(isthmusEntryPoint);
-    FeatureBoard features = isthmusEntryPoint.buildFeatureBoard();
-    assertFalse(features.allowsSqlBatch());
-  }
-
-  /** Test that the command line options are correctly parsed into the {@link FeatureBoard}. */
-  @Test
-  void customFeatureBoard() {
```
vbarua, line 23: Removed this test as it's not really verifying much anymore, and the FeatureBoard is not long for this world as well.

Blizzara (approved):

PR #373, opened 2025-04-04
feat(spark): add some date functions

andrew-coleman:
The date/time functions in Spark don’t map directly to the Substrait eqivalents. E.g.
- `date ± interval-days` are handled by the `DateAdd` & `DateSub` functions in Spark, but as a variant of the arithmetic `add` function in substrait.
- The date/time component extraction functions are all handled by different functions in Spark, but by a single `extract` function in Substrait with an `enum` argument to specify which component.

Neither of these could be handled using the existing function mapping capabilities in the `spark` module.

This commit exends this capability so that it can now handle these two scenarios in (I hope) a generic way.

I’ve added a few variants of the `extract` function - more can follow.

Adding this will give us 100% pass rate for all the TPC-DS querues.  The README is updated accordingly.

### DIFF
--- a/spark/src/main/scala/io/substrait/spark/expression/Enum.scala
+++ b/spark/src/main/scala/io/substrait/spark/expression/Enum.scala
@@ -1,0 +1,39 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to you under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.substrait.spark.expression
+
+import org.apache.spark.sql.catalyst.expressions.{LeafExpression, Unevaluable}
+import org.apache.spark.sql.types.{DataType, NullType}
+
+/**
+ * For internal use only. This represents the equivalent of a Substrait enum parameter type for use
+ * during conversion. It must not become part of a final Spark logical plan.
+ *
+ * @param value
+ *   The enum string value.
+ */
+case class Enum(value: String) extends LeafExpression with Unevaluable {
+  override def nullable: Boolean = false
+
+  override def dataType: DataType = NullType
+
+  override def equals(that: Any): Boolean = that match {
+    case Enum(other) => other == value
+    case _ => false
+  }
+}
--- a/spark/src/main/scala/io/substrait/spark/expression/FunctionConverter.scala
+++ b/spark/src/main/scala/io/substrait/spark/expression/FunctionConverter.scala
@@ -27,7 +27,7 @@ import org.apache.spark.sql.types.DataType
 
 import com.google.common.collect.{ArrayListMultimap, Multimap}
 import io.substrait.`type`.Type
-import io.substrait.expression.{Expression => SExpression, ExpressionCreator, FunctionArg}
+import io.substrait.expression.{EnumArg, Expression => SExpression, ExpressionCreator, FunctionArg}
 import io.substrait.expression.Expression.FailureBehavior
 import io.substrait.extension.SimpleExtension
 import io.substrait.function.{ParameterizedType, ToTypeString}
@@ -93,14 +93,28 @@ abstract class FunctionConverter[F <: SimpleExtension.Function, T](functions: Se
     (matcherMap, keyMap)
   }
 
-  def getSparkExpressionFromSubstraitFunc(key: String, outputType: Type): Option[Sig] = {
-    val sigs = substraitFuncKeyToSig.get(key)
-    sigs.size() match {
-      case 0 => None
-      case 1 => Some(sigs.iterator().next())
-      case _ => None
+  def getSparkExpressionFromSubstraitFunc(
+      key: String,
+      args: Seq[Expression]): Option[Expression] = {
+    val candidates = substraitFuncKeyToSig.get(key).asScala.toList
+    val sigs = if (candidates.length > 1) {
+      // attempt to disambiguate with the key (if it's been set)
+      val specific = candidates.filter {
+        case SpecialSig(_, _, Some(sig), _) if sig == key => true
+        case _ => false
+      }
+      if (specific.nonEmpty) {
+        specific
+      } else {
+        // no matching signature, so select the generic one(s)
+        candidates
+      }
+    } else {
+      candidates
     }
+    sigs.headOption.map(sig => sig.makeCall(args))
   }
+
   private def createFinder(name: String, functions: Seq[F]): FunctionFinder[F, T] = {
     new FunctionFinder[F, T](
       name,
@@ -237,10 +251,14 @@ class FunctionFinder[F <: SimpleExtension.Function, T](
     val singularInputType: Option[SingularArgumentMatcher[F]],
     val parent: FunctionConverter[F, T]) {
 
-  def attemptMatch(expression: Expression, operands: Seq[SExpression]): Option[T] = {
-    val opTypes = operands.map(_.getType)
+  def attemptMatch(expression: Expression, operands: Seq[FunctionArg]): Option[T] = {
     val outputType = ToSubstraitType.apply(expression.dataType, expression.nullable)
-    val opTypesStr = opTypes.map(t => t.accept(ToTypeString.INSTANCE))
+
+    val opTypesStr = operands.map {
+      case e: SExpression => e.getType.accept(ToTypeString.INSTANCE)
+      case t: Type => t.accept(ToTypeString.INSTANCE)
+      case _: EnumArg => "req"
+    }
 
     val possibleKeys =
       Util.crossProduct(opTypesStr.map(s => Seq(s))).map(list => list.mkString("_"))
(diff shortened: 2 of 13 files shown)


spark/src/main/scala/io/substrait/spark/expression/FunctionConverter.scala
```diff
@@ -293,14 +309,29 @@ class FunctionFinder[F <: SimpleExtension.Function, T](
...
           a
         } else {
           ExpressionCreator.cast(t, a, FailureBehavior.THROW_EXCEPTION)
         }
-      })
+      case other => other
+    }
+  }
+
+  private def validateOutputType(
```
Blizzara, line 324: I'd just get rid of these calls, maybe leave a TODO in place instead

spark/src/test/scala/io/substrait/spark/TPCDSPlan.scala
```diff
@@ -31,22 +30,10 @@ class TPCDSPlan extends TPCDSBase with SubstraitPlanTestBase {
     spark.conf.set("spark.sql.readSideCharPadding", "false")
   }
 
-  // spotless:off
-  val failingSQL: Set[String] = Set(
```
Blizzara, line 35: nice! 🎉

spark/src/main/scala/io/substrait/spark/expression/FunctionConverter.scala
```diff
@@ -237,10 +251,14 @@ class FunctionFinder[F <: SimpleExtension.Function, T](
...
 
-  def attemptMatch(expression: Expression, operands: Seq[SExpression]): Option[T] = {
-    val opTypes = operands.map(_.getType)
+  def attemptMatch(expression: Expression, operands: Seq[FunctionArg]): Option[T] = {
     val outputType = ToSubstraitType.apply(expression.dataType, expression.nullable)
-    val opTypesStr = opTypes.map(t => t.accept(ToTypeString.INSTANCE))
+
+    val opTypesStr = operands.map {
+      case e: SExpression => e.getType.accept(ToTypeString.INSTANCE)
+      case t: Type => t.accept(ToTypeString.INSTANCE)
```
Blizzara, line 259: do you know if we have any tests for this (t: Type) case?

spark/src/main/scala/io/substrait/spark/expression/FunctionMappings.scala
```diff
@@ -19,19 +19,90 @@ package io.substrait.spark.expression
...
+    // The following map to the Substrait `extract` function.
+    case Year(date) => Some(Seq(Enum("YEAR"), date))
+    case Quarter(date) => Some(Seq(Enum("QUARTER"), Enum("ONE"), date))
+    case Month(date) => Some(Seq(Enum("MONTH"), Enum("ONE"), date))
+    case DayOfMonth(date) => Some(Seq(Enum("DAY"), Enum("ONE"), date))
+    case _ => None
+  }
+
+  def unapply(name_args: (String, Seq[Expression])): Option[Expression] = name_args match {
+    case ("add:date_iday", Seq(startDate, i @ Literal(_, DayTimeIntervalType.DEFAULT))) =>
```
Blizzara, line 76: I don't think these are correct, the `add:date_iday` has a return type of `timestamp`, while Spark returns `date`. We should rather add new definitions, either into substrait itself or into spark.yml here, for `add:date_i32` -> `date`

spark/src/main/scala/io/substrait/spark/expression/Enum.scala
```diff
@@ -0,0 +1,32 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.substrait.spark.expression
+
+import org.apache.spark.sql.catalyst.expressions.{LeafExpression, Unevaluable}
+import org.apache.spark.sql.types.{DataType, NullType}
+
+case class Enum(value: String) extends LeafExpression with Unevaluable {
```
Blizzara, line 23: makes sense, mind adding a docstring to explain what this is for though?

spark/src/main/scala/io/substrait/spark/expression/ToSubstraitExpression.scala
```diff
@@ -196,7 +199,10 @@ abstract class ToSubstraitExpression extends HasOutputStack[Seq[Attribute]] {
       case InSet(value, set) => translateIn(value, set.toSeq.map(v => Literal(v)))
       case scalar @ ScalarFunction(children) =>
         Util
-          .seqToOption(children.map(translateUp))
+          .seqToOption(children.map {
+            case Enum(value) => Some(ImmutableEnumArg.builder.value(Optional.of(value)).build)
```
Blizzara, line 203: nit: if EnumArg doesn't have the builder on itself, you could add it there so that you could do just `EnumArg.builder...`

spark/src/main/scala/io/substrait/spark/expression/ToSparkExpression.scala
```diff
@@ -162,10 +163,11 @@ class ToSparkExpression(
   override def visit(expr: SExpression.Cast): Expression = {
     val childExp = expr.input().accept(this)
     val tt = ToSparkType.convert(expr.getType)
-    val tz = childExp.dataType match {
-      case DateType => Some(SQLConf.get.getConf(SQLConf.SESSION_LOCAL_TIMEZONE))
-      case _ => None
-    }
+    val tz =
+      if (Cast.needsTimeZone(childExp.dataType, tt))
```
Blizzara, line 167: nice!

Blizzara (commented):
Thanks! Makes sense for the extract parts, but for the DateAdd and DateSub I don't think the signature is correct (output type would be wrong). I think we should instead add the correct signatures.

Blizzara:
FWIW, in our fork I've done something similar, but maybe even more generic still (yeah, I should pull it upstream too..):
```
  trait ToSpark {
    def apply(name: String, args: Seq[Expression], options: Seq[FunctionOption]): Expression
  }

  trait ToSubstrait {
    def apply(expr: Expression)
    : (String /* name */, Seq[Expression] /* args */, Seq[FunctionOption] /* options */ )
  }

  case class Mapping(
      sparkClasses: Seq[Class[_]],
      substraitNames: Seq[String],
      toSpark: ToSpark,
      toSubstrait: ToSubstrait) {}

  /** Helper for creating simple 1-to-1 mappings */
  def s[T <: Expression: ClassTag](name: String): Mapping = {
    val builder = FunctionRegistryBase.build[T](name, None)._2
    s[T](name, builder, (expr: Expression) => expr.children)
  }

  /** Helper for creating reasonably simple 1-to-1 mappings with custom handling for arguments */
  def s[T <: Expression: ClassTag](
      name: String,
      builder: Seq[Expression] => T,
      args: T => Seq[Expression]): Mapping = {
    // For cases requiring custom arguments handling
    Mapping(
      Seq(scala.reflect.classTag[T].runtimeClass),
      Seq(name),
      (_, args, _) => builder(args),
      expr => (name, args(expr.asInstanceOf[T]), Seq.empty)
    )
  }
```
this then allows you to do both simple and more complex mappings logic:
```
    s[IsNaN]("is_nan"),
    Mapping(
      Seq(classOf[Add], classOf[DateAddInterval]),
      Seq("add"),
      (_, args, _) => {
        (args.head.dataType, args.last.dataType) match {
          case (DataTypes.DateType, DataTypes.CalendarIntervalType) => DateAddInterval(args.head, args.last)
          case _ => Add(args.head, args.last)
        }
      },
      {
        case Add(left, right, _) => ("add", Seq(left, right), Seq.empty)
        case DateAddInterval(left, right, _, _) => ("add", Seq(left, right), Seq.empty)
      }
    ),
```

spark/src/main/scala/io/substrait/spark/expression/FunctionMappings.scala
nielspardon: Shouldn't we try to ensure that as many of the default Substrait functions are mapped to Spark and trying to keep the Spark specific mappings as minimal as possible? Especially for the Substrait -> Spark direction?

Blizzara: When they match, yes, but in this case they don't. In Substrait you can give here an interval like 1d 12h, so if the date is let's say today 2025-04-09, the result should be 2025-04-10T12:00:00, while this in Spark would return 2025-04-10.

nielspardon: So we agree we would need to ensure that for this mapping the behavior in Spark matches the behavior in Substrait for the Substrait -> Spark direction?

Blizzara: > would need to ensure that for this mapping the behavior in Spark matches the behavior in Substrait for the Substrait -> Spark direction?

Yes, which is currently not the case.

> I do understand that when we talk about the other direction Spark -> Substrait that we should make sure we are not losing something that can be expressed in Spark when mapping to Substrait.

In this case the Spark expression is more limited (takes number of days, while Substrait takes an interval), so there is no losing. However the return type is wrong - the Spark expr returns `date`, while the Substrait definition claims a `timestamp`.

> I'm just a bit concerned if we keep adding custom extensions for the subtrait-spark mappings that we are losing some of the utility of having a common format for query plans independent of engines.

Yeah, that's tricky. It's a nice goal, and worth going for when possible, but it just isn't always doable (or at least not easy) since the engines support different things and in different ways.

Internally, we've found it most useful to map to Substrait extensions what maps nicely, and for the rest do Spark specific mappings (and then replicate those in other engines we want to use). Not saying that's what this repo should do necessarily.



One way to solve this could be to map add:date_iday into `Add(Cast(X as TimestampNTZ), Y)`, I guess that should match for types.

Blizzara (approved):

spark/src/main/scala/io/substrait/spark/expression/FunctionConverter.scala
andrew-coleman: Yes, sorry.  The `Type` case is there for completeness.  As far as I can see, there are no currently supported functions that use type arguments.  I suppose we could throw an unsupported exception in this case, if you'd prefer.

PR #380, opened 2025-04-09
feat(isthmus): update to calcite 1.39

mbwhite:
(replacement for PR #361 

As a result of an (accidental) upgrade Calcite 1.39; conversions via Isthmus started to report that table names where unknown;

Isthmus had an external API that allows tables to be lookup via calls to a function - dynamic or 'lazy' lookup. A special subclass of the CalciteSimpleSchema was created to support this. This special subclass had to be in the same package as the CalciteSimpleSchema as it took advantage of being able to override a number of protected methods.

Calcite 1.39.0 introduced similar support for 'lazy' table lookups; that implementation, though changed the protected methods sufficiently Isthmus now failed (indeed rebuilding the code fails to compile).

This PR removes the special subclass, and allows Ishtmus to work as before; with the work down to remove the external function based APIs there was only 1 location left where this special schema class was used.

This has been altered to use a new class that implements a Calcite Schema; it does it in a much simpler way than the previously


vbarua:
I have an alternative proposal for building CalciteSchemas in https://github.com/substrait-io/substrait-java/pull/391.

That one also handles nested schemas better, which the existing LookupCalciteSchema didn't really do.

vbarua:
> Or if you prefer to go straight to your change, any outlook on when it can be merged?

I would prefer to do this to avoid merging code which will then just be deleted or heavily modified. I've asked for a review in the Slack to hopefully move that PR along.

mbwhite:
Closing this as the work has been done in other PRs.

PR #381, opened 2025-04-09
feat(core): add convenience methods Plan.toProto() and Plan.toJsonString()

nielspardon:
I noticed that whenever we wanted to serialize a Plan to Proto one would have to instantiate a `PlanToProtoConverter` and call the `PlanToProtoConverter.toProto()` instead I think we can offer some convenience by adding a `Plan.toProto()` method which generates the proto from the current plan.

This PR also changes the existing code to use the new `Plan.toProto()` method. In that context I'm also fixing an issue with the Spark example code using an older static version of Substrait instead of the most recent one.

Similarly, generating the proto JSON representation is quite complicated since it requires people to first convert a `Plan` to proto and then set up a protobuf JsonFormat printer to generate the proto JSON String which also requires people to declare an additional dependency. Instead we can hide all of this complexity and simply have a `Plan.toJsonString()` method which generates the proto JSON for the current plan.

### DIFF
--- a/core/src/main/java/io/substrait/plan/Plan.java
+++ b/core/src/main/java/io/substrait/plan/Plan.java
@@ -1,5 +1,7 @@
 package io.substrait.plan;
 
+import com.google.protobuf.InvalidProtocolBufferException;
+import com.google.protobuf.util.JsonFormat;
 import io.substrait.proto.AdvancedExtension;
 import io.substrait.relation.Rel;
 import java.util.List;
@@ -27,6 +29,28 @@ public abstract class Plan {
 
     public static ImmutableRoot.Builder builder() {
       return ImmutableRoot.builder();
+    }
+  }
+
+  /**
+   * Serializes this plan as protobuf.
+   *
+   * @return this plan in protobuf format
+   */
+  public io.substrait.proto.Plan toProto() {
+    return new PlanProtoConverter().toProto(this);
+  }
+
+  /**
+   * Serializes this plan as a protobuf JSON string.
+   *
+   * @return this plan as a protobuf JSON string
+   */
+  public String toJsonString() {
+    try {
+      return JsonFormat.printer().includingDefaultValueFields().print(this.toProto());
+    } catch (InvalidProtocolBufferException e) {
+      throw new IllegalStateException("Can not generate JSON from proto.", e);
     }
   }
 }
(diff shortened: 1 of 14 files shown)


vbarua:
> I noticed that whenever we wanted to serialize a Plan to Proto one would have to instantiate a PlanToProtoConverter and call the PlanToProtoConverter.toProto() instead I think we can offer some convenience by adding a Plan.toProto() method which generates the proto from the current plan.

I would push against this actually. Specifically, in your `Plan#toProto` method you create a `new PlanProtoConverter()`. At the moment, you can't really customize the `PlanProtoConverter` with a different `RelProtoConverter`, but it's something that we should eventually support, in the same way that the `RelProtoConverter` supports providing custom Expression and Type converters when you extend it.

https://github.com/substrait-io/substrait-java/blob/b7abddd4f05d9d45a37f163f398c611f39228f87/core/src/main/java/io/substrait/relation/RelProtoConverter.java#L42-L62

Thinking about the `PlanProtoConverter` as something that is intended to be extended, I'm wary of hardcoding it like you have `Plan#toProto`.

> Similarly, generating the proto JSON representation is quite complicated since it requires people to first convert a Plan to proto and then set up a protobuf JsonFormat printer to generate the proto JSON String which also requires people to declare an additional dependency. Instead we can hide all of this complexity and simply have a Plan.toJsonString() method which generates the proto JSON for the current plan.

My preference would be to not include this as it requires bringing in an extra dependency. Additionally, there's also lots of different ways to configure the JsonFormat used when marshalling, which will depend on your preferences. It is convenient, but it's not too difficult for end users to configure their own utility class for it.

nielspardon:
I do understand your concerns, however JSON is the recommended text serialization format of the Substrait specification according to this: https://substrait.io/serialization/text_serialization/
in my opinion there should be a built-in (default) way in substrait-java/core to generate both the protobuf and the JSON serialization in a convenient way. that does not exclude that substrait-java/core should also allow for customization/extension.

vbarua:
> I do understand your concerns, however JSON is the recommended text serialization format of the Substrait specification

The JSON format described in that documentation does not exist yet, and is not the same as the JSON that arises from converting protobuf messages to JSON.

nielspardon:
> The JSON format described in that documentation does not exist yet, and is not the same as the JSON that arises from converting protobuf messages to JSON.

We can skip the JSON part then. I understand you have further concerns about the `toProto()` methods. We can close this PR and revisit the topic at a later point time when things become clearer.

core/src/main/java/io/substrait/plan/Plan.java
```diff
@@ -29,4 +31,26 @@ public static ImmutableRoot.Builder builder() {
...
   }
+
+  /**
+   * Serializes this plan as protobuf.
+   *
+   * @return this plan in protobuf format
+   */
+  public io.substrait.proto.Plan toProto() {
+    return new PlanProtoConverter().toProto(this);
+  }
```
vbarua, line 42: Thinking on this some more, it's fine to include this.

Issue #382, opened 2025-04-10
[isthmus] subqueries/set predicates with field references outside of the subquery fail


mbwhite:
For TPC-H query 17 there is an issue with scalar subqueries referencing fields from the parent query:

```
select
  sum(l.l_extendedprice) / 7.0 as avg_yearly
from
  "lineitem" l,
  "part" p
where
  p.p_partkey = l.l_partkey
  and p.p_brand = 'Brand#13'
  and p.p_container = 'JUMBO CAN'
  and l.l_quantity < (
    select
      0.2 * avg(l2.l_quantity)
    from
      "lineitem" l2
    where
# the following line throws an exception in Calcite
# java.lang.AssertionError: RexInputRef index 16 out of range 0..15
      l2.l_partkey = p.p_partkey
  )
```

The way this works in Calcite is that Calcite assigns a `correlation variable` to the filter relation `$cor0` and then it references the field `p.p_partkey` via this correlation variable `$cor0.p_partkey`.

Now normally with the Calcite RelBuilder one would assign a correlation variable to a relation in the following way:

```java
RelBuilder relBuilder;

final Holder<@Nullable RexCorrelVariable> v = Holder.empty();

relBuilder.scan(...).variable(v::set).filter(...  builder.field(v.get(), "p_partkey") ...).;
```

Now the challenge with substrait-java is that it has two distinct visitors: one for converting Substrait relations to Calcite relations [SubstraitRelNodeConverter.java](https://github.com/substrait-io/substrait-java/blob/main/isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java) and one for converting Substrait expression to Calcite expressions [ExpressionRexConverter.java](https://github.com/substrait-io/substrait-java/blob/main/isthmus/src/main/java/io/substrait/isthmus/expression/ExpressionRexConverter.java). The expression conversion has currently no link bank to the relation it belongs to or the inputs of the relation.

Today, in Substrait the `p_partkey` field is being described as an [`outer reference`](https://github.com/substrait-io/substrait/blob/4a472447392d9beba0401987b4d845079241c30d/proto/substrait/algebra.proto#L1537-L1543) field in the following snippet which shows the filter condition of the subquery. The first field is `l2.l_partkey` and the second is `p.p_partkey`:

```json
"condition": {
  "scalarFunction": {
    "functionReference": 1,
    "outputType": {
      "bool": {
        "nullability": "NULLABILITY_REQUIRED"
      }
[...]

Issue #379, opened 2025-04-07
(isthmus) Calcite and Substrait types inference mismatches


vbarua:
When we convert from Substrait to Calcite, there are a number of places where Substrait provides a type that Calcite _could_ infer. These include:

#### Field Reference Conversion
https://github.com/substrait-io/substrait-java/blob/867697c030410a0334800e9c107733078f9c4805/isthmus/src/main/java/io/substrait/isthmus/expression/ExpressionRexConverter.java#L496-L499

#### Scalar Function Conversion
https://github.com/substrait-io/substrait-java/blob/867697c030410a0334800e9c107733078f9c4805/isthmus/src/main/java/io/substrait/isthmus/expression/ExpressionRexConverter.java#L324-L325

#### Window Function Conversion
https://github.com/substrait-io/substrait-java/blob/867697c030410a0334800e9c107733078f9c4805/isthmus/src/main/java/io/substrait/isthmus/expression/ExpressionRexConverter.java#L346

#### Aggregate Function Conversion
https://github.com/substrait-io/substrait-java/blob/867697c030410a0334800e9c107733078f9c4805/isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java#L303

## The Problem
When the Substrait type aligns with the type that Calcite would infer, there are no issue many issues.

However, it's possible for the Substrait type to not match the Calcite type. In these cases, subtle issues can arise and/or straight-up exceptions.

These mismatches can arise for a number of reasons:
1. A Substrait function is mapped to a Calcite function with similar-ish inference, except for nullability handling. An example of this was seen in https://github.com/substrait-io/substrait-java/pull/180
2. A Substrait function is mapped to the wrong Calcite function. A potential example of this is noted in https://github.com/substrait-io/substrait-java/issues/377
3. The stored type in a Substrait [POJO FieldReference](https://github.com/substrait-io/substrait-java/blob/42b87ae885e809404204a166d2addb5e4780150a/core/src/main/java/io/substrait/expression/FieldReference.java#L17) has a different type than the underlying field in Calcite. An example of when this happens is in https://github.com/substrait-io/substrait-java/issues/185.

## Aiding Detection
[...]

PR #383, opened 2025-04-10
fix(isthmus): handle Subqueries/set predicates with field references outside of the subquery

mbwhite:
This is resolving issue #382 found with TPC-H 17, when converting back to SQL from Substrait
The subquery references fields in an outer scope, the calcite correlation variables where referenced by Subsrtrait's 'outer reference'

However when converting back to calcite from a plan with these 'outer references' they were ignored.

Notes::

- I've added two new TPC-x tests, these are similar the existing tests, but do conversions SQL to Substrait and Substrait to SQL
- Additional SQL using subqueries has been added; would like to add some more. Suggestions welcome. 
- Added these as 1xx.sql files but could move to a separate folder?

### DIFF
--- a/isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
@@ -1,6 +1,6 @@
 package io.substrait.isthmus;
 
-import static io.substrait.isthmus.SqlToSubstrait.EXTENSION_COLLECTION;
+import static io.substrait.isthmus.SqlConverterBase.EXTENSION_COLLECTION;
 
 import com.google.common.collect.ImmutableList;
 import io.substrait.expression.Expression;
@@ -44,6 +44,7 @@ import org.apache.calcite.rel.type.RelDataType;
 import org.apache.calcite.rel.type.RelDataTypeFactory;
 import org.apache.calcite.rel.type.RelDataTypeField;
 import org.apache.calcite.rex.RexBuilder;
+import org.apache.calcite.rex.RexCorrelVariable;
 import org.apache.calcite.rex.RexInputRef;
 import org.apache.calcite.rex.RexNode;
 import org.apache.calcite.rex.RexSlot;
@@ -51,6 +52,7 @@ import org.apache.calcite.sql.SqlAggFunction;
 import org.apache.calcite.sql.parser.SqlParser;
 import org.apache.calcite.tools.Frameworks;
 import org.apache.calcite.tools.RelBuilder;
+import org.apache.calcite.util.Holder;
 
 /**
  * RelVisitor to convert Substrait Rel plan to Calcite RelNode plan. Unsupported Rel node will call
@@ -139,8 +141,12 @@ public class SubstraitRelNodeConverter extends AbstractRelVisitor<RelNode, Runti
   @Override
   public RelNode visit(Filter filter) throws RuntimeException {
     RelNode input = filter.getInput().accept(this);
+    final Holder<RexCorrelVariable> v = Holder.empty();
+    expressionRexConverter.addCorrelVariable(v);
+
+    RelBuilder r1 = relBuilder.push(input).variable(v::set);
     RexNode filterCondition = filter.getCondition().accept(expressionRexConverter);
-    RelNode node = relBuilder.push(input).filter(filterCondition).build();
+    RelNode node = r1.filter(List.of(v.get().id), filterCondition).build();
     return applyRemap(node, filter.getRemap());
   }
 
--- a/isthmus/src/main/java/io/substrait/isthmus/expression/ExpressionRexConverter.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/expression/ExpressionRexConverter.java
@@ -22,8 +22,10 @@ import io.substrait.type.StringTypeVisitor;
 import io.substrait.type.Type;
 import io.substrait.util.DecimalUtil;
 import java.math.BigDecimal;
+import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
+import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.TimeUnit;
 import java.util.stream.Collectors;
@@ -34,6 +36,7 @@ import org.apache.calcite.rel.RelNode;
 import org.apache.calcite.rel.type.RelDataType;
 import org.apache.calcite.rel.type.RelDataTypeFactory;
 import org.apache.calcite.rex.RexBuilder;
+import org.apache.calcite.rex.RexCorrelVariable;
 import org.apache.calcite.rex.RexFieldCollation;
 import org.apache.calcite.rex.RexInputRef;
 import org.apache.calcite.rex.RexNode;
@@ -46,6 +49,7 @@ import org.apache.calcite.sql.SqlKind;
 import org.apache.calcite.sql.SqlOperator;
 import org.apache.calcite.sql.fun.SqlStdOperatorTable;
 import org.apache.calcite.sql.parser.SqlParserPos;
+import org.apache.calcite.util.Holder;
 import org.apache.calcite.util.TimeString;
 import org.apache.calcite.util.TimestampString;
 
@@ -468,7 +472,7 @@ public class ExpressionRexConverter extends AbstractExpressionVisitor<RexNode, R
     return RexSubQuery.in(rel, ImmutableList.copyOf(needles));
   }
 
-  static class ToRexWindowBound
+  static final class ToRexWindowBound
       implements WindowBound.WindowBoundVisitor<RexWindowBound, RuntimeException> {
 
     static RexWindowBound lowerBound(RexBuilder rexBuilder, WindowBound bound) {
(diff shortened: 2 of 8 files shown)


isthmus/src/test/resources/tpch/queries/101.sql
mbwhite, line 1: This is the wrong way to do this - will create a separate 'extras' directory of sql

isthmus/src/test/java/io/substrait/isthmus/TestTpcdsQuery.java
```diff
@@ -0,0 +1,85 @@
...
+
+    assumeFalse(fromSubstraitKnownFails.contains(query));
+    assumeTrue(allPlans.get(query).isPresent());
+
+    Optional<Plan> possible = allPlans.get(query);
+
+    io.substrait.plan.Plan plan = new ProtoPlanConverter().from(possible.get());
+    SubstraitToCalcite substraitToCalcite = new SubstraitToCalcite(extensions, typeFactory);
+    RelNode relRoot = substraitToCalcite.convert(plan.getRoots().get(0)).project(true);
+    System.out.println(SubstraitToSql.toSql(relRoot));
```
vbarua, line 83: What's the property that you're trying to verify here, that we can generated the SQL without exploding or the output SQL matches the input SQL?

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
```diff
@@ -136,11 +138,27 @@ public static RelNode convert(
...
 
+  public static final class RelationVariableTuple {
+    public Rel relation;
+    public Holder<RexCorrelVariable> v;
+
+    RelationVariableTuple(Rel relation, Holder<RexCorrelVariable> v) {
+      this.relation = relation;
+      this.v = v;
+    }
+  }
```
vbarua, line 149: I don't believe this is used anywhere in your PR? Do we need it for something?

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
```diff
@@ -136,11 +138,27 @@ public static RelNode convert(
...
+
   @Override
   public RelNode visit(Filter filter) throws RuntimeException {
     RelNode input = filter.getInput().accept(this);
+
+    final Holder<RexCorrelVariable> v = Holder.empty();
+    expressionRexConverter.addCorrelVariable(v);
```
vbarua, line 156: Are correlation variable something that can only arise from filters, or do we need to be able to handle them anywhere that we can include expressions?

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
```diff
@@ -136,11 +138,27 @@ public static RelNode convert(
...
+
+    RelBuilder r1 = relBuilder.push(input).variable(v::set);
     RexNode filterCondition = filter.getCondition().accept(expressionRexConverter);
```
vbarua, line 159: As part of https://github.com/substrait-io/substrait-java/issues/379 a coworker is looking into creating a new ExpressionRexConverter for each relation that we are converting in order to pass in the Calcite input schema. It sounds like correlation variable might also require something like this, as I assume they are unique per relation.

vbarua (commented):
Left some questions

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
mbwhite: Yes I believe that is the case; I'd like to get some more examples of these correlation variables

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
mbwhite: It was used in an earlier version; originally I suspected that both the correlation variable and the relation needed to passed; but current thinking is no;  

Will drop this to avoid confusions.

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
mbwhite: A good question I wondered the same thing; the calcite API is only referencing variables for filters.  and some projects.  

However I've not been able to create a SQL query to use anything other than them in filters.

Will keep an eye out for other cases.

isthmus/src/test/java/io/substrait/isthmus/TestTpcdsQuery.java
mbwhite: That we can generate the SQL without exploding.

mbwhite:
@vbarua questions have been answered; I've updated the test cases to better separate concerns. 

found the hidden setting in vscode to change the formatted back to spotless!

isthmus/src/test/java/io/substrait/isthmus/TestExtendedTpchQuery.java
```diff
@@ -0,0 +1,67 @@
...
+  public void extendedTpchFromSubstrait(int query) throws Exception {
+    assumeFalse(fromSubstraitKnownFails.contains(query));
+    assumeTrue(allPlans.containsKey(query));
+
+    Plan possible = allPlans.get(query);
+
+    io.substrait.plan.Plan plan = new ProtoPlanConverter().from(possible);
+    SubstraitToCalcite substraitToCalcite = new SubstraitToCalcite(extensions, typeFactory);
+    RelNode relRoot = substraitToCalcite.convert(plan.getRoots().get(0)).project(true);
+    System.out.println(SubstraitToSql.toSql(relRoot));
```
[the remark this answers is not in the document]
mbwhite: also changed the other existing tests where this occurred.

isthmus/src/main/java/io/substrait/isthmus/expression/ExpressionRexConverter.java
```diff
@@ -487,25 +492,44 @@ public RexNode visit(Expression.Cast expr) throws RuntimeException {
...
+
   @Override
   public RexNode visit(FieldReference expr) throws RuntimeException {
     if (expr.isSimpleRootReference()) {
+      Optional<Integer> outerref = expr.outerReferenceStepsOut();
       var segment = expr.segments().get(0);
+      if (outerref.isPresent()) {
+        if (segment instanceof FieldReference.StructField) {
+          FieldReference.StructField f = (FieldReference.StructField) segment;
+          var node = referenceRelList.get(outerref.get() - 1).get();
```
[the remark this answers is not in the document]
mbwhite: Implications is that each relation and it's immediate subexpression is the same namespace.

isthmus/src/test/java/io/substrait/isthmus/TestTpcdsQuery.java
```diff
@@ -0,0 +1,88 @@
...
+import org.junit.jupiter.api.MethodOrderer.OrderAnnotation;
+import org.junit.jupiter.api.Order;
+import org.junit.jupiter.api.TestInstance;
+import org.junit.jupiter.api.TestInstance.Lifecycle;
+import org.junit.jupiter.api.TestMethodOrder;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.ValueSource;
+
+/**
+ * Updated TPC-H test to convert SQL to Substrait and replay those plans back to SQL Validating that
```
vbarua, line 23: I assume you mean TPC-D here?

isthmus/src/test/java/io/substrait/isthmus/TestTpcdsQuery.java
```diff
@@ -0,0 +1,88 @@
...
+ * the conversions can operate without exceptions
+ */
+@TestMethodOrder(OrderAnnotation.class)
+@TestInstance(Lifecycle.PER_CLASS)
+public class TestTpcdsQuery extends PlanTestBase {
```
vbarua, line 28: We should add these tests separately from your PR if possible, as it would be reasonable to have as many of these that already work running while we figure out the outer reference issues.

isthmus/src/test/java/io/substrait/isthmus/TestTpchQuery.java
```diff
@@ -0,0 +1,69 @@
...
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.ValueSource;
+
+/**
+ * Updated TPC-H test to convert SQL to Substrait and replay those plans back to SQL Validating that
+ * the conversions can operate without exceptions
+ */
+@TestMethodOrder(OrderAnnotation.class)
+@TestInstance(Lifecycle.PER_CLASS)
+public class TestTpchQuery extends PlanTestBase {
```
vbarua, line 27: We should add these tests separately from your PR if possible, as it would be reasonable to have as many of these that already work running while we figure out the outer reference issues.

vbarua (commented):
There's a couple of things I'm not sure about in this PR.

When processing the Filter
```java
@Override
  public RelNode visit(Filter filter) throws RuntimeException {
    RelNode input = filter.getInput().accept(this);
    final Holder<RexCorrelVariable> v = Holder.empty();
    expressionRexConverter.addCorrelVariable(v);

    RelBuilder r1 = relBuilder.push(input).variable(v::set);
    RexNode filterCondition = filter.getCondition().accept(expressionRexConverter);
    RelNode node = r1.filter(List.of(v.get().id), filterCondition).build();
    return applyRemap(node, filter.getRemap());
  }
```
How does this work if there is more than 1 FieldReference to a variable outside the subquery? It's perfectly valid to construct a Substrait Filter like `OUTER_REF_1.foo = 'a' AND OUTER_REF_2.bar = bar` in Substrait, and I believe you can do something similar in SQL if you reference multiple fields from outside the subquery boundary.

Additionally, when constructing a Substrait subquery you could have multiple Filters. Your resolution logic uses
```java
var node = referenceRelList.get(outerref.get() - 1).get();
```
but we only ever append to `referenceRelList`. If multiple Filters have outer references with outer step 1, the first one processed would point to `referenceRelList[0]` which would be correct, but the second Filter processed would also point to `referenceRelList[0]`, which would be wrong.

What happens if you have multiple subquery boundaries and you have an outer reference more than 1 step out? For example if we have  2 subqueries and we only add 1 correlation variable to `referenceRelList`, then you would try to fetch `referenceRelList[1]`, which would throw a NullPointerException.

I'm not confident that these changes are able to handle outer references more generally beyond the specific `01.sql` test you've added.

What I would like to see is more testing, both from the SQL side but also from the Substrait side directly.

mbwhite:
Hi @vbarua - thanks for the additional thoughts; would share the concern that there might be cases such that this change isn't general enough.

Though I do admit I've found it hard to create SQL queries that use these references!

based on the last comments - is it worth doing that separate PR updating the testing and come back to this fix?

PR #384, opened 2025-04-10
fix(spark): use ImmutableFileFormat builders

andrew-coleman:
Round-trip testing was failing for query plans that involved reading files from the filesystem.
This was due to the conversion code instantiating objects directly rather than using the immutable builders. The round-trip object comparison was then failing. This commit fixes that by using the ImmutableFileFormat builders, and hence the generated equals() methods.

### DIFF
--- a/spark/src/main/scala/io/substrait/spark/logical/ToSubstraitRel.scala
+++ b/spark/src/main/scala/io/substrait/spark/logical/ToSubstraitRel.scala
@@ -38,7 +38,7 @@ import io.substrait.debug.TreePrinter
 import io.substrait.expression.{Expression => SExpression, ExpressionCreator}
 import io.substrait.extension.ExtensionCollector
 import io.substrait.hint.Hint
-import io.substrait.plan.{ImmutablePlan, ImmutableRoot, Plan}
+import io.substrait.plan.{ImmutableRoot, Plan}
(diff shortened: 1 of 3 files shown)


spark/src/main/scala/io/substrait/spark/logical/ToSubstraitRel.scala
```diff
@@ -440,20 +440,19 @@ class ToSubstraitRel extends AbstractLogicalPlanVisitor with Logging {
...
-            if (fsRelation.options.getOrElse("header", false) == false) 0 else 1
-          override def getEscape: String = fsRelation.options.getOrElse("escape", "\\")
-          override def getValueTreatedAsNull: Optional[String] =
-            Optional.ofNullable(fsRelation.options.get("nullValue").orNull)
-        }
-      case _: ParquetFileFormat => new FileFormat.ParquetReadOptions {}
-      case _: OrcFileFormat => new FileFormat.OrcReadOptions {}
+        // default values for options specified here:
+        // https://spark.apache.org/docs/latest/sql-data-sources-csv.html#data-source-option
+        ImmutableFileFormat.DelimiterSeparatedTextReadOptions
```
Blizzara, line 445: mind adding a builder() method on the `FileFormat` itself while you're at it?

Blizzara (approved):

PR #392, opened 2025-04-24
chore(isthmus): enable working TPC-DS query test

bestbeforetoday:
Other changes have allowed TPC-DS query 2 to be converted to Substrait without error. This query previously failed due to lack of support for rouding decimals.

This change enables the unit test for TPC-DS query 2.


Blizzara (approved):
Thanks!

PR #393, opened 2025-05-02
fix(isthmus): upgrade to Calcite 1.39.0

bestbeforetoday:
This version of Calcite improves support for nested window aggregation, which allows Isthmus to support several TPC-DS queries that were previously failing.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -15,7 +15,7 @@ com.github.vlsi.vlsi-release-plugins.version=1.74
 
 # library version
 antlr.version=4.13.1
-calcite.version=1.38.0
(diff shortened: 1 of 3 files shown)


vbarua:
The vulnerability in question is in https://mvnrepository.com/artifact/org.apache.httpcomponents.client5/httpclient5

Calcite 1.39 pulls in 5.4.1, but we should be able to bump it in our build to 5.4.4 which does not appear to have the vulnerability. 

We've done this before to deal with an vuln in Calcite 1.38 as well https://github.com/substrait-io/substrait-java/pull/329/files, and actually we should probably remove that workaround now.

PR #394, opened 2025-05-02
chore: update Gradle wrapper version to 8.14

bestbeforetoday:
Also remove an unused plugin dependency.


bestbeforetoday:
It might be a coincidence but updating the Gradle wrapper version seems to avoid — at least for now — the build errors I frequently saw due to Java source and build version mismatches.

vbarua (approved):
🧹 🔧

PR #397, opened 2025-05-02
chore: update OSV-Scanner to v2.0.2

bestbeforetoday:
Also add a Makefile to allow vulnerability scans of dependencies to be run locally more easily.

- `make scan` to scan all components.
- `make component=core scan` to scan only the core component.
- `make clean` to remove Gradle lockfiles used for scans.


Makefile
```diff
@@ -0,0 +1,20 @@
...
+	docker run --rm --volume './$(component)/gradle.lockfile:/gradle.lockfile' $(OSV_SCANNER_IMAGE) scan --lockfile /gradle.lockfile
+else
+	$(MAKE) component=core scan
+	$(MAKE) component=isthmus scan
+	$(MAKE) component=isthmus-cli scan
+endif
+
+.PHONY: clean
+clean:
+	find . -depth 2 -type f -name gradle.lockfile -delete -print
```
vbarua, line 20: I really like using Makefiles for stuff like this 🔧 

This makes me want to add (in the future) targets like `make build`, `make test` and `make format` to invoke Gradle as it makes it easier for new contributors to run the codebase for the first time. When we do that, I would also want to have make clean invoke `./gradlew clean`. Would that cause any issues for your work here?

vbarua (approved):
Changes look great! Did leave one question, mostly because I want to piggyback off of your work and add new targets later.

Makefile
bestbeforetoday: Not at all. It sounds like a great idea to me. If you wanted to keep the Makefile targets more granular as it is extended, the `clean` target in this change could be renamed to `scan-clean`, and the (new) `clean` target could depend on `scan-clean`.

PR #398, opened 2025-05-09
style: add project icon for idea

bvolpato:
Very nit, but adding an icon that will show up in JetBrains' toolbox and around the project name.

image

Inspiration from Trino https://github.com/trinodb/trino/blob/master/.idea/icon.png and Calcite https://github.com/apache/calcite/blob/main/.idea/icon.png shown in the screenshot :-)

image


vbarua (approved):
✨

PR #401, opened 2025-05-30
feat(isthmus): support for SQL TRIM function

### LINKED ISSUE #399: Support for SQL TRIM function in Isthmus (TPC-DS query 98)
Attempting to use Isthmus (v0.56.0) to convert TPC-DS query 98, which contains the SQL `trim(TRAILING from i_category)`, results in the following exception:

```
Unable to handle symbol: TRAILING
java.lang.UnsupportedOperationException: Unable to handle symbol: TRAILING
	at io.substrait.isthmus.expression.LiteralConverter.lambda$convert$0(LiteralConverter.java:135)
	at java.base/java.util.Optional.orElseThrow(Optional.java:408)
	at io.substrait.isthmus.expression.LiteralConverter.convert(LiteralConverter.java:134)
	at io.substrait.isthmus.expression.RexExpressionConverter.visitLiteral(RexExpressionConverter.java:99)
	at io.substrait.isthmus.expression.RexExpressionConverter.visitLiteral(RexExpressionConverter.java:33)
	at org.apache.calcite.rex.RexLiteral.accept(RexLiteral.java:1308)
	at io.substrait.isthmus.expression.RexExpressionConverter.lambda$callConversionFailureMessage$1(RexExpressionConverter.java:93)
	[...]
	at io.substrait.isthmus.expression.RexExpressionConverter.callConversionFailureMessage(RexExpressionConverter.java:94)
	at io.substrait.isthmus.expression.RexExpressionConverter.visitCall(RexExpressionConverter.java:85)
	at io.substrait.isthmus.expression.RexExpressionConverter.visitCall(RexExpressionConverter.java:33)
	at org.apache.calcite.rex.RexCall.accept(RexCall.java:208)
	at io.substrait.isthmus.SubstraitRelVisitor.toExpression(SubstraitRelVisitor.java:102)
	[...]
	at io.substrait.isthmus.SubstraitRelVisitor.visit(SubstraitRelVisitor.java:157)
	at io.substrait.isthmus.SubstraitRelVisitor.visit(SubstraitRelVisitor.java:45)
	at io.substrait.isthmus.RelNodeVisitor.reverseAccept(RelNodeVisitor.java:108)
	at io.substrait.isthmus.SubstraitRelVisitor.apply(SubstraitRelVisitor.java:374)
	at io.substrait.isthmus.SubstraitRelVisitor.convert(SubstraitRelVisitor.java:392)
	at io.substrait.isthmus.SubstraitRelVisitor.convert(SubstraitRelVisitor.java:384)
```

Calcite [specifies](https://calcite.apache.org/docs/reference.html) the TRIM function as:

```
TRIM( { BOTH | LEADING | TRAILING } [[ string1 ] FROM ] string2)
```

Given input SQL statements, the following Rex calls are produced:

- `TRIM(column_name)` -> `TRIM(FLAG(BOTH), ' ', $0)`
- `TRIM(BOTH FROM column_name)` -> `TRIM(FLAG(BOTH), ' ', $0)`
- `TRIM(LEADING FROM column_name)` -> `TRIM(FLAG(LEADING), ' ', $0)`
- `TRIM(TRAILING FROM column_name)` -> `TRIM(FLAG(TRAILING), ' ', $0)`

Substrait defines three functions that match these call variants, all of which require exactly two parameters:

1. [trim](https://github.com/substrait-io/substrait/blob/d01089bfba0353319915d90ab3bc4ff479e3b0eb/extensions/functions_string.yaml#L1342)
2. [ltrim](https://github.com/substrait-io/substrait/blob/d01089bfba0353319915d90ab3bc4ff479e3b0eb/extensions/functions_string.yaml#L1298)
3. [rtrim](https://github.com/substrait-io/substrait/blob/d01089bfba0353319915d90ab3bc4ff479e3b0eb/extensions/functions_string.yaml#L1320)

This conversion presents two challenges:

1. There is not a one-to-one mapping between Rex call operation and Substrait function name — depending on its first argument, Rex `TRIM` maps to Substrait `trim`, `rtrim` and `ltrim`.
2. The number of arguments to the Rex call (3) do not match the number parameters expected by the Substrait functions (2).

bestbeforetoday:
This enables support for TPC-DS query 98.

Closes #399

### DIFF
--- a/isthmus/src/main/java/io/substrait/isthmus/expression/EnumConverter.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/expression/EnumConverter.java
@@ -14,6 +14,7 @@ import org.apache.calcite.avatica.util.TimeUnitRange;
 import org.apache.calcite.rex.RexBuilder;
 import org.apache.calcite.rex.RexLiteral;
 import org.apache.calcite.rex.RexNode;
+import org.apache.calcite.sql.fun.SqlTrimFunction.Flag;
 import org.apache.calcite.sql.type.SqlTypeName;
 
 /**
@@ -52,12 +53,21 @@ public class EnumConverter {
     calciteEnumMap.put(
         argAnchor(DefaultExtensionCatalog.FUNCTIONS_DATETIME, "extract:req_time", 0),
         TimeUnitRange.class);
+
+    calciteEnumMap.put(
+        argAnchor(DefaultExtensionCatalog.FUNCTIONS_STRING, "trim:vchar_vchar", 0), Flag.class);
+    calciteEnumMap.put(
+        argAnchor(DefaultExtensionCatalog.FUNCTIONS_STRING, "ltrim:vchar_vchar", 0), Flag.class);
+    calciteEnumMap.put(
+        argAnchor(DefaultExtensionCatalog.FUNCTIONS_STRING, "rtrim:vchar_vchar", 0), Flag.class);
   }
 
   private static Optional<Enum<?>> constructValue(
       Class<? extends Enum<?>> cls, Supplier<Optional<String>> option) {
     if (cls.isAssignableFrom(TimeUnitRange.class)) {
       return option.get().map(TimeUnitRange::valueOf);
+    } else if (cls.isAssignableFrom(Flag.class)) {
+      return option.get().map(Flag::valueOf);
     } else {
       return Optional.empty();
     }
--- a/isthmus/src/main/java/io/substrait/isthmus/expression/EnumConverter.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/expression/EnumConverter.java
@@ -55,17 +55,23 @@ public class EnumConverter {
         TimeUnitRange.class);
 
     calciteEnumMap.put(
-        argAnchor(DefaultExtensionCatalog.FUNCTIONS_STRING, "trim:vchar_vchar", 0), SqlTrimFunction.Flag.class);
+        argAnchor(DefaultExtensionCatalog.FUNCTIONS_STRING, "trim:vchar_vchar", 0),
+        SqlTrimFunction.Flag.class);
     calciteEnumMap.put(
-        argAnchor(DefaultExtensionCatalog.FUNCTIONS_STRING, "trim:str_str", 0), SqlTrimFunction.Flag.class);
+        argAnchor(DefaultExtensionCatalog.FUNCTIONS_STRING, "trim:str_str", 0),
+        SqlTrimFunction.Flag.class);
     calciteEnumMap.put(
-        argAnchor(DefaultExtensionCatalog.FUNCTIONS_STRING, "ltrim:vchar_vchar", 0), SqlTrimFunction.Flag.class);
+        argAnchor(DefaultExtensionCatalog.FUNCTIONS_STRING, "ltrim:vchar_vchar", 0),
+        SqlTrimFunction.Flag.class);
     calciteEnumMap.put(
-        argAnchor(DefaultExtensionCatalog.FUNCTIONS_STRING, "ltrim:str_str", 0), SqlTrimFunction.Flag.class);
+        argAnchor(DefaultExtensionCatalog.FUNCTIONS_STRING, "ltrim:str_str", 0),
+        SqlTrimFunction.Flag.class);
     calciteEnumMap.put(
-        argAnchor(DefaultExtensionCatalog.FUNCTIONS_STRING, "rtrim:vchar_vchar", 0), SqlTrimFunction.Flag.class);
+        argAnchor(DefaultExtensionCatalog.FUNCTIONS_STRING, "rtrim:vchar_vchar", 0),
+        SqlTrimFunction.Flag.class);
     calciteEnumMap.put(
-        argAnchor(DefaultExtensionCatalog.FUNCTIONS_STRING, "rtrim:str_str", 0), SqlTrimFunction.Flag.class);
+        argAnchor(DefaultExtensionCatalog.FUNCTIONS_STRING, "rtrim:str_str", 0),
+        SqlTrimFunction.Flag.class);
   }
 
   private static Optional<Enum<?>> constructValue(
@@ -139,9 +145,7 @@ public class EnumConverter {
   }
 
   static boolean isEnumValue(RexNode value) {
-    return value != null
-        && (value instanceof RexLiteral)
-        && value.getType().getSqlTypeName() == SqlTypeName.SYMBOL;
+    return value instanceof RexLiteral && value.getType().getSqlTypeName() == SqlTypeName.SYMBOL;
   }
 
   private static class ArgAnchor {
--- a/isthmus/src/main/java/io/substrait/isthmus/expression/EnumConverter.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/expression/EnumConverter.java
@@ -78,11 +78,13 @@ public class EnumConverter {
       Class<? extends Enum<?>> cls, Supplier<Optional<String>> option) {
     if (cls.isAssignableFrom(TimeUnitRange.class)) {
       return option.get().map(TimeUnitRange::valueOf);
-    } else if (cls.isAssignableFrom(SqlTrimFunction.Flag.class)) {
-      return option.get().map(SqlTrimFunction.Flag::valueOf);
-    } else {
-      return Optional.empty();
     }
+
+    if (cls.isAssignableFrom(SqlTrimFunction.Flag.class)) {
+      return option.get().map(SqlTrimFunction.Flag::valueOf);
+    }
+
+    return Optional.empty();
   }
 
   static Optional<RexLiteral> toRex(
--- a/isthmus/src/main/java/io/substrait/isthmus/expression/EnumConverter.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/expression/EnumConverter.java
@@ -14,7 +14,7 @@ import org.apache.calcite.avatica.util.TimeUnitRange;
 import org.apache.calcite.rex.RexBuilder;
 import org.apache.calcite.rex.RexLiteral;
 import org.apache.calcite.rex.RexNode;
-import org.apache.calcite.sql.fun.SqlTrimFunction.Flag;
+import org.apache.calcite.sql.fun.SqlTrimFunction;
 import org.apache.calcite.sql.type.SqlTypeName;
 
 /**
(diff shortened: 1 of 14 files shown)


bestbeforetoday:
@vbarua The only tests failing in this pull request are new ones I added for all the different permutations of the string trim function. The actually all successfully convert from SQL to Substrait and back again, but fail because somewhere in the round-trip there is some inconsistency to the types or type casts used for different string fields. I do not believe this is anything I have introduced in this change.

Can you help me spot whether this can be easily fixed so this pull request is ready in its current state? If not, I propose loosening my new unit tests to allow them to pass with the current behaviour.

Note that TPC-DS query 98 now passes (or rather no longer fails) in both directions, which was the main objective of this change.

isthmus/src/main/java/io/substrait/isthmus/expression/FunctionMappings.java
```diff
@@ -11,111 +11,104 @@
...
+              s(SqlStdOperatorTable.CONCAT, "concat"),
+              s(SqlStdOperatorTable.CHAR_LENGTH, "char_length"),
+              s(SqlStdOperatorTable.LOWER, "lower"),
+              s(SqlStdOperatorTable.UPPER, "upper"),
+              s(SqlStdOperatorTable.BETWEEN),
+              s(SqlStdOperatorTable.IS_NOT_DISTINCT_FROM, "is_not_distinct_from"),
+              s(SqlStdOperatorTable.COALESCE, "coalesce"),
+              s(SqlStdOperatorTable.TRIM, "trim"),
+              s(SqlStdOperatorTable.TRIM, "ltrim"),
+              s(SqlStdOperatorTable.TRIM, "rtrim"))
```
bestbeforetoday, line 66: These lines are the only actual change in this file. The rest is just slightly restructuring the existing code.

isthmus/src/main/java/io/substrait/isthmus/expression/EnumConverter.java
```diff
@@ -52,12 +53,21 @@ public class EnumConverter {
     calciteEnumMap.put(
         argAnchor(DefaultExtensionCatalog.FUNCTIONS_DATETIME, "extract:req_time", 0),
         TimeUnitRange.class);
+
+    calciteEnumMap.put(
+        argAnchor(DefaultExtensionCatalog.FUNCTIONS_STRING, "trim:vchar_vchar", 0), Flag.class);
+    calciteEnumMap.put(
+        argAnchor(DefaultExtensionCatalog.FUNCTIONS_STRING, "ltrim:vchar_vchar", 0), Flag.class);
+    calciteEnumMap.put(
+        argAnchor(DefaultExtensionCatalog.FUNCTIONS_STRING, "rtrim:vchar_vchar", 0), Flag.class);
```
vbarua: Should we also add the `str_str` variants?

isthmus/src/main/java/io/substrait/isthmus/expression/EnumConverter.java
```diff
@@ -52,12 +53,21 @@ public class EnumConverter {
     calciteEnumMap.put(
         argAnchor(DefaultExtensionCatalog.FUNCTIONS_DATETIME, "extract:req_time", 0),
         TimeUnitRange.class);
+
+    calciteEnumMap.put(
+        argAnchor(DefaultExtensionCatalog.FUNCTIONS_STRING, "trim:vchar_vchar", 0), Flag.class);
```
vbarua: minor: Instead of `Flag.class` can we use `SqlTrimFunction.Flag.class` to make it clearer that `Flag` is specific to `SqlTrimFunction`.

isthmus/src/main/java/io/substrait/isthmus/expression/ScalarFunctionConverter.java
```diff
@@ -34,6 +75,18 @@ public ScalarFunctionConverter(
...
       TypeConverter typeConverter) {
     super(functions, additionalSignatures, typeFactory, typeConverter);
+
+    var trims = new HashMap<Trim, ScalarFunctionVariant>();
+    for (var t : Trim.values()) {
+      var function = findFunction(t.getSubstraitName(), functions);
+      function.ifPresent(f -> trims.put(t, f));
+    }
+    trimFunctions = Collections.unmodifiableMap(trims);
+  }
```
vbarua, line 47: I think we should leave the ScalarFunctionConverter for functions that don't require this kind of logic, and pull the Trim handling code into it's own CallConverter like we do for other things like `FieldSelectionConverter` and `CallConverters.CAST`.

See https://github.com/substrait-io/substrait-java/blob/6bb121a386c3e1377c737cc0dc66b6ae90e71322/isthmus/src/main/java/io/substrait/isthmus/expression/CallConverters.java#L136-L143

isthmus/src/main/java/io/substrait/isthmus/expression/ScalarFunctionConverter.java
```diff
@@ -34,6 +75,18 @@ public ScalarFunctionConverter(
...
+
+  private Optional<ScalarFunctionVariant> findFunction(
+      String name, Collection<ScalarFunctionVariant> functions) {
+    return functions.stream().filter(f -> name.equals(f.name().toLowerCase(Locale.ROOT))).findAny();
```
vbarua: IMO we shouldn't lowercase names here because they are case-sensitive according the spec
> One or more user-friendly, case-sensitive UTF-8 strings that are used to reference this function.

See: https://substrait.io/expressions/scalar_functions/

isthmus/src/main/java/io/substrait/isthmus/expression/ScalarFunctionConverter.java
```diff
@@ -44,11 +97,59 @@ protected ImmutableList<FunctionMappings.Sig> getSigs() {
   @Override
   public Optional<Expression> convert(
       RexCall call, Function<RexNode, Expression> topLevelConverter) {
+    return customConvert(call, topLevelConverter)
+        .or(() -> standardConvert(call, topLevelConverter));
```
vbarua: This definitely makes me want to split this out into it's own converter, because this is effectively what the CallConverters do (chain conversion until one returns an Expression).

isthmus/src/test/java/io/substrait/isthmus/StringFunctionTest.java
```diff
@@ -4,7 +4,7 @@
 import org.junit.jupiter.params.ParameterizedTest;
 import org.junit.jupiter.params.provider.ValueSource;
 
-public class StringFunctionTest extends PlanTestBase {
+public final class StringFunctionTest extends PlanTestBase {
 
   static List<String> CREATES = List.of("CREATE TABLE strings (c16 CHAR(16), vc32 VARCHAR(32))");
```
vbarua: It would be good to test the `trim` functional against a column of type VARCHAR, which maps the Substrait `str` type. So something like
```sql
CREATE TABLE strings (c16 CHAR(16), vc32 VARCHAR(32), str VARCHAR)
```

isthmus/src/main/java/io/substrait/isthmus/expression/ScalarFunctionConverter.java
bestbeforetoday: I just did this because some of the existing code does. It should not be necessary here so I will remove the lower-casing as you suggest.

isthmus/src/main/java/io/substrait/isthmus/expression/ScalarFunctionConverter.java
bestbeforetoday: The trim handling still needs to use all the rest of the logic from the ScalarFunctionConverter. I want to avoid a cut & paste duplicate of the existing code just for a special case of scalar function conversion. I could implement a new converter, either as a subclass or using composition, to reuse all the ScalarFunctionConverter implementation instead if you prefer.

vbarua: > I could implement a new converter, either as a subclass or using composition,

I think that would be good. I can see this class starting to grow as we start processing other enum functions, and it could get unwieldy pretty fast.

bestbeforetoday:
I have:

- Ensured conversion to both variants (str and vchar) of the Substrait trim functions is successful.
- Factored our the trim handling to a separate class outside of ScalarFunctionConverter.

It was not practical to split the converter implementation further without rewriting the underlying function converter since:

1. The implementation is highly coupled with collaborating non-static internal nested classes.
2. There are several direct consumers of the ScalarFunctionConverter, which expect it to convert all scalar functions, including special cases such as trim.
3. The ScalarFunctionConverter is exposed in the public API so significant rework would be a breaking change.

I have not fixed the issue where casts are added during the Rex-to-Substrait conversion. Although this is a real issue that needs to be addressed, I do not believe it is caused by this change, and the PR is already bigger than I would like. Perhaps this change can be merged and the casting issue can be addressed later? I can raise an issue to describe the behaviour and your pointers to the problem area.

isthmus/src/main/java/io/substrait/isthmus/expression/SubstraitFunctionMapping.java
```diff
@@ -0,0 +1,33 @@
+package io.substrait.isthmus.expression;
+
+import io.substrait.extension.SimpleExtension.ScalarFunctionVariant;
+import java.util.List;
+import org.apache.calcite.rex.RexNode;
+
+class SubstraitFunctionMapping {
+
+  private final String name;
```
vbarua: What name is this supposed to correspond to? A substrait function name? A more specific name and/or light documentation would be helpful.

isthmus/src/main/java/io/substrait/isthmus/expression/SubstraitFunctionMapping.java
```diff
@@ -0,0 +1,33 @@
+package io.substrait.isthmus.expression;
+
+import io.substrait.extension.SimpleExtension.ScalarFunctionVariant;
+import java.util.List;
+import org.apache.calcite.rex.RexNode;
+
+class SubstraitFunctionMapping {
```
vbarua, line 11: Some light documentation about what this represents would be useful. It looks like it can represent multiple possible mappings?

isthmus/src/main/java/io/substrait/isthmus/expression/ScalarFunctionMapper.java
```diff
@@ -0,0 +1,13 @@
...
+import io.substrait.expression.Expression;
+import io.substrait.expression.FunctionArg;
+import java.util.List;
+import java.util.Optional;
+import org.apache.calcite.rex.RexCall;
+
+interface ScalarFunctionMapper {
+  Optional<SubstraitFunctionMapping> toSubstrait(RexCall call);
+
+  Optional<List<FunctionArg>> getRexArguments(Expression.ScalarFunctionInvocation expression);
```
vbarua: minor: `Rex` is the Calcite noun for "row-expression" so it's a little overloaded. What do you think about `getSubstraitArguments` instead to make it clear this is going from Substrait -> Substrait.

vbarua (commented):
I'm having a hard time figuring out how this all pieces together. Could you add some docs to the various things you've added to help with the review, and then future users.

Somewhat related. I think the way we're doing function conversion generally is way to convoluted. This was already pretty hard to understand, IMO, and the more indirection and special cases we add that harrder it gets to reason about.

At this point, I'm more inclined than not to merge this in, with the understanding that we may need to rework all function handling in the future.

isthmus/src/main/java/io/substrait/isthmus/expression/SubstraitFunctionMapping.java
bestbeforetoday: The containing class is called SubstraitFunctionMapping, which should give some context to the contained variables. I don't want to duplicate too much with the parent scope, but I'll change it to `substraitName` to improve clarity.

isthmus/src/main/java/io/substrait/isthmus/expression/ScalarFunctionMapper.java
bestbeforetoday: Good catch. This code was working with Calcite objects in an earlier iteration. The name should change to reflect the current implementation.

isthmus/src/main/java/io/substrait/isthmus/expression/TrimFunctionMapper.java
```diff
@@ -0,0 +1,136 @@
...
+ * operand indicates the Substrait function to which it should be mapped. The first operand is then
+ * omitted from the arguments supplied to the Substrait function.
+ *
+ * <ul>
+ *   <li>TRIM('BOTH', characters, string) -> trim(characters, string)
+ *   <li>TRIM('LEADING', characters, string) -> ltrim(characters, string)
+ *   <li>TRIM('TRAILING', .characters, string) -> rtrim(characters, string)
+ * </ul>
+ */
+final class TrimFunctionMapper implements ScalarFunctionMapper {
```
vbarua, line 33: We should make this public to allow users to create instances of it.

isthmus/src/main/java/io/substrait/isthmus/expression/ScalarFunctionConverter.java
```diff
@@ -34,6 +42,8 @@ public ScalarFunctionConverter(
       RelDataTypeFactory typeFactory,
       TypeConverter typeConverter) {
```
vbarua, line 43: What do you think about making the list of mappers an argument here, to allow users to provide their own mappers?

We can then move the below mapper initialization code the the default constructor above. The aim with these constructors is to have the boring constructor with sane default, and then the fully configurable constructor.

vbarua (commented):
One minor comment about the construction of the ScalaFunctionMapper. My aim is to merge this PR in this week.

isthmus/src/main/java/io/substrait/isthmus/expression/TrimFunctionMapper.java
bestbeforetoday: Why do we want this in the public API? It is an implementation detail of the ScalarFunctionConverter, which is public. There is no useful way for client code to use TrimFunctionMapper (or ScalarFunctionMapper) in isolation.

By making it public we just cause more significant breaking changes when function conversion is refactored.

isthmus/src/main/java/io/substrait/isthmus/expression/ScalarFunctionConverter.java
bestbeforetoday: I think that might be better for a refactor of function mapping in a separate PR where no new functionality is being added.

isthmus/src/main/java/io/substrait/isthmus/expression/TrimFunctionMapper.java
vbarua: > we just cause more significant breaking changes when function conversion is refactored.

Good point, let's hold off for now.

PR #402, opened 2025-06-03
feat: add new join types

nielspardon:
I noticed the Substrait specification has a couple more join types defined which were not supported yet by the substrait-java/core library. This PR adds support for these additional join types to substrait-java/core.

Additionally, it deprecates the old `JoinType` enum names `SEMI` and `ANTI` in order to be consistent with the Substrait specification names `LEFT_SEMI` and `LEFT_ANTI`.

This PR does neither add support for the new join types to the Isthmus mappings nor the Spark mappings.

### DIFF
--- a/core/src/main/java/io/substrait/relation/Join.java
+++ b/core/src/main/java/io/substrait/relation/Join.java
@@ -23,7 +23,20 @@ public abstract class Join extends BiRel implements HasExtension {
     OUTER(JoinRel.JoinType.JOIN_TYPE_OUTER),
     LEFT(JoinRel.JoinType.JOIN_TYPE_LEFT),
     RIGHT(JoinRel.JoinType.JOIN_TYPE_RIGHT),
+    LEFT_SEMI(JoinRel.JoinType.JOIN_TYPE_LEFT_SEMI),
+    LEFT_ANTI(JoinRel.JoinType.JOIN_TYPE_LEFT_ANTI),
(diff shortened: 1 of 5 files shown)


core/src/main/java/io/substrait/relation/Join.java
```diff
@@ -51,15 +64,26 @@ public static JoinType fromProto(JoinRel.JoinType proto) {
...
+          case LEFT, OUTER, LEFT_SINGLE -> getRight().getRecordType().fields().stream()
               .map(TypeCreator::asNullable);
-          case SEMI, ANTI -> Stream.of(); // these are left joins which ignore right side columns
+          case SEMI, ANTI, LEFT_SEMI, LEFT_ANTI -> Stream
+              .of(); // these are left joins which ignore right side columns
+          case LEFT_MARK -> Stream.of(
+              TypeCreator.REQUIRED
+                  .BOOLEAN); // left mark join keeps all fields from left and adds a boolean mark
+            // field
           default -> getRight().getRecordType().fields().stream();
```
vbarua, line 87: The output type derivations you've added make sense to me, but it's also made me realize that we should probably include reference examples for the various join types in the spec to make it easier for others to implement the correct behaviour.

Nothing for you to do here, but I've made in issue to track this: https://github.com/substrait-io/substrait/issues/818

vbarua (approved):
Changes look good to me. Thanks for updating these.

PR #405, opened 2025-06-04
build: update Gradle and plugin versions

bestbeforetoday:
Update:

- Gradle to 8.14.1
- io.github.gradle-nexus.publish-plugin to 2.0.0
- com.github.johnrengelman.shadow to com.gradleup.shadow 8.3.6

This avoids warnings about deprecated Gradle features being used in the build scripts.


build.gradle.kts
```diff
@@ -8,7 +8,7 @@ plugins {
   id("idea")
   id("com.github.vlsi.gradle-extensions") version "1.74"
   id("com.diffplug.spotless") version "6.19.0"
-  id("io.github.gradle-nexus.publish-plugin") version "1.1.0"
+  id("io.github.gradle-nexus.publish-plugin") version "2.0.0"
```
vbarua, line 11: It looks like 2.0.0 has deprecated, but not yet removed, support for Java 8.
https://github.com/gradle-nexus/publish-plugin/releases/tag/v2.0.0

If this prevents us from building for Java 8 we may need to revert this, but I'm happy to try it for now.

core/build.gradle.kts
```diff
@@ -7,7 +7,7 @@ plugins {
   id("antlr")
   id("com.google.protobuf") version "0.9.4"
   id("com.diffplug.spotless") version "6.19.0"
-  id("com.github.johnrengelman.shadow") version "8.1.1"
+  id("com.gradleup.shadow") version "8.3.6"
```
vbarua, line 10: TIL there's a new owner for this https://github.com/GradleUp/shadow?tab=readme-ov-file#shadow

vbarua (approved):
Changes look reasonable to me. Thanks for updating this. We'll need to keep an eye out to make sure we can build for Java 8.

build.gradle.kts
bestbeforetoday: We are running Gradle with Java 17 so I don't expect this to be an issue

PR #406, opened 2025-06-04
feat(core): set Substrait version and producer

nielspardon:
This PR sets the Substrait specification version in the Plan by default which is supposed to be required since Substrait 0.17.0 according to [this](https://github.com/substrait-io/substrait/blob/d430e521f203aec6a4e06731d4bfd68cdf61f443/proto/substrait/plan.proto#L27-L29).

1. The Substrait specification version is being retrieved from the `substrait` Git submodule by running `git describe --tags` via Gradle.

2. The Substrait specification version is then written into a `MANIFEST.MF` file together with the substrait-java version following the Java manifest conventions:

```
Manifest-Version: 1.0
Implementation-Title: substrait-java
Implementation-Version: 0.57.0
Specification-Title: substrait
Specification-Version: 0.69.0

```

3. During runtime the specification version is either retrieved:
  - via the JVM built-in mechanism `Version.class.getPackage().getSpecificationVersion()` which is only available when the `Version` class is being used from a JAR containing the `MANIFEST.MF`
  - or we are loading the `MANIFEST.MF` file from the classpath as a resource if we are using the `Version` class from within the Gradle build of substrait-java since then the class is not part of a JAR.
  
In order to make the unit tests work I simplified the `SqlToSubstrait.executeInner()` method by first creating a Substrait Java SDK `Plan` which then gets converted to a proto Plan instead of creating the proto Plan directly. Otherwise the proto Plan would not have contained the default version and producer values and some of the round tripping tests failed.

Finally, I'm adding one more entry to gitignore which is created by VSCode's Scala tooling.


.github/workflows/pr.yml
```diff
@@ -69,7 +69,11 @@ jobs:
       - name: Setup Gradle
         uses: gradle/actions/setup-gradle@v4
       - name: Build with Gradle
-        run: gradle build --rerun-tasks
+        run: |
+          # fetch submodule tags since actions/checkout@v4 does not
+          git submodule foreach 'git fetch --unshallow || true'
```
bestbeforetoday, line 74: I haven't tried it but I wonder if the `fetch-tags: true` option for actions/checkout would achieve similar results?

nielspardon: no, I tried. it only fetches tags for the parent repo and not for git submodules. there's an open issue for the checkout github action: https://github.com/actions/checkout/issues/435

isthmus/src/main/java/io/substrait/isthmus/SqlToSubstrait.java
```diff
@@ -56,22 +55,16 @@ List<RelRoot> sqlToRelNode(String sql, List<String> tables) throws SqlParseExcep
...
-            });
-    functionCollector.addExtensionsToPlan(plan);
-    return plan.build();
+    sqlToRelNode(sql, validator, catalogReader).stream()
+        .map(root -> SubstraitRelVisitor.convert(root, EXTENSION_COLLECTION, featureBoard))
+        .forEach(root -> builder.addRoots(root));
+
+    PlanProtoConverter planToProto = new PlanProtoConverter();
+
+    return planToProto.toProto(builder.build());
```
vbarua, line 71: Good call on wiring this through the POJOs

core/src/main/java/io/substrait/plan/Plan.java
```diff
@@ -19,6 +27,58 @@ public static ImmutablePlan.Builder builder() {
     return ImmutablePlan.builder();
   }
 
+  @Value.Immutable
+  public abstract static class Version {
```
vbarua, line 31: One small thing I would suggest here is splitting the general concept of the Version POJO from the substrait-java specific Version that we want to create. That is, other java based systems might use the POJOs to manipulate Substrait plans and might want to set their own versions, which gets a bit muddied if we include defaults in the definition IMO.

What do you think about something like:
```java
  @Value.Immutable
  public abstract static class Version {
    abstract int getMajor();

    abstract int getMinor();

    abstract int getPatch();

    abstract Optional<String> getGitHash();

    abstract Optional<String> getProducer();

    public static ImmutableVersion.Builder builder() {
      return ImmutableVersion.builder();
    }
  }

  public static final Version SUBSTRAIT_JAVA;

  static {
    final String[] VERSION_COMPONENTS = loadVersion();
      SUBSTRAIT_JAVA =
        Version.builder()
            .minor(Integer.parseInt(VERSION_COMPONENTS[1]))
            .patch(Integer.parseInt(VERSION_COMPONENTS[2]))
            .producer("substrait-java")
            .build();
  }
```

which keeps the POJO object generic in that it models _any_ Version, but we provide a pre-populated Version for folks that want to use it?

core/src/main/java/io/substrait/plan/Plan.java
```diff
@@ -19,6 +27,58 @@ public static ImmutablePlan.Builder builder() {
...
+    @SuppressWarnings("immutables:untype")
+    @Value.Default
+    public Optional<String> getGitHash() {
+      return Optional.ofNullable(null);
+    }
+
+    @SuppressWarnings("immutables:untype")
+    @Value.Default
+    public Optional<String> getProducer() {
+      return Optional.of("substrait-java");
```
vbarua: meta/philosophical: Should we actually consider `substrait-java` a producer, or is it a library _used_ by producers to generate plans?

Another way of thinking about it. The plans created by `SqlToSubstrait` are produced by `isthmus` using the substrait-java library, or at least that's how I think about it.

core/src/main/java/io/substrait/plan/ProtoPlanConverter.java
```diff
@@ -39,11 +39,29 @@ public Plan from(io.substrait.proto.Plan plan) {
...
     }
+
+    ImmutableVersion.Builder versionBuilder =
+        ImmutableVersion.builder()
+            .major(plan.getVersion().getMajorNumber())
+            .minor(plan.getVersion().getMinorNumber())
+            .patch(plan.getVersion().getPatchNumber());
+
+    // protobuf field 'git_hash' is an empty string by default
+    if (!"".equals(plan.getVersion().getGitHash())) {
```
vbarua: minor: I think we can simplify this to
```java
if (!plan.getVersion().getGitHash().isEmpty()) ...
```

which is slightly easier to read IMO. Same below.

core/src/main/java/io/substrait/plan/Plan.java
nielspardon: yeah, I also was a bit uncertain when writing this part. I thought that having a default might be good in case somebody would use the POJO classes in core directly. then we would set it to `isthmus` and `substrait-spark` in the respective submodules. the other thought I had was whether the producer string should also contain the producer version.

the producer field would anyway be optional so I guess we could also not set it in `core` and only set it in `isthmus` or `spark`

core/src/main/java/io/substrait/plan/Plan.java
nielspardon: we can change that sure. I did set a default mainly because the `substrait-validator` complains if it's not set at all so I thought it makes sense to set it by default in any case. the way it's implemented with `@Value.Default` would still allow users to override the values. instead of doing it for the individual version components we can also do it at the full version object level.

core/src/main/java/io/substrait/plan/ProtoPlanConverter.java
nielspardon: thanks, makes given that it shouldn't be null but always an empty string

core/src/main/java/io/substrait/plan/Plan.java
nielspardon: I pushed an update to this code:

```
@Value.Immutable
public abstract class Plan {

  @Value.Default
  public Version getVersion() {
    return Version.DEFAULT_VERSION;
  }
..
  @Value.Immutable
  public abstract static class Version {
    public static final Version DEFAULT_VERSION;

    static {
      final String[] versionComponents = loadVersion();
      DEFAULT_VERSION =
          ImmutableVersion.builder()
              .major(Integer.parseInt(versionComponents[0]))
              .minor(Integer.parseInt(versionComponents[1]))
              .patch(Integer.parseInt(versionComponents[2]))
              .producer(Optional.of("substrait-java"))
              .build();
    }
```

core/src/main/java/io/substrait/plan/ProtoPlanConverter.java
nielspardon: I changed the two occurrences as suggested.

vbarua (approved):

core/src/main/java/io/substrait/plan/Plan.java
nielspardon: I'm also setting the producer to `isthmus` / `substrait-spark` now in the corresponding module

PR #408, opened 2025-06-06
fix(core): start with functionAnchor one instead of zero

nielspardon:
When validating Substrait plans generated by substrait-java with the current substrait-validator (which is pretty out of date) the validator gives the following warning for all plans:

```
Warning (code 3005):
at plan.extensions[0].mapping_type<extension_function>.function_anchor:
use of anchor zero is discouraged, as references set to zero may be confused with "unspecified". (code 3005)
```

This can be easily fixed by assigning functionAnchor values starting from one instead of zero.


Blizzara (approved):
Thanks! I think this should be fine but will leave open for a day or so in case someone else (mainly @vbarua) would think of a reason why this shouldn't change

vbarua (approved):
LGTM

PR #409, opened 2025-06-06
build: explicitly declare test framework deps

bestbeforetoday:
This change avoids the Gradle deprecation warning:

> The automatic loading of test framework implementation dependencies
> has been deprecated. This is scheduled to be removed in Gradle 9.0.


core/build.gradle.kts
```diff
@@ -81,9 +81,9 @@ configurations[JavaPlugin.COMPILE_ONLY_CONFIGURATION_NAME].extendsFrom(shadowImp
 configurations[JavaPlugin.TEST_IMPLEMENTATION_CONFIGURATION_NAME].extendsFrom(shadowImplementation)
 
 dependencies {
-  testImplementation("org.junit.jupiter:junit-jupiter-api:${JUNIT_VERSION}")
-  testImplementation("org.junit.jupiter:junit-jupiter-params:${JUNIT_VERSION}")
-  testRuntimeOnly("org.junit.jupiter:junit-jupiter-engine:${JUNIT_VERSION}")
+  testImplementation(platform("org.junit:junit-bom:${JUNIT_VERSION}"))
```
vbarua, line 84: TIL that `junit-bom` let's us set the JUNIT_VERSION once, and then it will align all the other junit imports ✨ 

https://stackoverflow.com/questions/67328406/what-is-junit-bom-and-junit-platform-for-and-should-i-include-them-in-gradle-de

PR #410, opened 2025-06-10
feat(core): add DdlRel

andrew-coleman:
Add the POJOs for the DdlRel proto message
Similar pattern to the previous WriteRel bindings.

### DIFF
--- a/core/src/main/java/io/substrait/relation/AbstractDdlRel.java
+++ b/core/src/main/java/io/substrait/relation/AbstractDdlRel.java
@@ -1,0 +1,77 @@
+package io.substrait.relation;
+
+import io.substrait.expression.Expression;
+import io.substrait.proto.DdlRel;
+import io.substrait.type.NamedStruct;
+import io.substrait.type.Type;
+import java.util.Optional;
+
+public abstract class AbstractDdlRel extends ZeroInputRel {
+  public abstract NamedStruct getTableSchema();
+
+  public abstract Expression.StructLiteral getTableDefaults();
+
+  public abstract DdlObject getObject();
+
+  public abstract DdlOp getOperation();
+
+  public abstract Optional<Rel> getViewDefinition();
+
+  public enum DdlObject {
+    UNSPECIFIED(DdlRel.DdlObject.DDL_OBJECT_UNSPECIFIED),
+    TABLE(DdlRel.DdlObject.DDL_OBJECT_TABLE),
+    VIEW(DdlRel.DdlObject.DDL_OBJECT_VIEW);
+
+    private final DdlRel.DdlObject proto;
+
+    DdlObject(DdlRel.DdlObject proto) {
+      this.proto = proto;
+    }
+
+    public DdlRel.DdlObject toProto() {
+      return proto;
+    }
+
+    public static DdlObject fromProto(DdlRel.DdlObject proto) {
+      for (var v : values()) {
+        if (v.proto == proto) {
+          return v;
+        }
+      }
+      throw new IllegalArgumentException("Unknown type: " + proto);
+    }
+  }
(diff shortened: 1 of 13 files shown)


core/src/main/java/io/substrait/relation/AbstractDdlRel.java
```diff
@@ -0,0 +1,78 @@
...
+public abstract class AbstractDdlRel extends ZeroInputRel {
+  public abstract NamedStruct getTableSchema();
+
+  public abstract Expression.StructLiteral getTableDefaults();
+
+  public abstract DdlObject getObject();
+
+  public abstract DdlOp getOperation();
+
+  public abstract Optional<Rel> getViewDefinition();
```
vbarua, line 18: If I understand the core spec, this is intended as the definition for a view for `CREATE VIEW` style statements.

There's nothing indicating that it is optional, or that it should only be set when `object` DDL_OBJECT_VIEW, but I think that's consistent with the current definition.

core/src/main/java/io/substrait/relation/AbstractDdlRel.java
```diff
@@ -0,0 +1,78 @@
...
+        }
+      }
+      throw new IllegalArgumentException("Unknown type: " + proto);
+    }
+  }
+
+  @Override
+  public Type.Struct deriveRecordType() {
+    // return getViewDefinition().getRecordType();
+    return getTableSchema().struct();
```
vbarua, line 75: I _think_ the table schema is probably the correct return type. But this also seems underspecified in the core spec.

Can we remove the commented code, or make it a TODO to check if we need to consider the view.

core/src/main/java/io/substrait/relation/Extension.java
```diff
@@ -58,4 +58,11 @@ public interface WriteExtensionObject extends ToProto<com.google.protobuf.Any> {
      */
     NamedStruct deriveSchema();
   }
+
+  public interface DdlExtensionObject extends ToProto<com.google.protobuf.Any> {
+    /**
+     * @return the table schema for the associated {@link ExtensionWrite} relation
+     */
+    NamedStruct deriveSchema();
```
vbarua: Thinking about this, I don't think we actually need a deriveSchema method here, because the schema is given directly as the `table_schema`.

The same applies to the WriteExtensionObject as well.

What do you think about removing these?

core/src/test/java/io/substrait/type/proto/DdlRelRoundtripTest.java
```diff
@@ -0,0 +1,76 @@
...
+        VirtualTableScan.builder().initialSchema(schema).addRows(defaults).build();
+
+    ExtensionDdl command =
+        ExtensionDdl.builder()
+            .viewDefinition(virtTable)
+            .tableSchema(schema)
+            .tableDefaults(defaults)
+            .detail(detail)
+            .operation(ExtensionDdl.DdlOp.ALTER)
+            .object(ExtensionDdl.DdlObject.VIEW)
```
vbarua, line 69: When we use `VIEW` do we need to set the `viewDefinition` as well?

vbarua (commented):
Left some minor comments. Some of them are related to ambiguities in the spec around the DdlRel definition, which I don't consider blockers for this work.

core/src/test/java/io/substrait/type/proto/DdlRelRoundtripTest.java
andrew-coleman: According to the docs we do (https://substrait.io/relations/logical_relations/#ddl-properties).
I set it on line 64.

vbarua (approved):

Issue #411, opened 2025-06-11
SPECIFICATION_VERSION missing in manifest


vbarua:
Following https://github.com/substrait-io/substrait-java/pull/406, @ingomueller-net reported issues when updating the consumer-testing library to use

See: https://github.com/substrait-io/substrait-java/commit/de8bf2abce302dadbcead2364f6f950fc5ad6673#r159771706

It may make sense to revert the changes before the next release.

nielspardon:
@ingomueller-net can you share a bit more context about your build environment? are you experiencing this locally or in Github Actions?

nielspardon:
Please still provide some reproduction steps just to be sure I'm looking at the right problem.

nielspardon:
Good, then I think my PR will fix your issue.

vbarua:
@ingomueller-net I've merged Niels' PR which should fix your issue hopefully. Let us know if it works (or doesn't).

ingomueller-net:
Thanks a lot for taking care of this issue promptly! However, it doesn't seem to work yet; I still get the same error. A few things I observed:

* The manifest file is now in the jar and the specification version is set:
   ```
   $ jar -xf jars/isthmus-0.58.0-all.jar META-INF/MANIFEST.MF
   $ grep Specification META-INF/MANIFEST.MF 
   Specification-Title: substrait
   Specification-Version: 0.69.0
   ```
* I am not sure if that was the problem to begin with. [`getResourceAsStream`](https://docs.oracle.com/javase/8/docs/api/java/lang/ClassLoader.html#getResourceAsStream-java.lang.String-) returns `null` if no manifest file is found, so if that were the case, we should have gotten an `NullPointerException` on using the *manifest*, not on using the value looked up in the manifest. (It could be, it finds some other manifest though that I am not aware of...)
* [`getValue`](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/jar/Attributes.html#getValue(java.util.jar.Attributes.Name)) on the manifest attributes returns `null` if the attribute wasn't found. That seems to be happening still. Maybe it'd be worth testing that and throwing an exception to improve the error message?

nielspardon:
[...]

PR #412, opened 2025-06-11
fix(isthmus,spark): add MANIFEST.MF specification version to JAR

nielspardon:
fixes https://github.com/substrait-io/substrait-java/issues/411


vbarua (approved):
Thanks for the quick fix 🙇

PR #414, opened 2025-06-12
build: use Central Publisher API for Maven publish

### LINKED ISSUE #407: build: migrate to new Maven Central Portal
@jacques-n has migrated the Maven Central namespace for Substrait from the to be shut down OSSRH system to the new Central Portal according to this thread on the [mailing list](https://groups.google.com/g/substrait/c/dFiizCUJEEU/m/eEEEi_yGAwAJ).

Now in order for the next release to be published correctly to Maven Central we would need to change the Gradle build to use the new Central Portal credentials and configuration which also involves switching to a different Gradle plugin than the default `maven-publish` plugin.

More info: https://central.sonatype.org/faq/what-is-different-between-central-portal-and-legacy-ossrh/

bestbeforetoday:
Publishing to Maven Central using OSSRH is deprecated and is sunset on 2025-06-30. The Maven namespace has been migrated to the Central Publishing Portal. There is currently no support for publishing using the Central Publisher API in Gradle's own publishing capability, and no official Sonatype Gradle publishing plugin. JReleaser is currently recommended by Sonatype for Gradle publishing to Maven Central.

Closes #407

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -17,15 +17,15 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [ ubuntu-latest, macOS-latest ]
+        os: [ubuntu-latest, macOS-latest]
     steps:
       - uses: actions/checkout@v4
         with:
           submodules: recursive
       - uses: graalvm/setup-graalvm@v1
         with:
-          java-version: '17'
-          distribution: 'graalvm'
+          java-version: "17"
+          distribution: "graalvm"
           # helps avoid rate-limiting issues
           github-token: ${{ secrets.GITHUB_TOKEN }}
       - name: Setup Gradle
@@ -65,11 +65,11 @@ jobs:
       - name: Set up JDK 17
         uses: actions/setup-java@v4
         with:
-          java-version: '17'
-          distribution: 'temurin'
+          java-version: "17"
+          distribution: "temurin"
       - uses: actions/setup-node@v4
         with:
-          node-version: '20'
+          node-version: "20"
       - name: Setup Gradle
         uses: gradle/actions/setup-gradle@v4
       - name: Download isthmus-ubuntu-latest binary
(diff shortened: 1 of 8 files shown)


.github/workflows/release.yml
```diff
@@ -86,8 +86,10 @@ jobs:
         run: ./ci/release/run.sh
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-          SONATYPE_USER: ${{ secrets.SONATYPE_USER }}
-          SONATYPE_PASSWORD: ${{ secrets.SONATYPE_PASSWORD }}
           SIGNING_KEY_ID: ${{ secrets.SIGNING_KEY_ID }}
           SIGNING_PASSWORD: ${{ secrets.SIGNING_PASSWORD }}
           SIGNING_KEY: ${{ secrets.SIGNING_KEY }}
+          JRELEASER_GPG_SECRET_KEY: ${{ secrets.SIGNING_KEY }}
```
nielspardon, line 92: shouldn't this be `secrets.SIGNING_KEY_ID` like in line 89?

```suggestion
          JRELEASER_GPG_SECRET_KEY: ${{ secrets.SIGNING_KEY_ID }}
```

nielspardon: nevermind, I see `secrets.SIGNING_KEY` also exists.

.github/workflows/release.yml
```diff
@@ -86,8 +86,10 @@ jobs:
...
+          JRELEASER_GPG_PASSPHRASE: ${{ secrets.SIGNING_PASSWORD }}
+          JRELEASER_MAVENCENTRAL_USERNAME: ${{ secrets.MAVEN_CENTRAL_USER }}
+          JRELEASER_MAVENCENTRAL_PASSWORD: ${{ secrets.MAVEN_CENTRAL_KEY }}
```
bestbeforetoday, line 95: @jacques-n these are the new secrets you shared for publishing to the Maven Central Publisher API.

The `SIGNING_KEY` is the GPG / PGP key used to create digital signatures for the assets that will then be published to Maven Central using the publishing secrets. The signing secrets are not changed from what was there before. Only the mechanism used to do the signing: JReleaser instead of the Gradle signing plugin.

jacques-n: LOL. That is what I get for not looking closely I suppose.

I don't have the keys as once they were only help long enough to input into GitHub. 

Best suggestion I have is for you to test locally using a a different registry and/or locally hosted one. If that works I'm inclined to merge and cross fingers. Thoughts?

vbarua: > If that works I'm inclined to merge and cross fingers.

Given that the current [release](https://github.com/substrait-io/substrait-java/actions/runs/15658895473/job/44114146716) is failing already, this can't make it worse 😅

jacques-n (approved):
let's see if this makes anything better :P

if it makes things worse blame @vbarua for suggesting a not-run merge :D

PR #415, opened 2025-06-12
feat: expose Nullability and Decomposability enums

vbarua:


vbarua:
Was trying to use the `Nullability` enum externally, and noticed that it and `Decomposability` are not public and as such cannot be used in external code.

Blizzara (approved):

PR #417, opened 2025-06-16
Add expressions support for Virtual Tables

moraistejerina:


core/src/main/java/io/substrait/relation/ProtoRelConverter.java
```diff
@@ -418,9 +418,11 @@ protected FileOrFiles newFileOrFiles(ReadRel.LocalFiles.FileOrFiles file) {
   protected VirtualTableScan newVirtualTable(ReadRel rel) {
     var virtualTable = rel.getVirtualTable();
     var virtualTableSchema = newNamedStruct(rel);
+
     var converter =
         new ProtoExpressionConverter(lookup, extensions, virtualTableSchema.struct(), this);
-    List<Expression.StructLiteral> structLiterals = new ArrayList<>(virtualTable.getValuesCount());
+
+    List<Expression.StructLiteral> structLiterals = new ArrayList<>(virtualTable.getValuesCount() + virtualTable.getExpressionsCount());
```
vbarua: To support expression lists, this should just be `List<Expression>`.

core/src/main/java/io/substrait/relation/ProtoRelConverter.java
```diff
@@ -430,6 +432,16 @@ protected VirtualTableScan newVirtualTable(ReadRel rel) {
                       .collect(java.util.stream.Collectors.toList()))
               .build());
     }
+    for (var struct : virtualTable.getExpressionsList()) {
+      structLiterals.add(
+          ImmutableExpression.StructLiteral.builder()
+              .fields(
+                  struct.getFieldsList().stream()
+                      .map(io.substrait.proto.Expression::getLiteral)
```
vbarua: If you want to add support for expression lists, we should not assume that they are literals here.

To handle this fully, we'll need to update the POJO (Plain Old Java Object) representation of VirtualTable to allow `rows` to contain full expressions, and not just literals.

vbarua:
Superseded by https://github.com/substrait-io/substrait-java/pull/579#pullrequestreview-3475068310

PR #420, opened 2025-06-19
fix(core): convert hints in ProtoRelConverter

strausza:
Most of the mappings were missing the hints. This PR re-uses the already existing `optionalHint` method and includes the hints for relations where this was missing. No other functional changes.

### DIFF
--- a/core/src/main/java/io/substrait/relation/ProtoRelConverter.java
+++ b/core/src/main/java/io/substrait/relation/ProtoRelConverter.java
@@ -179,7 +179,8 @@ public class ProtoRelConverter {
 
     builder
         .commonExtension(optionalAdvancedExtension(rel.getCommon()))
-        .remap(optionalRelmap(rel.getCommon()));
+        .remap(optionalRelmap(rel.getCommon()))
+        .hint(optionalHint(rel.getCommon()));
     return builder.build();
   }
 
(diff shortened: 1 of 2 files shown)


core/src/main/java/io/substrait/relation/ProtoRelConverter.java
```diff
@@ -175,7 +175,8 @@ protected NamedWrite newNamedWrite(WriteRel rel) {
             .tableSchema(newNamedStruct(rel.getTableSchema()))
             .createMode(NamedWrite.CreateMode.fromProto(rel.getCreateMode()))
             .outputMode(NamedWrite.OutputMode.fromProto(rel.getOutput()))
-            .operation(NamedWrite.WriteOp.fromProto(rel.getOp()));
+            .operation(NamedWrite.WriteOp.fromProto(rel.getOp()))
+            .hint(optionalHint(rel.getCommon()));
```
vbarua: Super minor, but for consistency could you move the `.hint` calls to be with the `.commonExtension` and `.remap` calls here and in the other relations where that pattern is used. The intent is to group the relation specific calls together, and then have the calls associated with the RelCommon data together.

strausza: This is now adjusted for all rels. Thanks for the review.

PR #421, opened 2025-06-19
feat(core): add UpdateRel

andrew-coleman:
Add the POJOs for the UpdateRel proto message

### DIFF
--- a/core/src/main/java/io/substrait/relation/AbstractUpdate.java
+++ b/core/src/main/java/io/substrait/relation/AbstractUpdate.java
@@ -1,0 +1,32 @@
+package io.substrait.relation;
+
+import io.substrait.expression.Expression;
+import io.substrait.type.NamedStruct;
+import io.substrait.type.Type;
+import java.util.List;
+import org.immutables.value.Value;
+
+public abstract class AbstractUpdate extends ZeroInputRel implements HasExtension {
+
+  public abstract NamedStruct getTableSchema();
+
+  public abstract Expression getCondition();
+
+  public abstract List<TransformExpression> getTransformations();
+
+  @Value.Immutable
+  public abstract static class TransformExpression {
+    public abstract Expression getTransformation();
+
+    public abstract int getColumnTarget();
+
+    public static ImmutableTransformExpression.Builder builder() {
+      return ImmutableTransformExpression.builder();
+    }
+  }
+
+  @Override
+  public Type.Struct deriveRecordType() {
+    return getTableSchema().struct();
+  }
+}
--- a/core/src/main/java/io/substrait/relation/ProtoRelConverter.java
+++ b/core/src/main/java/io/substrait/relation/ProtoRelConverter.java
@@ -27,6 +27,7 @@ import io.substrait.proto.ProjectRel;
 import io.substrait.proto.ReadRel;
 import io.substrait.proto.SetRel;
 import io.substrait.proto.SortRel;
+import io.substrait.proto.UpdateRel;
 import io.substrait.proto.WriteRel;
 import io.substrait.relation.extensions.EmptyDetail;
 import io.substrait.relation.extensions.EmptyOptimization;
@@ -127,6 +128,9 @@ public class ProtoRelConverter {
       }
       case DDL -> {
         return newDdl(rel.getDdl());
+      }
+      case UPDATE -> {
+        return newUpdate(rel.getUpdate());
       }
       default -> {
         throw new UnsupportedOperationException("Unsupported RelTypeCase of " + relType);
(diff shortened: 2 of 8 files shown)


core/src/main/java/io/substrait/relation/Update.java
```diff
@@ -0,0 +1,44 @@
+package io.substrait.relation;
+
+import io.substrait.expression.Expression;
+import io.substrait.type.NamedStruct;
+import io.substrait.type.Type;
+import java.util.List;
+import org.immutables.value.Value;
+
+@Value.Immutable
+public abstract class Update extends ZeroInputRel implements HasExtension {
```
vbarua: So at the moment, there is only one type of update, which is the named_table
https://github.com/substrait-io/substrait/blob/793c64ba26e337c22f5e91b658be58b1eea7efd3/proto/substrait/algebra.proto#L722-L724

In the future there could be more though. Keeping that in mind, and for consistency with NamedWrite and NamedDDL, what do you think about removing the `@Value.Immutable` from this and adding a separate `NamedUpdate` class extending `Update`, and generating the POJO from that?

core/src/main/java/io/substrait/relation/ProtoRelConverter.java
```diff
@@ -258,6 +262,26 @@ protected Expression.StructLiteral tableDefaults(
...
+              .transformation(converter.from(transformation.getTransformation()))
+              .columnTarget(transformation.getColumnTarget())
+              .build());
+    }
+    return Update.builder()
+        .names(rel.getNamedTable().getNamesList())
+        .tableSchema(tableSchema)
+        .addAllTransformations(transformations)
+        .condition(converter.from(rel.getCondition()))
+        .build();
```
vbarua: We should be setting `.extensions` here as well.

That's what the
```java
    if (rel.hasAdvancedExtension()) {
      builder.extension(advancedExtension(rel.getAdvancedExtension()));
    }
```
line does in the other creation methods.

This has made me realize that we didn't handle this for the Write and DDL variants. Could I ask you to handle that either here, or in a follow-up PR?

This has also made me realize that UpdateRel doesn't have RelCommon field at all. Filed https://github.com/substrait-io/substrait/issues/821 for that.

core/src/main/java/io/substrait/relation/RelProtoConverter.java
```diff
@@ -420,6 +422,26 @@ public Rel visit(ExtensionDdl ddl) throws RuntimeException {
...
+    var builder =
+        UpdateRel.newBuilder()
+            .setNamedTable(NamedTable.newBuilder().addAllNames(update.getNames()))
+            .setTableSchema(update.getTableSchema().toProto(typeProtoConverter))
+            .addAllTransformations(
+                update.getTransformations().stream()
+                    .map(this::toProto)
+                    .collect(Collectors.toList()))
+            .setCondition(toProto(update.getCondition()));
+    return Rel.newBuilder().setUpdate(builder).build();
```
vbarua, line 436: We should be checking the `update.getExtension` field to set it in the proto.

core/src/main/java/io/substrait/relation/ProtoRelConverter.java
andrew-coleman: I'll do the Write and DDL variants in a separate PR to minimise the merge conflicts with https://github.com/substrait-io/substrait-java/pull/420

core/src/main/java/io/substrait/relation/Update.java
andrew-coleman: I've also renamed `Update` to `AbstractUpdate` for consistency with the other two.

On a vaguely related note, I notice that the [NamedTable](https://github.com/substrait-io/substrait/blob/main/proto/substrait/algebra.proto#L742) message structure gets repeated in several places in the proto definitions. The same structure is [inlined](https://github.com/substrait-io/substrait/blob/main/proto/substrait/algebra.proto#L122) in the ReadRel message, and is structurally the same as the [NamedObjectWrite](https://github.com/substrait-io/substrait/blob/main/proto/substrait/algebra.proto#L593) message used by the WriteRel and DDlRel.  They all share the same description in their comments.  
Is this intentional or could they be refactored?

vbarua: > I notice that the [NamedTable](https://github.com/substrait-io/substrait/blob/main/proto/substrait/algebra.proto#L742) message structure gets repeated in several places in the proto definitions.

That probably could be factored out. It might even be backwards compatible.

> is structurally the same as the [NamedObjectWrite](https://github.com/substrait-io/substrait/blob/main/proto/substrait/algebra.proto#L593) message used by the WriteRel and DDlRel. They all share the same description in their comments.

That also seems redundant. I'm not sure if that was intentional or not.

core/src/main/java/io/substrait/relation/AbstractUpdate.java
```diff
@@ -0,0 +1,32 @@
+package io.substrait.relation;
+
+import io.substrait.expression.Expression;
+import io.substrait.type.NamedStruct;
+import io.substrait.type.Type;
+import java.util.List;
+import org.immutables.value.Value;
+
+public abstract class AbstractUpdate extends ZeroInputRel implements HasExtension {
```
vbarua, line 9: Good call updating the name of this too.

core/src/main/java/io/substrait/relation/ProtoRelConverter.java
```diff
@@ -258,6 +262,30 @@ protected Expression.StructLiteral tableDefaults(
...
+    for (var transformation : rel.getTransformationsList()) {
+      transformations.add(
+          NamedUpdate.TransformExpression.builder()
+              .transformation(converter.from(transformation.getTransformation()))
+              .columnTarget(transformation.getColumnTarget())
+              .build());
+    }
+    var builder =
+        NamedUpdate.builder()
+            .names(rel.getNamedTable().getNamesList())
```
vbarua, line 294: It might be worth checking the value of `rel.getUpdateTypeCase` further up to make sure it's a named table, and throw an error otherwise.

As is, I think this will NPE if the upstream ever adds a new `update_type` to the `oneof` and we get sent one before updating the code here. This does technically signal our lack of support for it, but an explicit `UnsupportedOperationException` would be clearer.

andrew-coleman: Yes, good catch.  I meant to structure the code to be the same as WriteRel and DdlRel, but I missed this bit.  I've used a switch statement which is a bit overkill for just one case, but it probably makes it more readable if we do add more cases.

vbarua (approved):

Issue #419, opened 2025-06-18
Maven signing not working correctly after switch to jreleaser


jacques-n:
I merged #407 and manually kicked off the release process. I see this warning: 

[sign] Did not find public key for signing. Ensure a value has been configured and properly formatted without indentation when reviewing the [build](https://github.com/substrait-io/substrait-java/actions/runs/15722081675/job/44305609728)

@nielspardon, can you take a look?

jacques-n:
I manual cancelled. I'll run it again :P

bestbeforetoday:
The new run is not doing anything useful since the release of 0.60.0 has already been successfully completed. There does not seem to be any issue here since the artifact signing and publishing to Maven Central all worked OK.

I have raised jreleaser/jreleaser#1904 to suggest not printing a warning message about missing public key when a public key is not required.

PR #422, opened 2025-06-20
build: use Gradle signing instead of JReleaser

bestbeforetoday:
JReleaser expects to have the corresponding public key available when signing artifacts so that it can verify that the public key is published to known key registries. While publishing works without the public key, it outputs a warning message, which can cause confusion when viewing build logs.

This change disables JReleaser artifact signing and reverts to using Gradle's own signing plugin. This also provides a smoother migration path to an official Sonatype Gradle publishing plugin, if/when one becomes available in the future.

Gradle wrapper is also updated to the latest patch release.

Relates to #419


bestbeforetoday:
@vbarua @jacques-n for your review.

bestbeforetoday:
Rebased this change on latest main branch content.

vbarua (approved):
Looks reasonable enough to me. Thanks for updating this.

PR #424, opened 2025-06-25
feat(spark): bitwise shift functions

andrew-coleman:
Many of the TPC-DS tests generate optimised plans that use ShiftRight function.  These were working because shift_right had been declared as an extension function in spark.yaml. Now that the bit shift functions are included in the base substrait definitions, it can be removed from spark.yaml.


vbarua (approved):
Thanks for following up with this update @andrew-coleman

PR #425, opened 2025-06-25
chore: avoid using generated immutables

andrew-coleman:
Some of the non-core code is directly importing and invoking the object factories that are generated at build time from the immutables annotations.  This commit changes the code to (where possible) use the parent class methods.

### DIFF
--- a/core/src/main/java/io/substrait/dsl/SubstraitBuilder.java
+++ b/core/src/main/java/io/substrait/dsl/SubstraitBuilder.java
@@ -3,16 +3,15 @@ package io.substrait.dsl;
 import com.github.bsideup.jabel.Desugar;
 import io.substrait.expression.AggregateFunctionInvocation;
 import io.substrait.expression.Expression;
+import io.substrait.expression.Expression.Cast;
 import io.substrait.expression.Expression.FailureBehavior;
 import io.substrait.expression.Expression.IfClause;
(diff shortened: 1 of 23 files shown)


core/src/main/java/io/substrait/dsl/SubstraitBuilder.java
```diff
@@ -678,11 +676,7 @@ public Expression.WindowFunctionInvocation windowFn(
   // Types
 
   public Type.UserDefined userDefinedType(String namespace, String typeName) {
-    return ImmutableType.UserDefined.builder()
-        .uri(namespace)
-        .name(typeName)
-        .nullable(false)
-        .build();
+    return Type.UserDefined.builder().uri(namespace).name(typeName).nullable(false).build();
   }
```
kadinrabo, line 680: Nice cleanup, this pretty much summarizes the PR

kadinrabo (approved):
Don't see anything wrong with this, looks good to me 👍

Found some other opportunities like: `ImmutableExpressionReference`, `ImmutableExtensionLookup`, `ImmutableEnumArg`

andrew-coleman:
> Found some other opportunities like: ImmutableExpressionReference, ImmutableExtensionLookup, ImmutableEnumArg

I've now added these.  Thanks!

PR #426, opened 2025-06-27
fix(isthmus): handle subqueries with outer field references

nielspardon:
changes `SubstraitRelNodeConverter` and `ExpressionRexConverter` to support handling subqueries with outer field references

- BREAKING CHANGE: adds a new context argument to `RelVisitor`, `ExpressionVisitor` and `FuncArgVisitor` interfaces which allows to share an additional context object between the visitors when traversing a Substrait plan
- makes TPC-H query 17 and TPC-DS query 6 work
- breaks TPC-DS queries 1, 30, 81 when going from Calcite RelNode to SQL since they are now getting differently from Substrait to Calcite (will address these in a follow-up PR)

additional cleanup:
- replaces a couple of generic `RuntimeException` throws with more specific exceptions
- removes no-op test class `CalciteFieldSelectionTest`
- renames a couple long generic type names to single letter names to be compliant with Java naming conventions
- removes unused `logger` constants which where also not following the Java naming convention for constants
- removes unnecessary `static` keywords for classes/enums within interfaces
- makes private inner classes with only private constructors final since they can't be extended anyway

fixes https://github.com/substrait-io/substrait-java/issues/382

replaces https://github.com/substrait-io/substrait-java/pull/383


nielspardon:
> For the sake of reviewability could I ask you split this into 3-ish PRs. One for your initial cleanups/improvements, one for just the visitor changes, and one for subquery changes.

Sure, I already thought that this might be necessary.

nielspardon:
First extracted PR for the visitor changes: https://github.com/substrait-io/substrait-java/pull/427

nielspardon:
updated to use the new `VisitationContext` interface from #427

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
```diff
@@ -183,12 +188,17 @@ public RelNode visit(Project project, EmptyVisitationContext context) throws Run
...
+    context.popParentRelNodes();
     return applyRemap(node, project.getRemap());
   }
 
   @Override
-  public RelNode visit(Cross cross, EmptyVisitationContext context) throws RuntimeException {
+  public RelNode visit(Cross cross, Context context) throws RuntimeException {
     RelNode left = cross.getLeft().accept(this, context);
     RelNode right = cross.getRight().accept(this, context);
     // Calcite represents CROSS JOIN as the equivalent INNER JOIN with true condition
```
vbarua: If we're going to handle subqueries, don't we need to propagate the parents through every relation type?

isthmus/src/main/java/io/substrait/isthmus/expression/ExpressionRexConverter.java
```diff
@@ -640,14 +651,15 @@ public RexNode visitEnumArg(
...
+    context.incrementSubqueryDepth();
     RelNode inputRelnode = expr.input().accept(relNodeConverter, context);
+    context.decrementSubqueryDepth();
     return RexSubQuery.scalar(inputRelnode);
   }
 
   @Override
-  public RexNode visit(SetPredicate expr, EmptyVisitationContext context) throws RuntimeException {
+  public RexNode visit(SetPredicate expr, Context context) throws RuntimeException {
     RelNode inputRelnode = expr.tuples().accept(relNodeConverter, context);
```
vbarua, line 687: Would we need to increment and decrement the subquery depth here? [SetPredicate](https://github.com/substrait-io/substrait/blob/793c64ba26e337c22f5e91b658be58b1eea7efd3/proto/substrait/algebra.proto#L1563C7-L1563C38) is a Subquery relation.

isthmus/src/main/java/io/substrait/isthmus/expression/ExpressionRexConverter.java
```diff
@@ -502,8 +484,7 @@ public RexNode visit(Expression.WindowFunctionInvocation expr, EmptyVisitationCo
   }
 
   @Override
-  public RexNode visit(Expression.InPredicate expr, EmptyVisitationContext context)
-      throws RuntimeException {
+  public RexNode visit(Expression.InPredicate expr, Context context) throws RuntimeException {
     List<RexNode> needles =
         expr.needles().stream().map(e -> e.accept(this, context)).collect(Collectors.toList());
     RelNode rel = expr.haystack().accept(relNodeConverter, context);
```
vbarua, line 520: Would we need to increment and decrement the subquery depth here? [InPredicate](https://github.com/substrait-io/substrait/blob/793c64ba26e337c22f5e91b658be58b1eea7efd3/proto/substrait/algebra.proto#L1561) is a Subquery relation.

vbarua (commented):
Did a very cursory first pass.

Something that I think would be helpful would be some tests based directly on Substrait plans, and not just the TPC based tests. This would help contributors, myself included, understand what exactly the structure of these subqueries looks like in Substrait, and how your processing is being applied on top of it.

nielspardon:
> Something that I think would be helpful would be some tests based directly on Substrait plans, and not just the TPC based tests. This would help contributors, myself included, understand what exactly the structure of these subqueries looks like in Substrait, and how your processing is being applied on top of it.

I drew this diagram for myself to reason about how outer field references work in Substrait vs. Calcite:
image

- Subqueries are an expression element so they only appear within relations that are configured using expressions e.g. `FilterRel`, `ProjectRel`, `JoinRel`. 
- Substrait addresses field references to the parent relation operations by a number of subquery boundaries one has to cross to find the parent relation of the referenced field which means we need to keep track of the subquery depth. 
- Calcite uses correlation variables e.g. `$cor0` to address a field from a parent relation e.g. `$cor0.P_PARTKEY`. The correlation variable needs to be added to the `variablesSet` of the parent relation the field is coming from `variablesSet=[$cor0]`. Additionally when creating the correlation variable we need to get the row type of the parent relation providing the field. When creating the field access in the subquery we need to have access to the parent relation row type and we need to ensure we add all the correlation variables pointing to this parent relation to the `variablesSet` of this parent relation.
- We do not need to track all parent relations but we just need to track the parent relations which support expressions. Like in the example above we need to know the 3 `FilterRels` and don't really care about the other relations in the tree.

I hope this helps.

isthmus/src/main/java/io/substrait/isthmus/expression/ExpressionRexConverter.java
nielspardon: good catch, I think that should be added

isthmus/src/main/java/io/substrait/isthmus/expression/ExpressionRexConverter.java
nielspardon: good catch, I think that should be added

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
nielspardon: not entirely sure how you mean that. we need to propagate the context through every relation type but we do not need to track every relation as a parent but only those relations that can contain subqueries in their expressions.

isthmus/src/main/java/io/substrait/isthmus/expression/ExpressionRexConverter.java
nielspardon: should be fixed now

isthmus/src/main/java/io/substrait/isthmus/expression/ExpressionRexConverter.java
nielspardon: should be fixed now

isthmus/src/test/java/io/substrait/isthmus/expression/SubqueryConversionTest.java
```diff
@@ -0,0 +1,164 @@
...
+import io.substrait.type.TypeCreator;
+import java.util.List;
+import org.apache.calcite.rel.RelNode;
+import org.junit.jupiter.api.Test;
+
+class SubqueryConversionTest extends PlanTestBase {
+  protected final SubstraitToCalcite converter = new SubstraitToCalcite(extensions, typeFactory);
+
+  @Test
+  void testOuterFieldReferenceOneStep() {
```
nielspardon, line 39: @vbarua what do you think about this testing approach?

nielspardon: I do have tests for scalar subqueries, in predicate and set predicate with various levels of depth. I guess the test coverage should be good enough. let me know if there's anything else I should add

core/src/main/java/io/substrait/dsl/SubstraitBuilder.java
```diff
@@ -706,4 +715,15 @@ public Plan plan(Plan.Root root) {
   public Rel.Remap remap(Integer... fields) {
     return Rel.Remap.of(Arrays.asList(fields));
   }
+
+  public Expression scalarSubquery(Rel input, Type type) {
+    return Expression.ScalarSubquery.builder().input(input).type(type).build();
+  }
+
+  public Expression exists(Rel project) {
```
vbarua: We should use`rel` instead of `project` for the argument name, as there's no requirement that this be a ProjectRel.

isthmus/src/test/java/io/substrait/isthmus/expression/SubqueryConversionTest.java
vbarua: This helps a lot in understanding how the references are constructed in Substrait 🙇

isthmus/src/test/java/io/substrait/isthmus/expression/SubqueryConversionTest.java
```diff
@@ -0,0 +1,371 @@
...
+                                    substraitBuilder.equal(
+                                        // customer.c_custkey
+                                        substraitBuilder.fieldReference(input2, 0),
+                                        // orders.o_custkey
+                                        FieldReference.newRootStructOuterReference(
+                                            1, TypeCreator.REQUIRED.I64, 1)),
+                                substraitBuilder.namedScan(
+                                    List.of("customer"),
+                                    List.of("c_custkey", "c_nationkey"),
+                                    List.of(TypeCreator.REQUIRED.I64, TypeCreator.REQUIRED.I64)))),
```
vbarua: minor: there are a couple of tables that you reuse in your tests that could be extracted out.

If you introduced something like:
```java
    final Rel customers =
        b.namedScan(
            List.of("customer"),
            List.of("c_custkey", "c_nationkey"),
            List.of(TypeCreator.REQUIRED.I64, TypeCreator.REQUIRED.I64));
```
it would make the plan building code a bit more concise.

isthmus/src/test/java/io/substrait/isthmus/expression/SubqueryConversionTest.java
```diff
@@ -0,0 +1,371 @@
...
+  @Test
+  void testOuterFieldReferenceOneStep() {
+    /*
+     * SELECT
+     *   orders.o_orderkey,
+     *   (SELECT customer.c_nationkey FROM customer WHERE customer.c_custkey = orders.o_custkey)
+     * FROM orders
+     */
+    final Rel root =
+        substraitBuilder.project(
```
vbarua, line 47: minor: I often alias `substraitBuilder` to `b` in order to minimize it's presence when building plans like this and making them easier to read.

isthmus/src/test/java/io/substrait/isthmus/TpcdsQueryTest.java
```diff
@@ -13,7 +14,7 @@
 /** TPC-DS test to convert SQL to Substrait and then convert those plans back to SQL. */
 public class TpcdsQueryTest extends PlanTestBase {
   private static final Set<Integer> toSubstraitExclusions = Set.of(9, 27, 36, 70, 86);
-  private static final Set<Integer> fromSubstraitExclusions = Set.of(6, 8, 67);
+  private static final Set<Integer> fromSubstraitExclusions = Set.of(1, 8, 30, 67, 81);
```
vbarua: Did 1, 30 and 81 work before, and now they don't?

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
```diff
@@ -144,8 +147,11 @@ public static RelNode convert(
   @Override
   public RelNode visit(Filter filter, Context context) throws RuntimeException {
     RelNode input = filter.getInput().accept(this, context);
+    context.pushParentRelNodes(input);
     RexNode filterCondition = filter.getCondition().accept(expressionRexConverter, context);
-    RelNode node = relBuilder.push(input).filter(filterCondition).build();
+    RelNode node =
+        relBuilder.push(input).filter(context.popCorrelationIds(), filterCondition).build();
+    context.popParentRelNodes();
```
vbarua: This might be clearer if we called `pushParentRelNodes` something like `pushExpressionBase`* and added some docs to the method to indicate that this is the Calcite relation on which the expressions we are converting are built**, and to which subquery references are resolved. 

The converted `input` is a parent of relations in any subqueries we might encounter when converting expressions. But that's the uncommon case, and more commonly I think people will see this and think of it as the parent of expression, which is what I did at first and found somewhat confusing. The name also made me think that we had to push it parent in every relation.

This also comes up when looking at the join which has `context.pushParentRelNodes(left, right)` which was weird in a "how can these expression have 2 parents" kind of way, and when we call `getParentRelation` and get multiple parents it's not immediately clear that these are 2 parents at the same level vs a chain of parents.

\* I don't know if this is actually a better name, I just find the use of "Parent" here confusing.

\** The work you're doing here will actually provide a mechanism to verify the agreement of Calcite and Substrait type inference because we can extract the Calcite schema from the context https://github.com/substrait-io/substrait-java/issues/379.

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
vbarua: Ah, I think I understand this better now. The language of `pushParentRelNodes` confused me. Made a separate comment about this.

vbarua (commented):
Now that we've merged the precursor PR, this is much easier to look at. Your comments and tests help a lot as well.

We've only have tests that go from Substrait -> Calcite. Does this work going the other way? Not something that I think needs to happen as part of this PR, but possibly something to document as an issue.

To your points about
> Subqueries are an expression element so they only appear within relations that are configured using expressions e.g. FilterRel, ProjectRel, JoinRel.

> Calcite only supports providing a variablesSet for Project, Filter and Join if you search for variablesSet in the Calcite RelBuilder.

SortRels also contain expressions which could contain subqueries. Would the conversion just blow up now if we encountered that?

AggregateRels also contains expressions, but a because of the https://github.com/substrait-io/substrait-java/blob/main/isthmus/src/main/java/io/substrait/isthmus/PreCalciteAggregateValidator.java that may Just Work (TM) because they would get moved into a Project before the Aggregate.

I don't expect subqueries in those relations to be that common, but it might be worth a quick check to see what happens and filing a follow-up issue.

isthmus/src/test/java/io/substrait/isthmus/TpcdsQueryTest.java
nielspardon: They did convert without throwing an exception because they were converted incorrectly. Those queries also have subqueries with outer field references which previously were not converted to outer field references (correlation variables) but as direct field references and they just happened to not throw an exception during conversion.

If you look at e.g. TPC-DS query 1:
```sql
with customer_total_return as
    (select 
        sr_customer_sk as ctr_customer_sk
        ,sr_store_sk as ctr_store_sk
        ,sum(SR_RETURN_AMT_INC_TAX) as ctr_total_return
        from store_returns
        ,date_dim
    where sr_returned_date_sk = d_date_sk
        and d_year =2001
    group by sr_customer_sk
        ,sr_store_sk)
        
select  c_customer_id
from customer_total_return ctr1
,store
,customer
where ctr1.ctr_total_return > (select avg(ctr_total_return)*1.2     <--- has an aggregate
      from customer_total_return ctr2
      where ctr1.ctr_store_sk = ctr2.ctr_store_sk)     <--- filter has an outer reference to ctr1
and s_store_sk = ctr1.ctr_store_sk
and s_state = 'distmember(fips_county, 61, 3)'
and ctr1.ctr_customer_sk = c_customer_sk
order by c_customer_id
LIMIT 100
```

Calcite explain after going from Substrait back to Calcite:

```
LogicalProject(C_CUSTOMER_ID=[$0])
  LogicalSort(sort0=[$0], dir0=[ASC], fetch=[100])
    LogicalProject(C_CUSTOMER_ID0=[$33])
      LogicalFilter(condition=[AND(>(CAST($2):DECIMAL(9, 3), $SCALAR_QUERY({
LogicalProject($f1=[*($0, 1.2:DECIMAL(2, 1))])
  LogicalAggregate(group=[{}], agg#0=[AVG($0)])
    LogicalProject($f20=[$2])
      LogicalFilter(condition=[=($cor0.SR_STORE_SK0, $1)])         <-- LogicalFilter with correl var
        LogicalAggregate(group=[{0, 1}], agg#0=[SUM($2)])           <-- directly on top of LogicalAggregate
          LogicalProject(SR_CUSTOMER_SK0=[$3], SR_STORE_SK0=[$7], SR_RETURN_AMT_INC_TAX0=[$13])
            LogicalFilter(condition=[AND(=($0, $20), =($26, 2001))])
              LogicalJoin(condition=[true], joinType=[inner])
                LogicalTableScan(table=[[tpcds, STORE_RETURNS]])
                LogicalTableScan(table=[[tpcds, DATE_DIM]])
})), =($3, $1), =(CAST($27):VARCHAR(30) NOT NULL, 'distmember(fips_county, 61, 3)'), =($0, $32))], variablesSet=[[$cor0]])
        LogicalJoin(condition=[true], joinType=[inner])
          LogicalJoin(condition=[true], joinType=[inner])
            LogicalAggregate(group=[{0, 1}], agg#0=[SUM($2)])
              LogicalProject(SR_CUSTOMER_SK0=[$3], SR_STORE_SK0=[$7], SR_RETURN_AMT_INC_TAX0=[$13])
                LogicalFilter(condition=[AND(=($0, $20), =($26, 2001))])
                  LogicalJoin(condition=[true], joinType=[inner])
                    LogicalTableScan(table=[[tpcds, STORE_RETURNS]])
                    LogicalTableScan(table=[[tpcds, DATE_DIM]])
            LogicalTableScan(table=[[tpcds, STORE]])
          LogicalTableScan(table=[[tpcds, CUSTOMER]])
```

This is hitting some bug in Calcite since the queries turn out to have a `LogicalFilter` directly on top of a `LogicalAggregate` where the filter now contains a correlation variable and Calcite has special handling of a `LogicalFilter` on top of a `LogicalAggregate` which treats the filter as a `HAVING` clause when converting Rel to SQL. For some reason unknown to me right now the `org.apache.calcite.rel.rel2sql.SqlImplementor$Context` used here is not a `org.apache.calcite.rel.rel2sql.SqlImplementor$AliasContext` but a context which has no implementation of the `getAliasContext()` method. I was planning do dig into the Calcite fix separately.

```
Caused by: java.lang.UnsupportedOperationException
	at org.apache.calcite.rel.rel2sql.SqlImplementor$Context.getAliasContext(SqlImplementor.java:1074)
	at org.apache.calcite.rel.rel2sql.SqlImplementor$Context.toSql(SqlImplementor.java:682)
	at org.apache.calcite.rel.rel2sql.SqlImplementor$Context.toSql(SqlImplementor.java:1196)
	at org.apache.calcite.rel.rel2sql.SqlImplementor$Context.callToSql(SqlImplementor.java:880)
	at org.apache.calcite.rel.rel2sql.SqlImplementor$Context.toSql(SqlImplementor.java:853)
	at org.apache.calcite.rel.rel2sql.RelToSqlConverter.visit(RelToSqlConverter.java:477)
	...
```

https://github.com/apache/calcite/blob/3719447e125982aaa47f0a7b2564f94b9900539e/core/src/main/java/org/apache/calcite/rel/rel2sql/RelToSqlConverter.java#L465-L485

I did check whether the Calcite `RelNode` structure produced by `SqlToRelNodeConverter` can be translated back to SQL and it does. It only breaks when we go through Substrait and the reason seems to be the missing intermediate field names are leading to missing intermediate `LogicalProject` rels compared to the Calcite explain before the Substrait conversion:

```
LogicalProject(C_CUSTOMER_ID=[$0])
  LogicalSort(sort0=[$0], dir0=[ASC], fetch=[100])
    LogicalProject(C_CUSTOMER_ID=[$33])
      LogicalFilter(condition=[AND(>(CAST($2):DECIMAL(9, 3), $SCALAR_QUERY({
LogicalProject(EXPR$0=[*($0, 1.2:DECIMAL(2, 1))])
  LogicalAggregate(group=[{}], agg#0=[AVG($0)])
    LogicalProject(CTR_TOTAL_RETURN=[$2])
      LogicalFilter(condition=[=($cor0.CTR_STORE_SK, $1)])                 <-- HERE is our LogicalFilter
        LogicalProject(CTR_CUSTOMER_SK=[$0], CTR_STORE_SK=[$1], CTR_TOTAL_RETURN=[$2])
          LogicalAggregate(group=[{0, 1}], CTR_TOTAL_RETURN=[SUM($2)])         <-- HERE is the LogicalAggregate
            LogicalProject(SR_CUSTOMER_SK=[$3], SR_STORE_SK=[$7], SR_RETURN_AMT_INC_TAX=[$13])
              LogicalFilter(condition=[AND(=($0, $20), =($26, 2001))])
                LogicalJoin(condition=[true], joinType=[inner])
                  LogicalTableScan(table=[[tpcds, STORE_RETURNS]])
                  LogicalTableScan(table=[[tpcds, DATE_DIM]])
})), =($3, $1), =(CAST($27):VARCHAR(30) NOT NULL, 'distmember(fips_county, 61, 3)'), =($0, $32))], variablesSet=[[$cor0]])
        LogicalJoin(condition=[true], joinType=[inner])
          LogicalJoin(condition=[true], joinType=[inner])
            LogicalProject(CTR_CUSTOMER_SK=[$0], CTR_STORE_SK=[$1], CTR_TOTAL_RETURN=[$2])
              LogicalAggregate(group=[{0, 1}], CTR_TOTAL_RETURN=[SUM($2)])
                LogicalProject(SR_CUSTOMER_SK=[$3], SR_STORE_SK=[$7], SR_RETURN_AMT_INC_TAX=[$13])
                  LogicalFilter(condition=[AND(=($0, $20), =($26, 2001))])
                    LogicalJoin(condition=[true], joinType=[inner])
                      LogicalTableScan(table=[[tpcds, STORE_RETURNS]])
                      LogicalTableScan(table=[[tpcds, DATE_DIM]])
            LogicalTableScan(table=[[tpcds, STORE]])
          LogicalTableScan(table=[[tpcds, CUSTOMER]])
```

For TPC-H query 17 which works now the Calcite explain coming from Substrait looks like this. It has a different pattern not hitting the Calcite bug:
```
LogicalProject(AVG_YEARLY=[$0])
  LogicalProject($f1=[/($0, 7.0:DECIMAL(2, 1))])
    LogicalAggregate(group=[{}], agg#0=[SUM($0)])
      LogicalProject(L_EXTENDEDPRICE0=[$5])
        LogicalFilter(condition=[AND(=($16, $1), =(CAST($19):VARCHAR(10), 'Brand#13'), =(CAST($22):VARCHAR(10), 'JUMBO CAN'), <($4, CAST($SCALAR_QUERY({
LogicalProject($f1=[*(0.2:DECIMAL(2, 1), $0)])
  LogicalAggregate(group=[{}], agg#0=[AVG($0)])              <-- LogicalAggregate indirectly on top of
    LogicalProject(L_QUANTITY0=[$4])
      LogicalFilter(condition=[=($1, $cor0.P_PARTKEY)])       <-- LogicalFilter with correl var
        LogicalTableScan(table=[[LINEITEM]])
})):DECIMAL(38, 0)))], variablesSet=[[$cor0]])
          LogicalJoin(condition=[true], joinType=[inner])
            LogicalTableScan(table=[[LINEITEM]])
            LogicalTableScan(table=[[PART]])
```

nielspardon:
> We've only have tests that go from Substrait -> Calcite. Does this work going the other way? Not something that I think needs to happen as part of this PR, but possibly something to document as an issue.

we have test coverage since we round trip for the TPC-H / TPC-DS queries but we can also add unit tests from Calcite to Substrait

nielspardon:
> SortRels also contain expressions which could contain subqueries. Would the conversion just blow up now if we encountered that?

Good question. My goal was to get to the bottom of why TPC-H query 17 didn't convert back to Calcite in order to wrap up converting all 22 TPC-H queries successfully. Turns out I opened a huge can of worms which I would like to address step by step. In my opinion this PR is a step in the right direction but definitely not a fix-it-all.

core/src/main/java/io/substrait/dsl/SubstraitBuilder.java
nielspardon: done

isthmus/src/test/java/io/substrait/isthmus/expression/SubqueryConversionTest.java
nielspardon: done

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
nielspardon: I changed the code by moving the `RangeMap` creation into the `Context` class. We don't need the full outer input relations but only the row types so I changed the method name to `pushOuterRowType` which is more accurate. I also added javadoc comments for all the methods in the `Context` class which should make it a bit easier to understand what's going on.

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
nielspardon: I improved the naming

isthmus/src/test/java/io/substrait/isthmus/TpcdsQueryTest.java
nielspardon: I found the issue: https://github.com/apache/calcite/pull/4481

nielspardon: the Calcite fix has been merged meanwhile so we should get the TPC-DS queries working again with the next Calcite release

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
```diff
@@ -502,9 +525,101 @@ private RelNode applyRemap(RelNode relNode, Rel.Remap remap) {
...
       return new Context();
     }
+
+    /**
+     * Adds the outer row types to the top of the stack of outer row types.
+     *
+     * <p>Row types are stored as a {@link RangeMap} with field indices as keys and the {@link
+     * RelDataType} row type containing the field at the field index by continuously numbering the
+     * field indices from 0 across all provided row types in the order the row types are passed as
+     * arguments.
```
vbarua, line 675: The RangeMap is super nice for this ✨

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
```diff
@@ -502,9 +525,101 @@ private RelNode applyRemap(RelNode relNode, Rel.Remap remap) {
...
+     *
+     * @param inputs the row types to add
+     */
+    public void pushOuterRowType(final RelDataType... inputs) {
```
vbarua, line 679: `pushOuterRowType` is a good name for this. We should update the `popParentRelNodes` to match.

isthmus/src/test/java/io/substrait/isthmus/TpcdsQueryTest.java
vbarua: > but as direct field references and they just happened to not throw an exception during conversion.

Ah. Classic. Thanks for digging into this 🙇

vbarua (approved):
Once thing I would like changed before merging is the name of `popParentRelNodes`, which should probably be `popOuterRowType`. But other than that, these changes look good. As you point out
> this PR is a step in the right direction but definitely not a fix-it-all.

and I'm happy to have this merged at this point and iterate upon it in future PRs.

PR #427, opened 2025-06-27
feat(core): adds context to RelVisitor, ExpressionVisitor and FuncArgVisitor

nielspardon:
BREAKING CHANGE: adds a new context argument to RelVisitor, ExpressionVisitor and FuncArgVisitor

Preparation for #426


core/src/main/java/io/substrait/relation/RelVisitor.java
```diff
@@ -4,56 +4,56 @@
 import io.substrait.relation.physical.MergeJoin;
 import io.substrait.relation.physical.NestedLoopJoin;
 
-public interface RelVisitor<OUTPUT, EXCEPTION extends Exception> {
-  OUTPUT visit(Aggregate aggregate) throws EXCEPTION;
+public interface RelVisitor<O, C, E extends Exception> {
```
vbarua: One thing that might be nice here is to have a tagging interface like
```java
interface VisitationContext { }
```
and have a bound like `C extends VisitationContext` along with a class like
```java
class EmptyVisitationContext extends VisitationContext
```
with a default instance to avoid having `null` everywhere.

This does just end up being equivalent to the `Void` with extra steps, but the interface would give us somewhere to attach documentation to and the bound would help users understand the intent a bit better.

vbarua (commented):
Had a chance to take a look, and have some high-level notes.

I'm generally a fan of this, but I do worry that it could introduce a lot of noise for visitations that don't require context.

Your usage in [`a2f0c8e` (#426)](https://github.com/substrait-io/substrait-java/pull/426/commits/a2f0c8e38d4d5ef1b4e65c9d4b1e470c57244e4c) doesn't _strictly_ require a context parameter like you're introducing, as you could achieve a similar result by just pushing the context before calls to accept, and popping it after. _However_, that is more error prone and having it as parameter makes it easier to understand the dataflow and makes it harder for users to not propagate the context because they're forced to.

I'm going to ping the Slack to get additional eyes for this, because it does end up being an fairly intrusive API change and I'd like to hear what others think.

core/src/main/java/io/substrait/relation/RelVisitor.java
nielspardon: Sure, we can do that.

nielspardon:
> I'm going to ping the Slack to get additional eyes for this, because it does end up being an fairly intrusive API change and I'd like to hear what others think.

make sense. that's why I tagged it as a breaking change since users of substrait-java might be implementing the current visitor and it might be nice to get informed in the release notes that such an intrusive API change has happened.

core/src/main/java/io/substrait/relation/RelVisitor.java
nielspardon: I added the `VisitationContext` as suggested

core/src/test/java/io/substrait/type/proto/GenericRoundtripTest.java
```diff
@@ -57,7 +57,7 @@ public void roundtripTest(Method m, List<Object> paramInst, UnsupportedTypeGener
             null,
             EMPTY_TYPE,
             new ProtoRelConverter(new ExtensionCollector(), defaultExtensionCollection));
-    assertEquals(val, from.from(val.accept(to)));
+    assertEquals(val, from.from(val.accept(to, null)));
```
vbarua: I would prefer to use an instance of the empty context here, and in other places we use null, but also that's super minor.

vbarua (approved):
We've had this open for a bit and haven't had any other feedback, so I'm prepared to merge it this week.

My original plan had been to wait for https://github.com/substrait-io/substrait-java/pull/426 to be ready as well to merge both of them at the same time and minimize API churn, however I think that might open us up to merge conflicts.

Something I think we could do to address would be to bring the `SubstraitRelNodeConverter.Context` class into this PR, just like
```java
public static class Context {
    public static Context newContext() {
      return new Context();
    }
}
```
and update the `SubstraitRelNodeConverter` to use it:
```java
public class SubstraitRelNodeConverter
    extends AbstractRelVisitor<RelNode, SubstraitRelNodeConverter.Context, RuntimeException>
```

This way, we only break the API once with this merge, and then https://github.com/substrait-io/substrait-java/pull/426 will add the functionality to the Context.

What do you think @nielspardon?

core/src/test/java/io/substrait/type/proto/GenericRoundtripTest.java
nielspardon: done

nielspardon:
I pushed a few more of the API changes from including the `SubstraitRelNodeConverter.Context` in this PR, especially for the unit test class `accept()` calls which means #426 is now super clean.

isthmus/src/main/java/io/substrait/isthmus/SubstraitToCalcite.java
```diff
@@ -176,13 +178,13 @@ private NamedStructGatherer() {
 
     public static Map<List<String>, NamedStruct> gatherTables(Rel rel) {
       var visitor = new NamedStructGatherer();
-      rel.accept(visitor);
+      rel.accept(visitor, null);
```
vbarua, line 181: We should update all the usage of null to use the EmptyVisitationContext.INSTANCE you introduced, but we can do that as a follow-up.

vbarua (approved):
There's still a couple of usages of `null` where we could use the EmptyVisitationContext instead, but that can be handled as a follow-up change.

For the future, if you could avoid squashing and force pushing PR feedback, it would help with reviewing. Specifically, if you make make changes based on PR feedback their own commit, I can just look at the commit during review. We squash merge into main, so they won't show up in the final history anyway.

isthmus/src/main/java/io/substrait/isthmus/SubstraitToCalcite.java
nielspardon: I created #434 for this

PR #428, opened 2025-07-01
Add support of ddl and dml sql statements to SqlToSubstrait

ZorinAnton:
DDL:

For the above mentioned sql statements create the following relations:
- `NamedWrite` for `create table as`
- `NamedDdl` for `create view`
All other types of sql statements are handled as before.

DML:

The class `SubstraitRelVisitor` now supports the following operations for the `TableModify` node:
- insert
- update
- delete

### DIFF
--- a/isthmus/src/main/java/io/substrait/isthmus/SqlToSubstrait.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/SqlToSubstrait.java
@@ -1,9 +1,12 @@
 package io.substrait.isthmus;
 
 import com.google.common.annotations.VisibleForTesting;
+import io.substrait.isthmus.expression.DdlRelBuilder;
+import io.substrait.plan.ImmutablePlan;
 import io.substrait.plan.Plan.Version;
 import io.substrait.plan.PlanProtoConverter;
 import io.substrait.proto.Plan;
+import java.util.LinkedList;
 import java.util.List;
 import org.apache.calcite.plan.hep.HepPlanner;
 import org.apache.calcite.plan.hep.HepProgram;
@@ -12,6 +15,7 @@ import org.apache.calcite.prepare.Prepare;
 import org.apache.calcite.rel.RelRoot;
 import org.apache.calcite.schema.Schema;
 import org.apache.calcite.sql.SqlNode;
+import org.apache.calcite.sql.SqlNodeList;
 import org.apache.calcite.sql.parser.SqlParseException;
 import org.apache.calcite.sql.parser.SqlParser;
 import org.apache.calcite.sql.validate.SqlValidator;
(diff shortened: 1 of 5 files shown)


isthmus/src/main/java/io/substrait/isthmus/SqlToSubstrait.java
```diff
@@ -53,32 +57,84 @@ List<RelRoot> sqlToRelNode(String sql, List<String> tables) throws SqlParseExcep
...
+      String sql,
+      SqlValidator validator,
+      Prepare.CatalogReader catalogReader,
+      ImmutablePlan.Builder builder)
+      throws SqlParseException {
+    SqlParser parser = SqlParser.create(sql, parserConfig);
+    var parsedList = parser.parseStmtList();
+    SqlToRelConverter converter = createSqlToRelConverter(validator, catalogReader);
+    // IMPORTANT: parsedList gets filtered in the call below
+    List<io.substrait.plan.Plan.Root> ddlRelRoots = ddlSqlToRootNodes(parsedList, converter);
```
vbarua, line 89: Do we need to process DDL statements separately? Would it be possible to just parse _all_ statements, and then just use the conversion code in SubstraitRelVisitor?

vbarua (commented):
This PR motivated me to continue some ongoing work for https://github.com/substrait-io/substrait-java/issues/362, specifically https://github.com/substrait-io/substrait-java/pull/430.

Effectively, the plan is to break SqlToSubstrait down into a conversion of SQL into Calcite, and then Calcite into Substrait. The `DdlRelBuilder` you've introduced goes against this work because it couples the SQL to the Substrait translation. It also seems somewhat duplicative, because I think we should be able to use the Calcite machinary to convert DDL SqlNodes into RelNodes, and then all we would need would be your SubstraitRelVisitor updates.

ZorinAnton:
I think it's a good idea to break down SqlToSubstrait into Sql->Calcite and Calcite->Substrait, as it will give better flexibility.
However, DDL statements will probably require a special processing. Calcite can parse them, but it doesn't translate them to Relational Algebra, that's why I created DdlRelBuilder that translates SqlNode directly to substrait classes. SqlValidator also doesn't support DDL statements and I wasn't sure if it makes sense to add such a support.

vbarua:
>DDL statements will probably require a special processing. Calcite can parse them, but it doesn't translate them to Relational Algebra

Ah, TIL. Poking around your PR, the DML statements `INSERT`, `UPDATE` and `DELETE` get turned into a LogicalTableModify relations, but the DDL statements don't have Calcite equivalents. I guess that's because Calcite expects/wants tables and views to be available in the catalog, so it doesn't add them.

I want to think more about the DDL API, and honestly kind of want to land it _after_ https://github.com/substrait-io/substrait-java/pull/430.

That being said, would you be open to splitting the DML and DDL processing into separate PRs? The DML works is pretty straightforward and should be easier to review on it's own.

PR #429, opened 2025-07-04
feat(spark): support insert/append operations

andrew-coleman:
Partial implementation of the substrait WriteRel relation in order to support conversion of Spark insert and append operations.

Added tests for inserting to in-memory database tables and appending to local files.

A few things to note / discuss:
- Writing to the filesystem requires information of file path, format, etc.  To do this, I used the `ExtensionWrite` relation, and embedded a `ReadRel.LocalFiles.FileOrFiles` message as the extension detail.  This then requires the use of the extension's ProtoRelConverter to unpack the proto message.
- Since this might be a common scenario, it might be cleaner to extend the `WriteRel` proto message to also support a `LocalFiles` write_type in the same way that the `ReadRel` has a `LocalFiles` read_type. What do you think?  
- Would it also make sense to make `FileOrFiles` a top-level message so that it could be re-used in this way?

### DIFF
--- a/spark/src/main/scala/io/substrait/spark/logical/ToLogicalPlan.scala
+++ b/spark/src/main/scala/io/substrait/spark/logical/ToLogicalPlan.scala
@@ -16,19 +16,21 @@
  */
 package io.substrait.spark.logical
 
-import io.substrait.spark.{DefaultRelVisitor, SparkExtension, ToSparkType, ToSubstraitType}
+import io.substrait.spark.{DefaultRelVisitor, FileHolder, SparkExtension, ToSparkType, ToSubstraitType}
 import io.substrait.spark.expression._
 
-import org.apache.spark.sql.SparkSession
-import org.apache.spark.sql.catalyst.InternalRow
+import org.apache.spark.sql.{SaveMode, SparkSession}
+import org.apache.spark.sql.catalyst.{InternalRow, TableIdentifier}
 import org.apache.spark.sql.catalyst.analysis.{caseSensitiveResolution, MultiInstanceRelation, UnresolvedRelation}
+import org.apache.spark.sql.catalyst.catalog.{CatalogStorageFormat, CatalogTable, CatalogTableType}
 import org.apache.spark.sql.catalyst.expressions._
 import org.apache.spark.sql.catalyst.expressions.aggregate.{AggregateExpression, AggregateFunction}
 import org.apache.spark.sql.catalyst.plans.{FullOuter, Inner, LeftAnti, LeftOuter, LeftSemi, RightOuter}
 import org.apache.spark.sql.catalyst.plans.logical._
 import org.apache.spark.sql.catalyst.util.toPrettySQL
 import org.apache.spark.sql.execution.QueryExecution
-import org.apache.spark.sql.execution.datasources.{HadoopFsRelation, InMemoryFileIndex, LogicalRelation}
+import org.apache.spark.sql.execution.command.{CreateDataSourceTableAsSelectCommand, DataWritingCommand, LeafRunnableCommand}
+import org.apache.spark.sql.execution.datasources.{FileFormat => SparkFileFormat, HadoopFsRelation, InMemoryFileIndex, InsertIntoHadoopFsRelationCommand, LogicalRelation, V1Writes}
 import org.apache.spark.sql.execution.datasources.csv.CSVFileFormat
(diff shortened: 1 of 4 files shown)


spark/src/main/scala/io/substrait/spark/logical/ToLogicalPlan.scala
```diff
@@ -397,23 +421,94 @@ class ToLogicalPlan(spark: SparkSession = SparkSession.builder().getOrCreate())
...
-      isStreaming = false
+  }
+
+  override def visit(write: NamedWrite): LogicalPlan = {
+    val child = write.getInput.accept(this)
+    val names = write.getNames
+    if (names.size() < 3) {
+      throw new UnsupportedOperationException(
+        s"NamedWrite requires three names (catalog, database, table): $names")
+    }
```
nielspardon: Does Spark always require exactly 3 names? Shouldn't the if then check for exactly 3 names?

andrew-coleman: No, Spark only needs the table name, and optionally the db and catalog names.
The tree name requirement comes from the equivalent ReadRel [code](https://github.com/substrait-io/substrait-java/blob/main/spark/src/main/scala/io/substrait/spark/logical/ToSubstraitRel.scala#L511) that creates the array of three names by splitting the identifier `catalog.database.table`. So I thought I'd make the WriteRel consistent with this.

Having said that, the Substrait [proto definition](https://github.com/substrait-io/substrait/blob/main/proto/substrait/algebra.proto#L594) gives the example `mydb.mytable` (two names), so I've updated the code to be more tolerant.

nielspardon: I thought that the Substrait specification was way more tolerant on how many names one may use for identifying a table so I was curious where the requirement was coming from and the conclusion is there is a constraint in Spark

spark/src/main/scala/io/substrait/spark/logical/ToLogicalPlan.scala
```diff
@@ -409,23 +433,98 @@ class ToLogicalPlan(spark: SparkSession = SparkSession.builder().getOrCreate())
...
+    val child = write.getInput.accept(this, context)
+    val names = write.getNames
+    val (table, database, catalog) = names.size match {
+      case 1 => (names.get(0), None, None)
+      case 2 => (names.get(1), Some(names.get(0)), None)
+      case 3 => (names.get(2), Some(names.get(1)), Some(names.get(0)))
+      case _ =>
+        throw new UnsupportedOperationException(
+          s"NamedWrite requires up to three names ([[catalog,] database,] table): $names")
+    }
```
bestbeforetoday: I'm not sure it is better but I _think_ you could avoid the index access and give the elements suitable names with:

```scala
    val (table, database, catalog) = names.asScala.toList match {
      case List(table) => (table, None, None)
      case List(database, table) => (table, Some(database), None)
      case List(catalog, database, table) => (table, Some(database), Some(catalog))
      case _ =>
          s"NamedWrite requires up to three names ([[catalog,] database,] table): $names")
    }
```

bestbeforetoday (approved):
Looks good to me.

spark/src/main/scala/io/substrait/spark/logical/ToLogicalPlan.scala
andrew-coleman: Good suggestion, thanks.
I've put in a slight variation on this ;)

bestbeforetoday: Yep, that's nicer 👍

bestbeforetoday (approved):

spark/src/main/scala/io/substrait/spark/logical/ToLogicalPlan.scala
```diff
@@ -501,6 +600,9 @@ class ToLogicalPlan(spark: SparkSession = SparkSession.builder().getOrCreate())
           aggregate.groupingExpressions,
           renameAndCastExprs(aggregate.aggregateExpressions),
           aggregate.child)
+      // if the plan represents a 'write' command, then leave as is
```
Blizzara, line 603: This makes sense, but would we need to instead do the renaming at the write node? Or does that happen automatically due to `outputColumnNames = write.getTableSchema.names.asScala` or something?

PR #430, opened 2025-07-05
feat(isthmus): additional parsing and unparsing utils for SQL

vbarua:
The SubstraitCreateStatementParser can be used to turn one or more SQL CREATE
statements into a CalciteCatalogReader

It is based on internal logic that was used to convert CREATE statements in
methods like SqlToSubstrait#execute(String sql, List<String> tables) into
catalogs for processing

This method has been introduced to allow users to access this logic so that they
can provide Isthmus with catalogs directly, enabling the removal of methods that
performed extra processing on user input to produce catalogs

The generation of catalogs is a Calcite concern, and as such is considered out
of scope of the API for Isthmus. SubstraitCreateStatementParser is provided
as a convenience to users.

The SubstraitSqlDialect is now provided as a convenience for producing SQL from
a Calcite RelNode. It replaces SubstraitToSql#toSql(RelNode root).

BREAKING CHANGE: removed SqlToSubstrait#execute(String sql, List<String> tables) 
BREAKING CHANGE: removed SqlToSubstrait#execute(String sql, String name, Schema schema) 
BREAKING CHANGE: removed SubstraitToSql#substraitRelToCalciteRel(Rel relRoot, List<String> tables)
BREAKING CHANGE: removed SubstraitToSql#toSql(RelNode root)


vbarua:
This work is being done as part of https://github.com/substrait-io/substrait-java/issues/362

isthmus/src/test/java/io/substrait/isthmus/PlanTestBase.java
```diff
@@ -46,45 +53,45 @@ public static String asString(String resource) throws IOException {
...
-              .collect(Collectors.toList());
+  protected static CalciteCatalogReader TPCH_CATALOG;
+
+  static {
+    try {
+      List<String> tpchCreateStatements = List.of(asString("tpch/schema.sql").split(";"));
+      TPCH_CATALOG =
+          SubstraitCreateStatementParser.processCreateStatementsToCatalog(tpchCreateStatements);
+    } catch (IOException | SqlParseException e) {
+      throw new RuntimeException(e);
```
vbarua, line 64: Instead of just loading the TPCH schema strings, we can turn them into a catalog as well. This allows us to remove a bunch of boilerplate in various tests.

isthmus/src/test/java/io/substrait/isthmus/PlanTestBase.java
```diff
@@ -277,6 +299,18 @@ protected String toSql(Plan plan) {
...
+  protected static CalciteCatalogReader schemaToCatalog(String schemaName, Schema schema) {
+    CalciteSchema rootSchema = CalciteSchema.createRootSchema(false);
+    rootSchema.add(schemaName, schema);
+    List<String> defaultSchema = List.of(schemaName);
+    return new CalciteCatalogReader(
+        rootSchema,
+        defaultSchema,
+        new JavaTypeFactoryImpl(SubstraitTypeSystem.TYPE_SYSTEM),
+        CalciteConnectionConfig.DEFAULT.set(
+            CalciteConnectionProperty.CASE_SENSITIVE, Boolean.FALSE.toString()));
```
vbarua, line 314: This code is, admittedly, somewhat misplaced. It's effectively the extracted code of `SqlToSubstrait#execute(String sql, String name, Schema schema)` for use in 2 places.

I did not create a public alternative to this as I don't think it's used externally, and if it is it's easy to recreate, unlike the processing for turning SQL CREATE statements into a catalog.

isthmus/src/test/java/io/substrait/isthmus/OneWaySimpleTest.java
```diff
@@ -1,35 +0,0 @@
...
-import com.google.protobuf.util.JsonFormat;
-import io.substrait.proto.Plan;
-import java.io.IOException;
-import java.util.List;
-import org.apache.calcite.sql.parser.SqlParseException;
-import org.junit.jupiter.api.Test;
-
-public class OneWaySimpleTest extends PlanTestBase {
-  @Test
-  public void windowFunctionsSql2Sub() throws IOException, SqlParseException {
```
vbarua, line 14: The tests in `WindowFunctionTest` effectively verify this, and much more thoroughly so.

isthmus/src/main/java/io/substrait/isthmus/SubstraitToSql.java
```diff
@@ -1,65 +1,16 @@
...
-import org.apache.calcite.sql.SqlWriterConfig;
-import org.apache.calcite.sql.parser.SqlParseException;
 
 public class SubstraitToSql extends SqlConverterBase {
 
   public SubstraitToSql() {
     super(FEATURES_DEFAULT);
   }
 
-  public RelNode substraitRelToCalciteRel(Rel relRoot, List<String> tables)
```
vbarua, line 21: Users can now use the `SubstraitCreateStatementParser` to creates a catalog directly.

isthmus/src/main/java/io/substrait/isthmus/SubstraitToSql.java
```diff
@@ -1,65 +1,16 @@
...
-      new SqlDialect(SqlDialect.EMPTY_CONTEXT) {
-        @Override
-        public boolean supportsApproxCountDistinct() {
-          return true;
-        }
-      };
-
-  public static String toSql(RelNode root) {
-    return toSql(root, DEFAULT_SQL_DIALECT);
-  }
```
vbarua, line 46: SqlSubstraitDialect is now provided as a utility for generating SQL

isthmus/src/main/java/io/substrait/isthmus/SqlConverterBase.java
```diff
@@ -71,106 +52,4 @@ protected SqlConverterBase(FeatureBoard features) {
...
-
-  protected static SqlParseException fail(String text, SqlParserPos pos) {
-    return new SqlParseException(text, pos, null, null, new RuntimeException("fake lineage"));
-  }
-
-  protected static SqlParseException fail(String text) {
-    return fail(text, SqlParserPos.ZERO);
-  }
-
-  protected static final class Validator extends SqlValidatorImpl {
```
vbarua, line 159: This class was extracted into SubstraitSqlValidator.

isthmus/src/main/java/io/substrait/isthmus/SqlToSubstrait.java
```diff
@@ -29,27 +28,15 @@ public SqlToSubstrait(FeatureBoard features) {
     super(features);
   }
 
-  public Plan execute(String sql, List<String> tables) throws SqlParseException {
-    CalciteCatalogReader catalogReader = registerCreateTables(tables);
-    SqlValidator validator = Validator.create(factory, catalogReader, SqlValidator.Config.DEFAULT);
-    return executeInner(sql, validator, catalogReader);
-  }
```
vbarua, line 36: Users can now use the SubstraitCreateStatementParser to creates a catalog directly.

vbarua:
cc: @andrew-coleman @bestbeforetoday @nielspardon

isthmus/src/main/java/io/substrait/isthmus/SqlExpressionToSubstrait.java
```diff
@@ -140,10 +142,10 @@ private Result registerCreateTablesForExtendedExpression(List<String> tables)
     CalciteSchema rootSchema = CalciteSchema.createRootSchema(false);
     CalciteCatalogReader catalogReader =
         new CalciteCatalogReader(rootSchema, List.of(), factory, config);
-    SqlValidator validator = Validator.create(factory, catalogReader, SqlValidator.Config.DEFAULT);
     if (tables != null) {
       for (String tableDef : tables) {
-        List<SubstraitTable> tList = parseCreateTable(factory, validator, tableDef);
+        List<SubstraitTable> tList =
+            SubstraitCreateStatementParser.processCreateStatements(tableDef);
```
nielspardon, line 148: this method might (in a future PR) also be a candidate for refactoring. I'm wondering if we really need the variant of `processCreateStatements()` returning a list of Substrait tables to be public or whether the logic of the `registerCreateTablesForExtendedExpression` method could be rewritten to work with the `CatalogReader` instead of the list of tables.

isthmus/src/main/java/io/substrait/isthmus/sql/SubstraitCreateStatementParser.java
```diff
@@ -0,0 +1,136 @@
...
+import org.apache.calcite.sql.ddl.SqlCreateTable;
+import org.apache.calcite.sql.ddl.SqlKeyConstraint;
+import org.apache.calcite.sql.parser.SqlParseException;
+import org.apache.calcite.sql.parser.SqlParser;
+import org.apache.calcite.sql.parser.SqlParserPos;
+import org.apache.calcite.sql.parser.ddl.SqlDdlParserImpl;
+import org.apache.calcite.sql.validate.SqlConformanceEnum;
+import org.apache.calcite.sql.validate.SqlValidator;
+
+/** Utility class for parsing CREATE statements into a {@link CalciteSchema} */
```
nielspardon: wouldn't this be more accurate?
```suggestion
/** Utility class for parsing CREATE statements into a {@link CalciteCatalogReader} */
```

isthmus/src/main/java/io/substrait/isthmus/sql/SubstraitCreateStatementParser.java
```diff
@@ -0,0 +1,136 @@
...
+          CalciteSchema.createRootSchema(false), List.of(), TYPE_FACTORY, CONNECTION_CONFIG);
+
+  // A validator is needed to convert the types in column declarations to Calcite types
+  private static final SqlValidator VALIDATOR =
+      new SubstraitSqlValidator(
+          // as we are validating CREATE statements, an empty catalog suffices
+          EMPTY_CATALOG);
+
+  /**
+   * @param createStatements a SQL string containing only CREATE statements
```
nielspardon, line 59: for a valid javadoc it would make sense to also have a method description
```suggestion
   * Parses a SQL string containing only CREATE statements into a list of {@link SubstraitTable}s.
   *
   * @param createStatements a SQL string containing only CREATE statements
```

isthmus/src/main/java/io/substrait/isthmus/sql/SubstraitCreateStatementParser.java
```diff
@@ -0,0 +1,136 @@
...
+      tableList.add(
+          new SubstraitTable(
+              create.name.names.get(0), TYPE_FACTORY.createStructType(columnTypes, names)));
+    }
+
+    return tableList;
+  }
+
+  /**
+   * @param createStatements a SQL string containing only CREATE statements
```
nielspardon, line 116:
```suggestion
   * Parses a list of SQL strings containing only CREATE statements into a {@link CalciteCatalogReader}.
   *
   * @param createStatements a list of SQL strings containing only CREATE statements
```

isthmus/src/test/java/io/substrait/isthmus/KeyConstraintsTest.java
```diff
@@ -12,11 +13,8 @@ public class KeyConstraintsTest extends PlanTestBase {
   public void tpcds(int query) throws Exception {
     SqlToSubstrait s = new SqlToSubstrait();
     String[] values = asString("keyconstraints_schema.sql").split(";");
```
nielspardon: since the `public static List<SubstraitTable> processCreateStatements(String createStatements)` method seems to be able to parse multiple CREATE statements from a single string I'm wondering whether we really need to split the schema SQL file content anymore. I do understand we are missing a `public static CatalogReader processCreateStatements(String createStatements)` method to properly replace this.

isthmus/src/test/java/io/substrait/isthmus/NameRoundtripTest.java
```diff
@@ -3,16 +3,20 @@
...
 import org.junit.jupiter.api.Test;
 
 public class NameRoundtripTest extends PlanTestBase {
 
   @Test
   void preserveNamesFromSql() throws Exception {
-    List<String> creates = List.of("CREATE TABLE foo(a BIGINT, b BIGINT)");
+    List<String> createStatements = List.of("CREATE TABLE foo(a BIGINT, b BIGINT)");
+    CalciteCatalogReader catalogReader =
+        SubstraitCreateStatementParser.processCreateStatementsToCatalog(createStatements);
```
nielspardon: couldn't we skip wrapping a single create statement into a `List.of()` if `processCreateStatementsToCatalog()` accepted a varargs argument instead of a list?

isthmus/src/test/java/io/substrait/isthmus/PlanTestBase.java
```diff
@@ -46,45 +53,45 @@ public static String asString(String resource) throws IOException {
...
-      } catch (IOException e) {
-        throw new UncheckedIOException(e);
-      }
-
-      createStatements =
-          Arrays.stream(values)
-              .map(String::trim)
-              .filter(s -> !s.isBlank())
-              .collect(Collectors.toList());
+  protected static CalciteCatalogReader TPCH_CATALOG;
```
nielspardon: this can also be `final`
```suggestion
  protected static final CalciteCatalogReader TPCH_CATALOG;
```

isthmus/src/test/java/io/substrait/isthmus/PlanTestBase.java
nielspardon: maybe we should have a `TpcDsCatalogReader` and a `TpcHCatalogReader` class for testing which could encapsulate such a logic

nielspardon (approved):
I like the change in general. just added some thoughts/minor improvements

isthmus/src/test/java/io/substrait/isthmus/PlanTestBase.java
vbarua: There is one instance of a test that fully creates a custom schema
https://github.com/substrait-io/substrait-java/blob/b53aed903f9785d06a1fb5c0582711fd743bc5f1/isthmus/src/test/java/io/substrait/isthmus/NestedStructQueryTest.java#L60-L61
otherwise I would just expose those 2 schemas as suggested.

There might be a better way to handle this in the future, but I'm not too worried about it right now because this ends up being an internal API we can tweak however we want.

isthmus/src/main/java/io/substrait/isthmus/SqlExpressionToSubstrait.java
vbarua: Indeed, I was focusing on SqlToSubstrait here, but SqlExpressionToSubstrait should eventually get the same treatment.

isthmus/src/test/java/io/substrait/isthmus/NameRoundtripTest.java
vbarua: Good idea. I switched the method to use varargs. This did required some additional test changes, but I think it's better overall.

PR #431, opened 2025-07-08
feat(core,isthmus): add dml support to SqlToSubstrait

ZorinAnton:
As discussed, splitting the PR https://github.com/substrait-io/substrait-java/pull/428 into dml and ddl parts. This PR is for dml support.

The class `SubstraitRelVisitor` now supports the following operations for the `TableModify` node:

- insert
- update
- delete

### DIFF
--- a/isthmus/src/main/java/io/substrait/isthmus/SubstraitRelVisitor.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/SubstraitRelVisitor.java
@@ -11,6 +11,7 @@ import io.substrait.isthmus.expression.RexExpressionConverter;
 import io.substrait.isthmus.expression.ScalarFunctionConverter;
 import io.substrait.isthmus.expression.WindowFunctionConverter;
 import io.substrait.plan.Plan;
+import io.substrait.relation.AbstractWriteRel;
 import io.substrait.relation.Aggregate;
 import io.substrait.relation.Cross;
 import io.substrait.relation.EmptyScan;
@@ -18,6 +19,8 @@ import io.substrait.relation.Fetch;
 import io.substrait.relation.Filter;
 import io.substrait.relation.Join;
 import io.substrait.relation.NamedScan;
+import io.substrait.relation.NamedUpdate;
+import io.substrait.relation.NamedWrite;
 import io.substrait.relation.Project;
 import io.substrait.relation.Rel;
 import io.substrait.relation.Set;
@@ -37,6 +40,7 @@ import org.apache.calcite.rel.RelNode;
 import org.apache.calcite.rel.RelRoot;
 import org.apache.calcite.rel.core.AggregateCall;
 import org.apache.calcite.rel.core.JoinRelType;
+import org.apache.calcite.rel.core.TableModify;
 import org.apache.calcite.rel.type.RelDataTypeFactory;
 import org.apache.calcite.rex.RexBuilder;
 import org.apache.calcite.rex.RexFieldAccess;
@@ -374,8 +378,76 @@ public class SubstraitRelVisitor extends RelNodeVisitor<Rel, RuntimeException> {
   }
 
   @Override
-  public Rel visit(org.apache.calcite.rel.core.TableModify modify) {
-    return super.visit(modify);
+  public Rel visit(TableModify modify) {
+    switch (modify.getOperation()) {
+      case INSERT:
+      case DELETE:
+        {
+          final Rel input = apply(modify.getInput());
+          final AbstractWriteRel.WriteOp op =
+              modify.getOperation() == TableModify.Operation.INSERT
+                  ? AbstractWriteRel.WriteOp.INSERT
+                  : AbstractWriteRel.WriteOp.DELETE;
+
+          assert modify.getTable() != null;
+          return NamedWrite.builder()
+              .input(input)
+              .tableSchema(typeConverter.toNamedStruct(modify.getTable().getRowType()))
+              .operation(op)
+              .createMode(AbstractWriteRel.CreateMode.UNSPECIFIED)
+              .outputMode(AbstractWriteRel.OutputMode.MODIFIED_RECORDS)
+              .names(modify.getTable().getQualifiedName())
+              .build();
+        }
+
+      case UPDATE:
+        {
+          assert modify.getTable() != null;
+
+          Expression condition;
+          if (modify.getInput() instanceof org.apache.calcite.rel.core.Filter) {
+            org.apache.calcite.rel.core.Filter filter =
+                (org.apache.calcite.rel.core.Filter) modify.getInput();
+            condition = toExpression(filter.getCondition());
+          } else {
+            condition = Expression.BoolLiteral.builder().nullable(false).value(true).build();
+          }
+
+          List<String> updateColumnNames = modify.getUpdateColumnList();
+          List<RexNode> sourceExpressions = modify.getSourceExpressionList();
+          List<String> allTableColumnNames = modify.getTable().getRowType().getFieldNames();
+          List<NamedUpdate.TransformExpression> transformations = new ArrayList<>();
+
+          for (int i = 0; i < updateColumnNames.size(); i++) {
+            String colName = updateColumnNames.get(i);
+            RexNode rexExpr = sourceExpressions.get(i);
+
+            int columnIndex = allTableColumnNames.indexOf(colName);
+            if (columnIndex == -1) {
+              throw new IllegalStateException(
+                  "Updated column '" + colName + "' not found in table schema.");
+            }
+
+            Expression substraitExpr = toExpression(rexExpr);
+
+            transformations.add(
+                NamedUpdate.TransformExpression.builder()
+                    .columnTarget(columnIndex)
+                    .transformation(substraitExpr)
+                    .build());
+          }
+
+          return NamedUpdate.builder()
+              .tableSchema(typeConverter.toNamedStruct(modify.getTable().getRowType()))
+              .names(modify.getTable().getQualifiedName())
+              .condition(condition)
+              .transformations(transformations)
+              .build();
+        }
+
+      default:
+        return super.visit(modify);
+    }
   }
 
   @Override
--- a/core/src/main/java/io/substrait/relation/RelCopyOnWriteVisitor.java
+++ b/core/src/main/java/io/substrait/relation/RelCopyOnWriteVisitor.java
@@ -215,7 +215,15 @@ public class RelCopyOnWriteVisitor<E extends Exception>
 
   @Override
   public Optional<Rel> visit(NamedWrite write, EmptyVisitationContext context) throws E {
-    throw new UnsupportedOperationException();
+
+    Optional<Rel> input = write.getInput().accept(this, context);
+
+    if (allEmpty(input)) {
+      return Optional.empty();
+    }
+
+    return Optional.of(
+        NamedWrite.builder().from(write).input(input.orElse(write.getInput())).build());
   }
 
   @Override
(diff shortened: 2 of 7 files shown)


isthmus/src/test/java/io/substrait/isthmus/SqlToSubstraitTest.java
```diff
@@ -0,0 +1,37 @@
+package io.substrait.isthmus;
+
+import io.substrait.plan.Plan;
+import java.io.IOException;
+import java.util.List;
+import java.util.stream.Collectors;
+import org.apache.calcite.sql.parser.SqlParseException;
+import org.junit.jupiter.api.Test;
+
+public class SqlToSubstraitTest extends PlanTestBase {
```
vbarua: I would suggest naming this something like DmlRoundtripTest. While this is testing the SqlToSubstrait to some extent, that's not what we really care about.

isthmus/src/test/java/io/substrait/isthmus/SqlToSubstraitTest.java
```diff
@@ -0,0 +1,37 @@
...
+
+  @Test
+  void testDml() throws SqlParseException, IOException {
```
vbarua: We should split this into 1 test per statement type.

isthmus/src/test/java/io/substrait/isthmus/SqlToSubstraitTest.java
```diff
@@ -0,0 +1,37 @@
...
+    final String sqlStatements = asString("sqltosubstrait/sqltosubstrait.sql");
```
vbarua: Instead of using a resource file for this, I would just include the queries inline below. It avoids the indirection, makes the tests easier to read and is consistent with our other tests generally.

isthmus/src/test/java/io/substrait/isthmus/SqlToSubstraitTest.java
```diff
@@ -0,0 +1,37 @@
...
+                root ->
+                    SubstraitRelVisitor.convert(
+                        root, SqlConverterBase.EXTENSION_COLLECTION, sql2subst.featureBoard))
+            .collect(Collectors.toList());
+    var builder = io.substrait.plan.Plan.builder();
+    for (final io.substrait.plan.Plan.Root planRoot : relRoots) {
+      builder.addRoots(planRoot);
+    }
+    final Plan plan = builder.build();
+    assertPlanRoundtrip(plan);
```
vbarua: If we test each statement separately, we can also use
```java
protected void assertFullRoundTrip(String sqlQuery, List<String> createStatements)
```
which verifies the conversion of `POJO -> Proto -> POJO -> Calcite -> POJO`, which is a stronger test than `assertPlanRoundtrip` which only checks `POJO -> Proto -> POJO`.

That being said, that would introduce more work because you would need to handle more of the conversion, specifically the Substrait to Calcite conversion in SubstraitRelNodeConverter. If that's beyond the capacity you have available, let me know. A simpler improvement would be to use `assertProtoPlanRoundrip`, which still only checks the protobuf conversions but does more of the wiring.

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelVisitor.java
```diff
@@ -352,7 +354,44 @@ public Rel visit(org.apache.calcite.rel.core.Exchange exchange) {
...
+            .tableSchema(typeConverter.toNamedStruct(modify.getRowType()))
+            .operation(AbstractWriteRel.WriteOp.INSERT)
+            .createMode(AbstractWriteRel.CreateMode.UNSPECIFIED)
+            .outputMode(AbstractWriteRel.OutputMode.NO_OUTPUT)
+            .names(modify.getTable().getQualifiedName())
+            .build();
+      }
+      case UPDATE -> {
+        assert modify.getTable() != null;
+        yield NamedWrite.builder()
```
vbarua: From the `TableModify` itself, we also need to consider the `updateColumnList` and `sourceExpressionList` which, [from the docs](https://github.com/apache/calcite/blob/df3b438ef124838f338fbbd1866c6da61b036890/core/src/main/java/org/apache/calcite/rel/core/TableModify.java#L104-L107), should be set for the `UPDATE` operation.

```java
   * @param updateColumnList List of column identifiers to be updated
   *           (e.g. ident1, ident2); null if not UPDATE
   * @param sourceExpressionList List of value expressions to be set
   *           (e.g. exp1, exp2); null if not UPDATE
```

Bit of weirdness in the spec I've noticed from this as well. We have both a WriteRel with an update mode, and a dedicated UpdateRel in the specification. I'm not sure which one we want to use here. I'm asking the community for clarification on this in the Slack.

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelVisitor.java
```diff
@@ -352,7 +354,44 @@ public Rel visit(org.apache.calcite.rel.core.Exchange exchange) {
...
+    final Rel input = apply(modify.getInput());
+    return switch (modify.getOperation()) {
+      case INSERT -> {
+        assert modify.getTable() != null;
+        yield NamedWrite.builder()
+            .input(input)
+            .tableSchema(typeConverter.toNamedStruct(modify.getRowType()))
+            .operation(AbstractWriteRel.WriteOp.INSERT)
+            .createMode(AbstractWriteRel.CreateMode.UNSPECIFIED)
+            .outputMode(AbstractWriteRel.OutputMode.NO_OUTPUT)
```
vbarua: I _think_ that the Calcite TableModify relation does actually produce an output. The [return type](https://github.com/apache/calcite/blob/df3b438ef124838f338fbbd1866c6da61b036890/core/src/main/java/org/apache/calcite/rel/core/TableModify.java#L204-L207) for the TableModify relation is
```java
  @Override public RelDataType deriveRowType() {
    return RelOptUtil.createDmlRowType(
        SqlKind.INSERT, getCluster().getTypeFactory());
  }
```
and the docs for `createDmlRowType` indicate
```java
   * Creates the row type descriptor for the result of a DML operation, which
   * is a single column named ROWCOUNT of type BIGINT for INSERT;
   * a single column named PLAN for EXPLAIN.
```

So Calcite might return the row count by default

isthmus/src/test/java/io/substrait/isthmus/SqlToSubstraitTest.java
ZorinAnton: I'm trying to add things required for assertFullRoundTrip. One of them is to implement `visit` methods for NamedWrite and NamedUpdate (being discussed below) in the class `SubstraitRelNodeConverter`.  The `LogicalTable.create` method needs an instance of `Prepare.CatalogReader` that I'm not sure where to take from:

```java
public static LogicalTableModify create(RelOptTable table,
      Prepare.CatalogReader schema, RelNode input,
      Operation operation, @Nullable List<String> updateColumnList,
      @Nullable List<RexNode> sourceExpressionList, boolean flattened)
```

What would be your recommendations here? Pass it to the constructor and make it a private field of the class?
Thanks!

vbarua: Ah, of course there's no methods for the the DML and DDL relations in the RelBuilder 😢 

Looking at it, it's kind of a weird dependency for TableModify to have. Looking at the usage in [Calcite](https://github.com/apache/calcite/blob/dafb8d7a1fde884daf76d2a24aa6aea9c39ee5ca/core/src/main/java/org/apache/calcite/rel/core/TableModify.java#L224-L227)
```java
      inputRowType =
          typeFactory.createJoinType(rowType,
              getCatalogReader().createTypeFromProjection(rowType,
                  updateColumnList));
```
it's [calling into](https://github.com/apache/calcite/blob/dafb8d7a1fde884daf76d2a24aa6aea9c39ee5ca/core/src/main/java/org/apache/calcite/prepare/CalciteCatalogReader.java#L250-L254)
```java
  @Override public RelDataType createTypeFromProjection(final RelDataType type,
      final List<String> columnNameList) {
    return SqlValidatorUtil.createTypeFromProjection(type, columnNameList,
        typeFactory, nameMatcher.isCaseSensitive());
  }
```
which really doesn't need information specific to the catalog.

What I would suggest here is something like
```java
      RelOptTable table = relBuilder.getRelOptSchema().getTableForMember(write.getNames());
      Prepare.CatalogReader catalogReader = (Prepare.CatalogReader) table.getRelOptSchema();
```
which Calcite itself does in [TableModify.java ](https://github.com/apache/calcite/blob/dafb8d7a1fde884daf76d2a24aa6aea9c39ee5ca/core/src/main/java/org/apache/calcite/rel/core/TableModify.java#L152-L154)
```java
        requireNonNull(
            (Prepare.CatalogReader) input.getTable("table").getRelOptSchema(),
            "relOptSchema"),
```

This avoids needing to inject the `Prepare.CatalogReader`, and I _think_ should mostly work here. If it does cause problems now, or in the future, we can add the catalog reader to the API.

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelVisitor.java
```diff
@@ -375,7 +378,75 @@ public Rel visit(org.apache.calcite.rel.core.Exchange exchange) {
...
   @Override
   public Rel visit(org.apache.calcite.rel.core.TableModify modify) {
-    return super.visit(modify);
+    switch (modify.getOperation()) {
+      case INSERT:
+      case DELETE:
+        {
+          final Rel input = apply(modify.getInput());
+          final AbstractWriteRel.WriteOp op =
+              modify.getOperation() == org.apache.calcite.rel.core.TableModify.Operation.INSERT
```
nielspardon: it looks like for `org.apache.calcite.rel.core.TableModify` we don't have a name clash and could import this class

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelVisitor.java
```diff
@@ -375,7 +378,75 @@ public Rel visit(org.apache.calcite.rel.core.Exchange exchange) {
...
+          if (modify.getInput() instanceof org.apache.calcite.rel.core.Filter) {
+            org.apache.calcite.rel.core.Filter filter =
+                (org.apache.calcite.rel.core.Filter) modify.getInput();
+            condition = toExpression(filter.getCondition());
+          } else {
+            condition = Expression.BoolLiteral.builder().nullable(false).value(true).build();
+          }
+
+          List<String> updateColumnNames = modify.getUpdateColumnList();
+          List<org.apache.calcite.rex.RexNode> sourceExpressions = modify.getSourceExpressionList();
```
nielspardon: it looks like for `org.apache.calcite.rex.RexNode` we don't have a name clash and could import this class

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelVisitor.java
```diff
@@ -374,8 +378,76 @@ public Rel visit(org.apache.calcite.rel.core.Exchange exchange) {
...
+          }
+
+          List<String> updateColumnNames = modify.getUpdateColumnList();
+          List<RexNode> sourceExpressions = modify.getSourceExpressionList();
+          List<String> allTableColumnNames = modify.getTable().getRowType().getFieldNames();
+          List<NamedUpdate.TransformExpression> transformations = new ArrayList<>();
+
+          for (int i = 0; i < updateColumnNames.size(); i++) {
+            String colName = updateColumnNames.get(i);
+            org.apache.calcite.rex.RexNode rexExpr = sourceExpressions.get(i);
```
nielspardon: I think this also can use the simple class name:
```suggestion
            RexNode rexExpr = sourceExpressions.get(i);
```

isthmus/src/test/java/io/substrait/isthmus/SqlToSubstraitTest.java
ZorinAnton: Addressed

isthmus/src/test/java/io/substrait/isthmus/SqlToSubstraitTest.java
ZorinAnton: Addressed

isthmus/src/test/java/io/substrait/isthmus/SqlToSubstraitTest.java
ZorinAnton: Addressed

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelVisitor.java
ZorinAnton: Addressed

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelVisitor.java
ZorinAnton: Addressed

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelVisitor.java
ZorinAnton: Addressed

isthmus/src/test/java/io/substrait/isthmus/SqlToSubstraitTest.java
nielspardon: this has also been addressed from what I can see in the code

nielspardon (approved):
LGTM

PR #432, opened 2025-07-08
feat: support DDL statements in SqlToSubstrait

ZorinAnton:
As discussed, splitting the PR https://github.com/substrait-io/substrait-java/pull/428 into dml and ddl parts. This PR is for ddl support.

For the ddl sql statements the following relations are created:

- `NamedWrite` for create table as
- `NamedDdl` for create view

All other types of sql statements are handled as before.

Since DDL statements have no representation in Calcite's relational algebra, these changes conflict with the proposed split of SqlToSubstrait into SqlToCalcite and CalciteToSubstrait: https://github.com/substrait-io/substrait-java/issues/362. This is a key architectural decision that requires careful design.


isthmus/src/main/java/io/substrait/isthmus/SqlToSubstrait.java
```diff
@@ -42,32 +45,87 @@ List<RelRoot> sqlToRelNode(String sql, Prepare.CatalogReader catalogReader)
...
+    for (RelRoot relRoot : sqlNodesToRelNode(parsedList, converter)) {
+      io.substrait.plan.Plan.Root convert =
+          SubstraitRelVisitor.convert(relRoot, EXTENSION_COLLECTION, featureBoard);
+      sqlRelRoots.add(convert);
+    }
+
+    ddlRelRoots.addAll(sqlRelRoots);
+    ddlRelRoots.forEach(builder::addRoots);
+    return ddlRelRoots;
+  }
```
nielspardon: I find it a bit strange that we are returning a `List<io.substrait.plan.Plan.Root>` while at the same time adding the `Root`s to the `ImmutablePlan.Builder`. I would propose we return `void` and just add the plans to the `ImmutablePlan.Builder`.

During testing you can still retrieve the `Root`s from the `Plan` instance and do your assertions.

```suggestion
  private void sqlToPlanNodes(
      String sql,
      SqlValidator validator,
      Prepare.CatalogReader catalogReader,
      ImmutablePlan.Builder builder)
      throws SqlParseException {
    SqlParser parser = SqlParser.create(sql, parserConfig);
    SqlNodeList parsedList = parser.parseStmtList();
    SqlToRelConverter converter = createSqlToRelConverter(validator, catalogReader);
    // IMPORTANT: parsedList gets filtered in the call below
    List<io.substrait.plan.Plan.Root> ddlRelRoots = ddlSqlToRootNodes(parsedList, converter);
    ddlRelRoots.forEach(builder::addRoots);

    List<io.substrait.plan.Plan.Root> sqlRelRoots = new LinkedList<>();

    for (RelRoot relRoot : sqlNodesToRelNode(parsedList, converter)) {
      io.substrait.plan.Plan.Root convert =
          SubstraitRelVisitor.convert(relRoot, EXTENSION_COLLECTION, featureBoard);
      builder.addRoots(convert);
    }
  }
```

isthmus/src/main/java/io/substrait/isthmus/SqlToSubstrait.java
```diff
@@ -42,32 +45,87 @@ List<RelRoot> sqlToRelNode(String sql, Prepare.CatalogReader catalogReader)
     return sqlToRelNode(sql, validator, catalogReader);
   }
 
+  List<io.substrait.plan.Plan.Root> sqlToPlanNodes(String sql, Prepare.CatalogReader catalogReader)
+      throws SqlParseException {
+
+    SqlValidator validator = new SubstraitSqlValidator(catalogReader);
+    return sqlToPlanNodes(sql, validator, catalogReader, io.substrait.plan.Plan.builder());
+  }
+
```
nielspardon: this seems to be mainly required for your unit tests and won't work anymore if `sqlToPlanNodes` returns `void`. as I indicated in the other comment you can still get the `Root`s from the `Plan` instance during unit testing.

isthmus/src/main/java/io/substrait/isthmus/expression/DdlRelBuilder.java
```diff
@@ -0,0 +1,123 @@
...
+    super();
+    this.converter = converter;
+    this.bestExpRelRootGetter = bestExpRelRootGetter;
+    this.extensionCollection = extensionCollection;
+    this.featureBoard = featureBoard;
+
+    createHandlers.put(
+        SqlCreateTable.class, sqlCall -> handleCreateTable((SqlCreateTable) sqlCall));
+    createHandlers.put(SqlCreateView.class, sqlCall -> handleCreateView((SqlCreateView) sqlCall));
+  }
```
nielspardon: can we move the constructor to the top of the class before the `findCreateHandler` method?
```suggestion
  public DdlRelBuilder(
      final SqlToRelConverter converter,
      final BiFunction<SqlToRelConverter, SqlNode, RelRoot> bestExpRelRootGetter,
      final SimpleExtension.ExtensionCollection extensionCollection,
      final FeatureBoard featureBoard) {
    super();
    this.converter = converter;
    this.bestExpRelRootGetter = bestExpRelRootGetter;
    this.extensionCollection = extensionCollection;
    this.featureBoard = featureBoard;

    createHandlers.put(
        SqlCreateTable.class, sqlCall -> handleCreateTable((SqlCreateTable) sqlCall));
    createHandlers.put(SqlCreateView.class, sqlCall -> handleCreateView((SqlCreateView) sqlCall));
  }

  private Function<SqlCall, Plan.Root> findCreateHandler(final SqlCall call) {
    Class<?> currentClass = call.getClass();
    while (SqlCall.class.isAssignableFrom(currentClass)) {
      final Function<SqlCall, Plan.Root> found = createHandlers.get(currentClass);
      if (found != null) {
        return found;
      }
      currentClass = currentClass.getSuperclass();
    }
    return null;
  }
```

isthmus/src/main/java/io/substrait/isthmus/expression/DdlRelBuilder.java
```diff
@@ -0,0 +1,123 @@
...
+
+  private NamedStruct getSchema(final RelRoot queryRelRoot) {
+    final RelDataType rowType = queryRelRoot.rel.getRowType();
+
+    final TypeConverter typeConverter = TypeConverter.DEFAULT;
+    return typeConverter.toNamedStruct(rowType);
+  }
+
+  public Plan.Root handleCreateTable(final SqlCreateTable sqlCreateTable) {
```
nielspardon: does this method need to be public or would be sufficient to be protected or private? I don't see it called outside this class.

isthmus/src/main/java/io/substrait/isthmus/expression/DdlRelBuilder.java
```diff
@@ -0,0 +1,123 @@
...
+            .operation(AbstractWriteRel.WriteOp.CTAS)
+            .createMode(AbstractWriteRel.CreateMode.REPLACE_IF_EXISTS)
+            .outputMode(AbstractWriteRel.OutputMode.NO_OUTPUT)
+            .names(sqlCreateTable.name.names)
+            .build();
+
+    return Plan.Root.builder().input(namedWrite).build();
+  }
+
+  Plan.Root handleCreateView(final SqlCreateView sqlCreateView) {
```
nielspardon: I don't see this method being called outside this class so I guess it could be private or explicitly protected.

isthmus/src/test/java/io/substrait/isthmus/DdlRoundtripTest.java
```diff
@@ -0,0 +1,57 @@
...
+import io.substrait.plan.ImmutablePlan;
+import io.substrait.plan.Plan;
+import io.substrait.plan.PlanProtoConverter;
+import io.substrait.plan.ProtoPlanConverter;
+import java.util.List;
+import org.apache.calcite.prepare.Prepare;
+import org.apache.calcite.sql.parser.SqlParseException;
+import org.junit.jupiter.api.Test;
+
+public class DdlRoundtripTest extends PlanTestBase {
```
nielspardon: I believe JUnit 5 test classes should ideally be package private:
```suggestion
class DdlRoundtripTest extends PlanTestBase {
```

isthmus/src/test/java/io/substrait/isthmus/DdlRoundtripTest.java
```diff
@@ -0,0 +1,57 @@
...
+  final Prepare.CatalogReader catalogReader =
+      SubstraitCreateStatementParser.processCreateStatementsToCatalog(
+          "create table src1 (intcol int, charcol varchar(10))",
+          "create table src2 (intcol int, charcol varchar(10))");
+
+  public DdlRoundtripTest() throws SqlParseException {}
+
```
nielspardon: do we need this empty default constructor?
```suggestion
```

isthmus/src/main/java/io/substrait/isthmus/SqlToSubstrait.java
```diff
@@ -42,32 +45,87 @@ List<RelRoot> sqlToRelNode(String sql, Prepare.CatalogReader catalogReader)
...
+    sqlToPlanNodes(sql, validator, catalogReader, builder);
     PlanProtoConverter planToProto = new PlanProtoConverter();
 
     return planToProto.toProto(builder.build());
   }
 
+  // TBD: this method bypasses conversion to calcite's RelRoot breaks
+  // the idea of splitting this class into SqlToCalcite and CalciteToSubstrait,
+  // as there is no relational algebra representations for DDL statements.
+  private List<io.substrait.plan.Plan.Root> sqlToPlanNodes(
```
nielspardon: I'm wondering whether this method should be better called `sqlToPlanRoots` given that it converts SQL to `Plan.Root`

isthmus/src/main/java/io/substrait/isthmus/SqlToSubstrait.java
```diff
@@ -42,32 +45,87 @@ List<RelRoot> sqlToRelNode(String sql, Prepare.CatalogReader catalogReader)
...
+          SubstraitRelVisitor.convert(relRoot, EXTENSION_COLLECTION, featureBoard);
+      sqlRelRoots.add(convert);
+    }
+
+    ddlRelRoots.addAll(sqlRelRoots);
+    ddlRelRoots.forEach(builder::addRoots);
+    return ddlRelRoots;
+  }
+
+  private List<io.substrait.plan.Plan.Root> ddlSqlToRootNodes(
```
nielspardon: also this method is not converting to `nodes` but to `roots` so I would rename it to `ddlSqlToPlanRoots`

isthmus/src/main/java/io/substrait/isthmus/SqlToSubstrait.java
```diff
@@ -42,32 +45,87 @@ List<RelRoot> sqlToRelNode(String sql, Prepare.CatalogReader catalogReader)
...
-
+    sqlToPlanNodes(sql, validator, catalogReader, builder);
     PlanProtoConverter planToProto = new PlanProtoConverter();
 
     return planToProto.toProto(builder.build());
   }
 
+  // TBD: this method bypasses conversion to calcite's RelRoot breaks
+  // the idea of splitting this class into SqlToCalcite and CalciteToSubstrait,
+  // as there is no relational algebra representations for DDL statements.
```
nielspardon: do we need to keep this comment? it looks like the public API did not change so I guess there is no issue here?

isthmus/src/main/java/io/substrait/isthmus/SqlToSubstrait.java
```diff
@@ -42,32 +45,87 @@ List<RelRoot> sqlToRelNode(String sql, Prepare.CatalogReader catalogReader)
...
+  private List<io.substrait.plan.Plan.Root> sqlToPlanNodes(
+      String sql,
+      SqlValidator validator,
+      Prepare.CatalogReader catalogReader,
+      ImmutablePlan.Builder builder)
+      throws SqlParseException {
+    SqlParser parser = SqlParser.create(sql, parserConfig);
+    SqlNodeList parsedList = parser.parseStmtList();
+    SqlToRelConverter converter = createSqlToRelConverter(validator, catalogReader);
+    // IMPORTANT: parsedList gets filtered in the call below
```
nielspardon: can we make this comment a bit more specific?
```suggestion
    // IMPORTANT: the following call removes the converted DDL SqlNode entries from parsedList
```

isthmus/src/main/java/io/substrait/isthmus/SqlToSubstrait.java
ZorinAnton: addressed

isthmus/src/main/java/io/substrait/isthmus/SqlToSubstrait.java
ZorinAnton: addressed

isthmus/src/main/java/io/substrait/isthmus/expression/DdlRelBuilder.java
ZorinAnton: addressed

isthmus/src/main/java/io/substrait/isthmus/expression/DdlRelBuilder.java
ZorinAnton: addressed

isthmus/src/main/java/io/substrait/isthmus/expression/DdlRelBuilder.java
ZorinAnton: addressed

isthmus/src/main/java/io/substrait/isthmus/SqlToSubstrait.java
ZorinAnton: I refactored this method out

isthmus/src/test/java/io/substrait/isthmus/DdlRoundtripTest.java
ZorinAnton: This was intended to draw the reviewers' attention to this specific place in the code.

isthmus/src/main/java/io/substrait/isthmus/SqlToSubstrait.java
ZorinAnton: addressed

isthmus/src/main/java/io/substrait/isthmus/SqlToSubstrait.java
ZorinAnton: This was intended to draw the reviewers' attention this specific place in the code. Removed now.

isthmus/src/main/java/io/substrait/isthmus/SqlToSubstrait.java
ZorinAnton: the method has been refactored to avoid implicit removals.

isthmus/src/test/java/io/substrait/isthmus/DdlRoundtripTest.java
ZorinAnton: yes, as it can throw an exception during the `catalogReader` field initialization.

isthmus/src/main/java/io/substrait/isthmus/SqlToSubstrait.java
```diff
@@ -48,26 +51,66 @@ private Plan executeInner(String sql, SqlValidator validator, Prepare.CatalogRea
...
-        .map(root -> SubstraitRelVisitor.convert(root, EXTENSION_COLLECTION, featureBoard))
-        .forEach(root -> builder.addRoots(root));
-
+    sqlToPlanRoots(sql, validator, catalogReader, builder);
     PlanProtoConverter planToProto = new PlanProtoConverter();
 
     return planToProto.toProto(builder.build());
   }
 
+  void sqlToPlanRoots(
```
nielspardon: shouldn't this also get `private` or `protected` visibility like the others?
```suggestion
  private void sqlToPlanRoots(
```

ZorinAnton: Similarly to `sqlToRelNode`, this method gets called directly in a test case, so I gave it the same visibility.

nielspardon: why is the public API not sufficient aka `SqlToSubstrait.execute()`? shouldn't that also trigger the new code and give you a `Plan` with the DDL `Root`s?

isthmus/src/test/java/io/substrait/isthmus/DdlRoundtripTest.java
nielspardon: thanks, I see. then I would add a call to the super constructor since that's a best practice
```suggestion
  public DdlRoundtripTest() throws SqlParseException {
    super();
  }

```

isthmus/src/test/java/io/substrait/isthmus/DdlRoundtripTest.java
```diff
@@ -0,0 +1,57 @@
...
+  }
+
+  void testPlanRoundTrip(String sqlStatement) throws SqlParseException {
+    SqlToSubstrait sql2subst = new SqlToSubstrait();
+    ImmutablePlan.Builder builder = io.substrait.plan.Plan.builder();
+    SqlValidator validator = new SubstraitSqlValidator(catalogReader);
+
+    sql2subst.sqlToPlanRoots(sqlStatement, validator, catalogReader, builder);
+
+    final Plan plan = builder.build();
```
nielspardon: wouldn't this be much simpler?

```suggestion
    final Plan plan = sql2subst.execute(sqlStatement, catalogReader);
```

isthmus/src/test/java/io/substrait/isthmus/DdlRoundtripTest.java
ZorinAnton: fixed

isthmus/src/test/java/io/substrait/isthmus/DdlRoundtripTest.java
nielspardon: ah, I see. `execute()` is returning `io.substrait.proto.Plan` and not `io.substrait.plan.Plan` which is something we wanted to change in the isthmus API refactoring

nielspardon: I created this PR to make this code simpler: https://github.com/substrait-io/substrait-java/pull/473

nielspardon: the refactoring PR #473 has been merged yesterday. you can do this now:
```suggestion
    final Plan plan = sql2subst.convert(sqlStatement, catalogReader);
```

ZorinAnton: fixed

nielspardon:
@vbarua can you also take a look at this. I want to ensure the further refactoring to isthmus is somewhat aligned with these changes.

isthmus/src/main/java/io/substrait/isthmus/expression/DdlRelBuilder.java
```diff
@@ -0,0 +1,123 @@
...
+import org.apache.calcite.rel.RelRoot;
+import org.apache.calcite.rel.type.RelDataType;
+import org.apache.calcite.sql.SqlCall;
+import org.apache.calcite.sql.SqlNode;
+import org.apache.calcite.sql.ddl.SqlCreateTable;
+import org.apache.calcite.sql.ddl.SqlCreateView;
+import org.apache.calcite.sql.util.SqlBasicVisitor;
+import org.apache.calcite.sql2rel.SqlToRelConverter;
+
+public class DdlRelBuilder extends SqlBasicVisitor<Plan.Root> {
```
vbarua: The one thing that might be at odds with some of the work I've been doing on the API is that this returns a `Plan.Root` instead of a `RelRoot` or `RelNode`.

What I've been trying to do is modularize the API along two lines, roughly:
* `SQL -> Calcite RelRoot`
* `Calcite RelRoot -> Isthmus Plan`

The aim of this is to enable the workflows of users who are starting from SQL, and users who are starting directly with Calcite RelNodse, but share that majority of code for them. With the API modularized like this, SqlToSubstrait can effectively become the composition of these 2 conversion APIs. However, because what you're proposing here is effectively going from `SQL -> Isthmus POJO` directly, it makes that a bit trickier.

Aside from that goal though, a better argument for changing this is that this approach only works when going from SQL to POJO. If we assume that we want to be able to read in Substrait plans that create tables, then we'll need to have a first class entity in _Calcite_ to model this. The built-in LogicalTableModify doesn't work for this, but we can add our own TableCreate relation, along with a LogicalTableCreate. That can be used when going from Sql to Calcit, from Calcite to Isthmus, and also from Isthmus back to Calcite.

Does that make sense?

ZorinAnton: I think that the overall approach makes sense. The question is how to design this properly on the calcite's side.
I see the following options:

1. Use separate class for DDL relations. 
It means we can 
   - add the DdlRel class
   - implement required methods for visitors (not sure what's better here - introduce a new abstract visitor class for ddl or extend the RelVisitor)
   - add the DdlPlanRoot class (with the semantic similar to the RelRoot)
    Probably the DdlPlanRoot class should have the same base class with the RelRoot class for uniform handling in Sql->Calcite conversion.
   
2. Extend the RelNode class to support DDL. This simplifies the required changes, but I guess it diverges from Calcite conventions and add unnecessary complexity to RelNode 

Please let me know your thoughts.

vbarua: 1 makes the most sense to me. I think CREATE is the only operation that we have in Substrait that's not supported as a Calcite relation. I would re-use Calcite's LogicalModify relation for INSERT, DELETE and UPDATE like we are already doing, and the create a relation specifically for CREATE.

> implement required methods for visitors

We should be able to add a visit method for our new relation in the Isthmu s[RelNodeVisitor](https://github.com/substrait-io/substrait-java/blob/68c50c37b71bc4ae289be95b65fe18bff1c07d51/isthmus/src/main/java/io/substrait/isthmus/RelNodeVisitor.java#L21-L22) utility. The Calcite [RelShuttle](https://github.com/apache/calcite/blob/f4f0e978dc26739b8e966706bc1401ed40ded61b/core/src/main/java/org/apache/calcite/rel/RelShuttle.java#L78) also has a
```java
RelNode visit(RelNode other);
```
method for cases like this if we need that.

Would we need a custom RelRoot class, or could we just use the existing RelRoot with the SqlKind set to [CREATE_TABLE](https://github.com/apache/calcite/blob/f4f0e978dc26739b8e966706bc1401ed40ded61b/core/src/main/java/org/apache/calcite/sql/SqlKind.java#L1290-L1291) or [OTHER_DDL](https://github.com/apache/calcite/blob/f4f0e978dc26739b8e966706bc1401ed40ded61b/core/src/main/java/org/apache/calcite/sql/SqlKind.java#L1353-L1360)

vbarua: To further elucidate, this conversion would slot into here roughly: https://github.com/substrait-io/substrait-java/pull/474/files#diff-9977fe9108f68631bd554929d8e12242611c49b22534be2b701494ac133aa874R117-R140

ZorinAnton: I guess making changes in calcite will take time, what should we do with this PR? 
- Merge now as temporary solution
- Hold it off until those changes will be made approved and released
- Wait until https://github.com/substrait-io/substrait-java/pull/474 will be merged and adjust these changes accordingly as  a temporary solution

vbarua: > I guess making changes in calcite will take time, 

Ah, I was not suggesting we make the changes in Calcite. We should be able to define our own RelNode in Isthmus with something like:
```java
package io.substrait.isthmus.calcite.relations;

import org.apache.calcite.rel.AbstractRelNode;

public abstract class TableCreate extends AbstractRelNode {
//    ...
}

public class LogicalTableCreate extends TableCreate {
//    ...
}
```

ZorinAnton: ok, I thought we need to extend a hierarchy that's a part of calcite. If we can do that in the isthmus code it will be much easier.

vbarua: Ok yeah we can do something much simpler than extending Calcite. I was pointing out the
```java
RelNode visit(RelNode other);
```
because that's the mechanism by which Calcite let's you visit nodes that you define outside of Calcite core.

ZorinAnton: It seems with this approach we still have to make DdlRel (Create) a child of RelNode, but I have a couple of concerns with it:
- RelNode represents a relational expression, that processes data (records), while DdlRel does something else (it modifies the database catalog with a new entry for table, view or mqt)
- RelNode has the `getRowType` method which represents the type of the rows returned by the relational expression, which is not applicable for DdlRel

So I'm not sure if it is correct to say that DdlRel **is** RelNode. In my opinion, it would be better if they have the same parent class. But this probably requires changes in calcite.

ZorinAnton: @vbarua I introduced a new class `CalciteOperation` to represent relational algebra and ddl operations. Please take a look when you get a chance. If the suggested approach makes sense I will update `SubstraitToCalcite` accordingly.

isthmus/src/main/java/io/substrait/isthmus/sql/SubstraitSqlStatementParser.java
```diff
@@ -31,4 +31,18 @@ public static List<SqlNode> parseStatements(String sqlStatements) throws SqlPars
...
+
+  /**
+   * Parse one or more SQL statements to a list of {@link SqlNode}s.
+   *
+   * @param sqlStatements a string containing one or more SQL statements
+   * @param parserConfig sql parser configuration
+   * @return a list of {@link SqlNode}s corresponding to the given statements
+   * @throws SqlParseException if there is an error while parsing the SQL statements
+   */
+  public static List<SqlNode> parseStatements(String sqlStatements, SqlParser.Config parserConfig)
```
vbarua: Thinking out loud about this in the context of https://github.com/substrait-io/substrait-java/issues/362. We've spent some effort removing SQL parser configurability in the context of the FeatureBoard API that we had. The FeatureBoard was effectively a Substrait wrapper around Calcite configurations

Here though, we're introducing some degree of configuration back in via `parserConfig` parameter. However, in this case we're directly exposing the Calcite configuration instead of wrapping it, which I think is in line with those goals.

That being said, the intent with this class originally was to introduce a default "Substrait flavoured SQL parser". If we're allowing the injection of custom parser configs, that's no longer the case really.

isthmus/src/main/java/io/substrait/isthmus/sql/SubstraitSqlStatementParser.java
```diff
@@ -31,4 +31,18 @@ public static List<SqlNode> parseStatements(String sqlStatements) throws SqlPars
     SqlParser parser = SqlParser.create(sqlStatements, PARSER_CONFIG);
     return parser.parseStmtList();
```
vbarua: minor: if we're going to introduce the method below, we should re-use it here by passing in the sqlStatements along with the default config.

isthmus/src/main/java/io/substrait/isthmus/sql/SubstraitSqlToCalcite.java
```diff
@@ -26,6 +33,62 @@
  */
 public class SubstraitSqlToCalcite {
 
+  protected static SqlParser.Config createDefaultParserConfig() {
+    return SqlParser.Config.DEFAULT
+        .withUnquotedCasing(createDefaultFeatureBoard().unquotedCasing())
+        .withParserFactory(SqlDdlParserImpl.FACTORY)
+        .withConformance(SqlConformanceEnum.LENIENT);
+  }
```
vbarua: We should define a global re-usable default parser config somewhere. Maybe SubstraitSqlStatementParser would be a good place for this.

isthmus/src/main/java/io/substrait/isthmus/sql/SubstraitSqlToCalcite.java
```diff
@@ -26,6 +33,62 @@
...
+
+  protected static FeatureBoard createDefaultFeatureBoard() {
+    return ImmutableFeatureBoard.builder().build();
+  }
```
vbarua: We're actively removing the FeatureBoard, I would not include it here.

isthmus/src/main/java/io/substrait/isthmus/sql/SubstraitSqlToCalcite.java
```diff
@@ -26,6 +33,62 @@
...
+
+  protected static SqlToRelConverter createSqlToRelConverter(
```
vbarua: Is the intent of making this protected to allow users to override the SqlToRelConverter used in their own system?

isthmus/src/main/java/io/substrait/isthmus/operation/SqlKindFromRel.java
```diff
@@ -0,0 +1,264 @@
...
+
+/**
+ * A visitor to infer the general SqlKind from the root of a Substrait Rel tree. Note: This infers
+ * the general operation type, as the original SQL syntax is not preserved in the Substrait plan.
+ */
+public class SqlKindFromRel
+    implements RelVisitor<SqlKind, EmptyVisitationContext, RuntimeException> {
+
+  // Most common query operations map to SELECT.
+  private static final SqlKind QUERY_KIND = SqlKind.SELECT;
```
vbarua: I think we should just inline this as I don't think we get any extra clarity from defining it statically.

isthmus/src/test/java/io/substrait/isthmus/DdlRoundtripTest.java
```diff
@@ -0,0 +1,86 @@
...
+          "create table src1 (intcol int, charcol varchar(10))",
+          "create table src2 (intcol int, charcol varchar(10))");
+
+  public DdlRoundtripTest() throws SqlParseException {
+    super();
+  }
+
+  void testSqlToSubstrait(String sqlStatement) throws SqlParseException {
+    SqlToSubstrait sqlToSubstrait = new SqlToSubstrait();
+    io.substrait.proto.Plan protoPlan = sqlToSubstrait.execute(sqlStatement, catalogReader);
```
vbarua: We should use `sqlToSubstrait.convert` here now that `execute` is deprecated.

ZorinAnton:
@vbarua please check out this alternative approach, where DDL operations are treated as RelNodes. While the code looks simpler, I think it might be confusing because these DDL 'relations' don't actually consume or produce rows. I'm not sure if there's a solution that is both simple and conceptually correct. The absence of native DDL support in Calcite seems to confirm this difficulty

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelVisitor.java
```diff
@@ -445,8 +450,52 @@ public Rel visit(TableModify modify) {
     }
   }
 
+  private NamedStruct getSchema(final RelRoot queryRelRoot) {
+    final RelDataType rowType = queryRelRoot.rel.getRowType();
+
+    final TypeConverter typeConverter = TypeConverter.DEFAULT;
+    return typeConverter.toNamedStruct(rowType);
```
vbarua, line 455: We should use the `typeConverter` that's set on the SubstraitRelVisitor for this so that user-defined types are handled correctly.

isthmus/src/main/java/io/substrait/isthmus/SubstraitToSql.java
```diff
@@ -13,4 +16,12 @@ public SubstraitToSql() {
   public RelNode substraitRelToCalciteRel(Rel relRoot, Prepare.CatalogReader catalog) {
     return SubstraitRelNodeConverter.convert(relRoot, relOptCluster, catalog, parserConfig);
   }
+
+  public RelRoot subsrtaitRelToCalciteRelRoot(Rel relRoot, Prepare.CatalogReader catalogReader) {
```
vbarua: minor: typo in `subsrtait`

isthmus/src/main/java/io/substrait/isthmus/calcite/rel/DdlRelDataType.java
```diff
@@ -0,0 +1,13 @@
+package io.substrait.isthmus.calcite.rel;
+
+import java.util.LinkedList;
+import org.apache.calcite.rel.type.RelDataTypeImpl;
+
+public class DdlRelDataType extends RelDataTypeImpl {
```
vbarua: This isn't used anywhere. It's also probably better to stick to the built-in Calcite types as well. Can you remove this.

vbarua (commented):
Didn't have time for a full review today, but I think this approach is a reasonable compromise.

I broadly agree with
> I'm not sure if there's a solution that is both simple and conceptually correct.

but I think this is a good enough API for now. We can live with it for a bit, identify where it doesn't work for us and then change it in the future if need be.

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelVisitor.java
ZorinAnton: fixed

isthmus/src/main/java/io/substrait/isthmus/SubstraitToSql.java
ZorinAnton: this is a leftover from previous experiments, so dropping it

isthmus/src/main/java/io/substrait/isthmus/calcite/rel/DdlRelDataType.java
ZorinAnton: removed

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
```diff
@@ -547,6 +552,17 @@ public RelNode visit(NamedUpdate update, Context context) {
...
 
+  @Override
+  public RelNode visit(NamedDdl namedDdl, Context context) {
+    if (namedDdl.getViewDefinition().isEmpty()) {
+      throw new IllegalArgumentException("no view definition found");
+    }
+    Rel viewDefinition = namedDdl.getViewDefinition().get();
+    RelNode relNode = viewDefinition.accept(this, context);
+    RelRoot relRoot = RelRoot.of(relNode, SqlKind.SELECT);
+    return new CreateView(namedDdl.getNames(), relRoot);
```
vbarua: It looks like we've only added support for NamedDdls when they are creating views. I think that's reasonable, but if that's what we want to do we should:
1. Check the values of `getObject` and `getOperation` to make sure the NamedDDL is a CREATE VIEW
2. Make the error indicate that we only support this for now.

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
```diff
@@ -584,15 +600,19 @@ public RelNode visit(VirtualTableScan virtualTableScan, Context context) {
         relBuilder.getCluster(), rowTypeWithNames, ImmutableList.copyOf(tuples));
   }
 
+  private RelNode handleCreateTableAs(NamedWrite namedWrite, Context context) {
+    Rel input = namedWrite.getInput();
+    RelNode relNode = input.accept(this, context);
+    RelRoot relRoot = RelRoot.of(relNode, SqlKind.SELECT);
+    return new CreateTable(namedWrite.getNames(), relRoot);
```
vbarua: Do we need to capture/handle the various CreateModes as well?

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelVisitor.java
```diff
@@ -445,8 +450,50 @@ public Rel visit(TableModify modify) {
...
+  public Rel handleCreateTable(CreateTable createTable) {
+    RelRoot input = createTable.getInput();
+    Rel inputRel = apply(input.project());
+    NamedStruct schema = getSchema(input);
+    return NamedWrite.builder()
+        .input(inputRel)
+        .tableSchema(schema)
+        .operation(AbstractWriteRel.WriteOp.CTAS)
+        .createMode(AbstractWriteRel.CreateMode.REPLACE_IF_EXISTS)
+        .outputMode(AbstractWriteRel.OutputMode.NO_OUTPUT)
```
vbarua, line 467: It's okay if this is the only type of CREATE TABEL operation, create and output mode we want to support for now, but we should match this restriction when we create a CreateTable in the SubstraitRelNodeConverter.

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelVisitor.java
```diff
@@ -445,8 +450,50 @@ public Rel visit(TableModify modify) {
...
+    final Expression.StructLiteral defaults = ExpressionCreator.struct(false);
+
+    return NamedDdl.builder()
+        .viewDefinition(inputRel)
+        .tableSchema(getSchema(input))
+        .tableDefaults(defaults)
+        .operation(AbstractDdlRel.DdlOp.CREATE)
+        .object(AbstractDdlRel.DdlObject.VIEW)
+        .names(createView.getNames())
+        .build();
```
vbarua, line 485: Same comment as above but for CreateView.

isthmus/src/main/java/io/substrait/isthmus/calcite/rel/CreateTable.java
```diff
@@ -0,0 +1,41 @@
...
+  private List<String> names;
+  private RelRoot input;
+
+  public RelRoot getInput() {
+    return input;
+  }
+
+  public void setInput(RelRoot input) {
+    this.input = input;
+  }
```
vbarua, line 19: Is there any reason to expose setters for input and name? They can both be configured directly when the CreateTable is constructed. I would recommend removing them unless for them.

isthmus/src/main/java/io/substrait/isthmus/calcite/rel/CreateView.java
```diff
@@ -0,0 +1,39 @@
...
+    this.names = names;
+  }
+
+  public RelRoot getInput() {
+    return input;
+  }
+
+  public void setInput(RelRoot input) {
+    this.input = input;
+  }
```
vbarua, line 40: Same comment as above for the setters in CreateView.

isthmus/src/main/java/io/substrait/isthmus/calcite/rel/CreateView.java
```diff
@@ -0,0 +1,39 @@
...
+
+  public CreateView(List<String> names, RelRoot input) {
+    super(input.rel.getCluster(), input.rel.getTraitSet());
+    this.names = names;
+    this.input = input;
+  }
+
+  @Override
+  protected RelDataType deriveRowType() {
+    // return new DdlRelDataType();
```
vbarua: We should remove this comment.

isthmus/src/main/java/io/substrait/isthmus/calcite/rel/CreateTable.java
```diff
@@ -0,0 +1,41 @@
...
+  public CreateTable(List<String> names, RelRoot input) {
+    super(input.rel.getCluster(), input.rel.getTraitSet());
+
+    this.names = names;
+    this.input = input;
+  }
+
+  @Override
+  protected RelDataType deriveRowType() {
+    // return new DdlRelDataType();
```
vbarua: We should remove this comment.

vbarua (commented):
Left some more comments.

The biggest takeaway is that when converting from SQL to Substrait, we only generate a subset of valid Substrait objects. When converting back from Substrait to Calcite, we need to be aware that there are other sources of Substrait objects beyond those generated from the SQL path here, so we should restrict our processing to only handle what we've chosen to support explicitly for now.

vbarua (commented):
Left some more comments.

The biggest takeaway is that when converting from SQL to Substrait, we only generate a subset of valid Substrait objects. When converting back from Substrait to Calcite, we need to be aware that there are other sources of Substrait objects beyond those generated from the SQL path here, so we should restrict our processing to only handle what we've chosen to support explicitly for now.

ZorinAnton:
>  so we should restrict our processing to only handle what we've chosen to support explicitly for now.

Makes sense. I've added a validator to SubstraitRelNodeConverter to ensure incoming relations have a supported configuration. If this approach is acceptable, we can move the validation logic for non-DDL and non-DML relations there as well (probably in a separate PR).

vbarua:
@ZorinAnton could you avoid rebasing your PR every time you update it. It makes it harder to review because I can't just look at the new commits, and it messes with Github PR view a bit. The end result will be squash merged as well, so we don't need to keep the branch fresh with main.

isthmus/src/main/java/io/substrait/isthmus/calcite/rel/CreateTable.java
```diff
@@ -0,0 +1,41 @@
...
+
+  public RelRoot getInput() {
+    return input;
+  }
+
+  public void setInput(RelRoot input) {
+    this.input = input;
+  }
+
+  public CreateTable(List<String> names, RelRoot input) {
```
vbarua: It would be more ergonomic for this and CreateView to consume a RelNode and not a RelRoot. Does that seem reasonable / feasible?

isthmus/src/main/java/io/substrait/isthmus/calcite/rel/CreateTable.java
vbarua: Handled in my suggestion PR

isthmus/src/main/java/io/substrait/isthmus/calcite/rel/CreateView.java
vbarua: Handled in my suggestion PR

isthmus/src/main/java/io/substrait/isthmus/calcite/rel/CreateView.java
vbarua: Handled in my suggestion PR.

isthmus/src/main/java/io/substrait/isthmus/calcite/rel/CreateTable.java
vbarua: Handled in my suggestion PR.

isthmus/src/test/java/io/substrait/isthmus/PlanTestBase.java
```diff
@@ -178,6 +178,7 @@ protected void assertFullRoundTrip(String sqlQuery, Prepare.CatalogReader catalo
     ExtensionCollector extensionCollector = new ExtensionCollector();
 
     // SQL -> Calcite 1
+    //    RelRoot calcite1 = SubstraitSqlToCalcite.convertRelationalQuery(sqlQuery, catalogReader);
```
vbarua: Removed this comment in my PR.

vbarua (commented):
> I just didn't want to have a separate commit per every singe change.

Because it gets squashed at the end this won't matter. When I merge as well, I summarize the individual commits into a coherent message.

I've made some suggestions for this a separate PR targetting your PR
https://github.com/ZorinAnton/substrait-java/pull/1

> I've added a validator to SubstraitRelNodeConverter to ensure incoming relations have a supported configuration.

I don't think we need that level of machinary for this work. For one thing, in the long-term we shouldn't need those validations at all because we should be able to convert ALL valid Substrait messages to Calcite and back. In my suggestions, I've inlined very basic restrictions.

I'm not opposed to validations in general, but I don't want to make them part of this PR because if we start fiddling with a public API for validations it will extend how long this PR remains open.

ZorinAnton:
> I've made some suggestions for this a separate PR targetting your PR

Thanks, merged it.

> in the long-term we shouldn't need those validations at all because we should be able to convert ALL valid Substrait messages to Calcite and back.

before it happens, a dedicated validator would give a quick view of what is supported and what's not, but I agree that it needs to be addressed in a separate PR.

isthmus/src/main/java/io/substrait/isthmus/calcite/rel/CreateTable.java
vbarua: @ZorinAnton Github marked this comment as outdated, but it's still relevant.

vbarua: Oh wait, I missed your updated. Thanks for handling that 🙇

vbarua (approved):

PR #434, opened 2025-07-11
chore: use EmptyVisitationContext.INSTANCE instead of null

nielspardon:
Follow-up to #427


vbarua (approved):
Thanks for following up with this 🙇

PR #437, opened 2025-07-13
style: replace RuntimeException with more specific exceptions

nielspardon:
replaces throwing a generic `RuntimeException` with a more specific exception like `IllegalStateException`, `IllegalArgumentException` or `UnsupportedOperationException`

### DIFF
--- a/core/src/main/java/io/substrait/extension/SimpleExtension.java
+++ b/core/src/main/java/io/substrait/extension/SimpleExtension.java
@@ -736,7 +736,7 @@ public class SimpleExtension {
                   try (InputStream stream = ExtensionCollection.class.getResourceAsStream(path)) {
                     return load(path, stream);
                   } catch (IOException e) {
-                    throw new RuntimeException(e);
+                    throw new IllegalStateException(e);
                   }
                 })
             .collect(Collectors.toList());
@@ -752,7 +752,7 @@ public class SimpleExtension {
       ExtensionSignatures doc = objectMapper(namespace).readValue(str, ExtensionSignatures.class);
       return buildExtensionCollection(namespace, doc);
     } catch (JsonProcessingException e) {
-      throw new RuntimeException(e);
+      throw new IllegalStateException(e);
     }
   }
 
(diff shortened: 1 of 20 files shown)


core/src/main/java/io/substrait/extension/SimpleExtension.java
```diff
@@ -729,7 +729,7 @@ public static ExtensionCollection load(List<String> resourcePaths) {
                   try (var stream = ExtensionCollection.class.getResourceAsStream(path)) {
                     return load(path, stream);
                   } catch (IOException e) {
-                    throw new RuntimeException(e);
+                    throw new IllegalStateException(e);
```
bestbeforetoday: Since IOException is the underlying exception, you might consider [UncheckedIOException](https://docs.oracle.com/javase/8/docs/api/java/io/UncheckedIOException.html) here.

bestbeforetoday (commented):
One suggestion. Otherwise this looks good to me.

core/src/main/java/io/substrait/extension/SimpleExtension.java
nielspardon: thanks, I changed the two occurrences.

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
```diff
@@ -1,5 +1,6 @@
 package io.substrait.isthmus;
 
+import static io.substrait.isthmus.SqlConverterBase.EXTENSION_COLLECTION;
 import static io.substrait.isthmus.SqlToSubstrait.EXTENSION_COLLECTION;
```
bestbeforetoday: I am surprised there wasn't a compile error with two static imports of the same name. I guess because they actually refer to the same constant. I think there should only be one import here though.

bestbeforetoday (changes requested):
One import issue (see comment), but otherwise looks good to me.

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
nielspardon: thanks, this is a bit annoying. this has been happening a couple of times now. I think it's VSCode somehow understanding that `EXTENSION_COLLECTION` is actually defined in `SqlConverterBase` which `SqlToSubstrait` inherits from.

nielspardon: I pushed a commit switching to `SqlConverterBase.EXTENSION_COLLECTION` consistently which should help to avoid having VSCode change this line over and over again

PR #433, opened 2025-07-10
chore: avoid Java version mismatch

bestbeforetoday:
When developing, the build (both on the command-line and in the IDE) frequently fails due to usage of unsupported Java language features for the target Java runtime. This is due to the the Java source version being set to a newer version than the target Java version. While this sometimes works, this combination is explicitly disallowed by the Java compiler. Source version must be equal to or less than the target version.

This change uses the correct Java toolchain version for each sub-project. The exception being the core sub-project, since it targets Java 8 whereas ANTLR requires Java 11+ to run. Instead, the Java compiler release option is set to Java 8, ensuring that Java 8 bytecode is produced and that no APIs not present in Java 8 are used.

While numerous code changes are required to adhere to the target Java language version, there are no functional changes.

PR #438, opened 2025-07-13
style: avoid using star imports

nielspardon:
This makes it easier to understand which types/constants/methods are actually used in the code.


bestbeforetoday:
The latest version of the Spotless plugin has a `removeWildcardImports()` step. Probably this change should include adoption of that Spotless capability; otherwise later changes will just end up regressing these changes and/or adding more wildcard imports.

nielspardon:
I rebased on latest main and pushed an additional commit upgrading spotless and enabling the removeWildcardImports option. The Spotless update did trigger some reformatting of existing code.

bestbeforetoday (approved):
This change looks good to me.

It would be nice to figure out if we will proceed with #433 before merging this since it will introduce a fourth round of merge conflicts with that PR.

vbarua (approved):
I've updated these in the past as I've seen them, so it's great that we can just enforce this via Spotless.

PR #439, opened 2025-07-13
style: add missing @Override annotations

nielspardon:


nielspardon:
> It would be good to couple this change with some linting / static analysis to flag or enforce missing `@Override` annotations. For example, [PMD](https://pmd.github.io/pmd/) with _only_ the [MissingOverride](https://pmd.github.io/pmd/pmd_rules_java_bestpractices.html#missingoverride) rule applied. Without this, it becomes a manual process to check and correct missing annotations. It also actively helps contributors to ensure those annotations are included in their code.

Yes, I also would like to introduce PMD though I'm waiting for the `build.gradle` changes for the java version fix to be merged. I then would like to move create Gradle convention plugins so we can share build logic across modules and then I would introduce PMD.

bestbeforetoday (approved):
A nice improvement that should help avoid breakages if any of the overridden methods are refactored / renamed.

vbarua (approved):

PR #444, opened 2025-07-17
build: use JSpecify @Nullable instead of JSR305

nielspardon:
JSR305 has been dormant for years and was never finalized. It would be better to use alternative approaches like the checkerframework which is e.g. what Apache Calcite is doing.


nielspardon:
looks like a transient build error due to our faulty Java version config. should be fixed by #443 or #433

bestbeforetoday:
Based on a little searching, I am not sure that checkerframework is something we should be buying in to:

- https://github.com/google/guava/issues/3031
- https://github.com/google/guava/issues/2960
- https://github.com/jspecify/jspecify/issues/239

JSpecify seems to have [a lot of industry backing](https://jspecify.dev/about/) (Google, Jetbrains, Meta, Microsoft, Oracle, PMD, ...) so maybe that is a better direction?

nielspardon:
> Based on a little searching, I am not sure that checkerframework is something we should be buying in to:

I don't mind which alternative we use we just shouldn't use JSR305 in my opinion.

nielspardon:
I rebased on current main and pushed a commit switching to JSpecify

vbarua (approved):
LGTM

PR #445, opened 2025-07-17
chore(isthmus,isthmus-cli): remove unused dependencies

nielspardon:
removes unused dependencies from isthmus and isthmus-cli

- Jackson is neither used in isthmus nor in isthmus-cli
- caffeine is not used
- reflections is not used in isthmus but only in isthmus-cli


bestbeforetoday (approved):
Good catch. Nice to slim down the dependency tree.

Issue #447, opened 2025-07-18
Use canonical URIs for default extensions


benbellick:
### Problem
Currently, we're using relative URIs for default extensions. We should be consistent with substrait's official recommendation. 

### Solution
Switch to using the recommended canonical URIs for all default extensions. 

### Dependencies
https://github.com/substrait-io/substrait/issues/828

PR #448, opened 2025-07-18
Closes gh-447

benbellick:


core/src/main/java/io/substrait/extension/DefaultExtensionCatalog.java
```diff
@@ -1,17 +1,30 @@
...
+  public static final String FUNCTIONS_LOGARITHMIC =
+      "https://github.com/substrait-io/substrait/blob/main/extensions/functions_logarithmic.yaml";
+  public static final String FUNCTIONS_ROUNDING =
+      "https://github.com/substrait-io/substrait/blob/main/extensions/functions_rounding.yaml";
+  public static final String FUNCTIONS_ROUNDING_DECIMAL =
+      "https://github.com/substrait-io/substrait/blob/main/extensions/functions_rounding_decimal.yaml";
+  public static final String FUNCTIONS_SET =
+      "https://github.com/substrait-io/substrait/blob/main/extensions/functions_set.yaml";
+  public static final String FUNCTIONS_STRING =
+      "https://github.com/substrait-io/substrait/blob/main/extensions/functions_string.yaml";
```
nielspardon, line 29: since substrait-java typically uses a specific tag from the substrait repository as a Git submodule I'm wondering whether these URLs should reference the Git tag, e.g.:
`https://github.com/substrait-io/substrait/blob/v0.74.0/extensions/functions_boolean.yaml`

The Substrait version of the current Git tag is available in `io.substrait.SubstraitVersion.VERSION` without the `v` prefix though.

benbellick: Hi @nielspardon, thanks for the comment. I perhaps opened this PR a bit overeagerly, as https://github.com/substrait-io/substrait/issues/827 should probably be resolved first. That being said, your suggestion sounds reasonable. 

A third option could be to allow both, i.e have an argument passed to the constructor of this class of boolean called `versioned`. If true, we can make the string e.g. "https://github.com/substrait-io/substrait/blob/v{io.substrait.SubstraitVersion.VERSION}/extensions/functions_string.yaml", otherwise, it is just the `main` one. 

To be clear, I'm not sure what the right approach is 🤷 but I do think there needs to be consistency among all of the other libraries, which is why solving the problem in the upstream `substrait` repo first is probably the right approach.

nielspardon: no worries, I agree with your issue. I was also thinking about how to improve the URI scheme for Substrait extensions and opening up the discussion was a good move by you so that we can get different perspectives on the issue and hopefully come to a consensus on a consistent solution.

benbellick:
Closed, instead look to https://github.com/substrait-io/substrait-java/pull/470

PR #449, opened 2025-07-24
chore: update spark example to use latest substrait-spark

nielspardon:
The substrait spark example was using a static outdated version of the substrait spark mappings. This PR changes the substrait spark example to always use the latest substrait spark mappings and updates the code according to the latest changes.

Additionally, it changes the substrait spark mapping Gradle build to declare the spark-core and spark-sql dependencies as `api` dependencies so we don't have to redeclare them for the substrait spark example. This also required to change the `java` Gradle plugin to the `java-library` Gradle plugin.

### DIFF
--- a/examples/substrait-spark/build.gradle.kts
+++ b/examples/substrait-spark/build.gradle.kts
@@ -10,10 +10,7 @@ repositories {
 }
 
 dependencies {
-  implementation("org.apache.spark:spark-core_2.12:3.5.1")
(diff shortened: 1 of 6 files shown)


vbarua (approved):
>This PR changes the substrait spark example to always use the latest substrait spark mappings and updates the code according to the latest changes.

We should definitely aim to keep this example up to date. Thanks for updating this.

PR #451, opened 2025-07-29
feat(spark): add LogicalRDD support

andrew-coleman:
Support conversion of dataframes that are created using the Spark createDataFrame() method.
This produces a LogicalRDD in the query plan which can be converted to a substrait VirtualTableScan.
Introduces overridable `rddLimit` to guard against serialising very large datasets.

### DIFF
--- a/spark/src/main/scala/io/substrait/spark/logical/ToSubstraitRel.scala
+++ b/spark/src/main/scala/io/substrait/spark/logical/ToSubstraitRel.scala
@@ -21,11 +21,13 @@ import io.substrait.spark.expression._
 
 import org.apache.spark.internal.Logging
 import org.apache.spark.sql.SaveMode
+import org.apache.spark.sql.catalyst.InternalRow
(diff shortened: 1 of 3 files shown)


spark/src/main/scala/io/substrait/spark/logical/ToSubstraitRel.scala
```diff
@@ -64,6 +66,10 @@ class ToSubstraitRel extends AbstractLogicalPlanVisitor with Logging {
 
   private val existenceJoins = scala.collection.mutable.Map[Long, SExpression.InPredicate]()
 
+  private var _rddLimit = 100
+  def rddLimit: Int = _rddLimit
```
nielspardon, line 70: it might be good idea to check whether the RDD has more than the `rddLimit` rows and log a warning message in order to notify users of the fact and inform them that there is a modifiable limit

nielspardon (approved):
LGTM if we can add a log message

PR #452, opened 2025-07-30
style: use explicit types instead of var keyword

nielspardon:
In my opinion using explicit types instead of the `var` keyword makes the code more readable since all types used in a source file are visible through import statements and it is clear what the type of a field/variable is just by looking at the code.

This is why I'm proposing to switch to using explicit types and enforcing the use of explicit types through the corresponding PMD rule.

### DIFF
--- a/isthmus/src/test/java/io/substrait/isthmus/WindowFunctionTest.java
+++ b/isthmus/src/test/java/io/substrait/isthmus/WindowFunctionTest.java
@@ -37,7 +37,7 @@ public class WindowFunctionTest extends PlanTestBase {
     @ParameterizedTest
     @ValueSource(strings = {"rank", "dense_rank", "percent_rank"})
     void rankFunctions(String rankFunction) throws IOException, SqlParseException {
-      var query =
+      String query =
           String.format(
               "select O_ORDERKEY, %s() over (order by O_SHIPPRIORITY) from ORDERS", rankFunction);
       assertFullRoundTrip(query);
@@ -46,7 +46,7 @@ public class WindowFunctionTest extends PlanTestBase {
     @ParameterizedTest
     @ValueSource(strings = {"rank", "dense_rank", "percent_rank"})
     void rankFunctionsWithPartitions(String rankFunction) throws IOException, SqlParseException {
-      var query =
+      String query =
           String.format(
               "select O_ORDERKEY, %s() over (partition by O_CUSTKEY order by O_SHIPPRIORITY) from ORDERS",
               rankFunction);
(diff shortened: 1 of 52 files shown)


bestbeforetoday:
I agree that having the types there give more context, particularly when just viewing in the GitHub UI rather than in an IDE that can tell you the types. On the other hand, I am pretty used to types being inferred rathern than spelled out from Golang, TypeScript and Python (with type hints), so omitting the explicit typing doesn't bother me. There are certainly a few cases in this codebase where generic type names are so long that spelling them out actively decreases readability though the clutter they introduce.

There is a pretty balanced perpective on pros and cons of using type inference in Java [here](https://openjdk.org/projects/amber/guides/lvti-style-guide). A takeaway seems to be that in some cases using `var` is beneficial, while in others it is not. I don't know if imposing a hard rule one way or another is the right answer since it removes the flexibility to pick the better option for a given case. On the other hand, different people have different views on when one option is preferred so it is a potential source of pointless frustration.

I guess what I am trying to say is that, while I don't personally feel a need for `var` to be avoided, I am happy with whatever the community feel is the better option. Sitting right on the fence. 😄

isthmus/src/test/java/io/substrait/isthmus/WindowFunctionTest.java
```diff
@@ -156,10 +156,10 @@ void rangePrecedingToCurrent() throws IOException, SqlParseException {
     @Test
     void rangeCurrentToFollowing() throws IOException, SqlParseException {
       /*
-      LogicalProject(EXPR$0=[MIN($7) OVER (PARTITION BY $1 ORDER BY $3 RANGE BETWEEN CURRENT ROW AND 11 FOLLOWING)])
+      LogicalProject(EXPR=[MIN() OVER (PARTITION BY String ORDER BY  RANGE BETWEEN CURRENT ROW AND 11 FOLLOWING)])
```
vbarua: Did the PMD formatting change this?

nielspardon: PMD does not format code must have been something else in my toolchain. Let me revert that change.

nielspardon: rebased and reverted the comment changes

vbarua (approved):
I'm on board with this. I find it's more helpful generally to have types annotated explicitly. I do like using var when developing code and figuring out the shape of what I'm building, but once that's done the explicit types are very helpful, especially during rewiews.

PR #453, opened 2025-07-30
style: cleanup unused code

nielspardon:
- removes unused local variables
- removes unused assignments
- removes unused private fields
- removes unused private methods
- configures a PMD rule for each

### DIFF
--- a/core/src/main/java/io/substrait/type/parser/TypeStringParser.java
+++ b/core/src/main/java/io/substrait/type/parser/TypeStringParser.java
@@ -60,19 +60,13 @@ public class TypeStringParser {
         final int charPositionInLine,
         final String msg,
         final RecognitionException e) {
-      throw new ParseError(msg, e, line, charPositionInLine);
+      throw new ParseError(msg, e);
     }
   }
 
   public static class ParseError extends RuntimeException {
-    private final int line;
-    private final int posInLine;
-
-    public ParseError(
-        final String message, final Throwable cause, final int line, final int posInLine) {
+    public ParseError(final String message, final Throwable cause) {
       super(message, cause);
-      this.line = line;
-      this.posInLine = posInLine;
     }
   }
 }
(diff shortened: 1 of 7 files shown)


core/src/main/java/io/substrait/type/parser/TypeStringParser.java
```diff
@@ -60,19 +60,13 @@ public void syntaxError(
...
     }
   }
 
   public static class ParseError extends RuntimeException {
-    private final int line;
-    private final int posInLine;
-
-    public ParseError(
-        final String message, final Throwable cause, final int line, final int posInLine) {
+    public ParseError(final String message, final Throwable cause) {
```
bestbeforetoday, line 68: Annoyingly, since this is a public class, this is a (breaking) public API change. Likely nobody is using it — or needs to use it — so perhaps that doesn't matter. If not, maybe it should not be public at all.

I wonder if the ParseError might be more helpful if it incorporated the line and position (along with the message) into the message for the underlying exception?

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelVisitor.java
```diff
@@ -207,22 +207,9 @@ public Rel visit(org.apache.calcite.rel.core.Correlate correlate) {
     // right input of correlated-join is similar to a correlated sub-query
     apply(correlate.getRight());
 
-    var joinType = asJoinType(correlate);
```
bestbeforetoday: I wonder if this call was intended to have a side-effect of throwing an exception if the join type was not INNER or LEFT? I guess if the tests still pass then we can assume it wasn't intended behaviour though, and removing this is fine.

core/src/main/java/io/substrait/type/parser/TypeStringParser.java
nielspardon: good point, though this seems to be only used within the generated ANTLR parser code

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelVisitor.java
nielspardon: Good observation though I think it does not make a huge difference since currently it calls `return super.visit(correlate)` in the next line which then calls `visitOther()` which throws an `UnsupportedOperationException` for `org.apache.calcite.rel.core.Correlate`. So this is effectively replacing an `IllegalArgumentException` with an `UnsupportedOperationException`.

PR #457, opened 2025-08-05
feat(isthmus): add dynamic function conversion for Substrait<->Calcite

ZorinAnton:
This PR introduces support for dynamic scalar User-Defined Functions (UDFs) for round-trip conversion between Substrait plans and Calcite RelNodes.

It establishes the Substrait YAML extension files as the single source of truth (SSoT) for all function signatures. The primary addition is a new utility, [YamlToSqlOperator](https://github.com/ZorinAnton/substrait-java/blob/ad11957f3c02f1383dc5128fad83763431422341/isthmus/src/main/java/io/substrait/isthmus/YamlToSqlOperator.java#L25), which acts as a bridge between the Substrait and Calcite worlds by dynamically generating Calcite SqlOperator objects from the parsed YAML definitions.

The implementation is optimized to only generate these operators for functions that are truly dynamic —that is, functions present in the provided extensions but not already defined in the default [SubstraitOperatorTable](https://github.com/ZorinAnton/substrait-java/blob/ad11957f3c02f1383dc5128fad83763431422341/isthmus/src/main/java/io/substrait/isthmus/calcite/SubstraitOperatorTable.java#L26).

Key Changes:
- New [YamlToSqlOperator](https://github.com/ZorinAnton/substrait-java/blob/ad11957f3c02f1383dc5128fad83763431422341/isthmus/src/main/java/io/substrait/isthmus/YamlToSqlOperator.java#L25) Utility: A type-safe converter that translates [SimpleExtension.ExtensionCollection](https://github.com/ZorinAnton/substrait-java/blob/ad11957f3c02f1383dc5128fad83763431422341/core/src/main/java/io/substrait/extension/SimpleExtension.java#L569) objects into a List<SqlOperator>. 

- Enhanced SqlToSubstrait (SQL → Substrait): The [SqlToSubstrait ](https://github.com/ZorinAnton/substrait-java/blob/ad11957f3c02f1383dc5128fad83763431422341/isthmus/src/main/java/io/substrait/isthmus/SqlToSubstrait.java#L40) constructor is updated to:

  - Identify dynamic UDFs by filtering the loaded extensions against the known operators in [SubstraitOperatorTable](https://github.com/ZorinAnton/substrait-java/blob/ad11957f3c02f1383dc5128fad83763431422341/isthmus/src/main/java/io/substrait/isthmus/calcite/SubstraitOperatorTable.java#L26).

  - Use [YamlToSqlOperator](https://github.com/ZorinAnton/substrait-java/blob/ad11957f3c02f1383dc5128fad83763431422341/isthmus/src/main/java/io/substrait/isthmus/YamlToSqlOperator.java#L25) to generate operators for these dynamic UDFs.

  - Inject the new operators into [Calcite's validator](https://github.com/ZorinAnton/substrait-java/blob/ad11957f3c02f1383dc5128fad83763431422341/isthmus/src/main/java/io/substrait/isthmus/SqlToSubstrait.java#L48) via a сhained SqlOperatorTable.

- Enhanced [SubstraitToCalcite ](https://github.com/ZorinAnton/substrait-java/blob/ad11957f3c02f1383dc5128fad83763431422341/isthmus/src/main/java/io/substrait/isthmus/SubstraitToCalcite.java#L96) (Substrait → SQL): The [SubstraitRelNodeConverter ](https://github.com/ZorinAnton/substrait-java/blob/ad11957f3c02f1383dc5128fad83763431422341/isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java#L141)is similarly enhanced to:

  - Also identify dynamic UDFs from the provided [SimpleExtension.ExtensionCollection](https://github.com/ZorinAnton/substrait-java/blob/ad11957f3c02f1383dc5128fad83763431422341/core/src/main/java/io/substrait/extension/SimpleExtension.java#L569).

  - Dynamically create [FunctionMappings.Sig](https://github.com/ZorinAnton/substrait-java/blob/ad11957f3c02f1383dc5128fad83763431422341/isthmus/src/main/java/io/substrait/isthmus/expression/FunctionMappings.java#L125) objects for these UDFs, enabling the [ScalarFunctionConverter ](https://github.com/ZorinAnton/substrait-java/blob/ad11957f3c02f1383dc5128fad83763431422341/isthmus/src/main/java/io/substrait/isthmus/expression/ScalarFunctionConverter.java#L21) to correctly map them back to their SqlOperator during plan conversion.

- Architectural Refactoring: The EXTENSION_COLLECTION in [SqlConverterBase ](https://github.com/ZorinAnton/substrait-java/blob/ad11957f3c02f1383dc5128fad83763431422341/isthmus/src/main/java/io/substrait/isthmus/SqlConverterBase.java#L22) has been changed from static final to an instance field to allow for the dynamic configuration required by this feature.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,12 @@
 Release Notes
 ---
 
+## [0.72.0](https://github.com/substrait-io/substrait-java/compare/v0.71.0...v0.72.0) (2025-12-07)
+
+### Features
+
+* support Nested Structs ([#579](https://github.com/substrait-io/substrait-java/issues/579)) ([8845bbf](https://github.com/substrait-io/substrait-java/commit/8845bbff6954c2ba5105469401a94bf54c686f49))
+
 ## [0.71.0](https://github.com/substrait-io/substrait-java/compare/v0.70.0...v0.71.0) (2025-11-30)
 
 ### Features
--- a/gradle.properties
+++ b/gradle.properties
@@ -8,7 +8,7 @@ org.gradle.parallel=true
 org.gradle.caching=true
 
 #version that is going to be updated automatically by releases
-version = 0.71.0
+version = 0.72.0
 
 #signing
 SIGNING_KEY_ID = 193EAE47


isthmus/src/main/java/io/substrait/isthmus/sql/SubstraitSqlValidator.java
```diff
@@ -12,4 +13,8 @@ public class SubstraitSqlValidator extends SqlValidatorImpl {
   public SubstraitSqlValidator(Prepare.CatalogReader catalogReader) {
     super(SubstraitOperatorTable.INSTANCE, catalogReader, catalogReader.getTypeFactory(), CONFIG);
   }
+
+  public SubstraitSqlValidator(SqlOperatorTable opTable, Prepare.CatalogReader catalogReader) {
```
vbarua: minor: It's slightly nicer API-wise if we re-order the arguments so that the constructors are:
```java
public SubstraitSqlValidator(Prepare.CatalogReader catalogReader)
public SubstraitSqlValidator(Prepare.CatalogReader catalogReader, SqlOperatorTable opTable)
```

isthmus/src/main/java/io/substrait/isthmus/YamlToSqlOperator.java
vbarua, line 1: We should call this `SimpleExtensionToSqlOperator` to more accurately capture what the conversion is.

isthmus/src/main/java/io/substrait/isthmus/SubstraitToSql.java
```diff
@@ -10,7 +12,21 @@ public SubstraitToSql() {
     super(FEATURES_DEFAULT);
   }
 
+  public SubstraitToSql(List<String> yamlFunctionFiles) {
```
vbarua: I think this would be better to keep this API simple and just make it:
```java
SubstraitToSql(SimpleExtension.ExtensionCollection extensions)
```

Users can then handle loading and merging extensions any way they see fit.

isthmus/src/main/java/io/substrait/isthmus/SqlToSubstrait.java
```diff
@@ -14,31 +16,48 @@
...
+    this(SimpleExtension.loadDefaults(), features);
+  }
+
+  public SqlToSubstrait(SimpleExtension.ExtensionCollection extensions, FeatureBoard features) {
+    super(features, extensions);
+
+    SimpleExtension.ExtensionCollection customExtensionCollection =
+        ExtensionUtils.getCustomExtensions(extensions);
+    List<SqlOperator> generatedCustomOperators =
+        YamlToSqlOperator.from(customExtensionCollection, this.factory);
```
vbarua: Do we actually need to distinguish between the SqlOperators defined explicitly in the `SubstraitOperatorTable` and those that you are generating here?

I think we can just generate mappings for every function found in the extension collection, and then chain the operator tables like:
```java
SqlOperatorTables.chain(
  SubstraitOperatorTable.INSTANCE,
  SqlOperatorTables.of(generatedCustomOperators)
);
```
we automatically give precedence to the non-custom ones.

isthmus/src/main/java/io/substrait/isthmus/SqlToSubstrait.java
```diff
@@ -14,31 +16,48 @@
...
-    this(null);
+    this(SimpleExtension.loadDefaults(), null);
   }
 
   public SqlToSubstrait(FeatureBoard features) {
-    super(features);
+    this(SimpleExtension.loadDefaults(), features);
+  }
+
+  public SqlToSubstrait(SimpleExtension.ExtensionCollection extensions, FeatureBoard features) {
```
vbarua, line 30: I think this should just be
```java
public SqlToSubstrait(SqlOperatorTable operatorTable, FeatureBoard features)
```
 to give users direct control over the operator table they are using.

vbarua (commented):
I started taking a look at this today, but didn't have time to finish.

I'm going to restate what I think your doing with these changes to make sure that we're on the same page.

To start, you're generating SqlOperators dynamically from the function definitions in SimpleExtensions. You then use these to generate a SqlOperatorTable that covers functions that aren't already in the SubstraitOperatorTable.

This let's you do things like parse SQL queries with functions that are defined only in the SimpleExtension, and aren't part of the SubstraitOperatorTable.

With these definitions, you can also add additional mappings when converting from Substrait to Calcite, and back, for functions that aren't explicitly mapped in the static `SCALAR_SIGS` mapping.

## Meta Comment
I find the usage of "custom" for this concept somewhat confusing. These aren't really custom functions because they are part of the core Substrait spec. A better word for this might be "dynamic". In my head, you're effectively generating SqlOperators dynamically based on the Substrait definitions, and then using them as fall-backs when explicit mappings.

isthmus/src/main/java/io/substrait/isthmus/SqlToSubstrait.java
ZorinAnton: I'm not sure that precedence should be given the to default (non-custom) operators. What if users want to override them?

If I understood correctly, you are suggesting to generate SqlOperators for all extensions, the default ones and the dynamic:

```java
List<SqlOperator> generatedCustomOperators =
        SimpleExtensionToSqlOperator.from(dynamicExtensionCollection, this.factory);

```
I see two issues here:
- duplicated work (default operators already exist in SubstraitOperatorTable.INSTANCE
- it requires the fully fledged [TypeExpressionEvaluator](https://github.com/substrait-io/substrait-java/blob/3707fe231a011cf18d96eb35c080a2b4dbd5b0a9/core/src/main/java/io/substrait/type/TypeExpressionEvaluator.java#L7) that will probably be needed at some point anyway, but I guess this should be done in a different PR

isthmus/src/main/java/io/substrait/isthmus/SqlToSubstrait.java
ZorinAnton: In this constructor we need to call the constructor of the base class, [SqlConverterBase](https://github.com/ZorinAnton/substrait-java/blob/ad11957f3c02f1383dc5128fad83763431422341/isthmus/src/main/java/io/substrait/isthmus/SqlConverterBase.java#L21) that requires an instance of SimpleExtension.ExtensionCollection as an argument. Otherwise it uses SimpleExtension.loadDefaults() which will cause an inconsistency between the `operatorTable` and the `extensionCollection` fields, as the former will contain operators that don't have corresponding extensions in the latter.

isthmus/src/main/java/io/substrait/isthmus/SimpleExtensionToSqlOperator.java
```diff
@@ -0,0 +1,319 @@
...
+        .collect(Collectors.toList());
+  }
+
+  private static SqlFunction toSqlFunction(
+      SimpleExtension.Function function,
+      RelDataTypeFactory typeFactory,
+      TypeConverter typeConverter) {
+    List<SimpleExtension.ValueArgument> requiredArgs =
+        function.args().stream()
+            .filter(SimpleExtension.Argument::required)
```
vbarua: I believe this is meaningless because it's always `true` if you look at the implementations. I'm not even sure what `require` is used for actually.

isthmus/src/main/java/io/substrait/isthmus/SimpleExtensionToSqlOperator.java
```diff
@@ -0,0 +1,319 @@
...
+            .filter(t -> t instanceof SimpleExtension.ValueArgument)
```
vbarua: I'm not sure if it's valid to ignore Enum arguments when generating functions.

For example in [extract](https://github.com/substrait-io/substrait/blob/a5fbc0654cd77c9fa7a3ad308159dc534d7ea05f/extensions/functions_datetime.yaml#L69-L80) 
```
    impls:
      - args:
          - name: component
            options: [ YEAR, ISO_YEAR, US_YEAR, HOUR, MINUTE, SECOND,
                       MILLISECOND, MICROSECOND, SUBSECOND, PICOSECOND, UNIX_TIME, TIMEZONE_OFFSET ]
            description: The part of the value to extract.
          - name: x
            value: timestamp_tz
          - name: timezone
            description: Timezone string from IANA tzdb.
            value: string
        return: i64
```
the enumeration argument for component is part of the function signature. It might make sense to ignore function with enum arguments to start with.

isthmus/src/main/java/io/substrait/isthmus/SimpleExtensionToSqlOperator.java
```diff
@@ -0,0 +1,319 @@
...
+          opBinding.collectOperandTypes().stream()
+              .map(typeConverter::toSubstrait)
+              .collect(Collectors.toList());
+
+      TypeExpression returnExpression = function.returnType();
+      Type resolvedSubstraitType =
+          TypeExpressionEvaluator.evaluateExpression(
+              returnExpression, function.args(), substraitArgTypes);
+
+      return typeConverter.toCalcite(typeFactory, resolvedSubstraitType);
```
vbarua: You're not using the [nullability of the function](https://substrait.io/expressions/scalar_functions/#nullability-handling) to determine if the output is nullable or not. We should probably account for it.

isthmus/src/main/java/io/substrait/isthmus/SimpleExtensionToSqlOperator.java
```diff
@@ -0,0 +1,319 @@
...
+
+    SqlReturnTypeInference returnTypeInference =
+        new SubstraitReturnTypeInference(function, typeFactory, typeConverter);
+
+    return new SqlFunction(
+        function.name(),
+        SqlKind.OTHER_FUNCTION,
+        returnTypeInference,
+        null,
+        OperandTypes.family(argFamilies),
```
vbarua, line 81: I think this works fine for MIRROR and DECLARED_OUTPUT [nullabilities](https://substrait.io/expressions/scalar_functions/#nullability-handling), but will potentially overmatch on DISCRETE  which puts constraints on the nullability of _input_ arguments when matching functions.

isthmus/src/main/java/io/substrait/isthmus/SqlToSubstrait.java
vbarua: > I'm not sure that precedence should be given the to default (non-custom) operators. 

I would argue that it's preferrable to bind to the built-ins when possible, because internally in both the RelBuilder and optimizer rules Calcite looks for specific operator instances when performing optimization (e.g. comparison simplification, aggregate rewrititing, etc). If we use dynamic operators instead of built-in ones, Calcite can no longer perform these optimizations.

isthmus/src/main/java/io/substrait/isthmus/SimpleExtensionToSqlOperator.java
ZorinAnton: The `required` method is used here: https://github.com/substrait-io/substrait-java/blob/ec767204312f125155362b97ac9c07dc7e61a76c/core/src/main/java/io/substrait/extension/SimpleExtension.java#L235. I would think it is supposed to reflect whether the argument is required or optional. So probably it is not correct to always return true.

isthmus/src/main/java/io/substrait/isthmus/SimpleExtensionToSqlOperator.java
ZorinAnton: I added the code to handle EnumArgument, however I'm not sure if currently sql with functions using it can be translated to  substrait, at least I couldn't manage to do it. Comments in this [function ](https://github.com/substrait-io/substrait-java/blob/4e660971648ea2fb44df6a01e392e3c86de834c4/isthmus/src/main/java/io/substrait/isthmus/expression/FunctionConverter.java#L343) indirectly mention that.

isthmus/src/main/java/io/substrait/isthmus/SimpleExtensionToSqlOperator.java
ZorinAnton: Added handling of nullability.

isthmus/src/main/java/io/substrait/isthmus/SimpleExtensionToSqlOperator.java
ZorinAnton: Added handling of all three types of nullability.

isthmus/src/main/java/io/substrait/isthmus/SqlToSubstrait.java
ZorinAnton: Ok, I changed the order so that built-in operators have precedence.

isthmus/src/main/java/io/substrait/isthmus/SubstraitToSql.java
ZorinAnton: addressed

isthmus/src/main/java/io/substrait/isthmus/YamlToSqlOperator.java
ZorinAnton: addressed

isthmus/src/main/java/io/substrait/isthmus/sql/SubstraitSqlValidator.java
ZorinAnton: addressed

isthmus/src/main/java/io/substrait/isthmus/sql/SubstraitSqlToCalcite.java
```diff
@@ -41,6 +42,23 @@ public static RelRoot convertQuery(String sqlStatement, Prepare.CatalogReader ca
     return convertQuery(sqlStatement, catalogReader, validator, createDefaultRelOptCluster());
   }
 
+  /**
+   * Converts a SQL statement to a Calcite {@link RelRoot}.
+   *
+   * @param sqlStatement a SQL statement string
+   * @param catalogReader the {@link Prepare.CatalogReader} for finding tables/views referenced in
+   *     the SQL statement
+   * @param operatorTable the {@link SqlOperatorTable} for dynamic operators
```
vbarua: Strictly speaking, you can inject a SqlOperatorTable here that incorporates dynamic operators, but you can also inject a SqlOperatorTable that doesn't. What this allows you to do is have more fine-grained control over valid operators by allowing you to control what SqlOperatorTable you use, instead of just being forced to use the default SubstraitOperatorTable that we provide.

```suggestion
   * @param operatorTable the {@link SqlOperatorTable} for controlling valid operators
```

isthmus/src/main/java/io/substrait/isthmus/sql/SubstraitSqlToCalcite.java
```diff
@@ -72,6 +90,24 @@ public static RelRoot convertQuery(
...
   }
 
+  /**
+   * Converts one or more SQL statements to a List of {@link RelRoot}, with one {@link RelRoot} per
+   * statement.
+   *
+   * @param sqlStatements a string containing one or more SQL statements
+   * @param catalogReader the {@link Prepare.CatalogReader} for finding tables/views referenced in
+   *     the SQL statements
+   * @param operatorTable the {@link SqlOperatorTable} for dynamic operators
```
vbarua:
```suggestion
   * @param operatorTable the {@link SqlOperatorTable} for controlling valid operators
```

isthmus/src/main/java/io/substrait/isthmus/SqlExpressionToSubstrait.java
```diff
@@ -34,12 +35,12 @@ public class SqlExpressionToSubstrait extends SqlConverterBase {
...
 
   public SqlExpressionToSubstrait() {
-    this(FEATURES_DEFAULT, EXTENSION_COLLECTION);
+    this(FEATURES_DEFAULT, DefaultExtensionCatalog.DEFAULT_COLLECTION);
   }
 
   public SqlExpressionToSubstrait(
       FeatureBoard features, SimpleExtension.ExtensionCollection extensions) {
-    super(features);
+    super(features, extensions);
```
vbarua, line 43: Good fix. This would probably have resulted in some weirdness for user providing their own extensions collections.

isthmus/src/main/java/io/substrait/isthmus/SimpleExtensionToSqlOperator.java
```diff
@@ -0,0 +1,342 @@
...
+
+  private SimpleExtensionToSqlOperator() {}
+
+  public static List<SqlOperator> from(SimpleExtension.ExtensionCollection collection) {
+    return from(collection, DEFAULT_TYPE_FACTORY);
+  }
+
+  public static List<SqlOperator> from(
+      SimpleExtension.ExtensionCollection collection, RelDataTypeFactory typeFactory) {
+    TypeConverter typeConverter = TypeConverter.DEFAULT;
```
vbarua: I would suggest making the TypeConverter injectable as well to allow users with custom extension with custom types to use this capability more easily.

Effectively we can have these 2 constructors:
```java
public static List<SqlOperator> from(SimpleExtension.ExtensionCollection collection) {}
public static List<SqlOperator> from(
      SimpleExtension.ExtensionCollection collection, RelDataTypeFactory typeFactory, TypeConverter typeConverter)
```

The first for default behaviour, and the second for customizable behaviour.

isthmus/src/main/java/io/substrait/isthmus/SimpleExtensionToSqlOperator.java
```diff
@@ -0,0 +1,342 @@
...
+    List<SqlTypeFamily> argFamilies = new ArrayList<>();
+
+    for (SimpleExtension.Argument arg : function.requiredArguments()) {
+      if (arg instanceof SimpleExtension.ValueArgument) {
+        SimpleExtension.ValueArgument valueArg = (SimpleExtension.ValueArgument) arg;
+        SqlTypeName typeName = valueArg.value().accept(new CalciteTypeVisitor());
+        argFamilies.add(typeName.getFamily());
+      } else if (arg instanceof SimpleExtension.EnumArgument) {
+        // Treat an EnumArgument as a required string literal.
+        argFamilies.add(SqlTypeFamily.STRING);
```
vbarua, line 69: This seems reasonable enough for now. We don't have a good story for enum arguments at this point, but this can be revisited when we've figured that out.

isthmus/src/main/java/io/substrait/isthmus/SimpleExtensionToSqlOperator.java
```diff
@@ -0,0 +1,342 @@
...
+      SimpleExtension.Function function,
+      RelDataTypeFactory typeFactory,
+      TypeConverter typeConverter) {
+
+    List<SqlTypeFamily> argFamilies = new ArrayList<>();
+
+    for (SimpleExtension.Argument arg : function.requiredArguments()) {
+      if (arg instanceof SimpleExtension.ValueArgument) {
+        SimpleExtension.ValueArgument valueArg = (SimpleExtension.ValueArgument) arg;
+        SqlTypeName typeName = valueArg.value().accept(new CalciteTypeVisitor());
```
vbarua: minor: The CalciteTypeVisitor is stateless, so it might be worth creating one at the class level once rather than instantiating one of for every individual argument we process.

isthmus/src/main/java/io/substrait/isthmus/SimpleExtensionToSqlOperator.java
```diff
@@ -0,0 +1,342 @@
...
+      boolean finalIsNullable;
+      switch (function.nullability()) {
+        case MIRROR:
+          // If any input is nullable, the output is nullable.
+          finalIsNullable =
+              opBinding.collectOperandTypes().stream().anyMatch(RelDataType::isNullable);
+          break;
+        case DISCRETE:
+          // The function can return null even if inputs are not null.
+          finalIsNullable = true;
```
vbarua: DISCRETE should also use the nullability as declared in the Substrait type like DECLARED_OUTPUT does. It's a _stricter_ version of DECLARED_OUTPUT, because the argument nullabilities must also match.

See https://substrait.io/expressions/scalar_functions/#nullability-handling

isthmus/src/main/java/io/substrait/isthmus/SimpleExtensionToSqlOperator.java
```diff
@@ -0,0 +1,342 @@
...
+
+    @Override
+    public SqlTypeName visit(ParameterizedType.IntervalDay expr) {
+      return SqlTypeName.INTERVAL_DAY;
+    }
+
+    @Override
+    public SqlTypeName visit(ParameterizedType.IntervalCompound expr) {
+      // TODO: double check
+      return SqlTypeName.INTERVAL_DAY_HOUR;
```
vbarua: I don't thinks this is correct. [IntervalCompound](https://substrait.io/types/type_classes/#compound-types) is defined as
> A compound interval type that is composed of elements of the underlying elements and rules of both interval_month and interval_day to express arbitrary durations across multiple grains. Substrait gives no definition for the conversion of values between independent grains (e.g. months to days).

So it would be a combination of Calcite's `INTERVAL_DAY` and `INTERVAL_YEAR_MONTH`.

I would suggest avoiding dealing with this mapping for and and just not generating dynamic operators for Substrait functions that use it. For users trying to process SQL, this shouldn't cause any issues because Calcite can't generate that type. For users trying to process Substrait plans containing those functions into Calcite, those would fail anyways because the [TypeConverter](https://github.com/substrait-io/substrait-java/blob/main/isthmus/src/main/java/io/substrait/isthmus/TypeConverter.java) from Substrait to Calcite can't handle it currently anyways.

isthmus/src/test/java/io/substrait/isthmus/PlanTestBase.java
```diff
@@ -144,6 +154,39 @@ protected RelRoot assertSqlSubstraitRelRoundTrip(
...
+    // sql <--> substrait round trip test.
+    // Assert (sql -> calcite -> substrait) and (sql -> substrait -> calcite -> substrait) are same.
+    // Return list of sql -> Substrait rel -> Calcite rel.
+
+    SubstraitToCalcite substraitToCalcite = new SubstraitToCalcite(extensions, typeFactory);
+
+    SqlToSubstrait s = new SqlToSubstrait(extensions, null);
+
+    // 1. SQL -> Calcite RelRoot
+    Plan plan1 = s.convert(query, catalogReader);
```
vbarua, line 188: A number of the comments in this test utility don't line up with what it's actually doing. For example, this is a Substrait Plan, not a Calcite RelRoot.

isthmus/src/test/java/io/substrait/isthmus/PlanTestBase.java
```diff
@@ -144,6 +154,39 @@ protected RelRoot assertSqlSubstraitRelRoundTrip(
     return relRoot2;
   }
 
+  protected RelRoot assertSqlSubstraitRelRoundTripWorkaroundOptimizer(
```
vbarua: I might suggest a name like `assertSqlSubstraitRelRoundTripLoosePOJOComparison` along with a docstring to indicate that loose here means not comparing the POJOs to the first POJO due to optimizer differences.

isthmus/src/test/java/io/substrait/isthmus/SimpleExtensionToSqlOperatorTest.java
```diff
@@ -0,0 +1,46 @@
...
+    Optional<SqlOperator> function =
+        operators.stream()
+            .filter(op -> op.getName().equalsIgnoreCase("REGEXP_EXTRACT"))
+            .findFirst();
+
+    assertTrue(function.isPresent(), "The REGEXP_EXTRACT function should be present.");
+
+    SqlOperator op = function.get();
+    System.out.println("Successfully found and verified Custom UDF:");
+    System.out.printf("  - Name: %s%n", op.getName());
```
vbarua: We should avoid printing in tests to not clutter test runs.

isthmus/src/test/resources/extensions/scalar_functions_custom.yaml
```diff
@@ -0,0 +1,45 @@
...
+
+  - name: "format_text"
+    description: "Formats text based on a mode. The output is nullable if the input is."
+    impls:
+      - args:
+          - name: "mode"
+#            options: ["UPPER", "LOWER"]
+            value: string
+          - name: "input_text"
+#            options: ["UPPER", "LOWER"]
```
vbarua: Do these options still need to be uncommented?

isthmus/src/main/java/io/substrait/isthmus/SimpleExtensionToSqlOperator.java
```diff
@@ -0,0 +1,342 @@
...
+
+  public static List<SqlOperator> from(SimpleExtension.ExtensionCollection collection) {
+    return from(collection, DEFAULT_TYPE_FACTORY);
+  }
+
+  public static List<SqlOperator> from(
+      SimpleExtension.ExtensionCollection collection, RelDataTypeFactory typeFactory) {
+    TypeConverter typeConverter = TypeConverter.DEFAULT;
+    return Stream.concat(
+            collection.scalarFunctions().stream(), collection.aggregateFunctions().stream())
```
vbarua, line 50: I assume that plan is to add support for window functions as a followup?

isthmus/src/main/java/io/substrait/isthmus/SimpleExtensionToSqlOperator.java
```diff
@@ -0,0 +1,342 @@
...
+    }
+
+    @Override
+    public SqlTypeName visit(Type.IntervalDay day) {
+      return SqlTypeName.INTERVAL_DAY;
+    }
+
+    @Override
+    public SqlTypeName visit(Type.UUID expr) {
+      return SqlTypeName.VARCHAR;
```
vbarua: Can we use `SqlTypeName.UUID` here?

isthmus/src/main/java/io/substrait/isthmus/SimpleExtensionToSqlOperator.java
```diff
@@ -0,0 +1,342 @@
...
+        case "TIMESTAMP":
+          return SqlTypeName.TIMESTAMP;
+        case "TIMESTAMP_TZ":
+          return SqlTypeName.TIMESTAMP_WITH_LOCAL_TIME_ZONE;
+        case "DATE":
+          return SqlTypeName.DATE;
+        case "TIME":
+          return SqlTypeName.TIME;
+        case "UUID":
+          return SqlTypeName.VARCHAR;
```
vbarua: Could this be `SqlTypeName.UUID`?

isthmus/src/main/java/io/substrait/isthmus/ExtensionUtils.java
```diff
@@ -0,0 +1,38 @@
...
+import io.substrait.extension.SimpleExtension;
+import io.substrait.isthmus.calcite.SubstraitOperatorTable;
+import java.util.List;
+import java.util.Locale;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+public class ExtensionUtils {
+
+  public static SimpleExtension.ExtensionCollection getDynamicExtensions(
```
vbarua, line 32: We should add a docstring to this to explain what it's doing, and what is considered a dynamic extension.

isthmus/src/main/java/io/substrait/isthmus/ExtensionUtils.java
```diff
@@ -0,0 +1,38 @@
...
+      SimpleExtension.ExtensionCollection extensions) {
+    Set<String> knownFunctionNames =
+        SubstraitOperatorTable.INSTANCE.getOperatorList().stream()
+            .map(op -> op.getName().toLowerCase(Locale.ROOT))
+            .collect(Collectors.toSet());
+
+    List<SimpleExtension.ScalarFunctionVariant> customFunctions =
+        extensions.scalarFunctions().stream()
+            .filter(f -> !knownFunctionNames.contains(f.name().toLowerCase(Locale.ROOT)))
```
vbarua, line 41: I don't know if this is the correct way to filter for functions. In `createScalarFuntionConverter`, your filtering the functions using the mappings in `FunctionMappings.SCALAR_SIGS`. Is there are reason not to do that here? 

Also, would it make sense to consolidate that filtering functionality here as well?

isthmus/src/main/java/io/substrait/isthmus/ExtensionUtils.java
```diff
@@ -0,0 +1,38 @@
...
+        .build();
+  }
+
+  public static SimpleExtension.ExtensionCollection loadExtensions(List<String> yamlFunctionFiles) {
+    SimpleExtension.ExtensionCollection allExtensions = DefaultExtensionCatalog.DEFAULT_COLLECTION;
+    if (yamlFunctionFiles != null && !yamlFunctionFiles.isEmpty()) {
+      allExtensions = allExtensions.merge(SimpleExtension.load(yamlFunctionFiles));
+    }
+    return allExtensions;
+  }
```
vbarua: It doesn't appear that this is used. Do we need this or can we delete this?

isthmus/src/test/java/io/substrait/isthmus/SimpleExtensionToSqlOperatorTest.java
```diff
@@ -0,0 +1,46 @@
...
+import java.util.List;
+import java.util.Optional;
+import org.apache.calcite.sql.SqlOperandCountRange;
+import org.apache.calcite.sql.SqlOperator;
+import org.junit.jupiter.api.Test;
+
+class SimpleExtensionToSqlOperatorTest {
+
+  @Test
+  void test() throws IOException {
```
vbarua: This test does demonstrate that we can produce a SqlOperator for a function declared in an extension, but it doesn't necessarily show that the argument and return type inference align of the SqlOperator are aligned with the function declaration?

How feasible do you think it would be to check that the operators we are producing:
* Apply the correct argument restrictions
* Have the correct return type inference

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
```diff
@@ -153,11 +152,45 @@ public SubstraitRelNodeConverter(
     this.expressionRexConverter.setRelNodeConverter(this);
   }
 
+  private static ScalarFunctionConverter createScalarFunctionConverter(
```
vbarua, line 165: My preference would be to make a function like this available to users so that they can produce a scalar function converter with dynamic mappings, but not force users to use dynamic mappings.

vbarua (commented):
Found some time to take a look at this again.

Overall I think it's pretty reasonable feature, but I'm a little wary of enabling it by default everywhere like you have because I can't fully think of the implications of having dynamic mappings in places where they weren't expected before. Examples of things I'm worried about are:
* Systems that might actually prefer mappings to fail in order to warn users about missing features.
* Dynamic function generation failures with custom user catalogs.

Ideally, I would love to make this capability something that users can opt into. That way, we could release it and users could try it out in their system and report any issues. The way you've structured it now though, it's pretty difficult to do this because the dynamic mappings are created as part of the creating the various converters that are used to to convert between SQL, Substrait and Calcite.

How difficult do you think it would be to make the dynamic mapping behaviour opt-in?

isthmus/src/main/java/io/substrait/isthmus/sql/SubstraitSqlToCalcite.java
ZorinAnton: fixed

isthmus/src/main/java/io/substrait/isthmus/sql/SubstraitSqlToCalcite.java
ZorinAnton: fixed

isthmus/src/main/java/io/substrait/isthmus/SimpleExtensionToSqlOperator.java
ZorinAnton: fixed

isthmus/src/main/java/io/substrait/isthmus/SimpleExtensionToSqlOperator.java
ZorinAnton: fixed

isthmus/src/main/java/io/substrait/isthmus/SimpleExtensionToSqlOperator.java
ZorinAnton: fixed

isthmus/src/main/java/io/substrait/isthmus/SimpleExtensionToSqlOperator.java
ZorinAnton: dropped the overridden 'visit' method

isthmus/src/test/java/io/substrait/isthmus/PlanTestBase.java
ZorinAnton: fixed

isthmus/src/test/java/io/substrait/isthmus/PlanTestBase.java
ZorinAnton: fixed

isthmus/src/test/java/io/substrait/isthmus/SimpleExtensionToSqlOperatorTest.java
ZorinAnton: fixed

isthmus/src/test/resources/extensions/scalar_functions_custom.yaml
ZorinAnton: dropped

isthmus/src/main/java/io/substrait/isthmus/SimpleExtensionToSqlOperator.java
ZorinAnton: yes, I added a "TODO" comment

isthmus/src/main/java/io/substrait/isthmus/SimpleExtensionToSqlOperator.java
ZorinAnton: fixed

isthmus/src/main/java/io/substrait/isthmus/SimpleExtensionToSqlOperator.java
ZorinAnton: fixed

isthmus/src/main/java/io/substrait/isthmus/ExtensionUtils.java
ZorinAnton: fixed

isthmus/src/main/java/io/substrait/isthmus/ExtensionUtils.java
ZorinAnton: fixed

isthmus/src/main/java/io/substrait/isthmus/ExtensionUtils.java
ZorinAnton: dropped

isthmus/src/test/java/io/substrait/isthmus/SimpleExtensionToSqlOperatorTest.java
ZorinAnton: Extended to verify number of args, their types and the return type of dynamic functions.

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
ZorinAnton: fixed

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelVisitor.java
```diff
@@ -89,10 +91,25 @@ public SubstraitRelVisitor(
...
       SimpleExtension.ExtensionCollection extensions,
       FeatureBoard features) {
+
+    SimpleExtension.ExtensionCollection dynamicExtensionCollection =
+        ExtensionUtils.getDynamicExtensions(extensions);
+    List<SqlOperator> dynamicOperators =
+        SimpleExtensionToSqlOperator.from(dynamicExtensionCollection, typeFactory);
+
+    List<FunctionMappings.Sig> additionalSignatures =
+        dynamicOperators.stream()
```
vbarua: Dynamic operators are still enabled by default here.

isthmus/src/test/java/io/substrait/isthmus/PlanTestBase.java
```diff
@@ -144,6 +154,75 @@ protected RelRoot assertSqlSubstraitRelRoundTrip(
...
+    return relRoot2;
+  }
+
+  /**
+   * Convenience overload of {@link #assertSqlSubstraitRelRoundTripLoosePojoComparison(String,
+   * Prepare.CatalogReader, FeatureBoard)} with default FeatureBoard behavior (no dynamic UDFs).
+   */
+  protected RelRoot assertSqlSubstraitRelRoundTripLoosePojoComparison(
+      String query, Prepare.CatalogReader catalogReader) throws Exception {
+    return assertSqlSubstraitRelRoundTripLoosePojoComparison(query, catalogReader, null);
```
vbarua: Instead of using `null` here, I would just pass int eh default feature board and avoid the null checking code in the method above.

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelVisitor.java
ZorinAnton: fixed

isthmus/src/test/java/io/substrait/isthmus/PlanTestBase.java
ZorinAnton: fixed

vbarua (approved):
Thanks for working on this.

PR #466, opened 2025-08-06
build: ignore dependabot commits during commitlint

nielspardon:
This PR adds a dependabot config for Gradle and Github Actions dependencies.

I tested it on my fork: https://github.com/nielspardon/substrait-java/pulls

I tried the git submodule config but that is limited in that it does not support further configuration to e.g. only update to tags so I would not use dependabot for the git submodule updates.

I had to modify the `commitlint` configuration to allow the dependabot PRs to bypass the commit linting. The reason is that dependabot PRs create commits where the commit body contains lines longer than 100 characters which is otherwise not allowed by the current `commitlint` config.

### DIFF
--- a/.commitlintrc.js
+++ b/.commitlintrc.js
@@ -1,4 +1,4 @@
-{
+module.exports = {
   "rules": {
     "body-leading-blank": [1, "always"],
     "body-max-line-length": [2, "always", 100],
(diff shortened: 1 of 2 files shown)


nielspardon:
> You might want to configure Dependabot with a [commit message prefix](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/customizing-dependabot-prs#adding-a-prefix-to-commit-messages) to get closer to the conventional commits format, although it seems to be doing a pretty good job of deriving the format based on previous commits anyway.

good idea. done.

nielspardon:
do you prefer `chore(deps):` or `build(deps):` as the prefix?

vbarua:
`build(deps)` seem reasonable to me. Also TIL that `chore` is no longer part of the Angular convention https://stackoverflow.com/questions/26944762/when-to-use-chore-as-type-of-commit-message

nielspardon:
I changed the commit prefix config to `build(deps): ` and updated the commtilint config to ignore the prefixed commit message style

PR #467, opened 2025-08-07
build: use Gradle version catalog for dependencies and plugins

nielspardon:
The way dependency versions are currently declared is not working well with dependabot. To make it easier for dependabot to detect the dependency versions I suggest to move the dependency versions into a [Gradle version catalog](https://docs.gradle.org/current/userguide/version_catalogs.html) (`libs.versions.toml`).

This PR moves both the dependency and Gradle plugin versions into a version catalog.

### DIFF
--- a/bom/build.gradle.kts
+++ b/bom/build.gradle.kts
@@ -21,30 +21,12 @@ plugins {
 
 val String.v: String get() = rootProject.extra["$this.version"] as String
 
-// Note: Gradle allows to declare dependency on "bom" as "api",
-// and it makes the contraints to be transitively visible
-// However Maven can't express that, so the approach is to use Gradle resolution
-// and generate pom files with resolved versions
-// See https://github.com/gradle/gradle/issues/9866
-
-fun DependencyConstraintHandlerScope.apiv(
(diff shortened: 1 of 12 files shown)


bestbeforetoday (approved):
This is a really nice improvement to dependency version management, and something I was planning to get to myself.

If you haven’t already, it might be worth just checking the publish process (using JReleaser) runs locally without unexpected errors since it is sensitive to how the jreleaser plugin is loaded. You will need a real keys in gpg for signing but the Maven Central credentials should not matter since you don’t actually want to publish anyway; just confirm the process does not fail before it attempts to publish.

```bash
export SIGNING_KEY_ID=...
export SIGNING_PASSWORD=...
export SIGNING_KEY=...
export JRELEASER_MAVENCENTRAL_USERNAME=...
export JRELEASER_MAVENCENTRAL_PASSWORD=...

./gradlew clean
./gradlew publishAllPublicationsToStagingRepository
JRELEASER_MAVENCENTRAL_STAGE=UPLOAD ./gradlew jreleaserDeploy
```

Otherwise looks great to me.

PR #469, opened 2025-08-07
fix(core): update hints to match substrait spec

strausza:
This PR includes the missing fields for `Hint` and adjusts the proto<->java converters accordingly.


core/src/main/java/io/substrait/hint/Hint.java
```diff
@@ -1,21 +1,91 @@
...
+
+  public abstract List<SavedComputation> getSavedComputations();
+
+  public enum ComputationType {
+    COMPUTATION_TYPE_UNSPECIFIED(RelCommon.Hint.ComputationType.COMPUTATION_TYPE_UNSPECIFIED),
+    COMPUTATION_TYPE_HASHTABLE(RelCommon.Hint.ComputationType.COMPUTATION_TYPE_HASHTABLE),
+    COMPUTATION_TYPE_BLOOM_FILTER(RelCommon.Hint.ComputationType.COMPUTATION_TYPE_BLOOM_FILTER),
+    COMPUTATION_TYPE_UNKNOWN(RelCommon.Hint.ComputationType.COMPUTATION_TYPE_UNKNOWN);
+
+    private RelCommon.Hint.ComputationType proto;
```
vbarua: minor: this can be made `final`

core/src/main/java/io/substrait/relation/ProtoRelConverter.java
```diff
@@ -882,14 +887,56 @@ protected static Optional<Rel.Remap> optionalRelmap(io.substrait.proto.RelCommon
...
+    if (hint.hasConstraint()) {
+      io.substrait.proto.RelCommon.Hint.RuntimeConstraint constraint = hint.getConstraint();
+      io.substrait.hint.ImmutableRuntimeConstraint.Builder constraintBuilder =
+          RuntimeConstraint.builder();
+      if (constraint.hasAdvancedExtension()) {
+        constraintBuilder.extension(advancedExtension(constraint.getAdvancedExtension()));
+      }
+      builder.runtimeConstraint(constraintBuilder.build());
+    }
+    if (!hint.getLoadedComputationsList().isEmpty()) {
```
vbarua: minor: I think you can omit the `!hint.getLoadedComputationsList().isEmpty()` check here, because the default value for this when it is not set is the empty list. It should always be valid to call `.forEach`.

This also applies to the processing for `getSavedComputationsList()` below.

core/src/test/java/io/substrait/relation/ProtoRelConverterTest.java
```diff
@@ -130,28 +137,66 @@ void extensionTable() {
...
+      builder.rowCount(42).recordSize(42);
+      if (includeEmptyOptimization) {
+        builder.extension(AdvancedExtension.builder().addOptimizations().build());
+      }
+      return builder.build();
+    }
+
+    LoadedComputation getLoadedComputation() {
+      return LoadedComputation.builder()
+          .computationId(0)
```
vbarua: Let's use id `1` here instead of `0`. `0` is the default value for the field if it's not set in the protobuf, so a non-zero field is better for checking conversion logic. Also, `1` lines up better semantically with your saved computation below which uses value `1` (i.e. we should load an existing computation).

core/src/test/java/io/substrait/relation/ProtoRelConverterTest.java
```diff
@@ -130,28 +137,66 @@ void extensionTable() {
...
+          .computationType(ComputationType.COMPUTATION_TYPE_UNSPECIFIED)
```
vbarua: As above, let's use a non UNSPECIFIED value for this. UNSPECIFIED is also a default value if none is set. We should set the same computation type here as below.

core/src/test/java/io/substrait/relation/ProtoRelConverterTest.java
```diff
@@ -130,28 +137,66 @@ void extensionTable() {
...
+          .build();
+    }
+
+    RuntimeConstraint getRuntimeConstraint(boolean includeEmptyOptimization) {
+      ImmutableRuntimeConstraint.Builder builder = RuntimeConstraint.builder();
+      if (includeEmptyOptimization) {
+        builder.extension(AdvancedExtension.builder().addOptimizations().build());
+      }
+      return builder.build();
+    }
```
vbarua, line 169: minor: let's change all the `get` methods to `create`. You're not really getting any of these objects, you're creating them.

core/src/test/java/io/substrait/relation/ProtoRelConverterTest.java
```diff
@@ -130,28 +137,66 @@ void extensionTable() {
...
     @Test
-    void relWithHint() {
-      Rel relWithHints =
-          NamedScan.builder()
-              .from(commonTable)
-              .hint(Hint.builder().addOutputNames("Test hint").build())
+    void relWithCompleteHint() {
+      // Note, we must leave the advancedExtensions empty, otherwise the conversion fails (see tests
+      // for advancedExtensions above). However, this still tests whether they are correctly
+      // propagated for the hints.
```
vbarua: I don't fully understand this comment. You are setting advanced extensions in a bunch of the components below when you pass `true` to your constructors aren't you?

vbarua (commented):
Left some minor comments, but overall this looks great. Thanks for adding this, it's super thorough.

core/src/test/java/io/substrait/relation/ProtoRelConverterTest.java
strausza: I agree with not using defaults.

I've had different values there as Saved/LoadedComputation share the same field names which easily allows for typos (see [here](https://github.com/substrait-io/substrait-java/compare/c7cc2cd9a7114e0e541cb82a08c6fa7c1e99e8ec..e5a11955b779af798087eda36047d61023e36f31)...) and using the same values wouldn't catch this. I've added now two more tests, checking them separately.

core/src/main/java/io/substrait/relation/ProtoRelConverter.java
strausza: fixed

core/src/test/java/io/substrait/relation/ProtoRelConverterTest.java
strausza: fixed

core/src/test/java/io/substrait/relation/ProtoRelConverterTest.java
strausza: this was obselete, now removed, sorry.

core/src/test/java/io/substrait/relation/ProtoRelConverterTest.java
```diff
@@ -130,28 +137,100 @@ void extensionTable() {
   @Nested
   class HintsTest {
 
+    Stats getStats(boolean includeEmptyOptimization) {
```
vbarua: minor: `createStats` instead of `getStats`

strausza: I missed this, sorry. Fixed now.

PR #473, opened 2025-08-11
feat(isthmus): add method converting SQL to POJO Plan

nielspardon:
I noticed that #432 could benefit from this refactoring.

BREAKING CHANGE: deprecates SqlToSubstrait.execute()

- adds a new `convert()` method to `SqlToSubstrait` which converts a SQL statements string to a POJO `Plan` instead of a proto `Plan`
- deprecates the `execute()` method in `SqlToSubstrait` which converts a SQL statements string to a proto `Plan`
- changes any uses of `execute()` in the code to `convert()`
- removes the `@VisibleForTesting` annotation from `SqlToSubstrait.createSqlToRelConverter()` and `SqlToSubstrait.getBestExpRelRoot()` since they are not actually used in testing and makes them `protected` instead of package private
- adds `@VisibleForTesting` annotation to `SqlToSubstrait.sqlToRelNode()` which is actually still used in testing
- changes the `TpcdsQueryTest` and `TpchQueryTest` to test POJO to SQL and PROTO to SQL separately since I noticed that TPC-DS query 67 only fails for PROTO to SQL but started working to POJO to SQL

### DIFF
--- a/isthmus-cli/src/main/java/io/substrait/isthmus/cli/IsthmusEntryPoint.java
+++ b/isthmus-cli/src/main/java/io/substrait/isthmus/cli/IsthmusEntryPoint.java
@@ -10,6 +10,7 @@ import io.substrait.isthmus.ImmutableFeatureBoard;
 import io.substrait.isthmus.SqlExpressionToSubstrait;
 import io.substrait.isthmus.SqlToSubstrait;
 import io.substrait.isthmus.sql.SubstraitCreateStatementParser;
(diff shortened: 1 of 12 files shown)


vbarua (approved):
Changes look good to me. This in inpsired me to push through my own changes extending these. I'll open a PR for those shortly.

Also, I don't believe this a breaking change as you've only deprecated the method, not removed it.

PR #474, opened 2025-08-11
feat(isthmus): introduce SubstraitSqlToCalcite and SubstraitStatementParser

vbarua:


isthmus/src/test/java/io/substrait/isthmus/ApplyJoinPlanTest.java
```diff
@@ -11,10 +12,7 @@
 
 public class ApplyJoinPlanTest extends PlanTestBase {
```
nielspardon: when I touched this class yesterday I got the impression this class mainly tests the SQL to Calcite conversion and that SQL to Substrait does not work. especially the first part where it validates the SQL to Calcite conversion feels redundant since that should have been covered by the Apache Calcite project already.

then if the first part of the tests is redundant then we don't need certain SubstraitSqlToCalcite methods anymore

isthmus/src/main/java/io/substrait/isthmus/sql/SubstraitSqlToCalcite.java
```diff
@@ -0,0 +1,113 @@
...
+import org.apache.calcite.sql2rel.SqlToRelConverter;
+import org.apache.calcite.sql2rel.StandardConvertletTable;
+
+/**
+ * Substrait flavoured SQL processor provided as a utility for testing and experimentation,
+ * utilizing {@link SubstraitStatementParser} and {@link SubstraitSqlValidator}
+ */
+public class SubstraitSqlToCalcite {
+
+  public static RelRoot convertQuery(String statement, Prepare.CatalogReader catalogReader)
```
nielspardon: maybe we should start adding more javadoc comments to public methods when we introduce / touch them

vbarua: I think that's a great idea. I'll add some today.

isthmus/src/test/java/io/substrait/isthmus/ApplyJoinPlanTest.java
vbarua: > especially the first part where it validates the SQL to Calcite conversion feels redundant since that should have been covered by the Apache Calcite project already.

That does sound reasonable to me, but one thing to note is that the `OuterReferenceResolver` is in Isthmus so the checks for that aren't strictly Calcite related.

> I got the impression this class mainly tests the SQL to Calcite conversion and that SQL to Substrait does not work.

There are an awful lot of `Assertions.assertThrows` in this.

> is redundant then we don't need certain SubstraitSqlToCalcite methods anymore

Which methods specifically do you think are redundant?

nielspardon: > That does sound reasonable to me, but one thing to note is that the `OuterReferenceResolver` is in Isthmus so the checks for that aren't strictly Calcite related.

I see. Then I think this test class should be about testing the `OuterReferenceResolver` and not about the conversion of the various join type SQLs to Substrait.

I would move the first section of each of the test methods which builds up the fieldAccessDepthMap and test that the OuterReferenceResolver is doing its job into its own `OuterReferenceResolverTest` class.

nielspardon: Also I would find it much cleaner if the `OuterReferenceResolverTest` would start from a Calcite plan that was built from `RelBuilder` since effectively the `OuterReferenceResolver` works off of the Calcite `RelRoot/RelNode`. While the SQL to Substrait conversion works off of the SQL string.

nielspardon: this is my proposal for refactoring this test class: https://github.com/substrait-io/substrait-java/pull/498

isthmus/src/main/java/io/substrait/isthmus/sql/SubstraitSqlToCalcite.java
```diff
@@ -0,0 +1,163 @@
...
+      throw new IllegalArgumentException(
+          String.format("Expected one statement, found: %d", sqlNodes.size()));
+    }
+    List<RelRoot> relRoots = convert(sqlNodes, catalogReader, validator, cluster);
+    // as there was only 1 statement, there should only be 1 root
+    return relRoots.get(0);
+  }
+
+  /**
+   * Converts one or more SQL statement to a List of {@link RelRoot}, with one {@link RelRoot} per
```
nielspardon: minor: I noticed in `SubstraitSqlStatementParser` you were using the plural here:

```suggestion
   * Converts one or more SQL statements to a List of {@link RelRoot}, with one {@link RelRoot} per
```

isthmus/src/main/java/io/substrait/isthmus/sql/SubstraitSqlToCalcite.java
```diff
@@ -0,0 +1,163 @@
...
+   * @throws SqlParseException if there is an error while parsing the SQL statements
+   */
+  public static List<RelRoot> convertQueries(
+      String sqlStatements, Prepare.CatalogReader catalogReader) throws SqlParseException {
+    SqlValidator validator = new SubstraitSqlValidator(catalogReader);
+    return convertQueries(sqlStatements, catalogReader, validator, createDefaultRelOptCluster());
+  }
+
+  /**
+   * Converts one or more SQL statement to a List of {@link RelRoot}, with one {@link RelRoot} per
```
nielspardon: minor: I noticed in `SubstraitSqlStatementParser` you were using the plural here:

```suggestion
   * Converts one or more SQL statements to a List of {@link RelRoot}, with one {@link RelRoot} per
```

isthmus/src/main/java/io/substrait/isthmus/sql/SubstraitSqlStatementParser.java
```diff
@@ -0,0 +1,34 @@
...
+
+  private static final SqlParser.Config PARSER_CONFIG =
+      SqlParser.config()
+          // TODO: switch to Casing.UNCHANGED
+          .withUnquotedCasing(Casing.TO_UPPER)
+          // use LENIENT conformance to allow for parsing a wide variety of dialects
+          .withConformance(SqlConformanceEnum.LENIENT);
+
+  /**
+   * Parse one or more SQL statements to a list of {@link SqlNode}s
```
nielspardon: don't remember whether that was a javadoc or Google code style checkstyle thing but usually the description in a javadoc ends with a full stop:
```suggestion
   * Parses one or more SQL statements to a list of {@link SqlNode}s.
```

nielspardon (approved):
~LGTM with the few minor documentation fixes~ please see comment below

isthmus/src/test/java/io/substrait/isthmus/NameRoundtripTest.java
```diff
@@ -18,17 +19,18 @@ void preserveNamesFromSql() throws Exception {
...
         new SubstraitToCalcite(EXTENSION_COLLECTION, typeFactory);
 
     String query = "SELECT \"a\", \"B\" FROM foo GROUP BY a, b";
     List<String> expectedNames = List.of("a", "B");
 
-    Plan plan = s.convert(query, catalogReader);
-    assertEquals(1, plan.getRoots().size());
+    org.apache.calcite.rel.RelRoot calciteRelRoot1 =
+        SubstraitSqlToCalcite.convertQuery(query, catalogReader);
+    assertEquals(expectedNames, calciteRelRoot1.validatedRowType.getFieldNames());
```
nielspardon, line 30: This change I don't fully follow. How is it an improvement that we are testing SQL to Calcite before we test SQL to Substrait? Shouldn't it be sufficient if we test SQL to Substrait since that is our non-experimentation public API? Shouldn't it be obvious that Calcite's SQL To RelNode conversion preserves the names?

isthmus/src/test/java/io/substrait/isthmus/PlanTestBase.java
```diff
@@ -129,13 +130,11 @@ protected RelRoot assertSqlSubstraitRelRoundTrip(
...
+    // 1. SQL -> Calcite RelRoot
+    RelRoot relRoot1 = SubstraitSqlToCalcite.convertQuery(query, catalogReader);
 
-    // 1. SQL -> Substrait Plan
-    Plan plan1 = s.convert(query, catalogReader);
-
-    // 2. Substrait Plan  -> Substrait Rel
-    Plan.Root pojo1 = plan1.getRoots().get(0);
+    // 2. Calcite RelRoot  -> Substrait Rel
+    Plan.Root pojo1 = SubstraitRelVisitor.convert(relRoot1, extensions);
```
nielspardon: similarly here: what is the benefit of breaking out the test conversion into Sql to Calcite and Calcite to Substrait if we could simply use `SqlToSubstrait` which is performing these two steps for us in one go?

nielspardon (changes requested):
sorry, approved maybe a little to soon. can you enlighten me on the reasoning behind the changes to `NameRoundtripTest` and `PlanTestBase`. What are we gaining by breaking these into multiple steps instead of using the `SqlToSubstrait` wrapper?

isthmus/src/test/java/io/substrait/isthmus/PlanTestBase.java
vbarua: Ah, gotcha. You're right, in this case it has limited utility because we're not checking any properties between relRoot1 and relRoot2. I'll remove this.

isthmus/src/test/java/io/substrait/isthmus/NameRoundtripTest.java
vbarua: > Shouldn't it be obvious that Calcite's SQL To RelNode conversion preserves the names?

You would think so, but if you get rid of the quoting the (current) parser settings will uppercase all the identifiers.  This was checked indirectly, but because I had access to a convenient converter from Sql to Calcite now I felt that adding a direct check for this would be helpful.

isthmus/src/main/java/io/substrait/isthmus/sql/SubstraitSqlStatementParser.java
vbarua: Makes sense, I've gone ahead and updated it. Is this the kind of thing that could be enforced by PMD?

nielspardon: This is something that the Checkstyle framework can help with: https://checkstyle.org/checks/javadoc/javadocstyle.html#JavadocStyle

> Ensures the first sentence ends with proper punctuation (That is a period, question mark, or exclamation mark, by default).

PR #498, opened 2025-08-14
refactor(ishtmus): split ApplyJoinPlanTest

nielspardon:
splits ApplyJoinPlanTest into OuterReferenceResolverTest and SqlToSubstraitJoinConversionTest


isthmus/src/test/java/io/substrait/isthmus/RelCreator.java
vbarua, line 1: Now that I'm reminded of the existence of this class, I want to put it on the chopping block as well. Effectively, this class exists purely to orchestrate Calcite code and internally is only used in test. I don't think it should be part of our public API. Not something for this PR though.

nielspardon: the good thing is that `RelCreator` is at least only in `src/test` and not in `src/main` :-)

PR #503, opened 2025-08-26
fix(isthmus): tpcds q67

ZorinAnton:
The fix handles conversion between sql and substrait, when aggregation is done using `rollup` function and there is an outer projection. Minimal representative example:

Failing:

```sql

select i_category from (
select i_category, count(*) from item group by
rollup(
i_category
)
)
```
Workable:

```sql

select i_category from (
select i_category, count(*) from item group by
--rollup(
i_category
--)
)
```

### DIFF
--- a/core/src/main/java/io/substrait/relation/ProtoRelConverter.java
+++ b/core/src/main/java/io/substrait/relation/ProtoRelConverter.java
@@ -620,8 +620,19 @@ public class ProtoRelConverter {
 
   protected Aggregate newAggregate(AggregateRel rel) {
     Rel input = from(rel.getInput());
+    Type.Struct inputSchema;
+    if (input instanceof Project) {
+      List<Type> types =
+          ((Project) input)
+              .getExpressions().stream().map(Expression::getType).collect(Collectors.toList());
+      inputSchema = Type.Struct.builder().fields(types).nullable(false).build();
+    } else {
+      inputSchema = input.getRecordType();
+    }
+
     ProtoExpressionConverter protoExprConverter =
-        new ProtoExpressionConverter(lookup, extensions, input.getRecordType(), this);
+        new ProtoExpressionConverter(lookup, extensions, inputSchema, this);
+
     ProtoAggregateFunctionConverter protoAggrFuncConverter =
         new ProtoAggregateFunctionConverter(lookup, extensions, protoExprConverter);
 
--- a/isthmus/src/main/java/io/substrait/isthmus/SubstraitRelVisitor.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/SubstraitRelVisitor.java
@@ -244,9 +244,60 @@ public class SubstraitRelVisitor extends RelNodeVisitor<Rel, RuntimeException> {
     return Set.builder().inputs(inputs).setOp(setOp).build();
   }
 
+  /**
+   * Pre-processes the input to an Aggregate relation to handle nullability changes introduced by
+   * ROLLUP/CUBE/GROUPING SETS.
+   *
+   * @param aggregate The original Calcite aggregate node.
+   * @return A Substrait Rel node that is correctly typed to be the input to the Substrait
+   *     Aggregate.
+   */
+  private Rel handleRollupCorrection(org.apache.calcite.rel.core.Aggregate aggregate) {
+    Rel originalInput = apply(aggregate.getInput());
+
+    // Determine the correct final output type for the aggregate, which accounts for nullability.
+    NamedStruct aggregateOutputType = typeConverter.toNamedStruct(aggregate.getRowType());
+    List<Integer> groupKeyIndices = aggregate.getGroupSet().asList();
+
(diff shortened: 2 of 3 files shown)


core/src/main/java/io/substrait/relation/ProtoRelConverter.java
```diff
@@ -620,8 +620,19 @@ protected Expand newExpand(ExpandRel rel) {
 
   protected Aggregate newAggregate(AggregateRel rel) {
     Rel input = from(rel.getInput());
+    Type.Struct inputSchema;
+    if (input instanceof Project) {
+      List<Type> types =
+          ((Project) input)
+              .getExpressions().stream().map(Expression::getType).collect(Collectors.toList());
+      inputSchema = Type.Struct.builder().fields(types).nullable(false).build();
```
vbarua, line 628: A bit late to this, but why was this needed? This is the wrong output type for a Project, which by default should include all input columns and all expressions.

cc: @ZorinAnton

ZorinAnton: for the query above:

``` sql
select i_category from (
select i_category, count(*) from item group by
rollup(
i_category
)
)
```

the input relation looks as follows:
``` 
Project{
	input=Project{
		remap=Remap{indices=[22]}, input=NamedScan{initialSchema=NamedStruct{struct=Struct{nullable=false, fields=[I64{nullable=false}, FixedChar{nullable=false, length=16}, Date{nullable=false}, Date{nullable=false}, VarChar{nullable=false, length=200}, Decimal{nullable=false, scale=2, precision=7}, Decimal{nullable=false, scale=2, precision=7}, I32{nullable=false}, FixedChar{nullable=false, length=50}, I32{nullable=false}, FixedChar{nullable=false, length=50}, I32{nullable=false}, FixedChar{nullable=false, length=50}, I32{nullable=false}, FixedChar{nullable=false, length=50}, FixedChar{nullable=false, length=20}, FixedChar{nullable=false, length=20}, FixedChar{nullable=false, length=20}, FixedChar{nullable=false, length=10}, FixedChar{nullable=false, length=10}, I32{nullable=false}, FixedChar{nullable=false, length=50}]}, names=[I_ITEM_SK, I_ITEM_ID, I_REC_START_DATE, I_REC_END_DATE, I_ITEM_DESC, I_CURRENT_PRICE, I_WHOLESALE_COST, I_BRAND_ID, I_BRAND, I_CLASS_ID, I_CLASS, I_CATEGORY_ID, I_CATEGORY, I_MANUFACT_ID, I_MANUFACT, I_SIZE, I_FORMULATION, I_COLOR, I_UNITS, I_CONTAINER, I_MANAGER_ID, I_PRODUCT_NAME]}, names=[tpcds, ITEM]}, expressions=[FieldReference{segments=[StructField{offset=12}], type=FixedChar{nullable=false, length=50}}]
		}, 
	expressions=[Cast{type=FixedChar{nullable=true, length=50}, input=FieldReference{segments=[StructField{offset=0}], type=FixedChar{nullable=false, length=50}}, failureBehavior=RETURN_NULL}]
	}
```

`inputSchema = input.getRecordType()` gives the following schema as concatenation of input column types and types of the expressions:
`Struct{nullable=false, fields=[FixedChar{nullable=false, length=50}, FixedChar{nullable=true, length=50}]}`, which doesn't look correct, because the output of the projection has only one column. This causes a validation failure in calcite.

The code above (that collects only expressions types) produces the schema: `Struct{nullable=false, fields=[FixedChar{nullable=true, length=50}]}`, which is accepted by calcite.

Do you know why a projection's record type is derived as concatenation of types of input columns and expressions?

https://github.com/substrait-io/substrait-java/blob/516c0908e79cf55c1f238ed430be2584668befcc/core/src/main/java/io/substrait/relation/Project.java#L21

vbarua: > Do you know why a projection's record type is derived as concatenation of types of input columns and expressions?

Per the specification, a Substrait [Project](https://substrait.io/relations/logical_relations/#project-operation) outputs all of its inputs fields, followed by all of its expressions. A remap can be used to further control which fields are output. The change here contravenes the spec, and could potentially cause broader issues. We should undo this change.

Was
```
CHAR(50) NOT NULL
input:
CHAR(50)
```
the error your were seeing? That is likely due to differences null handling behaviour between Calcite and Substrait that we'll need to handle somewhere else.

ZorinAnton: >Per the specification, a Substrait [Project](https://substrait.io/relations/logical_relations/#project-operation) outputs all of its inputs fields, followed by all of its expressions.

Ok, thanks for pointing out, reverting for now.

> Was
CHAR(50) NOT NULL
input:
CHAR(50)
the error your were seeing?

Yes, projection returns the same column as non-nullable and nullable and calcite's validation fails on checking the former: the outer select returns the nullable column.

PR #508, opened 2025-09-11
fix(isthmus): improve derivation of AggregateRel return type

ZorinAnton:
Minimal reproducing query:
``` sql
select charcol from (
  select charcol, count(*) from (values('a')) as t(charcol) group by rollup(charcol)
)
```
If this query gets converted to `subtrait proto` and then to `substrait pojo`, the latter looks as follows:

>```Plan{version=Version{major=0, minor=74, patch=0, producer=isthmus}, roots=[Root{input=Project{remap=Remap{indices=[2]}, input=Aggregate{input=VirtualTableScan{initialSchema=NamedStruct{struct=Struct{nullable=false, fields=[FixedChar{nullable=false, length=1}]}, names=[CHARCOL]}, rows=[StructLiteral{nullable=false, fields=[FixedCharLiteral{nullable=false, value=a}]}]}, groupings=[Grouping{expressions=[FieldReference{segments=[StructField{offset=0}], type=FixedChar{nullable=false, length=1}}]}, Grouping{expressions=[]}], measures=[Measure{function=AggregateFunctionInvocation{declaration=count:, arguments=[], options=[], aggregationPhase=INITIAL_TO_RESULT, sort=[], outputType=I64{nullable=false}, invocation=ALL}}]}, expressions=[FieldReference{segments=[StructField{offset=0}], type=FixedChar{nullable=false, length=1}}]}, names=[CHARCOL]}], expectedTypeUrls=[]}```

The ROLLUP function makes grouping columns nullable. This was causing a Calcite validation error because the outer projection's CHARCOL column was incorrectly defined as non-nullable. This fix corrects the column type, allowing the conversion to succeed.

>```Plan{version=Version{major=0, minor=74, patch=0, producer=isthmus}, roots=[Root{input=Project{remap=Remap{indices=[2]}, input=Aggregate{input=VirtualTableScan{initialSchema=NamedStruct{struct=Struct{nullable=false, fields=[FixedChar{nullable=false, length=1}]}, names=[CHARCOL]}, rows=[StructLiteral{nullable=false, fields=[FixedCharLiteral{nullable=false, value=a}]}]}, groupings=[Grouping{expressions=[FieldReference{segments=[StructField{offset=0}], type=FixedChar{nullable=false, length=1}}]}, Grouping{expressions=[]}], measures=[Measure{function=AggregateFunctionInvocation{declaration=count:, arguments=[], options=[], aggregationPhase=INITIAL_TO_RESULT, sort=[], outputType=I64{nullable=false}, invocation=ALL}}]}, expressions=[FieldReference{segments=[StructField{offset=0}], type=FixedChar{nullable=true, length=1}}]}, names=[CHARCOL]}], expectedTypeUrls=[]}```


vbarua:
Thanks for the reproducer for this. I'm digging into this a bit because there's a general pattern of bugs we've also noticed due to disagreement between Calcite and Substrait as to nullabilities.

ZorinAnton:
I tried adding an extra integer, while it works with minimal reproducer, it somehow breaks the calcite validation when `partition by` is used together with `rollup`. I would suggest to address this issue in a separate PR.

core/src/main/java/io/substrait/relation/Aggregate.java
```diff
@@ -21,17 +21,42 @@ public abstract class Aggregate extends SingleInputRel implements HasExtension {
...
+    }
+
+    final LinkedHashSet<Expression> uniqueExpressions =
+        getGroupings().stream()
+            .flatMap(g -> g.getExpressions().stream())
+            .collect(Collectors.toCollection(LinkedHashSet::new));
+
+    // For each unique expression, determine its final nullability based on the spec.
+    final Stream<Type> groupingTypes =
+        uniqueExpressions.stream()
```
nielspardon: minor: I'm wondering whether we want to use the term "grouping expression" to be super precise.

```suggestion
    final LinkedHashSet<Expression> uniqueGroupingExpressions =
        getGroupings().stream()
            .flatMap(g -> g.getExpressions().stream())
            .collect(Collectors.toCollection(LinkedHashSet::new));

    // For each unique grouping expression, determine its final nullability based on the spec.
    final Stream<Type> groupingTypes =
        uniqueGroupingExpressions.stream()
```

core/src/main/java/io/substrait/relation/Aggregate.java
```diff
@@ -21,17 +21,42 @@ public abstract class Aggregate extends SingleInputRel implements HasExtension {
...
+              .flatMap(g -> g.getExpressions().stream())
+              .map(Expression::getType);
+      final Stream<Type> measureTypes = getMeasures().stream().map(t -> t.getFunction().getType());
+      return TypeCreator.REQUIRED.struct(Stream.concat(groupingTypes, measureTypes));
+    }
+
+    final LinkedHashSet<Expression> uniqueGroupingExpressions =
+        getGroupings().stream()
+            .flatMap(g -> g.getExpressions().stream())
+            .collect(Collectors.toCollection(LinkedHashSet::new));
```
vbarua, line 37: I'm not entirely sure this will work as expected because I'm not sure if expression equality  / hash coding is well defined. That being said, we're already working on switching to the new grouping field mechanism that will make this obsolete in https://github.com/substrait-io/substrait-java/pull/521

vbarua (approved):
Apologies for the delay. I agree that this is reasonable to merge in as its strictly improving the existing deficiency, even if it's not fully solving it. The followup issue tracks the remaining work https://github.com/substrait-io/substrait-java/issues/527

I'll merge this myself later today, I want to make include a succint description in the body for the release notes. Strictly speaking, while this fixes rollup queries, that's more a side-effect of correcting the type-derivation.

vbarua:
Ah, a test appears to have broken following the merge of https://github.com/substrait-io/substrait-java/pull/518 😢

PR #513, opened 2025-09-18
chore(isthmus): alternate forms for tpc-ds queries

bestbeforetoday:
Alternate forms for TPC-DS queries 27, 36, 70 and 86. These forms rewrite queries that use the GROUPING aggregate function, which does not have a direct Substrait equivalent. These test cases now pass.


nielspardon (approved):
LGTM, I think it is good that we can cover these TPC-DS queries already in the alternate forms specified as part of the TPC-DS benchmark while we are still figuring out how to support the original format with the `GROUPING` / `GROUPING_ID` function syntax.

PR #515, opened 2025-09-19
build(isthmus-cli): use GraalVM 25

bestbeforetoday:
Benefits from Java 25 performance optimizations. Also resolves an issue where SLF4J log statements were output by the CLI command since the JDK 1.4 logging implementation was specified as a runtime dependency.


nielspardon:
> Also resolves an issue where SLF4J log statements were output by the CLI command since the JDK 1.4 logging implementation was specified as a runtime dependency.

This makes me wonder what the behavior should be. Should the CLI have a `--debug` or `--verbose` flag to control whether and which log messages should be printed instead of dropping them?

nielspardon (approved):
LGTM

PR #516, opened 2025-09-19
chore(isthmus): debug log missing Calcite mapping

bestbeforetoday:
There is an info level log in FunctionConverter for every every Calcite operator for which there is no direct Substrait mapping. This message makes the logs extremely busy and provides no value to end users. This change reclassifies the message as debug log.


nielspardon (approved):
LGTM

PR #518, opened 2025-09-23
feat(spark): support Hive DDL / Insert operations

andrew-coleman:
Add support for DdlRel and WriteRel for Hive in Spark

### DIFF
--- a/spark/src/main/scala/io/substrait/spark/logical/ToLogicalPlan.scala
+++ b/spark/src/main/scala/io/substrait/spark/logical/ToLogicalPlan.scala
@@ -29,12 +29,13 @@ import org.apache.spark.sql.catalyst.plans.{FullOuter, Inner, LeftAnti, LeftOute
 import org.apache.spark.sql.catalyst.plans.logical._
 import org.apache.spark.sql.catalyst.util.toPrettySQL
 import org.apache.spark.sql.execution.QueryExecution
-import org.apache.spark.sql.execution.command.{CreateDataSourceTableAsSelectCommand, DataWritingCommand, LeafRunnableCommand}
+import org.apache.spark.sql.execution.command.{CreateDataSourceTableAsSelectCommand, CreateTableCommand, DataWritingCommand, DropTableCommand, LeafRunnableCommand}
 import org.apache.spark.sql.execution.datasources.{FileFormat => SparkFileFormat, HadoopFsRelation, InMemoryFileIndex, InsertIntoHadoopFsRelationCommand, LogicalRelation, V1Writes}
 import org.apache.spark.sql.execution.datasources.csv.CSVFileFormat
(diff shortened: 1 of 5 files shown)


spark/src/main/scala/io/substrait/spark/logical/ToLogicalPlan.scala
```diff
@@ -439,35 +443,44 @@ class ToLogicalPlan(spark: SparkSession = SparkSession.builder().getOrCreate())
...
+          } else {
+            CreateDataSourceTableAsSelectCommand(
+              table,
+              saveMode(write.getCreateMode),
+              child,
+              write.getTableSchema.names().asScala
+            )
+          }
+        }
+      case WriteOp.INSERT =>
```
bestbeforetoday: I notice no `isHive` check here. Might it be possible to arrive here when `isHive` is not true?

spark/src/test/scala/io/substrait/spark/HiveTableSuite.scala
```diff
@@ -0,0 +1,113 @@
...
+    val substraitPlan = toSubstrait.convert(plan)
+
+    // Serialize to proto buffer
+    val bytes = new PlanProtoConverter()
+      .toProto(substraitPlan)
+      .toByteArray
+
+    // Read it back
+    val protoPlan = io.substrait.proto.Plan
+      .parseFrom(bytes)
```
bestbeforetoday, line 37: It's valid but I'm wondering if there is any benefit to doing the work of de/serializing to bytes, rather than just using the protobuf object. Is it exercising anything other than protobuf's ability to de/serialize?

spark/src/test/scala/io/substrait/spark/HiveTableSuite.scala
```diff
@@ -0,0 +1,113 @@
...
+
+    // now execute the round-tripped plan and assert an extra row is appended
+    spark.sessionState.executePlan(plan).executedPlan.execute()
+    assertResult(4)(spark.sql("select * from test").count())
+    // and again...
+    spark.sessionState.executePlan(plan).executedPlan.execute()
+    assertResult(5)(spark.sql("select * from test").count())
+  }
+
+  test("CTAS") {
```
bestbeforetoday: For the less expert amongst us, it might be clearer as:

```suggestion
  test("Create table as select") {
```

spark/src/test/scala/io/substrait/spark/HiveTableSuite.scala
andrew-coleman: In general, I put all tests through the full roundtrip lifecycle, including down to the proto bytes to ensure that all the visitors within the proto/rel converters have been implemented.

bestbeforetoday (approved):

Issue #299, opened 2024-10-02
handle the new grouping behavior using references


kadinrabo:
We should update the behavior in substrait-java with the spec update [706](https://github.com/substrait-io/substrait/pull/706) changing the _definition of grouping sets in `AggregateRel` to consist of references into a list of grouping expressions instead of consisting of expressions directly._

This change keeps consumers from having to deduplicate the expressions in the grouping sets in order to execute the query or even derive the output schema. Explained in further detail in [#700](https://github.com/substrait-io/substrait/issues/700).

PR #521, opened 2025-09-24
fix(isthmus): adding support for new proto and deprecated form in AggregateRel

gord02:
Substrait is either produced using using the deprecated form where Groupings is set with a grouping_expressions list, or the new proto form where the grouping_expressions is defined on the AggregateRel itself with the Groupings having a list of references into the grouping_expressions list. 

This PR would add support for both of these forms resolving bugs that expected the deprecated form of AggregateRel.

Key Changes:
- added check for deprecated and new form of proto in ProtoRelConverter
- Added support for both proto forms to be outputted in RelProtoConverter
- added tests for checking both forms of proto and some lines for checking the output of RelProtoConverter


core/src/main/java/io/substrait/relation/ProtoRelConverter.java
```diff
@@ -627,15 +627,37 @@ protected Aggregate newAggregate(AggregateRel rel) {
...
-                      .map(protoExprConverter::from)
-                      .collect(java.util.stream.Collectors.toList()))
-              .build());
+
+    //      new proto form is used
+    if (!rel.getGroupingExpressionsList().isEmpty()) {
+
+      List<io.substrait.proto.Expression> allGroupingKeys = rel.getGroupingExpressionsList();
+
+      for (int i = 0; i < rel.getGroupingsList().size(); i++) {
```
vbarua: I don't think your looping here is quite right. For one, you declaring `i` and updating it, but you never use it in the body.  I would suggest simplifying this code. Effectively, the way the new grouping mechanism works is that we have:
1. A list of shared grouping expressions.
2. 1 or more Groupings, each of which contains a list of expression references which index into 1.

Something like the following might be a bit more straightforward:
```java
// Convert all the grouping expressions
List<Expression> groupingExpressions = rel.getGroupingExpressionsList()...
...
// Convert all the groupings
List<Aggregate.Grouping> groupings = new ArrayList<>();
for (AggregateRel.Grouping grouping : rel.getGroupingsList()) {    
  // Each value in the getExpressionReferencesList maps to an index in groupingExpressions.
  // Get all the expressions and build a new Aggregate.Grouping
  grouping.getExpressionReferencesList()...
}

```

core/src/main/java/io/substrait/relation/ProtoRelConverter.java
```diff
@@ -627,15 +627,37 @@ protected Aggregate newAggregate(AggregateRel rel) {
...
+
+    } else {
+      //        using the deprecated form of Grouping and Aggregate
+      for (AggregateRel.Grouping grouping : rel.getGroupingsList()) {
+        groupings.add(
+            Aggregate.Grouping.builder()
+                .expressions(
+                    grouping.getGroupingExpressionsList().stream()
+                        .map(protoExprConverter::from)
+                        .collect(Collectors.toList()))
```
vbarua:
```suggestion
                        .collect(java.util.stream.Collectors.toList()))
```
To maintain consistency with the rest of the project.

core/src/main/java/io/substrait/relation/RelProtoConverter.java
```diff
@@ -117,12 +120,45 @@ private io.substrait.proto.Expression.FieldReference toProto(FieldReference fiel
 
   @Override
   public Rel visit(Aggregate aggregate, EmptyVisitationContext context) throws RuntimeException {
+
+    List<io.substrait.proto.Expression> groupingExpressions = new ArrayList<>();
```
vbarua: Having thought on this some more, let's not handle the Rel to Proto side of this in this PR. There are some deeper things we need to check in the library when we do this, specifically when integrating with Calcite. I've added some comments to the tracking issue for this work https://github.com/substrait-io/substrait-java/issues/299#issuecomment-3358935137

With what you're doing now, substrait-java will be able to accept the new form, which will still be an improvement.

core/src/test/java/io/substrait/relation/AggregateRelTest.java
```diff
@@ -0,0 +1,140 @@
...
+        io.substrait.proto.Type.Struct.newBuilder().addTypes(i32Type).addTypes(i32Type).build();
+
+    return io.substrait.proto.NamedStruct.newBuilder()
+        .setStruct(structType)
+        .addNames("col1")
+        .addNames("col2")
+        .build();
+  }
+
+  public static io.substrait.proto.Expression createExpression(int col) {
```
vbarua: It makes sense for this to return an expression, but if we name it `createFieldReference` it will more precisely identify the kind of expression it returns.

Also good job constructing this with the protos directly, it's definitely not as nice as working with the library builders.

core/src/test/java/io/substrait/relation/AggregateRelTest.java
```diff
@@ -0,0 +1,140 @@
...
+  @Test
+  public void testNewAggregateProtoForm() {
+    Expression col1Ref = createExpression(0);
+    Expression col2Ref = createExpression(1);
+
+    AggregateRel.Grouping grouping =
+        AggregateRel.Grouping.newBuilder()
+            .addExpressionReferences(0) // new proto form
+            .addExpressionReferences(1)
+            .build();
```
vbarua, line 100: I think we can focus on testing the new form in these tests, and not worry about the deprecated form here. There are plenty of tests for that via roundtrip testing.

What I would suggest is adding a case with multiple grouping sets. I think that would have caught the issue you had in your loop.

core/src/main/java/io/substrait/relation/ProtoRelConverter.java
```diff
@@ -627,15 +627,34 @@ protected Aggregate newAggregate(AggregateRel rel) {
...
-    for (AggregateRel.Grouping grouping : rel.getGroupingsList()) {
-      groupings.add(
-          Aggregate.Grouping.builder()
-              .expressions(
-                  grouping.getGroupingExpressionsList().stream()
-                      .map(protoExprConverter::from)
-                      .collect(java.util.stream.Collectors.toList()))
-              .build());
+
+    //      new proto form is used
```
vbarua:
```suggestion
    // Groupings are set using the AggregateRel grouping_expression mechanism
```

core/src/main/java/io/substrait/relation/ProtoRelConverter.java
```diff
@@ -627,15 +627,34 @@ protected Aggregate newAggregate(AggregateRel rel) {
...
+        List<Integer> references = grouping.getExpressionReferencesList();
+        List<Expression> groupExpressions = new ArrayList<>();
+        for (int ref : references) {
+          groupExpressions.add(protoExprConverter.from(allGroupingKeys.get(ref)));
+        }
+        groupings.add(Aggregate.Grouping.builder().addAllExpressions(groupExpressions).build());
+      }
+
+    } else {
+      //        using the deprecated form of Grouping and Aggregate
```
vbarua:
```suggestion
      // Groupings are set using the deprecated Grouping grouping_expressions mechanism 
```

core/src/main/java/io/substrait/relation/ProtoRelConverter.java
```diff
@@ -627,15 +627,34 @@ protected Aggregate newAggregate(AggregateRel rel) {
...
-              .expressions(
-                  grouping.getGroupingExpressionsList().stream()
-                      .map(protoExprConverter::from)
-                      .collect(java.util.stream.Collectors.toList()))
-              .build());
+
+    //      new proto form is used
+    if (!rel.getGroupingExpressionsList().isEmpty()) {
+
+      List<io.substrait.proto.Expression> allGroupingKeys = rel.getGroupingExpressionsList();
```
vbarua: I would suggest pre-converting all the grouping expressions here. It's minor, but if the same expression is referenced in multiple groupings, we would convert it multiple times below. If we pre-convert them all here, we can guarantee that we only do it once.

core/src/test/java/io/substrait/relation/AggregateRelTest.java
```diff
@@ -0,0 +1,162 @@
...
+    io.substrait.relation.Rel resultRel = converter.from(relProto);
+
+    assertTrue(resultRel instanceof Aggregate);
+    Aggregate agg = (Aggregate) resultRel;
+    assertEquals(1, agg.getGroupings().size());
+    assertEquals(2, agg.getGroupings().get(0).getExpressions().size());
+  }
+
+  @Test
+  public void testNewAggregateProtoForm() {
```
vbarua:
```suggestion
  public void testAggregateWithSingleGrouping() {
```

core/src/test/java/io/substrait/relation/AggregateRelTest.java
```diff
@@ -0,0 +1,162 @@
...
+    Expression col1Ref = createFieldReference(0);
+    Expression col2Ref = createFieldReference(1);
+
+    AggregateRel.Grouping grouping =
+        AggregateRel.Grouping.newBuilder()
+            .addExpressionReferences(0) // new proto form
```
vbarua:
```suggestion
            .addExpressionReferences(0)
```
I don't think there is much value in calling this out here, and will create result in confusion when we eventually delete the old mechanism fully.

core/src/test/java/io/substrait/relation/AggregateRelTest.java
```diff
@@ -0,0 +1,162 @@
...
+            .setDirectReference(seg1)
+            .setRootReference(Expression.FieldReference.RootReference.getDefaultInstance())
+            .build();
+
+    // Wrap the FieldReference in an Expression.selection
+    return Expression.newBuilder().setSelection(fieldRef1).build();
+  }
+
+  @Test
+  public void testDeprecatedGroupingExpressionsAreMapped() {
```
vbarua:
```suggestion
  public void testDeprecatedGroupingExpressionConversion() {
```

core/src/test/java/io/substrait/relation/AggregateRelTest.java
```diff
@@ -0,0 +1,162 @@
...
+    io.substrait.relation.Rel resultRel = converter.from(relProto);
+
+    assertTrue(resultRel instanceof Aggregate);
+    Aggregate agg = (Aggregate) resultRel;
+    assertEquals(1, agg.getGroupings().size());
+    assertEquals(2, agg.getGroupings().get(0).getExpressions().size());
+  }
+
+  @Test
+  public void testNewAggregateProtoFormMultipleGroupings() {
```
vbarua:
```suggestion
  public void testAggregateWithMultipleGroupings() {
```

vbarua (commented):
Overall looks good. Left some final suggestions, along with one improvement.

vbarua (approved):
Looks to me. Thanks for working on this!

PR #522, opened 2025-09-26
feat: enable handling of URNs alongside URIs

### LINKED ISSUE #509: Migrate extensions from URIs to URNs
See https://github.com/substrait-io/substrait/issues/856 for context.

This will require bumping the version once a new release occurs.

benbellick:
Closes #509

# Summary
This PR introduces a graceful URI -> URN migration. [See this issue in the upstream substrait repo for details.](https://github.com/substrait-io/substrait/issues/856)

The gist is that the library will be able to accept both uris and urns in input plans, but will always output both when producing plans. This is facilitated by keeping a map internally which tracks a bijection between uris and urns. A later PR will drop support for uris. 

This is a breaking change. 

# Recommended PR Review Strategy
To facilitate easier review of this PR, I suggest breaking the commits into three chunks.
1. [The first chunk](https://github.com/substrait-io/substrait-java/pull/522/files/3ee953c67c92e8e648109aba9c04e2f5de36e9ea) fully migrates uri to urn without any graceful handling. You can think about this as the target state of the repository once the migration is fully complete. The only other feature added here is to add some light urn validation (to ensure the format is `extension:<organization>:<name>`).
2. [The second chunk](https://github.com/substrait-io/substrait-java/pull/522/commits/c7639b3743e1e42d10239fbeaf17dbfc469576f0) just moves the internally used `BidiMap` to another file so that it can later be used to internally track the uri<->urn mapping. 
3. [The third chunk](https://github.com/substrait-io/substrait-java/pull/522/commits/7fd825a1156ad01cd57b4f3b863a9f68199159fe) is the real meat of the PR. This is where the uri<->urn mapping is introduced. This mapping is maintained internally inside of an `ExtensionCollection` which is now passed around as an argument to many of the classes.

# Testing
To robustly test this implementation, a new (light) test framework is introduced, where input plans are presented in JSON format. They then undergo proto -> POJO -> proto roundtrip conversion, and then compared with an expected JSON-format plan. The goal here is to ensure that regardless of the usage of uri or urn in input, both are present in the output. As an example, see [here](https://github.com/substrait-io/substrait-java/pull/522/files#diff-43e9e368b020f892836b0692ef4c9718986b247dcbf2d6d69876e8465008efd5) for an expected input plan in which only uri is used, and [here](https://github.com/substrait-io/substrait-java/pull/522/files#diff-e9db7e43d0728fd92da47f961b476991251ff88dca0b4bd1c88aa4ccc43ccd81) for the expected output plan, in which both uri and urn are present.

This test framework may be dropped once the migration is complete, though perhaps it will prove useful for other reasons.

### DIFF
--- a/core/build.gradle.kts
+++ b/core/build.gradle.kts
@@ -108,6 +108,8 @@ configurations[JavaPlugin.TEST_IMPLEMENTATION_CONFIGURATION_NAME].extendsFrom(sh
 
 dependencies {
   testImplementation(platform(libs.junit.bom))
+  testImplementation(libs.protobuf.java.util)
+
   testImplementation(libs.junit.jupiter)
   testRuntimeOnly(libs.junit.platform.launcher)
 
--- a/core/src/main/java/io/substrait/extension/AbstractExtensionLookup.java
+++ b/core/src/main/java/io/substrait/extension/AbstractExtensionLookup.java
@@ -13,26 +13,6 @@ public abstract class AbstractExtensionLookup implements ExtensionLookup {
     this.typeAnchorMap = typeAnchorMap;
   }
 
-  /**
-   * Gets the function anchor for a given reference (primarily for testing).
-   *
-   * @param reference The function reference
-   * @return The function anchor, or null if not found
-   */
-  public SimpleExtension.FunctionAnchor getFunctionAnchor(int reference) {
-    return functionAnchorMap.get(reference);
-  }
-
-  /**
-   * Gets the type anchor for a given reference (primarily for testing).
-   *
-   * @param reference The type reference
-   * @return The type anchor, or null if not found
-   */
-  public SimpleExtension.TypeAnchor getTypeAnchor(int reference) {
-    return typeAnchorMap.get(reference);
-  }
-
   @Override
   public SimpleExtension.ScalarFunctionVariant getScalarFunction(
       int reference, SimpleExtension.ExtensionCollection extensions) {
--- a/core/src/main/java/io/substrait/extension/AbstractExtensionLookup.java
+++ b/core/src/main/java/io/substrait/extension/AbstractExtensionLookup.java
@@ -13,6 +13,26 @@ public abstract class AbstractExtensionLookup implements ExtensionLookup {
     this.typeAnchorMap = typeAnchorMap;
   }
 
+  /**
+   * Gets the function anchor for a given reference (primarily for testing).
+   *
+   * @param reference The function reference
+   * @return The function anchor, or null if not found
+   */
+  public SimpleExtension.FunctionAnchor getFunctionAnchor(int reference) {
+    return functionAnchorMap.get(reference);
+  }
+
+  /**
+   * Gets the type anchor for a given reference (primarily for testing).
+   *
+   * @param reference The type reference
+   * @return The type anchor, or null if not found
+   */
+  public SimpleExtension.TypeAnchor getTypeAnchor(int reference) {
+    return typeAnchorMap.get(reference);
+  }
+
   @Override
   public SimpleExtension.ScalarFunctionVariant getScalarFunction(
       int reference, SimpleExtension.ExtensionCollection extensions) {
--- a/core/src/main/java/io/substrait/extension/BidiMap.java
+++ b/core/src/main/java/io/substrait/extension/BidiMap.java
@@ -1,0 +1,65 @@
+package io.substrait.extension;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Set;
+
+/** We don't depend on guava... */
+class BidiMap<T1, T2> {
+  private final Map<T1, T2> forwardMap;
+  private final Map<T2, T1> reverseMap;
+
+  BidiMap(Map<T1, T2> forwardMap) {
+    this.forwardMap = forwardMap;
+    this.reverseMap = new HashMap<>();
+    for (Map.Entry<T1, T2> entry : forwardMap.entrySet()) {
+      reverseMap.put(entry.getValue(), entry.getKey());
+    }
+  }
+
+  BidiMap() {
+    this.forwardMap = new HashMap<>();
+    this.reverseMap = new HashMap<>();
+  }
+
+  T2 get(T1 t1) {
+    return forwardMap.get(t1);
+  }
+
+  T1 reverseGet(T2 t2) {
+    return reverseMap.get(t2);
+  }
+
+  /**
+   * Associates the specified values in both directions. Throws if either value is already mapped to
+   * a different value.
+   */
+  void put(T1 t1, T2 t2) {
+    T2 existingForward = forwardMap.get(t1);
+    T1 existingReverse = reverseMap.get(t2);
+
+    if (existingForward != null && !existingForward.equals(t2)) {
+      throw new IllegalArgumentException("Key already exists in map with different value");
+    }
+    if (existingReverse != null && !existingReverse.equals(t1)) {
+      throw new IllegalArgumentException("Key already exists in map with different value");
+    }
+
+    forwardMap.put(t1, t2);
+    reverseMap.put(t2, t1);
+  }
+
+  void merge(BidiMap<T1, T2> other) {
+    for (Map.Entry<T1, T2> entry : other.forwardEntrySet()) {
+      put(entry.getKey(), entry.getValue());
+    }
+  }
+
+  Set<Map.Entry<T1, T2>> forwardEntrySet() {
+    return forwardMap.entrySet();
+  }
+
+  Set<Map.Entry<T2, T1>> reverseEntrySet() {
+    return reverseMap.entrySet();
+  }
+}
--- a/core/src/main/java/io/substrait/extension/BidiMap.java
+++ b/core/src/main/java/io/substrait/extension/BidiMap.java
@@ -5,7 +5,7 @@ import java.util.Map;
 import java.util.Set;
 
 /** We don't depend on guava... */
-class BidiMap<T1, T2> {
+public class BidiMap<T1, T2> {
   private final Map<T1, T2> forwardMap;
   private final Map<T2, T1> reverseMap;
 
--- a/core/src/main/java/io/substrait/extension/ExtensionCollector.java
+++ b/core/src/main/java/io/substrait/extension/ExtensionCollector.java
@@ -3,6 +3,7 @@ package io.substrait.extension;
 import io.substrait.proto.ExtendedExpression;
 import io.substrait.proto.Plan;
 import io.substrait.proto.SimpleExtensionDeclaration;
+import io.substrait.proto.SimpleExtensionURI;
 import io.substrait.proto.SimpleExtensionURN;
 import java.util.ArrayList;
 import java.util.HashMap;
@@ -19,14 +20,27 @@ import java.util.concurrent.atomic.AtomicInteger;
 public class ExtensionCollector extends AbstractExtensionLookup {
   private final BidiMap<Integer, SimpleExtension.FunctionAnchor> funcMap;
   private final BidiMap<Integer, SimpleExtension.TypeAnchor> typeMap;
+  private final SimpleExtension.ExtensionCollection extensionCollection;
 
   // start at 0 to make sure functionAnchors start with 1 according to spec
   private int counter = 0;
 
+  private String getUriFromUrn(String urn) {
+    return extensionCollection.getUriFromUrn(urn);
+  }
+
   public ExtensionCollector() {
+    this(SimpleExtension.loadDefaults());
+  }
+
+  public ExtensionCollector(SimpleExtension.ExtensionCollection extensionCollection) {
     super(new HashMap<>(), new HashMap<>());
+    if (extensionCollection == null) {
+      throw new IllegalArgumentException("ExtensionCollection is required");
+    }
     funcMap = new BidiMap<>(functionAnchorMap);
     typeMap = new BidiMap<>(typeAnchorMap);
+    this.extensionCollection = extensionCollection;
   }
 
   public int getFunctionReference(SimpleExtension.Function declaration) {
@@ -53,6 +67,7 @@ public class ExtensionCollector extends AbstractExtensionLookup {
     SimpleExtensions simpleExtensions = getExtensions();
 
     builder.addAllExtensionUrns(simpleExtensions.urns.values());
+    builder.addAllExtensionUris(simpleExtensions.uris.values());
     builder.addAllExtensions(simpleExtensions.extensionList);
   }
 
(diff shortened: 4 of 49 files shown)


core/build.gradle.kts
```diff
@@ -116,6 +116,7 @@ configurations[JavaPlugin.TEST_IMPLEMENTATION_CONFIGURATION_NAME].extendsFrom(sh
 dependencies {
   testImplementation(platform("org.junit:junit-bom:${JUNIT_VERSION}"))
   testImplementation("org.junit.jupiter:junit-jupiter")
+  testImplementation("com.google.protobuf:protobuf-java-util:${PROTOBUF_VERSION}")
```
benbellick: This is necessary to allow us to read in JSON-based plans for testing purposes.

core/src/main/java/io/substrait/extension/BidiMap.java
```diff
@@ -0,0 +1,62 @@
+package io.substrait.extension;
```
benbellick, line 1: This already existed in ExtensionCollector.java as a private class. Instead, I have moved it into a separate file so that we can also leverage it for a bijection between uris and urns for migration purposes.

core/src/main/java/io/substrait/extension/ImmutableExtensionLookup.java
```diff
@@ -25,43 +25,189 @@ public static Builder builder() {
...
+     * Resolves URN from URI using the URI/URN mapping.
+     *
+     * @param uri The URI to resolve
+     * @return The corresponding URN, or null if no mapping exists
+     */
+    private String resolveUrnFromUri(String uri) {
+      return extensionCollection.getUrnFromUri(uri);
+    }
+
+    private SimpleExtension.FunctionAnchor resolveFunctionAnchor(
```
benbellick, line 54: [This is meant to model the migration behavior of substrait-go.](https://github.com/substrait-io/substrait-go/pull/166#discussion_r2370372451)

core/src/test/java/io/substrait/extension/UriUrnMigrationEndToEndTest.java
```diff
@@ -0,0 +1,110 @@
...
+      String jsonContent =
+          new BufferedReader(new InputStreamReader(inputStream, StandardCharsets.UTF_8))
+              .lines()
+              .collect(Collectors.joining("\n"));
+
+      Plan.Builder planBuilder = Plan.newBuilder();
+      JsonFormat.parser().merge(jsonContent, planBuilder);
+      return planBuilder.build();
+    }
+  }
```
nielspardon, line 45: @vbarua is the protobuf JSON format now an official format of Substrait? If not, isn't relying on the protobuf JSON format in the core unit tests making it more official than it used to be? What are your thoughts?

benbellick: In a prior iteration of this PR, I did construct the plans by hand via the proto builders, rather than using JSON files. However, they were getting quite unreadable, and so I opted for switching to the JSON approach. 

Happy to consider other approaches, but even if substrait doesn't have official JSON support, I don't see harm in using it _internally_ in testing this library. No library APIs are being exposed specifically for JSON.

bestbeforetoday: Protobuf specifies a [canonical JSON representation](https://protobuf.dev/programming-guides/json/) for proto3 messages. I guess that, even if there is no official JSON format specifically for Substrait, JSON is an official format for the protobuf objects used to represent a Substrait plan.

nielspardon: I was asking because a while back I proposed to add some convenience methods for JSON serialization in this PR https://github.com/substrait-io/substrait-java/pull/381 since I had assumed the protobuf JSON format was the official Substrait JSON format which back then turned out to not be the case. I do understand that most people resort to looking at the protobuf JSON format for readability and debugging.

core/src/main/java/io/substrait/extension/BidiMap.java
```diff
@@ -0,0 +1,62 @@
...
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Set;
+
+/** We don't depend on guava... */
+public class BidiMap<T1, T2> {
```
vbarua, line 8: I think this still works if we make it package private, which avoid having to expose it externally.
```suggestion
class BidiMap<T1, T2> {
```

If this makes sense to you, we should also remove the `public` prefix from the constructors.

Reference: https://docs.oracle.com/javase/tutorial/java/javaOO/accesscontrol.html

core/src/main/java/io/substrait/extension/SimpleExtension.java
```diff
@@ -42,12 +43,22 @@
...
-  public static final String URI_LOCATOR_KEY = "uri";
+  // Key for looking up URN in InjectableValues
+  public static final String URN_LOCATOR_KEY = "urn";
 
-  private static ObjectMapper objectMapper(String namespace) {
+  private static void validateUrn(String urn) {
+    if (urn == null || urn.trim().isEmpty()) {
+      throw new IllegalArgumentException("URN cannot be null or empty");
+    }
+    if (!urn.matches("^extension:[^:]+:[^:]+$")) {
```
vbarua: It's probably worth compiling this regex with something like:
```java
  private static final Predicate<String> URN_CHECKER = Pattern.compile("^extension:[^:]+:[^:]+$").asPredicate();
```

core/src/main/java/io/substrait/extension/ExtensionCollector.java
```diff
@@ -19,14 +20,27 @@
...
 
   // start at 0 to make sure functionAnchors start with 1 according to spec
   private int counter = 0;
 
+  private String getUriFromUrn(String urn) {
+    return extensionCollection.getUriFromUrn(urn);
+  }
+
   public ExtensionCollector() {
+    this(SimpleExtension.loadDefaults());
```
vbarua: I noticed that we were calling `loadDefaults` in a bunch of places. The way it works now, it reprocesses the files every time, which seems wasteful. I've proposed some changes I've been meaning to make for a while in https://github.com/substrait-io/substrait-java/pull/529 to make this better.

core/src/test/java/io/substrait/extension/UriUrnMigrationEndToEndTest.java
vbarua: Ah, my objection in that PR specifically was around exposing any kind of _public_ JSON producing method that outputs the protobuf JSON format (protojson). If/when we do decide on a first-class Substrait JSON format, I want to avoid the issue of having an existing API that produces protojson because it won't necessarily match the format that we decide on, and would result in another awkward migration.

That is a separate concern from working with / using protojson plans. I don't see any issues with working with protojson directly, especially if it's not exposed to end-users as part of our public API.

vbarua (commented):
Found time for a first pass. Will do another one tomorrow.

Also, I had a suggestion that I turned into a separate PR https://github.com/substrait-io/substrait-java/pull/529

core/src/main/java/io/substrait/extension/BidiMap.java
benbellick: Done! I also remove `public` from all of the methods just to be consistent (of course those couldn't be accessed without a reference to the object, but still probably better that way).

core/src/main/java/io/substrait/extension/ImmutableExtensionLookup.java
```diff
@@ -25,25 +25,181 @@ public static Builder builder() {
     return new Builder();
```
vbarua: What do you think about this calling `builder(DefaultExtensionCollector.DEFAULT_COLLECTION)` directly? See below comment as well.

core/src/main/java/io/substrait/extension/ImmutableExtensionLookup.java
```diff
@@ -25,25 +25,181 @@ public static Builder builder() {
...
+  }
+
   public static class Builder {
     private final Map<Integer, SimpleExtension.FunctionAnchor> functionMap = new HashMap<>();
     private final Map<Integer, SimpleExtension.TypeAnchor> typeMap = new HashMap<>();
+    private final SimpleExtension.ExtensionCollection extensionCollection;
+
+    public Builder() {
+      this.extensionCollection = SimpleExtension.loadDefaults();
+    }
```
vbarua: I don't think we need this constructor, especially if we provide a static one above that does the right thing. If you think it's worth keeping, what I would recommend is having this be
```java
this(DefaultExtensionCollector.DEFAULT_COLLECTION)
```
to better re-use the initialization logic.

core/src/test/java/io/substrait/extension/UriUrnMigrationEndToEndTest.java
vbarua: @nielspardon does this seem reasonable to you / make sense?

core/src/main/java/io/substrait/extension/AbstractExtensionLookup.java
```diff
@@ -13,6 +13,26 @@ public AbstractExtensionLookup(
...
+
+  /**
+   * Gets the type anchor for a given reference (primarily for testing).
+   *
+   * @param reference The type reference
+   * @return The type anchor, or null if not found
+   */
+  public SimpleExtension.TypeAnchor getTypeAnchor(int reference) {
+    return typeAnchorMap.get(reference);
+  }
```
vbarua: If these are intended for testing I would prefer to hide them entirely. From what I see, they're only used in `ImmutableExtensionLookupUriUrnTest`, and we can access the `functionAnchorMap` and `typeAnchorMap` directly there as well.

If you wanted to add a bit more semantic meaning in the test, we could add a method like
```java
    String getFunctionUrn(ImmutableExtensionLookup lookup, int anchor) {
        return lookup.functionAnchorMap.get(anchor).urn();
    }
```
in `ImmutableExtensionLookupUriUrnTest` and use that.

core/src/main/java/io/substrait/extension/SimpleExtension.java
```diff
@@ -532,11 +545,22 @@ public TypeAnchor getAnchor() {
...
 
+    @JsonProperty("urn")
+    public abstract String urn();
+
+    // URI is not from YAML, but from the loading context
+    // this only needs to be present temporarily to handle the URI -> URN migration
+    @Value.Default
+    public String uri() {
+      return "";
+    }
```
vbarua: Is this actually needed? I commented this out locally, as well as the one place this gets used when building an `ImmutableSimpleExtension.ExtensionSignature` and none of the tests failed.

core/src/main/java/io/substrait/extension/SimpleExtension.java
```diff
@@ -665,28 +694,52 @@ public AggregateFunctionVariant getAggregateFunction(FunctionAnchor anchor) {
...
+              anchor.key(), anchor.urn()));
+    }
+
+    /**
+     * Gets the URI for a given URN. This is for internal framework use during URI/URN migration.
+     *
+     * @param urn The URN to look up
+     * @return The corresponding URI, or null if not found
+     */
+    public String getUriFromUrn(String urn) {
```
vbarua: Does this need to be public?

core/src/main/java/io/substrait/extension/SimpleExtension.java
```diff
@@ -665,28 +694,52 @@ public AggregateFunctionVariant getAggregateFunction(FunctionAnchor anchor) {
...
+      return uriUrnMap().reverseGet(urn);
+    }
+
+    /**
+     * Gets the URN for a given URI. This is for internal framework use during URI/URN migration.
+     *
+     * @param uri The URI to look up
+     * @return The corresponding URN, or null if not found
+     */
+    public String getUrnFromUri(String uri) {
```
vbarua: Does this need to be public?

core/src/main/java/io/substrait/extension/SimpleExtension.java
```diff
@@ -723,41 +777,64 @@ public static ExtensionCollection load(List<String> resourcePaths) {
...
-      ExtensionSignatures doc =
-          objectMapper(namespace).readValue(stream, ExtensionSignatures.class);
-      return buildExtensionCollection(namespace, doc);
-    } catch (RuntimeException ex) {
-      throw ex;
-    } catch (Exception ex) {
-      throw new IllegalStateException("Failure while parsing " + namespace, ex);
+  public static ExtensionCollection load(String uri, InputStream stream) {
+    try (Scanner scanner = new Scanner(stream)) {
+      scanner.useDelimiter("\\A");
```
vbarua: What's this magic delimiter `\\A`? Probably a good candidate for a static string with a name attached to it.

core/src/main/java/io/substrait/extension/ImmutableExtensionLookup.java
```diff
@@ -25,25 +25,181 @@ public static Builder builder() {
...
+        SimpleExtensionDeclaration.ExtensionFunction func,
+        Map<Integer, String> urnMap,
+        Map<Integer, String> uriMap) {
+
+      // 1. Try non-zero URN reference
+      if (func.getExtensionUrnReference() != 0) {
+        String urnFromUrnRef = urnMap.get(func.getExtensionUrnReference());
+        if (urnFromUrnRef != null) {
+          return SimpleExtension.FunctionAnchor.of(urnFromUrnRef, func.getName());
+        }
```
vbarua, line 67: If the URN is non-zero, and we can't resolve it, doesn't that mean the plan is malformed and we can blow up immediately?

core/src/main/java/io/substrait/extension/ImmutableExtensionLookup.java
```diff
@@ -25,25 +25,181 @@ public static Builder builder() {
...
+
+      // 2. Try non-zero URI reference
+      if (func.getExtensionUriReference() != 0) {
+        String uriFromUriRef = uriMap.get(func.getExtensionUriReference());
+        if (uriFromUriRef != null) {
+          String urnFromUriRef = resolveUrnFromUri(uriFromUriRef);
+          if (urnFromUriRef != null) {
+            return SimpleExtension.FunctionAnchor.of(urnFromUriRef, func.getName());
+          }
+          // URI found but could not be resolved to URN
```
vbarua: Isn't this also a violation that we can blow up for as well?

core/src/main/java/io/substrait/extension/ImmutableExtensionLookup.java
```diff
@@ -25,25 +25,181 @@ public static Builder builder() {
...
+        }
+      }
+
+      /* At this point both URI and URN are 0, so we need to
+        first see if they both resolve.
+      */
```
vbarua: I would clarify this a little bit with something like:
> At this point, both URI and URN are known be 0. With protobufs, we cannot distinguish between 0 as a intentional value vs 0 as a default value. We perform some additional checks to below to handle this.

Specifally, we should call out why 0 is special.

core/src/main/java/io/substrait/extension/ImmutableExtensionLookup.java
```diff
@@ -25,25 +25,181 @@ public static Builder builder() {
...
+        SimpleExtensionDeclaration.ExtensionType type,
+        Map<Integer, String> urnMap,
+        Map<Integer, String> uriMap) {
+
+      // 1. Try non-zero URN reference
+      if (type.getExtensionUrnReference() != 0) {
+        String urnFromUrnRef = urnMap.get(type.getExtensionUrnReference());
+        if (urnFromUrnRef != null) {
+          return SimpleExtension.TypeAnchor.of(urnFromUrnRef, type.getName());
+        }
```
vbarua, line 140: Same as above, we can explode here I think.

core/src/main/java/io/substrait/extension/ImmutableExtensionLookup.java
```diff
@@ -25,25 +25,181 @@ public static Builder builder() {
...
+
+      // 2. Try non-zero URI reference
+      if (type.getExtensionUriReference() != 0) {
+        String uriFromUriRef = uriMap.get(type.getExtensionUriReference());
+        if (uriFromUriRef != null) {
+          String urnFromUriRef = resolveUrnFromUri(uriFromUriRef);
+          if (urnFromUriRef != null) {
+            return SimpleExtension.TypeAnchor.of(urnFromUriRef, type.getName());
+          }
+          // URI found but could not be resolved to URN
```
vbarua: Also boom.

vbarua (commented):
Left some comments, but changes look good overall.

core/src/test/java/io/substrait/extension/UriUrnMigrationEndToEndTest.java
nielspardon: From a general direction it stills feels a bit inconsistent to me. I would either endorse protobuf JSON as an official format or not make use of it.

Independent of that I'm wondering how the maintenance of these static files would work and how the intended maintenance process would look like. Would we ever need to modify/regenerate them and how?

benbellick: If we don't make use of JSON here, do you have a suggestions for an alternative testing strategy? I also would say that these JSON can be fully removed once the migration is complete. They are only there to test this specific uri / urn dual functionality.

nielspardon: I'm not saying I'm against the JSON. I just find the messaging around the role of the protobuf JSON in the Substrait project a bit inconsistent.

I think you said yourself you were previously using the proto builders which would be an alternative.

> In a prior iteration of this PR, I did construct the plans by hand via the proto builders, rather than using JSON files. However, they were getting quite unreadable, and so I opted for switching to the JSON approach.

Using those might not be super readable, on the other hand they might have the benefit that the proto builders would break immediately when the deprecated functionality is removed and which might have the benefit that whoever removes the deprecated functionality does not forget to also delete the all the JSON files.

Again, I'm not against using the JSON files. I would just like to make sure we have a consistent (and ideally simple) messaging around the use of the JSON format.

benbellick: Direct construction via the builders is an alternative, but using the builders results in unreadable test cases. In my view, the tests should be optimized for readability because their goal is to communicate to the reader the correctness of this implementation.

These tests will also break once the next step of the migration takes place, necessitating their removal. (This is because once the migration is complete, generated plans will no longer emit uri information, which the expected plans do have.)

Also, I will personally shepherd the migration through here, so I will make sure that they are removed later. 

I don't have the full context on the history of the JSON substrait format, but my understanding now is that: 
a) there is no public facing substrait JSON API that is considered part of the spec, and
b) we can internally use JSON representations of plans as long as we don't publicly expose a JSON API in this library

Those two things are not in conflict. If it would help clarify the matter, I would be happy to add a comment to the code that makes it clear that this JSON is for internal use only.

core/src/main/java/io/substrait/extension/ImmutableExtensionLookup.java
benbellick: good catch

core/src/test/java/io/substrait/extension/UriUrnMigrationEndToEndTest.java
nielspardon: > I don't have the full context on the history of the JSON substrait format, but my understanding now is that:
> a) there is no public facing substrait JSON API that is considered part of the spec, and

This is my understanding as well. There is a placeholder in the documentation mentioning that there will be a JSON format but in reality there is no _officially approved_ JSON format for Substrait: https://substrait.io/serialization/text_serialization/

> b) we can internally use JSON representations of plans as long as we don't publicly expose a JSON API in this library

This is what I'm not sure about and that's why I asked. Since there is no _officially approved_ JSON format but a lot of people are using the protobuf JSON format to work with Substrait plans then de facto it is part of Substrait since people use it. I would prefer if the Substrait project would for example specify: We acknowledge that Substrait uses protobuf and protobuf exists in a binary and JSON serialization form. We do recommend to use the binary serialization format wherever possible and suggest to limit the use of JSON to debugging and testing.

I'm not objecting to your PR and have no issue if it gets merged. I just find the fact that we as the Substrait project claim there will be an official JSON format but it's not the protobuf JSON format while almost everyone is using the protobuf JSON format a bit awkwar. The most straightforward resolution and my preferred would be to just acknowledge that the protobuf JSON is an official format to work with Substrait.

PR #528, opened 2025-09-30
build: config option renamed in spotless 8 plugin

bestbeforetoday:
The `removeWildcardImports` configuration option was renamed to `forbidWildcardImports` in v8.0.0 of the Spotless Gradle plugin.

https://github.com/diffplug/spotless/releases/tag/gradle/8.0.0

This change uses the new named to avoid a build time warning.


nielspardon (approved):
LGTM

PR #529, opened 2025-10-01
feat: introduce DefaultExtensionCatalog.DEFAULT_COLLECTION

vbarua:
Provide convenient static instance of default ExtensionCollection

BREAKING CHANGE: removed SimpleExtension.loadDefaults

### DIFF
--- a/core/src/main/java/io/substrait/extension/DefaultExtensionCatalog.java
+++ b/core/src/main/java/io/substrait/extension/DefaultExtensionCatalog.java
@@ -1,5 +1,9 @@
 package io.substrait.extension;
 
+import java.util.Arrays;
+import java.util.List;
+import java.util.stream.Collectors;
+
 public class DefaultExtensionCatalog {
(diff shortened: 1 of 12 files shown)


core/src/main/java/io/substrait/extension/DefaultExtensionCatalog.java
```diff
@@ -14,4 +18,28 @@ public class DefaultExtensionCatalog {
   public static final String FUNCTIONS_ROUNDING_DECIMAL = "/functions_rounding_decimal.yaml";
   public static final String FUNCTIONS_SET = "/functions_set.yaml";
   public static final String FUNCTIONS_STRING = "/functions_string.yaml";
+
+  public static final SimpleExtension.ExtensionCollection DEFAULT_COLLECTION =
+      loadDefaultCollection();
```
vbarua, line 23: We can load the default ExtensionCollection once, and then share it everywhere.

benbellick (approved):
Nice 👊

PR #530, opened 2025-10-01
feat(isthmus): support subquery nested in project

### LINKED ISSUE #70: [ISTHMUS] [TPCDS] Unsupported subquery nested in Project relational operator
See tpcds `q9`

How to reproduce:
 Add a test to `TpcdsQueryNoValidation` like this:

```java
@ParameterizedTest
  @ValueSource(ints = {9})
  public void tpcdsQ9(int query) throws Exception {
    testQuery(query);
  }
```

On fixing, please move the fixed queries from the `tpcdsFailure` list to the `tpcdsSuccess` list

bestbeforetoday:
Previously a subquery nested within a project was explicitly disallowed when converting from Calcite to Substrait. This is because outer references were not correctly resolved within nested subqueries.

This change tracks outer references so they can be successfully resolved in subqueries nested within a project.

Closes #70


nielspardon (approved):
LGTM

PR #341, opened 2025-03-08
feat: allow for more ergonomic conversion of Extension relations

vbarua:
An emergent pattern with Extension relations is for users to include
Substrait messages within the detail field.

To convert these Substrait messages, users need to access to conversion
machinery within the code for converting details. This PR enables that
by modifying the ToProto interface.

BREAKING CHANGE: interface method ToProto#toProto now consumes a RelProtoConverter
BREAKING CHANGE: constructor for ExpressionProtoConverter now requires a RelProtoConverter specifically

PR #533, opened 2025-10-03
feat(core): use AdvancedExtension POJO in Plan POJO

### LINKED ISSUE #210: [Isthmus] Plan POJO uses AdvancedExtension proto
The [Plan POJO](https://github.com/substrait-io/substrait-java/blob/e73c6cac7a5b62d026ccd13eca97bfae02395aab/core/src/main/java/io/substrait/plan/Plan.java#L3C2-L3C45) uses the AdvancedExtension proto instead of the [AdvancedExtension POJO](https://github.com/substrait-io/substrait-java/blob/e73c6cac7a5b62d026ccd13eca97bfae02395aab/core/src/main/java/io/substrait/extension/AdvancedExtension.java).

Ideally, the POJOs should not contain protos.

nielspardon:
This PR:
- fixes #210
- changes the `io.substrait.plan.Plan` POJO to use the `io.substrait.extension.AdvancedExtension` POJO instead of the `io.substrait.proto.AdvancedExtension` proto class
- changes the conversion code for `AdvancedExtension` by introducing `AdvancedExtensionProtoConverter` and `ProtoAdvancedExtensionConverter` making it independent of `RelProtoConverter`
- changes `RelProtoConverter`, `ProtoRelConverter`, `PlanProtoConverter` and `ProtoPlanConverter` to use the new `AdvancedExtensionProtoConverter` and `ProtoAdvancedExtensionConverter` and allow to provide a custom implementation for those two classes
- adds a `PlanConverterTest` test class covering the changes for `Plan` and the corresponding converters
- updates `ExtensionRoundtripTest` containing the `RelProtoConverter`/`ProtoRelConverter` extension roundtrip tests
- moves `Extension.Optimization` and `Extension.Enhancement` into `AdvancedExtension`
- renames `io.substrait.relation.utils` package in test to `io.substrait.utils` to allow reuse for any test utils


core/src/main/java/io/substrait/plan/Plan.java
```diff
@@ -1,7 +1,7 @@
 package io.substrait.plan;
 
 import io.substrait.SubstraitVersion;
-import io.substrait.proto.AdvancedExtension;
+import io.substrait.extension.AdvancedExtension;
```
vbarua, line 4: ✨

core/src/main/java/io/substrait/relation/Extension.java
```diff
@@ -7,21 +7,6 @@
...
-   */
-  public interface Optimization extends ToProto<com.google.protobuf.Any> {}
-
-  /**
-   * Enhancement associated with an {@link io.substrait.proto.AdvancedExtension}
-   *
-   * <p>An enhancement alter semantics. Cannot be ignored by a consumer.
-   */
-  public interface Enhancement extends ToProto<com.google.protobuf.Any> {}
-
```
vbarua, line 24: Good idea moving these to be alongside `AdvancedExtension`. I think thats much better in terms of organization, and accessing these via `AdvancedExtension.Optimization` and `AdvancedExtension.Optimization` is much nicer.

core/src/main/java/io/substrait/extension/AdvancedExtension.java
```diff
@@ -1,27 +1,31 @@
...
+   * consumer.
+   */
+  public interface Optimization {}
+
+  /**
+   * Enhancement associated with an {@link io.substrait.proto.AdvancedExtension}
+   *
+   * <p>An enhancement alters semantics. Cannot be ignored by a consumer.
+   */
+  public interface Enhancement {}
```
vbarua, line 32: Neither Optimization or Enhancement extend from `ToProto` in your approach here.

The nice thing about having a `toProto` method attached to each of these classes was that it made it really easy to implement new Optimization and Enhancement classes and make sure they had associated serialization logic.

That was not without it's own weirdness, because it tied the POJO class to a specific serialization format explicitly. Also, only the serialization was handled this way, because the deserialization code used the `ProtoRelConverter#advancedExtension()` method, which users could override.

I do think you're approach is better overall.

core/src/main/java/io/substrait/extension/ExtensionProtoConverter.java
```diff
@@ -0,0 +1,49 @@
...
+   * Converts an {@link AdvancedExtension.Enhancement} to proto.
+   *
+   * <p>Override to provide a custom converter for {@link AdvancedExtension.Enhancement} data.
+   *
+   * @param enhancement the {@link AdvancedExtension.Enhancement} to convert
+   * @return the converted proto
+   */
+  protected Any toProto(final AdvancedExtension.Enhancement enhancement) {
+    return com.google.protobuf.Any.pack(com.google.protobuf.Empty.getDefaultInstance());
+  }
```
vbarua, line 53: The way this worked before, because the `toProto` method was attached to the Enhancement class, it was impossible to forget to wire in the serialization code. Because these are now decoupled, a user can add an Enhancement and forget to modify the serialization code.

They could theoretically notice if they inspect the output plan and notice a bunch of empty messages where they expect enhancements, but I think it might be better if this blew up loudly with an exception. What do you think?

The same might make sense for optimizations, but failing to serialize an optimization is less severe because they don't change plan semantics like enhancements do.

core/src/main/java/io/substrait/extension/ExtensionProtoConverter.java
```diff
@@ -0,0 +1,49 @@
...
+   * @param advancedExtension the {@link AdvancedExtension} to convert
+   * @return the converted {@link io.substrait.proto.AdvancedExtension}
+   */
+  public io.substrait.proto.AdvancedExtension toProto(final AdvancedExtension advancedExtension) {
+    final io.substrait.proto.AdvancedExtension.Builder builder =
+        io.substrait.proto.AdvancedExtension.newBuilder();
+    advancedExtension.getEnhancement().ifPresent(e -> builder.setEnhancement(toProto(e)));
+    advancedExtension.getOptimizations().forEach(e -> builder.addOptimization(toProto(e)));
+    return builder.build();
+  }
```
vbarua, line 27: Structuring the conversion this way is nice because I think it results in a more elegant way to solve the issue I fixed in https://github.com/substrait-io/substrait-java/pull/341.

Effectively, instead of trying to inject the extra metadata needed for the conversion there as a param that all `toProto` implementors need to handle (and provide), as a user we can inject that metadata into the constructor for the class extending AdvancedExtensionProtoConverter.

core/src/test/java/io/substrait/plan/PlanConverterTest.java
```diff
@@ -0,0 +1,106 @@
...
+    final io.substrait.proto.Plan protoPlan = toProtoConverter.toProto(plan);
+
+    final ProtoPlanConverter fromProtoConverter = new ProtoPlanConverter();
+
+    // Enhancements are not handled by the default ProtoPlanConverter
+    assertThrows(IllegalStateException.class, () -> fromProtoConverter.from(protoPlan));
+  }
+
+  @Test
+  void customAdvancedExtensionSerDes() {
```
vbarua:
```suggestion
  void customAdvancedExtensionSerde() {
```

This is me being incredibly opinionated on a thing that doesn't really matter, but I saw `SerDes` and it made my brain stop. It feels wrong.

core/src/test/java/io/substrait/utils/StringHolderHandlingProtoExtensionConverter.java
```diff
@@ -0,0 +1,19 @@
...
+import io.substrait.extension.AdvancedExtension.Enhancement;
+import io.substrait.extension.AdvancedExtension.Optimization;
+import io.substrait.extension.ProtoAdvancedExtensionConverter;
+
+public class StringHolderHandlingProtoAdvancedExtensionConverter
+    extends ProtoAdvancedExtensionConverter {
+  @Override
+  protected Enhancement enhancementFromAdvancedExtension(final Any any) {
+    return StringHolder.fromProto(any);
+  }
```
vbarua, line 12: opportunistic: You didn't modify this, and in fact I wrote the original version of this, but if I knew then what I knew now I would have done the following.

```java
    switch (any.getTypeUrl()) {
      case "type.googleapis.com/google.protobuf.StringValue":
        return StringHolder.fromProto(any);
      default:
            String.format(
                "Missing handler for protobuf enhancement with type URL: %s", any.getTypeUrl()));
    }
```

From experience I think is a better best practice for this kind of work, and it would be good to capture in this test class, which I suspect is what people look to when writing their own versions of this.

core/src/test/java/io/substrait/utils/StringHolderHandlingProtoExtensionConverter.java
```diff
@@ -0,0 +1,19 @@
...
+
+  @Override
+  protected Optimization optimizationFromAdvancedExtension(final Any any) {
+    return StringHolder.fromProto(any);
+  }
```
vbarua, line 17: Same idea as above.

```java
    switch (any.getTypeUrl()) {
      case "type.googleapis.com/google.protobuf.StringValue":
        return StringHolder.fromProto(any);
      default:
        return new EmptyOptimization();
    }
```

core/src/test/java/io/substrait/utils/StringHolderHandlingAdvancedExtensionProtoConverter.java
```diff
@@ -0,0 +1,27 @@
...
+import com.google.protobuf.Any;
+import io.substrait.extension.AdvancedExtension.Enhancement;
+import io.substrait.extension.AdvancedExtension.Optimization;
+import io.substrait.extension.AdvancedExtensionProtoConverter;
+
+public class StringHolderHandlingAdvancedExtensionProtoConverter
+    extends AdvancedExtensionProtoConverter {
+  @Override
+  protected Any toProto(final Optimization optimization) {
+    if (optimization instanceof StringHolder) {
```
vbarua: This is the one thing I don't like about the removal of the `toProto` method from the Optimization and Enhancement classes. It's going to result in a lot instanceof checking which just feels gross, but I do think it's better than what we had before.

Thinking forward, this might be something we can improve with a visitor.

core/src/test/java/io/substrait/utils/StringHolderHandlingAdvancedExtensionProtoConverter.java
```diff
@@ -0,0 +1,27 @@
...
+      return ((StringHolder) optimization).toProto(null);
+    }
+
+    return null;
```
vbarua: In a real system, I would suggest that this throws. It's not super important because this is a test, but I would suggest modelling that practice in this code.

vbarua (commented):
Looking at your PR, I think it could lead to a much nicer solution to what I tried to solve in https://github.com/substrait-io/substrait-java/pull/341.

[Shoehorning](https://github.com/substrait-io/substrait-java/pull/341/files#diff-1b9b158bbc954152e7281e40762ad1b526c320b98e4057d46b3b44a8e046ae45R5-R7) the `RelProtoConverter` argument into the `toProto` in interface was short-sighted of me, especially in the face of trying to convert AdvancedExtensions in places that don't have access to one, like what you're trying to do now in the PlanProtoRelConverter. A better idea would have been to require only the `ExtensionCollector`, which is really the only shared state that needs to be passed around. An even better idea comes about with your dedicated conversion class. I think users should be able to inject the metadata / classes they need into that custom class to achieve the existing capability.

I like a lot about this, but there are two things I don't like:
1. Having to do instanceof checks in the serialization code makes me sad.
2. Optimization and Enhancement now don't implement ToProto, but a bunch of other extension points do. 

I don't consider these blocking, but it makes me think of future improvements.

For 2, fully removing the `ToProto` class and moving all deserialization to a big ExtensionProtoConverter class might make sense. I'm thinking of one big class to avoid needing to have constructors with params for AdvancedExtensionProtoConverter, DetailProtoConverter, etc.

For 1, maybe kind of visitor could be used.

I've written a lot, so to summarize: Broadly in favor of this, have some questions around future work I want to discuss before merging.

nielspardon:
> For 2, fully removing the `ToProto` class and moving all deserialization to a big ExtensionProtoConverter class might make sense. I'm thinking of one big class to avoid needing to have constructors with params for AdvancedExtensionProtoConverter, DetailProtoConverter, etc.

I first tried to preserve the `ToProto` structure but realized that this was a bit too difficult. I was also wondering whether there was going to be too much fragmentation having one big ExtensionProtoConverter class sounds like a good compromise.

> For 1, maybe kind of visitor could be used.

Let me give that a try.

core/src/test/java/io/substrait/plan/PlanConverterTest.java
nielspardon: I have no problem renaming this one

core/src/test/java/io/substrait/utils/StringHolderHandlingProtoExtensionConverter.java
nielspardon: in my opinion it would make sense to move this check into `StringHolder.fromProto()` in this case to reduce code duplication though the exception message would be more generic since it would not refer to `enhancement` anymore

core/src/main/java/io/substrait/extension/ExtensionProtoConverter.java
nielspardon: yes, there's a difference between the user explicitly ignoring optimizations or the SDK by default ignoring optmizations. I would also go for the explicit approach to let the user make a choice if they are parsing plans with optimizations by blowing up. I changed the code accordingly and switched from `IllegalStateException` to `UnsupportedOperationException` which I feel is semantically closer for the case where we want to signal that there is some logic missing to handle enhancements and optimizations.

core/src/test/java/io/substrait/plan/PlanConverterTest.java
nielspardon: renamed

core/src/test/java/io/substrait/utils/StringHolderHandlingProtoExtensionConverter.java
nielspardon: I made the modifications in `StringHolder.fromProto()`.

core/src/test/java/io/substrait/utils/StringHolderHandlingProtoExtensionConverter.java
nielspardon: I made the modifications in `StringHolder.fromProto()` but without the `new EmptyOptimizaton()` since I dropped that in favor of blowing up.

core/src/test/java/io/substrait/utils/StringHolderHandlingAdvancedExtensionProtoConverter.java
nielspardon: I changed this by using generics `AdvancedExtension<
    O extends AdvancedExtension.Optimization, E extends AdvancedExtension.Enhancement>` and in the `ExtensionProtoConverter<
    O extends AdvancedExtension.Optimization, E extends AdvancedExtension.Enhancement>` this then allows to just do this:

```java
public class StringHolderHandlingExtensionProtoConverter
    extends ExtensionProtoConverter<StringHolder, StringHolder> {
  @Override
  protected Any toProto(final StringHolder holder) {
    return holder.toProto(null);
  }
}
```

no more `instanceof` checks and since `StringHolder` is used for both optimization and enhancements it collapses to a single `toProto` method.

core/src/test/java/io/substrait/utils/StringHolderHandlingAdvancedExtensionProtoConverter.java
nielspardon: this no longer returns null but throws an exception

core/src/test/java/io/substrait/utils/StringHolderHandlingAdvancedExtensionProtoConverter.java
vbarua: That works well. It also means that users can implement their own interface like
```java
interface CustomOptimizations extends AdvancedExtension.Optimization {
  Any myToProto(...)
}
```
and wire it into the ExtensionProtoConverter.

core/src/main/java/io/substrait/extension/ExtensionProtoConverter.java
```diff
@@ -0,0 +1,51 @@
...
+
+import com.google.protobuf.Any;
+
+/**
+ * Converter from {@link AdvancedExtension} to proto.
+ *
+ * <p>Extend this class to customize and use with {@link io.substrait.relation.RelProtoConverter}
+ * or @{link {@link io.substrait.plan.PlanProtoConverter}.
+ */
+public class ExtensionProtoConverter<
```
vbarua, line 12: Not for this PR, but when we get around to updating this with more converters, we'll want to update the docstring to reflect that.

core/src/main/java/io/substrait/extension/ExtensionProtoConverter.java
vbarua: I'm not sure if what you're saying and what you've implemented are in line.

> I would also go for the explicit approach to let the user make a choice if they are parsing plans with optimizations by blowing up.

To confirm, you're suggesting that serialization to proto should blow up, but not deserialization from proto?

core/src/main/java/io/substrait/relation/ProtoRelConverter.java
```diff
@@ -53,16 +53,26 @@ public class ProtoRelConverter {
 
   protected final ExtensionLookup lookup;
   protected final SimpleExtension.ExtensionCollection extensions;
-  private final ProtoTypeConverter protoTypeConverter;
+  protected final ProtoTypeConverter protoTypeConverter;
+  protected final ProtoExtensionConverter extensionConverter;
```
vbarua: For consistency across the library, I would recommend that both of these have the `proto` prefix.
```java
  protected final ProtoTypeConverter protoTypeConverter;
  protected final ProtoExtensionConverter protoExtensionConverter;
```

I'm not opposed to
```java
  protected final ProtoTypeConverter typeConverter;
  protected final ProtoExtensionConverter extensionConverter;
```
but it's not consistent with the other converter classes.

core/src/main/java/io/substrait/relation/RelProtoConverter.java
```diff
@@ -60,11 +61,19 @@ public class RelProtoConverter
   protected final TypeProtoConverter typeProtoConverter;
 
   protected final ExtensionCollector extensionCollector;
+  protected final ExtensionProtoConverter extensionConverter;
```
vbarua: To be consistent with `typeProtoConverter` this should probably be `extensionProtoConverter`. I would also group it with the ProtoConverters above, and not here. with the ExtensionCollector.

vbarua (commented):
Left a couple more comments, but changes look good overall. I think we can get this merged in tomorrow.

core/src/main/java/io/substrait/extension/ExtensionProtoConverter.java
nielspardon: > To confirm, you're suggesting that serialization to proto should blow up, but not deserialization from proto?

You're right. I was thinking about changing that yesterday but forgot. It should throw an exception both ways to make it obvious that users need to handle this.

core/src/main/java/io/substrait/relation/ProtoRelConverter.java
nielspardon: I renamed that accordingly

core/src/main/java/io/substrait/relation/RelProtoConverter.java
nielspardon: I renamed renamed and grouped that accordingly

core/src/main/java/io/substrait/extension/ExtensionProtoConverter.java
nielspardon: changed this accordingly

vbarua (approved):
LGMT. Thanks for working on this.

PR #539, opened 2025-10-06
build: use GitHub app to automate releases

vbarua:


vbarua:
@nielspardon pointed me towards https://gonzalohirsch.com/blog/semantic-release-and-branch-protection-rules/#the-solution as a possible way to deal with our release failures due to branch protections rule.

I've created an app called "substrait-java releaser" in the substrait organization for this work, and wired it in here.

.github/workflows/release.yml
```diff
@@ -81,7 +86,7 @@ jobs:
       - name: Run semantic-release
         run: ./ci/release/run.sh
         env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          GITHUB_TOKEN: ${{ steps.app-token.output.token }}
```
nielspardon: I'm wondering whether this should be `outputs` since in the two examples I shared it was plural:

```suggestion
          GITHUB_TOKEN: ${{ steps.app-token.outputs.token }}
```

also the Github Actions docs seem to use the plural:
https://docs.github.com/en/actions/how-tos/write-workflows/choose-what-workflows-do/pass-job-outputs

nielspardon (approved):
LGTM with the typo fixed

PR #542, opened 2025-10-07
build: downgrade Spotless to 7.2.1

vbarua:


bestbeforetoday:
It does fix the problem for me running locally. I am investigating other options but I agree that this is the right move in the short term to get publication working again.

Issue #552, opened 2025-10-14
The class SubstraitRelVisitor is not open enough for convenient inheritance


ZorinAnton:
There are two main issues:
- the method [popFieldAccessDepthMap](https://github.com/substrait-io/substrait-java/blob/0ad0570a9653982c248ddf8ee0019c5c10f43e89/isthmus/src/main/java/io/substrait/isthmus/SubstraitRelVisitor.java#L500) is private while it is used in static `convert` methods that might need to be reimplemented by subclasses.
- the static `convert` methos instantiate the visitor variable as `SubstraitRelVisitor`. There might be the need to instantiate subclasses instead.

As of now, the workaround is to define required convert methods in subclasses and use java reflection to call `popFieldAccessDepthMap`.

Proposed solution:
- Increase Visibility of popFieldAccessDepthMap to protected. This will allow subclasses to override the method and provide their own logic for handling field access resolution.
- Introduce a Factory for Visitor Creation and refactor the static convert methods to accept the factory. This would allow users to pass a constructor reference for their custom visitor subclass.

PR #553, opened 2025-10-14
refactor(isthmus): improve extensibility of SubstraitRelVisitor

ZorinAnton:
relates to #552

### DIFF
--- a/isthmus/src/test/java/io/substrait/isthmus/SubtraitRelVisitorExtensionTest.java
+++ b/isthmus/src/test/java/io/substrait/isthmus/SubtraitRelVisitorExtensionTest.java
@@ -29,7 +29,7 @@ import org.apache.calcite.tools.Frameworks;
 import org.apache.calcite.tools.RelBuilder;
 import org.junit.jupiter.api.Test;
 
-public class SubtraitRelVisitorExtensionTest {
+class SubtraitRelVisitorExtensionTest {
 
   public class Employee {
     public final int DEPT_ID;


isthmus/src/main/java/io/substrait/isthmus/SubstraitRelVisitor.java
```diff
@@ -522,9 +522,12 @@ public static Plan.Root convert(RelRoot relRoot, SimpleExtension.ExtensionCollec
   }
 
   public static Plan.Root convert(
-      RelRoot relRoot, SimpleExtension.ExtensionCollection extensions, FeatureBoard features) {
+      RelRoot relRoot,
+      SimpleExtension.ExtensionCollection extensions,
+      FeatureBoard features,
+      SubstraitRelVisitorCreator visitorCreator) {
```
nielspardon: I think it would be a good idea to have a unit test for this new functionality. I guess the idea here is to allow handling custom extension relation operations when mapping from Calcite by extending the `SubstraitRelVisitor`? Can we add a unit test that demonstrates how this is meant to be used?

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelVisitorCreator.java
```diff
@@ -0,0 +1,11 @@
+package io.substrait.isthmus;
+
+import io.substrait.extension.SimpleExtension;
+import org.apache.calcite.rel.type.RelDataTypeFactory;
+
+public interface SubstraitRelVisitorCreator {
```
nielspardon: A minor thing which I was wondering about is the term `Creator` since in the ticket you mentioned you wanted to add a `Factory` and that's also where my mind went to initially. But then we do already have `TypeCreator`, `ExpressionCreator` in substrait-java so I guess `SubstraitRelVisitorCreator` is fine.

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelVisitor.java
ZorinAnton: fixed

isthmus/src/test/java/io/substrait/isthmus/SubtraitRelVisitorExtensionTest.java
```diff
@@ -0,0 +1,255 @@
...
+import org.apache.calcite.rel.type.RelDataTypeFactory;
+import org.apache.calcite.rex.RexNode;
+import org.apache.calcite.schema.SchemaPlus;
+import org.apache.calcite.sql.parser.SqlParser;
+import org.apache.calcite.tools.FrameworkConfig;
+import org.apache.calcite.tools.Frameworks;
+import org.apache.calcite.tools.RelBuilder;
+import org.junit.jupiter.api.Test;
+
+public class SubtraitRelVisitorExtensionTest {
```
nielspardon: I think this could be package private with JUnit 5:
```suggestion
class SubtraitRelVisitorExtensionTest {
```

PR #554, opened 2025-10-14
feat(spark): dialect YAML file for spark converter

andrew-coleman:
Adds the formal dialect definition for the Spark converter. The file is generated from code which extracts the relevant information from the exisitng FunctionMapper definitions. This ensures that the dialect YAML remains in-sync with the code.
Tests have been introduced to ensure that the dialect conforms to the schema (defined in teh base substrait library), and that the published dialect YAML is the equivalent to one that would be generated from the current code.

### DIFF
--- a/spark/src/test/scala/io/substrait/spark/DialectSuite.scala
+++ b/spark/src/test/scala/io/substrait/spark/DialectSuite.scala
@@ -1,0 +1,60 @@
+package io.substrait.spark
+
+import io.substrait.spark.utils.{Dialect, DialectGenerator}
+import io.substrait.spark.utils.DialectGenerator.schemaPath
+
+import org.apache.spark.SparkFunSuite
+import org.apache.spark.sql.test.SharedSparkSession
+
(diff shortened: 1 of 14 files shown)


spark/src/test/scala/io/substrait/spark/DialectSuite.scala
```diff
@@ -0,0 +1,60 @@
...
+  test("compare generated dialect") {
+    val mapper = new ObjectMapper(new YAMLFactory()).registerModules(DefaultScalaModule)
+
+    val genDialect = DialectGenerator.generate()
+    val publishedDialect =
+      mapper.readValue(new File(dialectPath), scala.reflect.classTag[Dialect].runtimeClass)
+    // The following will fail if the generated dialect differs from the published one.
+    // If this is caused by an intentional change, the published dialect should be regenerated using:
+    // `./gradlew dialect`
+    assertResult(publishedDialect)(genDialect)
```
nielspardon, line 57: great that you thought of that

nielspardon (approved):
LGTM

PR #557, opened 2025-10-17
fix(spark): convert UnsafeArrayData literal to substrait

cht42:
UnsafeArrayData doesnt support the `array` method https://github.com/apache/spark/blob/c0416713e8e1e5fb81f218feb339cfbeacb61b72/sql/catalyst/src/main/java/org/apache/spark/sql/catalyst/expressions/UnsafeArrayData.java#L105


cht42:
done !

PR #561, opened 2025-10-22
build: limit immutables dependencies

bestbeforetoday:
The build depends on org.immutables:value as an annotation processor, which contains components beyond those strictly required for annotation processing.

This change limits the immutables dependencies and their scope to minimise the dependency footprint and avoid runtime classpath pollution.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -34,7 +34,7 @@ dependencies {
   testImplementation(libs.junit.jupiter.api)
   testImplementation(libs.junit.jupiter.engine)
   implementation(libs.slf4j.api)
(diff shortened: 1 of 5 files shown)


nielspardon:
> This change limits the immutables dependencies and their scope to minimise the dependency footprint and avoid runtime classpath pollution.

To my knowledge the annotation processor dependencies do not end up on the runtime classpath but they are used with the javac compile only or am I missing something?

We could still do this change though I'm wondering why the instructions on immutables.org are defaulting to using the full `org.immutables:value` package and not using the processor package by default instead.

PR #564, opened 2025-10-22
fix: prevents exception on construction of `FunctionConverter` with duplicate functions

### LINKED ISSUE #562: `FunctionConverter` crashes when multiple extensions define functions with the same name and signature
When merging extension collections that contain functions with identical names and signatures but different URNs, `FunctionConverter` throws `IllegalArgumentException` during initialization. This prevents using multiple extension sets that reasonably might implement the same function (e.g., ltrim) with slightly different semantics or optimizations.

Example Error:
```
java.lang.IllegalArgumentException: Multiple entries with same key: ltrim:str_str
    at com.google.common.collect.ImmutableMap$Builder.build(ImmutableMap.java:...)
    at io.substrait.isthmus.expression.FunctionConverter$FunctionFinder.<init (FunctionConverter.java:179)
```

benbellick:
Closes #562 

This test resolves an issue in which `FunctionConverter`s constructed with `ExtensionCollection`s that contain multiple functions with different URNs results in an exception being thrown. 

The fix is to allow a list of functions to be associated with a particular `<name>:<sig>` pairing, and then to select the final option of those available. This is reflected in the comments, which clarify that the last added function takes precedence.

### DIFF
--- a/isthmus/src/main/java/io/substrait/isthmus/expression/FunctionConverter.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/expression/FunctionConverter.java
@@ -47,7 +47,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 /**
- * Abstract base class for converting between Calcite SqlOperators and Substrait function
+ * Abstract base class for converting between Calcite {@link SqlOperator}s and Substrait function
  * invocations.
  *
  * <p>This class handles bidirectional conversion:
@@ -56,7 +56,7 @@ import org.slf4j.LoggerFactory;
  *   <li><b>Calcite → Substrait:</b> Subclasses implement {@code convert()} methods to convert
  *       Calcite calls to Substrait function invocations
  *   <li><b>Substrait → Calcite:</b> {@link #getSqlOperatorFromSubstraitFunc} converts Substrait
- *       function keys to Calcite SqlOperators
+ *       function keys to Calcite {@link SqlOperator}s
  * </ul>
  *
  * <p>When multiple functions with the same name and signature are passed into the constructor, a
@@ -157,15 +157,15 @@ public abstract class FunctionConverter<
   }
 
   /**
-   * Converts a Substrait function to a Calcite SqlOperator (Substrait → Calcite direction).
+   * Converts a Substrait function to a Calcite {@link SqlOperator} (Substrait → Calcite direction).
    *
    * <p>Given a Substrait function key (e.g., "concat:str_str") and output type, this method finds
-   * the corresponding Calcite SqlOperator. When multiple operators match, the output type is used
-   * to disambiguate.
+   * the corresponding Calcite {@link SqlOperator}. When multiple operators match, the output type
+   * is used to disambiguate.
    *
    * @param key the Substrait function key (function name with type signature)
    * @param outputType the expected output type
-   * @return the matching SqlOperator, or empty if no match found
+   * @return the matching {@link SqlOperator}, or empty if no match found
    */
   public Optional<SqlOperator> getSqlOperatorFromSubstraitFunc(String key, Type outputType) {
     Map<SqlOperator, TypeBasedResolver> resolver = getTypeBasedResolver();


isthmus/src/main/java/io/substrait/isthmus/expression/FunctionConverter.java
```diff
@@ -75,9 +76,9 @@ public FunctionConverter(
     this.typeFactory = typeFactory;
     this.substraitFuncKeyToSqlOperatorMap = ArrayListMultimap.create();
 
-    ArrayListMultimap<String, F> alm = ArrayListMultimap.<String, F>create();
+    ArrayListMultimap<String, F> nameToFn = ArrayListMultimap.<String, F>create();
```
benbellick, line 122: This is unrelated to the PR technically, but I just thought that this was a clearer name.

isthmus/src/test/java/io/substrait/isthmus/DuplicateFunctionUrnTest.java
```diff
@@ -0,0 +1,127 @@
...
+      }
+
+      String urn1 = ltrimFunctions.get(0).getAnchor().urn();
+      String urn2 = ltrimFunctions.get(1).getAnchor().urn();
+      if (urn1.equals(urn2)) {
+        throw new IllegalStateException(
+            "Expected different URNs for the two ltrim functions, but both were: " + urn1);
+      }
+    } catch (IOException e) {
+      throw new RuntimeException(e);
```
nielspardon: FYI: there is `UncheckedIOException` specifically for this purpose which avoids throwing a raw exception type:
```suggestion
      throw new UncheckedIOException(e);
```
https://docs.oracle.com/javase/8/docs/api/java/io/UncheckedIOException.html

benbellick: Thanks! Just added that.

isthmus/src/test/java/io/substrait/isthmus/DuplicateFunctionUrnTest.java
```diff
@@ -0,0 +1,187 @@
...
+  static final SimpleExtension.ExtensionCollection collection1;
+  static final SimpleExtension.ExtensionCollection collection2;
+  static final SimpleExtension.ExtensionCollection collection;
+
+  static {
+    try {
+      String extensions1 = asString("extensions/functions_duplicate_urn1.yaml");
+      String extensions2 = asString("extensions/functions_duplicate_urn2.yaml");
+      collection1 = SimpleExtension.load("urn1://functions", extensions1);
+      collection2 = SimpleExtension.load("urn2://functions", extensions2);
```
nielspardon: I think given that URNs are valid URIs you could provide URNs here as the URI for consistency:

```suggestion
      collection1 = SimpleExtension.load("urn:extension:io.substrait:functions_string", extensions1);
      collection2 = SimpleExtension.load("urn:extension:com.domain:string", extensions2);
```

isthmus/src/main/java/io/substrait/isthmus/expression/FunctionConverter.java
```diff
@@ -349,6 +350,9 @@ public Optional<T> attemptMatch(C call, Function<RexNode, Expression> topLevelCo
        * Not enough context here to construct a substrait EnumArg.
        * Once a FunctionVariant is resolved we can map the String Literal
        * to a EnumArg.
+       *
+       * Note that if there are multiple registered function extensions which can match a particular Call,
+       * the last one added to the extension collection will be matched.
```
nielspardon, line 422: I'm wondering whether this comment should be also in a more prominent place for any consumers of isthmus and where they would find that. Would maybe at the top of the `FunctionConverter` class be more prominent?

isthmus/src/test/java/io/substrait/isthmus/DuplicateFunctionUrnTest.java
```diff
@@ -0,0 +1,187 @@
...
+import java.util.Optional;
+import org.apache.calcite.rex.RexBuilder;
+import org.apache.calcite.rex.RexCall;
+import org.apache.calcite.rex.RexNode;
+import org.apache.calcite.sql.fun.SqlStdOperatorTable;
+import org.apache.calcite.sql.fun.SqlTrimFunction.Flag;
+import org.junit.jupiter.api.Test;
+
+/** Tests to reproduce #562 */
+public class DuplicateFunctionUrnTest extends PlanTestBase {
```
nielspardon: I think for JUnit 5 the test class can be package private:

```suggestion
class DuplicateFunctionUrnTest extends PlanTestBase {
```

isthmus/src/test/java/io/substrait/isthmus/DuplicateFunctionUrnTest.java
```diff
@@ -0,0 +1,187 @@
...
+
+    assertNotNull(converter);
+  }
+
+  @Test
+  void testDuplicateWindowFunctionWithDifferentUrns() {
+    WindowFunctionConverter converter =
+        new WindowFunctionConverter(collection.windowFunctions(), typeFactory);
+
+    assertNotNull(converter);
```
nielspardon: interesting, is it really possible that a constructor returns null in Java?

nielspardon:
looks good in general. just some minor remarks from my side

isthmus/src/test/java/io/substrait/isthmus/DuplicateFunctionUrnTest.java
benbellick: Yeah that is probably a bit easier to read

isthmus/src/test/java/io/substrait/isthmus/DuplicateFunctionUrnTest.java
benbellick: Good point!

benbellick: So I had to actually use these values somewhere so that the linter doesn't complain. I ended up just wrapping the construction of these objects in `assertDoesNotThrow`.

isthmus/src/main/java/io/substrait/isthmus/expression/FunctionConverter.java
benbellick: I ended up adding quite a bit of comments. Let me know if anything added is incorrect or just too much :)

isthmus/src/main/java/io/substrait/isthmus/expression/FunctionConverter.java
```diff
@@ -112,6 +156,17 @@ public FunctionConverter(
...
+   * Converts a Substrait function to a Calcite SqlOperator (Substrait → Calcite direction).
+   *
+   * <p>Given a Substrait function key (e.g., "concat:str_str") and output type, this method finds
+   * the corresponding Calcite SqlOperator. When multiple operators match, the output type is used
+   * to disambiguate.
+   *
+   * @param key the Substrait function key (function name with type signature)
+   * @param outputType the expected output type
+   * @return the matching SqlOperator, or empty if no match found
+   */
```
nielspardon, line 169: not a must, I tend to add `{@link xxx}` to mentions of classes/methods hoping they might be useful for a reader of the javadocs

```suggestion
  /**
   * Converts a Substrait function to a Calcite {@link SqlOperator} (Substrait → Calcite direction).
   *
   * <p>Given a Substrait function key (e.g., "concat:str_str") and output type, this method finds
   * the corresponding Calcite {@link SqlOperator}. When multiple operators match, the output type is used
   * to disambiguate.
   *
   * @param key the Substrait function key (function name with type signature)
   * @param outputType the expected output type
   * @return the matching {@link SqlOperator}, or empty if no match found
   */
```

isthmus/src/main/java/io/substrait/isthmus/expression/FunctionConverter.java
nielspardon: looks great, I was suggesting to Victor a couple weeks in another PR that we maybe should try to add javadocs whenever we touch code to slowly build up javadoc documentation so this is a great contribution

nielspardon (approved):
LGTM

isthmus/src/main/java/io/substrait/isthmus/expression/FunctionConverter.java
benbellick: That was a good idea, just pushed a commit for it.

PR #565, opened 2025-10-22
feat(core,isthmus): support grouping set index in Aggregate

### LINKED ISSUE #527: Add support for grouping set index in Aggregate
The current `derivceRecordType()` method does not correctly add the grouping set index column as documented in the current specification for Aggregate:

> Direct Output Order: The list of grouping expressions in declaration order followed by the list of measures in declaration order, **followed by an i32 describing the associated particular grouping set the value is derived from (if applicable).**

https://substrait.io/relations/logical_relations/#aggregate-operation

https://github.com/substrait-io/substrait-java/blob/4ae8188588aa006958c8839fd0540d511102164f/core/src/main/java/io/substrait/relation/Aggregate.java#L23-L35

This was uncovered in https://github.com/substrait-io/substrait-java/pull/508 which fixes the nullability for grouping set columns.

nielspardon:
fixes #527 

- adds the int32 grouping set index to `Aggregate.deriveRecordType()`
- adds Calcite mappings to isthmus which maps a `GROUP_ID()` aggregate function call mapping for the grouping set index since it has the same semantic as the Substrait grouping set index
- ensures remap removes the grouping set index field if it there is no `GROUP_ID()` function call in the Calcite plan

### DIFF
--- a/isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
@@ -358,7 +358,7 @@ public class SubstraitRelNodeConverter
     // map grouping set index if it is not removed via remap
     final boolean emitDirect = remap.isEmpty();
     final boolean groupingSetIndexGetsRemapped =
-        remap.isPresent() && remap.get().indices().contains(lastFieldIndex);
+        remap.map(r -> r.indices().contains(lastFieldIndex)).orElse(false);
     if (aggregate.getGroupings().size() > 1 && (emitDirect || groupingSetIndexGetsRemapped)) {
       aggregateCalls.add(
           AggregateCall.create(
@@ -379,8 +379,7 @@ public class SubstraitRelNodeConverter
         for (int i = 0; i < remapList.size(); i++) {
           if (remapList.get(i).equals(lastFieldIndex)) {
             // replace last field index with field index of the GROUP_ID() function call
-            remapList.remove(i);
-            remapList.add(i, groupingCallIndex);
+            remapList.set(i, groupingCallIndex);
           }
         }
         remap = Optional.of(Remap.of(remapList));
--- a/isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
@@ -45,6 +45,7 @@ import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.HashSet;
+import java.util.LinkedList;
 import java.util.List;
 import java.util.Optional;
 import java.util.OptionalLong;
@@ -375,7 +376,7 @@ public class SubstraitRelNodeConverter
               null));
       final int groupingCallIndex = aggregateCalls.size() - 1;
       if (groupingSetIndexGetsRemapped) {
-        List<Integer> remapList = new ArrayList<>(remap.get().indices());
+        List<Integer> remapList = new LinkedList<>(remap.get().indices());
         for (int i = 0; i < remapList.size(); i++) {
           if (remapList.get(i).equals(lastFieldIndex)) {
             // replace last field index with field index of the GROUP_ID() function call
--- a/isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
@@ -34,10 +34,12 @@ import io.substrait.relation.NamedUpdate;
 import io.substrait.relation.NamedWrite;
 import io.substrait.relation.Project;
 import io.substrait.relation.Rel;
+import io.substrait.relation.Rel.Remap;
 import io.substrait.relation.Set;
 import io.substrait.relation.Sort;
 import io.substrait.relation.VirtualTableScan;
 import io.substrait.type.NamedStruct;
+import io.substrait.type.TypeCreator;
 import io.substrait.util.VisitationContext;
 import java.util.ArrayList;
 import java.util.Collection;
@@ -74,6 +76,7 @@ import org.apache.calcite.rex.RexNode;
 import org.apache.calcite.rex.RexSlot;
 import org.apache.calcite.sql.SqlAggFunction;
 import org.apache.calcite.sql.SqlOperator;
+import org.apache.calcite.sql.fun.SqlStdOperatorTable;
 import org.apache.calcite.sql.parser.SqlParser;
 import org.apache.calcite.tools.Frameworks;
 import org.apache.calcite.tools.RelBuilder;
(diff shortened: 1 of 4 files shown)


isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
```diff
@@ -348,8 +351,44 @@ public RelNode visit(Aggregate aggregate, Context context) throws RuntimeExcepti
...
             .map(measure -> fromMeasure(measure, context))
             .collect(java.util.stream.Collectors.toList());
+
+    Optional<Remap> remap = aggregate.getRemap();
+    final int lastFieldIndex = groupExprs.size() + aggregateCalls.size();
+
+    // map grouping set index if it is not removed via remap
+    final boolean emitDirect = remap.isEmpty();
+    final boolean groupingSetIndexGetsRemapped =
+        remap.isPresent() && remap.get().indices().contains(lastFieldIndex);
```
bestbeforetoday: Using `isPresent` and `get` is discouraged for Optionals.


```suggestion
        remap.map(r -> r.indices().contains(lastFieldIndex)).orElse(false);
```

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
```diff
@@ -348,8 +351,44 @@ public RelNode visit(Aggregate aggregate, Context context) throws RuntimeExcepti
...
+              typeConverter.toCalcite(typeFactory, TypeCreator.REQUIRED.I64),
+              null));
+      final int groupingCallIndex = aggregateCalls.size() - 1;
+      if (groupingSetIndexGetsRemapped) {
+        List<Integer> remapList = new ArrayList<>(remap.get().indices());
+        for (int i = 0; i < remapList.size(); i++) {
+          if (remapList.get(i).equals(lastFieldIndex)) {
+            // replace last field index with field index of the GROUP_ID() function call
+            remapList.remove(i);
+            remapList.add(i, groupingCallIndex);
```
bestbeforetoday: This is likely to perform really badly for an ArrayList. Removing and/or inserting is likely to be much more efficient with a LinkedList, but you just want to replace:


```suggestion
            remapList.set(i, groupingCallIndex);
```

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelVisitor.java
```diff
@@ -262,16 +266,69 @@ public Rel visit(org.apache.calcite.rel.core.Aggregate aggregate) {
...
+
+    if (groupings.size() > 1) {
+      // remove the grouping set index if there was no explicit GROUP_ID() function call
+      if (groupIdCalls.isEmpty()) {
+        int groupingExprSize =
+            groupings.stream()
+                .flatMap(g -> g.getExpressions().stream())
+                .distinct()
+                .collect(Collectors.toList())
+                .size();
```
bestbeforetoday: No need to create a list only to count the number of elements.

```suggestion
            Math.toIntExact(
                groupings.stream().flatMap(g -> g.getExpressions().stream()).distinct().count());
```

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelVisitor.java
```diff
@@ -262,16 +266,69 @@ public Rel visit(org.apache.calcite.rel.core.Aggregate aggregate) {
...
+        builder.remap(Remap.offset(0, groupingExprSize + aggCalls.size()));
+      } else {
+        // remap grouping set index at the field positions where the GROUP_ID() function calls were
+        final int groupingFieldCount =
+            groupings.stream()
+                .flatMap(g -> g.getExpressions().stream())
+                .collect(Collectors.toList())
+                .size();
```
bestbeforetoday: No need to create a list only to count the number of elements.

```suggestion
            Math.toIntExact(groupings.stream().flatMap(g -> g.getExpressions().stream()).count());
```

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelVisitor.java
```diff
@@ -262,16 +266,69 @@ public Rel visit(org.apache.calcite.rel.core.Aggregate aggregate) {
...
+        final int filterAggCallCount = aggCalls.size();
+        final Integer groupingSetIndex = groupingFieldCount + filterAggCallCount;
+
+        final List<Integer> groupingFieldIndices =
+            IntStream.range(0, groupingFieldCount).mapToObj(i -> i).collect(Collectors.toList());
+
+        final List<Integer> remap = new ArrayList<>(groupingFieldIndices);
```
bestbeforetoday: Only the `remap` ArrayList seems to be needed.

```suggestion
        final List<Integer> remap =
            IntStream.range(0, groupingFieldCount)
                .mapToObj(i -> i)
                .collect(Collectors.toCollection(ArrayList::new));
```

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelVisitor.java
```diff
@@ -262,16 +266,69 @@ public Rel visit(org.apache.calcite.rel.core.Aggregate aggregate) {
...
+        for (int i = 0; i < aggregate.getAggCallList().size(); i++) {
+          AggregateCall aggCall = aggregate.getAggCallList().get(i);
+          if (filteredAggCalls.contains(aggCall)) {
+            remap.add(
+                i + groupingFieldCount, filteredAggCalls.indexOf(aggCall) + groupingFieldCount);
+          } else if (groupIdCalls.contains(aggCall)) {
+            remap.add(i + groupingFieldCount, groupingSetIndex);
+          } else {
+            throw new IllegalStateException(
+                "encountered AggregateCall that is neither in filteredAggCalls nor in groupIdCalls");
```
bestbeforetoday: To aid debugging, maybe the error message can contain more information about the AggregateCall that caused the problem?

```suggestion
            tthrow new IllegalStateException(
                "encountered AggregateCall that is neither in filteredAggCalls nor in groupIdCalls: "
                    + aggCall);
```

bestbeforetoday (commented):
Generally this looks great with clearly laid out code. Just a few very minor _suggestions_ to improve the implementation. See in-line comments.

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
nielspardon: thanks, makes sense, I missed the `set()` method exists for a List

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelVisitor.java
nielspardon: it might be useful. I wouldn't expect this to ever get triggered. just added it for safety if someone modified the code

nielspardon: included this change, also added a comment stating that this exception should never get thrown

bestbeforetoday (approved):
👍

PR #568, opened 2025-10-23
feat(expression,isthmus): mapping of square root function from substrait to calcite

asuriy:
- Addition of the missing scalar function (sqrt) mapping from substrait to substrait-java
- Mapped from [https://github.com/substrait-io/substrait/blob/main/extensions/functions_arithmetic.yaml](https://github.com/substrait-io/substrait/blob/main/extensions/functions_arithmetic.yaml)
- Builds successfully with added test

- Added line 67 to `substrait-java/isthmus/src/main/java/io/substrait/isthmus/expression/FunctionMappings.java`
- Added to line 46 in `substrait-java/isthmus/src/main/java/io/substrait/isthmus/expression/ScalarFunctionConverter.java`
- Added `substrait-java/isthmus/src/main/java/io/substrait/isthmus/expression/SqrtFunctionMapper.java`
- Added test to `substrait-java/isthmus/src/test/java/io/substrait/isthmus/ArithmeticFunctionTest.java`

### DIFF
--- a/isthmus/src/main/java/io/substrait/isthmus/expression/SqrtFunctionMapper.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/expression/SqrtFunctionMapper.java
@@ -4,7 +4,6 @@ import io.substrait.expression.Expression;
 import io.substrait.expression.FunctionArg;
 import io.substrait.extension.SimpleExtension.ScalarFunctionVariant;
 import java.math.BigDecimal;
-import java.util.Collection;
 import java.util.List;
 import java.util.Optional;
 import java.util.stream.Collectors;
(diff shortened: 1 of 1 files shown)


bestbeforetoday (approved):
@asuriy I pushed an additional commit to your PR with suggested changes. The major change is the removal of the custom mapping for the sqrt function itself. The default mapping is sufficient here so the custom mapper can just ignore this case. Custom mapping is only required to convert power(x, 0.5) to sqrt(x). I made some very slight logical changes to (I think) make the intent clearer. Mainly around using `isPowerOfHalf` as a function name to check whether to apply custom mapping instead of the rather cryptic `includesPower`.

I also took the liberty of rebasing both of our commits on top of the current main branch content, removing the two merge commits you added on top of your commit. I would recommend `git rebase` on the remote main branch to get a pull request up-to-date instead of pulling merge commits. It keeps your PR history clean instead of intermingling commits from your branch and the upstream. Some repositories will not allow merge commits in pull requests because of this.

If you are happy with the change I have made, we could go ahead as-is. Or we could get somebody else to review it too.

PR #569, opened 2025-10-23
fix(core): close AdvancedExtension serde gaps

nielspardon:
I noticed there are multiple gaps in the serialization and deserialization of `AdvancedExtension which I'm trying to close with this PR:

- in `PlanProtoConverter` pass `ExtensionProtoConverter` to `RelProtoConverter`
- in `ProtoPlanConverter` pass `ProtoExtensionConverter` to `ProtoRelConverter`
- move/add `HasExtension` interface to `AbstractDdlRel`, `AbstractWriteRel`
- add missing serde logic for relations to `RelProtoConverter`, `ProtoRelConverter`
- adds a test that confirms the extension converters are passed through from plan converters to rel converters

### DIFF
--- a/core/src/main/java/io/substrait/relation/ProtoRelConverter.java
+++ b/core/src/main/java/io/substrait/relation/ProtoRelConverter.java
@@ -257,7 +257,7 @@ public class ProtoRelConverter {
     }
   }
 
-  protected NamedDdl newNamedDdl(DdlRel rel) {
+  protected NamedDdl newNamedDdl(final DdlRel rel) {
     final NamedStruct tableSchema = newNamedStruct(rel.getTableSchema());
     final ImmutableNamedDdl.Builder builder =
         NamedDdl.builder()
@@ -485,11 +485,11 @@ public class ProtoRelConverter {
     return builder.build();
   }
 
-  protected ExtensionTable newExtensionTable(ReadRel rel) {
-    NamedStruct namedStruct = newNamedStruct(rel);
-    Extension.ExtensionTableDetail detail =
+  protected ExtensionTable newExtensionTable(final ReadRel rel) {
+    final NamedStruct namedStruct = newNamedStruct(rel);
+    final Extension.ExtensionTableDetail detail =
         detailFromExtensionTable(rel.getExtensionTable().getDetail());
-    ImmutableExtensionTable.Builder builder =
+    final ImmutableExtensionTable.Builder builder =
         ExtensionTable.from(detail).initialSchema(namedStruct);
 
     builder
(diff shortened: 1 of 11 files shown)


core/src/main/java/io/substrait/relation/ProtoRelConverter.java
```diff
@@ -238,35 +259,47 @@ protected Rel newDdl(DdlRel rel) {
...
-        .tableDefaults(tableDefaults(rel.getTableDefaults(), tableSchema))
-        .operation(ExtensionDdl.DdlOp.fromProto(rel.getOp()))
-        .object(ExtensionDdl.DdlObject.fromProto(rel.getObject()))
-        .viewDefinition(optionalViewDefinition(rel))
-        .commonExtension(optionalAdvancedExtension(rel.getCommon()))
-        .remap(optionalRelmap(rel.getCommon()))
-        .hint(optionalHint(rel.getCommon()))
-        .build();
+    final NamedStruct tableSchema = newNamedStruct(rel.getTableSchema());
+    final Builder builder =
```
andrew-coleman: To be consistent with the other builders, can remove the `import ImmutableExtensionDdl.Builder;` at the top of the file and refer to it explicitly here
```suggestion
    final ImmutableExtensionDdl.Builder builder =
```

nielspardon: makes sense. I changed that accordingly

core/src/main/java/io/substrait/relation/ProtoRelConverter.java
```diff
@@ -175,8 +187,8 @@ protected Rel newRead(ReadRel rel) {
     }
   }
 
-  protected Rel newWrite(WriteRel rel) {
-    WriteRel.WriteTypeCase relType = rel.getWriteTypeCase();
+  protected Rel newWrite(final WriteRel rel) {
+    final WriteRel.WriteTypeCase relType = rel.getWriteTypeCase();
```
andrew-coleman, line 191: Out of curiosity, why did you add the `final` modifier to the Write/Update/DDL relations, but not all the other relations that use this same code pattern?

andrew-coleman (approved):

core/src/main/java/io/substrait/relation/ProtoRelConverter.java
nielspardon: I mainly did it for the code I thought I touched to not introduce too many unrelated changes

nielspardon: I see what you mean, let me update the PR to be a little more consistent

nielspardon: should be more consistent now

PR #575, opened 2025-10-28
feat(isthmus): support fully qualified table names in SubstraitCreateStatementParser

nielspardon:
Currently, the `SubstraitCreateStatementParser` can only parse `CREATE TABLE` statement with simple table names without any additional qualifiers. This PR changes the code to support parsing statements with fully qualified table names.

In order to support this I refactored `SchemaCollector` and extracted some functionality into `Utils.createCalciteSchemaFromNames` so I can reuse it in `SubstraitCreateStatementParser`.

I did not change `SubstraitCreateStatementParser.processCreateStatements` which is used with extended expressions where I wasn't sure whether supporting fully qualified table names would be useful or a breaking change.

I changed private fields and methods in `SubstraitCreateStatementParser` to protected to allow consumers to potentially extend it and reuse functionality.

Added a new `SubstraitCreateStatementParserTest` class testing various combinations of create statements with and without fully qualified names.

### DIFF
--- a/isthmus/src/main/java/io/substrait/isthmus/Utils.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/Utils.java
@@ -63,20 +63,20 @@ public class Utils {
   }
 
   /**
-   * Traverses all names identifying a table, creating any schemas if they are not present in the
-   * root schema, returning the final sub schema.
+   * Traverses a list of hierarchical schema names, creating any schemas if they are not present in
+   * the root schema, returning the final sub schema or the root schema if the list of names is
+   * empty.
    *
    * @param rootSchema the root schema to add the missing schemas to
-   * @param names the list of hierarchical schema and table names ordered from parent to child with
-   *     the last element containing the table name
-   * @return the final sub schema
+   * @param names the list of hierarchical schema names ordered from parent to child
+   * @return the final sub schema or the root schema
    * @see io.substrait.isthmus.sql.SubstraitCreateStatementParser#processCreateStatementsToSchema
    * @see io.substrait.isthmus.SchemaCollector#toSchema
    */
   public static CalciteSchema createCalciteSchemaFromNames(
       @NonNull final CalciteSchema rootSchema, @NonNull final List<String> names) {
     CalciteSchema schema = rootSchema;
-    for (final String schemaName : names.subList(0, names.size() - 1)) {
+    for (final String schemaName : names) {
       final CalciteSchema subSchema = schema.getSubSchema(schemaName, false);
       if (subSchema != null) {
         schema = subSchema;
--- a/isthmus/src/main/java/io/substrait/isthmus/Utils.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/Utils.java
@@ -3,11 +3,13 @@ package io.substrait.isthmus;
 import static java.util.Collections.unmodifiableList;
 
 import com.google.common.collect.Streams;
+import io.substrait.isthmus.calcite.SubstraitSchema;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Objects;
 import java.util.function.BiFunction;
 import java.util.stream.Stream;
+import org.apache.calcite.jdbc.CalciteSchema;
 
 public class Utils {
   /**
@@ -57,5 +59,32 @@ public class Utils {
         .filter(list -> !list.isEmpty()) //
         .reduce(startProduct, appendAndGenLists, (s1, s2) -> Streams.concat(s1, s2)) //
     ;
+  }
+
+  /**
+   * Traverses all names identifying a table, creating any schemas if they are not present in the
+   * root schema, returning the final sub schema.
+   *
+   * @param rootSchema the root schema to add the missing schemas to
+   * @param names the list of hierarchical schema and table names ordered from parent to child with
+   *     the last element containing the table name
+   * @return the final sub schema
+   * @see io.substrait.isthmus.sql.SubstraitCreateStatementParser#processCreateStatementsToSchema
+   * @see io.substrait.isthmus.SchemaCollector#toSchema
+   */
+  public static CalciteSchema createCalciteSchemaFromNames(
+      CalciteSchema rootSchema, List<String> names) {
+    CalciteSchema schema = rootSchema;
+    for (String schemaName : names.subList(0, names.size() - 1)) {
+      CalciteSchema subSchema = schema.getSubSchema(schemaName, false);
+      if (subSchema != null) {
+        schema = subSchema;
+      } else {
+        SubstraitSchema newSubSchema = new SubstraitSchema();
+        schema = schema.add(schemaName, newSubSchema);
+      }
+    }
+
+    return schema;
   }
 }
--- a/isthmus/src/main/java/io/substrait/isthmus/Utils.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/Utils.java
@@ -10,6 +10,7 @@ import java.util.Objects;
 import java.util.function.BiFunction;
 import java.util.stream.Stream;
 import org.apache.calcite.jdbc.CalciteSchema;
+import org.jspecify.annotations.NonNull;
 
 public class Utils {
   /**
@@ -73,14 +74,14 @@ public class Utils {
    * @see io.substrait.isthmus.SchemaCollector#toSchema
    */
   public static CalciteSchema createCalciteSchemaFromNames(
-      CalciteSchema rootSchema, List<String> names) {
+      @NonNull final CalciteSchema rootSchema, @NonNull final List<String> names) {
     CalciteSchema schema = rootSchema;
-    for (String schemaName : names.subList(0, names.size() - 1)) {
-      CalciteSchema subSchema = schema.getSubSchema(schemaName, false);
+    for (final String schemaName : names.subList(0, names.size() - 1)) {
+      final CalciteSchema subSchema = schema.getSubSchema(schemaName, false);
       if (subSchema != null) {
         schema = subSchema;
       } else {
-        SubstraitSchema newSubSchema = new SubstraitSchema();
+        final SubstraitSchema newSubSchema = new SubstraitSchema();
         schema = schema.add(schemaName, newSubSchema);
       }
     }
--- a/isthmus/src/main/java/io/substrait/isthmus/sql/SubstraitCreateStatementParser.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/sql/SubstraitCreateStatementParser.java
@@ -1,29 +1,22 @@
 package io.substrait.isthmus.sql;
 
+import io.substrait.isthmus.SqlConverterBase;
 import io.substrait.isthmus.SubstraitTypeSystem;
 import io.substrait.isthmus.Utils;
 import io.substrait.isthmus.calcite.SubstraitTable;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
-import org.apache.calcite.avatica.util.Casing;
-import org.apache.calcite.config.CalciteConnectionConfig;
-import org.apache.calcite.config.CalciteConnectionProperty;
 import org.apache.calcite.jdbc.CalciteSchema;
-import org.apache.calcite.jdbc.JavaTypeFactoryImpl;
 import org.apache.calcite.prepare.CalciteCatalogReader;
 import org.apache.calcite.rel.type.RelDataType;
-import org.apache.calcite.rel.type.RelDataTypeFactory;
 import org.apache.calcite.sql.SqlNode;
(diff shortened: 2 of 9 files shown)


isthmus/src/test/java/io/substrait/isthmus/sql/SubstraitCreateStatementParserTest.java
```diff
@@ -0,0 +1,118 @@
...
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+
+import org.apache.calcite.prepare.CalciteCatalogReader;
+import org.apache.calcite.sql.parser.SqlParseException;
+import org.junit.jupiter.api.Test;
+
+class SubstraitCreateStatementParserTest {
+  @Test
+  void testProcessCreateStatementsToCatalogWithSingleCreateTableWithTableNameOnly()
```
bestbeforetoday: The test are already scoped by the class name of SubstraitCreateStatementParserTest. All of them also start `testProcessCreateStatementsToCatalogWith`. It is nicely descriptive but makes the test names very long and harder to read, without adding significant value beyond the enclosing class name. Perhaps they could all be simplified?

```suggestion
  void testSingleCreateTableWithTableNameOnly()
```

isthmus/src/test/java/io/substrait/isthmus/sql/SubstraitCreateStatementParserTest.java
```diff
@@ -0,0 +1,118 @@
...
+      throws SqlParseException {
+    final CalciteCatalogReader catalogReader =
+        SubstraitCreateStatementParser.processCreateStatementsToCatalog(
+            "create table src1 (intcol int, charcol varchar(10))");
+
+    assertEquals(catalogReader.getRootSchema().getTableNames().size(), 1);
```
bestbeforetoday: Nothing functionally wrong with it but `catalogReader.getRootSchema().` gets repeated  in (almost) every assertion throughout the file. It might be worth pulling out as a variable so each assertion is easier to read.

```suggestion
    CalciteSchema schema = catalogReader.getRootSchema();
    assertEquals(schema.getTableNames().size(), 1);
```

isthmus/src/test/java/io/substrait/isthmus/sql/SubstraitCreateStatementParserTest.java
```diff
@@ -0,0 +1,118 @@
...
+
+  @Test
+  void testProcessCreateStatementsToCatalogWithMultipleCreateTableWithTableNameOnly()
+      throws SqlParseException {
+    final CalciteCatalogReader catalogReader =
+        SubstraitCreateStatementParser.processCreateStatementsToCatalog(
+            "create table src1 (intcol int, charcol varchar(10))",
+            "create table src2 (intcol int, charcol varchar(10))");
+
+    assertEquals(catalogReader.getRootSchema().getTableNames().size(), 2);
```
bestbeforetoday: The parameters to `assertEquals` are the wrong way around so failure messages are confusing. A failure also just gives a mismatched number and requires subsequent debugging to see the incorrect content. I suggest comparing content instead, which shows the actual table names present on failure. Same for all other similar assertions in this class.

```suggestion
    CalciteSchema schema = catalogReader.getRootSchema();
    assertEquals(Set.of("src1", "src2"), schema.getTableNames());
```

isthmus/src/main/java/io/substrait/isthmus/Utils.java
```diff
@@ -58,4 +61,31 @@ public static <T> Stream<List<T>> crossProduct(List<List<T>> lists) {
...
+   * @param names the list of hierarchical schema and table names ordered from parent to child with
+   *     the last element containing the table name
+   * @return the final sub schema
+   * @see io.substrait.isthmus.sql.SubstraitCreateStatementParser#processCreateStatementsToSchema
+   * @see io.substrait.isthmus.SchemaCollector#toSchema
+   */
+  public static CalciteSchema createCalciteSchemaFromNames(
+      @NonNull final CalciteSchema rootSchema, @NonNull final List<String> names) {
+    CalciteSchema schema = rootSchema;
+    for (final String schemaName : names.subList(0, names.size() - 1)) {
```
bestbeforetoday: It seems odd as a user that you need to pass in a list with a final element that is always ignored. Is it possible a future implementation might need to use this final element, or would it make more sense to have the caller pass in a list omitting the unnecessary final table name element?

isthmus/src/main/java/io/substrait/isthmus/sql/SubstraitCreateStatementParser.java
```diff
@@ -24,54 +25,59 @@
...
 import org.apache.calcite.sql.validate.SqlConformanceEnum;
 import org.apache.calcite.sql.validate.SqlValidator;
+import org.jspecify.annotations.NonNull;
+import org.jspecify.annotations.Nullable;
 
 /** Utility class for parsing CREATE statements into a {@link CalciteCatalogReader} */
 public class SubstraitCreateStatementParser {
 
-  private static final RelDataTypeFactory TYPE_FACTORY =
+  protected static final RelDataTypeFactory TYPE_FACTORY =
```
bestbeforetoday: This change doesn't seem to require non-private access to these members. If we have existing use-cases for them to be accessible, fine. If not, perhaps we should avoid increasing the public API footprint until it is necessary.

If we do have a concrete use-case for exposing them, would it be better to go ahead and make them public rather then restrict to protected? They are part of the public API either way but public access allows more flexibility in how they are used. For example, if using composition instead of inheritance.

isthmus/src/main/java/io/substrait/isthmus/sql/SubstraitCreateStatementParser.java
```diff
@@ -81,65 +87,125 @@ public static List<SubstraitTable> processCreateStatements(String createStatemen
...
-            // result in parse exceptions. Ignore the constraint declaration.
-            continue;
-          }
+    return tableList;
+  }
 
-          throw fail("Unexpected column list construction.", node.getParserPosition());
-        }
+  /**
+   * Parses one or more SQL strings containing only CREATE statements into a {@link
```
bestbeforetoday, line 77: For `processCreateStatements()`, the JavaDoc now calls out that it only supports simple table names without any additional qualifiers. Should the JavaDoc here call out that is _does_ support qualifiers, or is it safe to assume that unless explicitly mentioned?

isthmus/src/main/java/io/substrait/isthmus/sql/SubstraitCreateStatementParser.java
```diff
@@ -81,65 +87,125 @@ public static List<SubstraitTable> processCreateStatements(String createStatemen
...
-        if (col.name.names.size() != 1) {
-          throw fail("Expected simple column names.", col.name.getParserPosition());
+  /**
+   * Parses one or more SQL strings containing only CREATE statements into a {@link CalciteSchema}.
+   *
+   * @param createStatements a SQL string containing only CREATE statements, must not be null
+   * @return a {@link CalciteSchema} generated from the CREATE statements
+   * @throws SqlParseException
+   */
+  protected static CalciteSchema processCreateStatementsToSchema(
```
bestbeforetoday: If this is intended for end user use, I think it should be public, similar to `processCreateStatementsToCatalog()`. If not, I think private would be a better choice to avoid leaking into the public API.

Same for the other newly added protected methods in this class.

bestbeforetoday (changes requested):
Generally looks good and clearly structured code to me.

Although not functionally incorrect, I think there are some changes that should be made to the unit test code for correctness and usability of failure messages.

Other than that, just some minor questions on API for consideration. Your call on the best approach. See inline comments.

isthmus/src/main/java/io/substrait/isthmus/sql/SubstraitCreateStatementParser.java
nielspardon: we had multiple instances where private visibility made it impossible for end users to customize and we had to go back and change visibility to protected which is why I decided to make these all protected in case someone wants to extend the `SubstraitCreateStatementParser` class and modify it while still reusing some of the existing logic.

isthmus/src/main/java/io/substrait/isthmus/sql/SubstraitCreateStatementParser.java
nielspardon: makes sense. I'll add a comment

isthmus/src/test/java/io/substrait/isthmus/sql/SubstraitCreateStatementParserTest.java
nielspardon: thanks for catching this. I know this but tend to forget about it from time to time. will change it

isthmus/src/test/java/io/substrait/isthmus/sql/SubstraitCreateStatementParserTest.java
nielspardon: I can shorten it to `toCatalog` given that there are two public methods in `SubstraitCreateStatementParser` with `processCreateStatements` and `processCreateStatementsToCatalog` where I'm only testing the latter.

isthmus/src/main/java/io/substrait/isthmus/sql/SubstraitCreateStatementParser.java
nielspardon: I added a comment that this method expects fully qualified names.

isthmus/src/main/java/io/substrait/isthmus/Utils.java
nielspardon: I changed this accordingly moving the `names.subList(0, names.size() - 1)` into the callers of the function and changed the javadocs

isthmus/src/test/java/io/substrait/isthmus/sql/SubstraitCreateStatementParserTest.java
nielspardon: I changed this

isthmus/src/test/java/io/substrait/isthmus/sql/SubstraitCreateStatementParserTest.java
nielspardon: I changed this

isthmus/src/test/java/io/substrait/isthmus/sql/SubstraitCreateStatementParserTest.java
nielspardon: I shortened the method names as indicated to only use the `toCatalog` prefix

isthmus/src/main/java/io/substrait/isthmus/sql/SubstraitCreateStatementParser.java
nielspardon: some of those we actually repeat in multiple locations so it might make sense to standardize

isthmus/src/main/java/io/substrait/isthmus/sql/SubstraitCreateStatementParser.java
nielspardon: I guess since they are `static` this won't help in terms of overwriting them

nielspardon: I'll keep the methods private for now

isthmus/src/main/java/io/substrait/isthmus/sql/SubstraitCreateStatementParser.java
nielspardon: - I moved `TYPE_FACTORY` into `SubstraitTypeSystem` and replaced all occurrences which were not using the constant to use it
- I moved `CONNECTION_CONFIG` into `SqlConverterBase` and replaced all occurrences which were not using the constant to use it
- I removed `PARSER_CONFIG` and I'm now using `SubstraitSqlStatementParser.parseStatements()` instead of creating our own parser
- I made `EMPTY_CATALOG` and `VALIDATOR` public in case someone wants to reuse these two constants

bestbeforetoday (approved):
I still think that, rather than comparing size of sets to an expected value, it would be worthwhile comparing the actual values to an expected set. Instead of failures like:

> expected: <1> but was: <0>

this would give more helpful failures like:

> expected: <[src1]> but was: <[]>

It would also remove the need for some of the follow-on `assertNotNull` checks that just check the set contains the expected values.

Provided nobody modifies the code in a way that breaks the tests and needs to debug the failure, it's fine though.

Besides that one nit, it's a really nice improvement. 👍

nielspardon:
I changed the tests to assert equality with sets of schema / table names

PR #576, opened 2025-10-28
fix(spark): add versions for jackson dependencies

bestbeforetoday:
The spark module had unversioned dependencies on a bundle of Jackson modules. This caused publishing to Maven Central to break.

This change adds a Jackson BOM dependency to the spark module to provide version information.


nielspardon (approved):
LGTM

PR #579, opened 2025-10-28
feat: support Nested Structs

gord02:
This PR is to change the substrait-java virtualTableScan to use nested structs in place of the now deprecated struct literals. 
 
The [issue](https://github.com/substrait-io/substrait-java/issues/587)

Key changes:
- Add the visitation pattern logic  
- Proto to Rel conversion 
- Rel to proto conversion 
- Spark to Substrait 
- Substrait to Spark
- SubstraitRelNodeConverter: substrait virtualTableScan to calcite logicalValues, required creating a project for non-literal expressions 
- SubstraitRelVisitor: Calcite to substrait conversion that removes the extra project added to account for non-literal expressions 

Testing:
- LogicalValuesTest tested the conversion from pojo to calcite node back to pojo 
- VirtualTableScanTest tests the creation of a VirtualTableScan using nested struct, and the equivalence of the pojo objects created from the proto VirtualTableScan set using the deprecated values, to the new fields attribute
- VirtualTableScanTest also validates that VirtualTableScan proto rows cannot be set using both values and fields

Testing:
`./gradlew test --tests io.substrait.isthmus.LogicalValuesTest  --debug-jvm`

### DIFF
--- a/core/src/main/java/io/substrait/relation/ProtoRelConverter.java
+++ b/core/src/main/java/io/substrait/relation/ProtoRelConverter.java
@@ -605,7 +605,7 @@ public class ProtoRelConverter {
    * @param structs the StructLiteral instances to convert
    * @return a list of NestedStruct instances with the same field structure
    */
-  public static List<Expression.NestedStruct> nestedStruct(
+  private static List<Expression.NestedStruct> nestedStruct(
       boolean nullable, Expression.StructLiteral... structs) {
     List<Expression.NestedStruct> nestedStructs = new ArrayList<>();
     for (Expression.StructLiteral struct : structs) {
--- a/core/src/test/java/io/substrait/relation/VirtualTableScanTest.java
+++ b/core/src/test/java/io/substrait/relation/VirtualTableScanTest.java
@@ -28,16 +28,12 @@ import org.junit.jupiter.api.Test;
 
 class VirtualTableScanTest extends TestBase {
 
-  io.substrait.proto.Expression expression =
-      io.substrait.proto.Expression.newBuilder()
-          .setLiteral(io.substrait.proto.Expression.Literal.newBuilder().setI32(3))
-          .build();
-
   io.substrait.proto.Expression.Literal literal =
       io.substrait.proto.Expression.Literal.newBuilder().setI32(3).build();
+  io.substrait.proto.Expression expression =
+      io.substrait.proto.Expression.newBuilder().setLiteral(literal).build();
 
   Type type = Type.newBuilder().setI32(Type.I32.newBuilder().build()).build();
-
   io.substrait.proto.NamedStruct schema =
       io.substrait.proto.NamedStruct.newBuilder()
           .setStruct(Type.Struct.newBuilder().addTypes(type).build())
@@ -87,16 +83,20 @@ class VirtualTableScanTest extends TestBase {
   }
 
   @Test
-  void valuesAndFieldsComparisonTest() {
-    io.substrait.proto.Rel valuesStructProto = protoRelVirtualTableValues();
-    io.substrait.proto.Rel fieldsStructProto = protoRelVirtualTableFields();
-    Rel relWithValues = protoRelConverter.from(valuesStructProto);
-    Rel relWithFields = protoRelConverter.from(fieldsStructProto);
-    assertEquals(relWithValues, relWithFields);
+  void virtualTableUsingValuesShouldBeConvertedToFields() {
+    // Create 2 virtual tables with the same contents, but different encodings (values vs fields)
+    io.substrait.proto.Rel virtualTableWithValuesSet = virtualTableUsingValues(schema, literal);
+    io.substrait.proto.Rel virtualTableWithFieldsSet = virtualTableUsingFields(schema, expression);
+
+    Rel rel1 = protoRelConverter.from(virtualTableWithFieldsSet);
+    Rel rel2 = protoRelConverter.from(virtualTableWithValuesSet);
+
+    // The relations should be equivalent
+    assertEquals(rel1, rel2);
   }
 
   @Test
-  void setUsingOnlyValuesOrFieldsTest() {
+  void rejectVirtualTablesWhenBothValuesAndFieldsAreSet() {
     io.substrait.proto.Expression.Literal.Struct literalStruct =
         io.substrait.proto.Expression.Literal.Struct.newBuilder().addFields(literal).build();
     io.substrait.proto.Expression.Nested.Struct nestedStruct =
@@ -112,13 +112,12 @@ class VirtualTableScanTest extends TestBase {
             .setBaseSchema(schema)
             .build();
 
-    io.substrait.proto.Rel valuesAndFieldsProto =
-        io.substrait.proto.Rel.newBuilder().setRead(readRel).build();
-    assertThrows(
-        IllegalArgumentException.class, () -> protoRelConverter.from(valuesAndFieldsProto));
+    io.substrait.proto.Rel rel = io.substrait.proto.Rel.newBuilder().setRead(readRel).build();
+    assertThrows(IllegalArgumentException.class, () -> protoRelConverter.from(rel));
   }
 
-  io.substrait.proto.Rel protoRelVirtualTableFields() {
+  io.substrait.proto.Rel virtualTableUsingFields(
+      io.substrait.proto.NamedStruct schema, io.substrait.proto.Expression expression) {
     io.substrait.proto.Expression.Nested.Struct struct =
         io.substrait.proto.Expression.Nested.Struct.newBuilder().addFields(expression).build();
 
@@ -131,7 +130,8 @@ class VirtualTableScanTest extends TestBase {
     return io.substrait.proto.Rel.newBuilder().setRead(readRel).build();
   }
 
-  io.substrait.proto.Rel protoRelVirtualTableValues() {
+  io.substrait.proto.Rel virtualTableUsingValues(
+      io.substrait.proto.NamedStruct schema, io.substrait.proto.Expression.Literal value) {
     io.substrait.proto.Expression.Literal.Struct struct =
         io.substrait.proto.Expression.Literal.Struct.newBuilder().addFields(literal).build();
 
@@ -145,7 +145,7 @@ class VirtualTableScanTest extends TestBase {
   }
 
   @Test
-  void notNullableVirtualTable() {
+  void rejectInvalidRowNullability() {
     ImmutableVirtualTableScan.Builder bldr =
         VirtualTableScan.builder()
             .initialSchema(
--- a/isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
@@ -676,31 +676,44 @@ public class SubstraitRelNodeConverter
       }
       return LogicalValues.create(relBuilder.getCluster(), rowTypeWithNames, tuplesBuilder.build());
     } else {
-      // the virtualTable row is represented by the expressions in the logicalProject
-      List<RelNode> projectRow = new ArrayList<>();
+      // When a VirtualTable contains non-literal expressions, they cannot be put directly into a
+      // LogicalValues relation. Instead, we create a LogicalProject for each row to compute its
+      // values, and combine them together using a LogicalUnion. For example the following:
+      //
+      //   VirtualTable
+      //     (e1, e2)
+      //     (e3, e4)
+      //
+      //  Becomes:
+      //
+      //   LogicalUnion(all=[true])
+      //     LogicalProject(exprs=[e1, e2])
+      //       <Empty Row>
+      //     LogicalProject(exprs=[e3, e4])
+      //       <Empty Row>
+      //
 
       RelDataType emptyRowType = typeFactory.createStructType(List.of(), List.of());
-      ImmutableList<ImmutableList<RexLiteral>> emptyValues =
-          ImmutableList.of(ImmutableList.copyOf(new ArrayList<>())); // dummy Row
+      ImmutableList<ImmutableList<RexLiteral>> emptyRowValue = ImmutableList.of(ImmutableList.of());
 
+      List<RelNode> projects = new ArrayList<>();
       for (final Expression.NestedStruct rowExpr : virtualTableScan.getRows()) {
         List<RexNode> rexRow = new ArrayList<>();
         for (Expression field : rowExpr.fields()) {
           rexRow.add(field.accept(expressionRexConverter, context));
         }
-        RelNode values = LogicalValues.create(relBuilder.getCluster(), emptyRowType, emptyValues);
+        RelNode values = LogicalValues.create(relBuilder.getCluster(), emptyRowType, emptyRowValue);
         RelNode project =
             LogicalProject.create(
                 values, Collections.emptyList(), rexRow, rowType, Collections.emptySet());
-        projectRow.add(project);
+        projects.add(project);
       }
-      RelNode union =
-          LogicalUnion.create(
-              projectRow, true); // rows are logicalProjects and the table is union of all rows
+      RelNode union = LogicalUnion.create(projects, true);
 
+      // Apply a final LogicalProject on top to capture the field names from the VirtualTable
       List<RexNode> topProjectExprs = new ArrayList<>();
       for (int i = 0; i < rowType.getFieldCount(); i++) {
-        topProjectExprs.add(rexBuilder.makeInputRef(rowType.getFieldList().get(i).getType(), i));
+        topProjectExprs.add(rexBuilder.makeInputRef(union, i));
       }
       return LogicalProject.create(
           union,
--- a/isthmus/src/main/java/io/substrait/isthmus/SubstraitRelVisitor.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/SubstraitRelVisitor.java
@@ -56,10 +56,6 @@ import org.apache.calcite.rel.RelRoot;
 import org.apache.calcite.rel.core.AggregateCall;
 import org.apache.calcite.rel.core.JoinRelType;
 import org.apache.calcite.rel.core.TableModify;
-import org.apache.calcite.rel.core.Union;
-import org.apache.calcite.rel.core.Values;
-import org.apache.calcite.rel.logical.LogicalProject;
-import org.apache.calcite.rel.logical.LogicalUnion;
 import org.apache.calcite.rel.type.RelDataType;
 import org.apache.calcite.rel.type.RelDataTypeFactory;
 import org.apache.calcite.rex.RexBuilder;
@@ -174,46 +170,8 @@ public class SubstraitRelVisitor extends RelNodeVisitor<Rel, RuntimeException> {
     return super.visit(calc);
   }
 
-  /*
-       we represent a logicalValue table that can store non-literal expressions by a LogicalProject
-        with an input of a logicalUnion with inputs of logicalProjects. Each of the unioned logicalProjects
-        store the expressions in row of the table, and each is passed a dummy logicalValues input
-  */
-  boolean isLogicalValues(LogicalProject project) {
-    RelNode node = project.getInput(0);
-    if (node instanceof LogicalUnion && !node.getInputs().isEmpty()) {
-      for (RelNode child : node.getInputs()) {
-        //          each logicalProject has a dummy logicalValues input
-        if (!(child instanceof LogicalProject && child.getInputs().get(0) instanceof Values)) {
-          return false;
-        }
-      }
-    } else {
-      return false;
-    }
-    return true;
-  }
-
   @Override
   public Rel visit(org.apache.calcite.rel.core.Project project) {
-    NamedStruct type = typeConverter.toNamedStruct(project.getRowType());
-    if (isLogicalValues((LogicalProject) project)) {
-      List<Expression.NestedStruct> structs = new ArrayList<>();
-      Union union = (LogicalUnion) project.getInputs().get(0);
-      // each node is logicalProject storing all expressions of table row
-      for (RelNode node : union.getInputs()) {
-        LogicalProject projectRow = (LogicalProject) node;
-        structs.add(
-            Expression.NestedStruct.builder()
-                .addAllFields(
-                    projectRow.getProjects().stream()
-                        .map(this::toExpression)
-                        .collect(Collectors.toUnmodifiableList()))
-                .build());
-      }
-      return VirtualTableScan.builder().initialSchema(type).addAllRows(structs).build();
-    }
-
     List<Expression> expressions =
         project.getProjects().stream()
             .map(this::toExpression)
--- a/isthmus/src/test/java/io/substrait/isthmus/VirtualTableScanTest.java
+++ b/isthmus/src/test/java/io/substrait/isthmus/VirtualTableScanTest.java
@@ -61,9 +61,6 @@ class VirtualTableScanTest extends PlanTestBase {
             + "    LogicalProject(exprs=[[*(6, 2), 8.8E0:DOUBLE]])\n"
             + "      LogicalValues(type=[RecordType()], tuples=[[{  }]])\n",
         explain(relNode));
-
-    // Check full roundtrip conversion
-    assertFullRoundTrip(virtualTableScan);
   }
 
   @SafeVarargs
--- a/spark/src/main/scala/io/substrait/spark/logical/ToLogicalPlan.scala
+++ b/spark/src/main/scala/io/substrait/spark/logical/ToLogicalPlan.scala
@@ -367,14 +367,11 @@ class ToLogicalPlan(spark: SparkSession = SparkSession.builder().getOrCreate())
       virtualTableScan: relation.VirtualTableScan,
       context: EmptyVisitationContext): LogicalPlan = {
     val rows = virtualTableScan.getRows.asScala.map {
-      case nestedStruct: NestedStruct =>
+      nestedStruct =>
         InternalRow.fromSeq(
           nestedStruct.fields.asScala
             .map(expr => expr.accept(expressionConverter, context).asInstanceOf[Literal].value)
         )
-      case other =>
-        throw new UnsupportedOperationException(
-          s"Unsupported row type in VirtualTableScan: ${other.getClass}")
     }
     virtualTableScan.getInitialSchema match {
       case ns: NamedStruct if ns.names().isEmpty && rows.length == 1 =>


nielspardon:
Does this PR replace #417 which seems to have tried to do a similar change?

benbellick:
I created an issue to track this particular problem @ #587. @gord02 would you mind including it somewhere in your PR description so that it gets linked?

core/src/main/java/io/substrait/expression/ExpressionCreator.java
```diff
@@ -286,6 +287,24 @@ public static Expression.StructLiteral struct(
     return Expression.StructLiteral.builder().nullable(nullable).addAllFields(values).build();
   }
 
+  public static Expression.StructNested struct(Iterable<? extends Expression> values) {
+    if (values == null) {
+      return Expression.StructNested.builder().build();
+    }
+    return Expression.StructNested.builder().addAllFields(values).build();
```
benbellick: Is this actually different from just doing
```suggestion
    return Expression.StructNested.builder().addAllFields(values).build();
```
?
I am genuinely not sure.

core/src/test/java/io/substrait/type/proto/DdlRelRoundtripTest.java
```diff
@@ -46,31 +44,21 @@ void alter() {
...
-        ExtensionDdl.builder()
-            .viewDefinition(virtTable)
-            .tableSchema(schema)
-            .tableDefaults(defaults)
-            .detail(detail)
-            .operation(ExtensionDdl.DdlOp.ALTER)
-            .object(ExtensionDdl.DdlObject.VIEW)
-            .build();
-
-    io.substrait.proto.Rel protoRel = relProtoConverter.toProto(command);
```
benbellick, line 72: This test has become a test for virtual tables instead of the intended `Ddl` test. You can see that the virtual table was just fed in as an input the `ExtensionDdl` builder as a `viewDefinition`. Were these changes actually necessary?

core/src/main/java/io/substrait/expression/ExpressionCreator.java
```diff
@@ -286,6 +287,24 @@ public static Expression.StructLiteral struct(
...
+  }
+
+  public static List<Expression.StructNested> toNested(
+      Iterable<? extends Expression.StructLiteral> structLiterals) {
```
benbellick:
```suggestion
      Iterable<Expression.StructLiteral> structLiterals) {
```

benbellick: But also looks like this function isn't used anywhere. Do we need to have it?

core/src/main/java/io/substrait/relation/ProtoRelConverter.java
```diff
@@ -581,17 +582,33 @@ protected FileOrFiles newFileOrFiles(ReadRel.LocalFiles.FileOrFiles file) {
 
   protected VirtualTableScan newVirtualTable(ReadRel rel) {
     ReadRel.VirtualTable virtualTable = rel.getVirtualTable();
+    // If both values and expressions are set, raise an error
+    if (virtualTable.getValuesCount() > 0 && virtualTable.getExpressionsCount() > 0) {
```
benbellick, line 624: Just wanted to say, I thought it might make more sense to do a more a clever migration strategy, but it looks like `substrait-go` [handles it the same way](https://github.com/substrait-io/substrait-go/blob/89ee28134213c81559d72fe9b239b109c0b91c44/plan/plan.go#L370), so for consistency, I am okay with just enforcing only one will be present 👍

core/src/main/java/io/substrait/expression/Expression.java
```diff
@@ -662,6 +662,30 @@ public <R, C extends VisitationContext, E extends Throwable> R accept(
     }
   }
 
+  @Value.Immutable
+  abstract class StructNested implements Expression {
```
benbellick: `StructNested` is just one of the `oneof` options in message `Nested`. Could we instead follow the pattern of `Literal` above and construct an interface called `Nested`, and then have `StructNested` implement the `Nested` interface? Then we can also include nullability as a field of the `Nested` interface to more closely align with the protos. Thoughts?

core/src/main/java/io/substrait/expression/ExpressionCreator.java
```diff
@@ -286,6 +287,24 @@ public static Expression.StructLiteral struct(
     return Expression.StructLiteral.builder().nullable(nullable).addAllFields(values).build();
   }
 
+  public static Expression.StructNested struct(Iterable<? extends Expression> values) {
```
benbellick:
```suggestion
  public static Expression.StructNested struct(Iterable<Expression> fields) {
```

Lets call it `fields` [to be consistent with the proto](https://github.com/substrait-io/substrait/blob/main/proto/substrait/algebra.proto#L1173) (`values` above is used for the List message).

core/src/main/java/io/substrait/expression/ExpressionCreator.java
benbellick: Also as discussed in the previous comment, I think that there should be a place to add nullability considering it is part of the type.

core/src/main/java/io/substrait/relation/ProtoRelConverter.java
```diff
@@ -581,17 +582,33 @@ protected FileOrFiles newFileOrFiles(ReadRel.LocalFiles.FileOrFiles file) {
...
+      throw new IllegalArgumentException(
+          "Virtual table cannot have both values and expressions set");
+    }
+
     NamedStruct virtualTableSchema = newNamedStruct(rel);
+
     ProtoExpressionConverter converter =
         new ProtoExpressionConverter(lookup, extensions, virtualTableSchema.struct(), this);
-    List<Expression.StructLiteral> structLiterals = new ArrayList<>(virtualTable.getValuesCount());
+
```
benbellick, line 631: instead of handling below the converion of proto -> expression for `StructLiteral`s and `StrutNested`'s inline, could we introduce method to handle conversions of `Nested` in `ProtoExpressionConverter.java` (once you do the above of having a `Nested` interface)? [Similar to how it is handled for Literals. ](https://github.com/substrait-io/substrait-java/blob/4b8366675108d506fd29f1cd355bd27712cafe73/core/src/main/java/io/substrait/expression/proto/ProtoExpressionConverter.java#L364)

core/src/main/java/io/substrait/relation/VirtualTableScan.java
```diff
@@ -5,12 +5,13 @@
...
 import io.substrait.util.VisitationContext;
 import java.util.List;
+import java.util.Objects;
 import org.immutables.value.Value;
 
 @Value.Immutable
 public abstract class VirtualTableScan extends AbstractReadRel {
 
-  public abstract List<Expression.StructLiteral> getRows();
+  public abstract List<Expression> getRows();
```
benbellick:
```suggestion
  public abstract List<Expression.StructNested> getRows();
```
Doesn't that make more sense if we are transferring to the newer type? The idea then is that internally, regardless of if the initial proto had `StructLiteral`s or `StructNested`s, we treat them all as `StructNested`s regardless. I think this will also simplify your `RelProtoConverter.java` implementation.

core/src/main/java/io/substrait/relation/VirtualTableScan.java
```diff
@@ -28,13 +29,13 @@ protected void check() {
...
-    assert rows.size() > 0
-        && names.stream().noneMatch(s -> s == null)
-        && rows.stream().noneMatch(r -> r == null)
-        && rows.stream()
-            .allMatch(r -> NamedFieldCountingTypeVisitor.countNames(r.getType()) == names.size());
+    assert !rows.isEmpty();
+    assert names.stream().noneMatch(Objects::isNull);
+    assert rows.stream().noneMatch(Objects::isNull);
+    assert rows.stream()
+        .allMatch(r -> NamedFieldCountingTypeVisitor.countNames(r.getType()) == names.size());
```
benbellick: Did these lines have to change? If not, I would rather keep them the same as they were before so that the diff is smaller. If you think it's an improvement, we can iterate on it in another PR :)

core/src/main/java/io/substrait/expression/ExpressionCreator.java
```diff
@@ -286,6 +286,19 @@ public static Expression.StructLiteral struct(
     return Expression.StructLiteral.builder().nullable(nullable).addAllFields(values).build();
   }
 
+  public static Expression.StructNested nestedStruct(
+      boolean nullable, Iterable<Expression.Literal> fields) {
```
benbellick:
```suggestion
      boolean nullable, Iterable<Expression> fields) {
```
The `fields` param of a `Nested.Struct` takes in an `Expression`, rather than just specifically a `Literal`. 
```proto
    message Struct {
      // Zero or more possibly heterogeneously-typed list of expressions that
      // form the struct fields.
      repeated Expression fields = 1;
    }
```

Is it possible to model the same here?

core/src/main/java/io/substrait/expression/ExpressionCreator.java
```diff
@@ -286,6 +286,19 @@ public static Expression.StructLiteral struct(
...
+    return Expression.StructNested.builder().nullable(nullable).addAllFields(fields).build();
+  }
+
+  public static Expression.StructNested nestedStruct(
+      boolean nullable, Expression.Literal... fields) {
```
benbellick: Same thing here
```suggestion
      boolean nullable, Expression... fields) {
```

core/src/main/java/io/substrait/expression/Expression.java
```diff
@@ -662,6 +669,30 @@ public <R, C extends VisitationContext, E extends Throwable> R accept(
     }
   }
 
+  @Value.Immutable
+  abstract class StructNested implements Nested {
+    public abstract List<Expression> fields();
+
+    @Override
+    public Type getType() {
+      return Type.withNullability(false)
```
benbellick:
```suggestion
      return Type.withNullability(nullable())
```
We should respect the nullability instead of dropping that info.

core/src/main/java/io/substrait/expression/proto/ExpressionProtoConverter.java
```diff
@@ -357,6 +363,17 @@ public Expression visit(
...
   }
 
+  @Override
+  public Expression visit(
+      io.substrait.expression.Expression.StructNested expr, EmptyVisitationContext context) {
+    return nested(
+        bldr -> {
+          List<Expression> values =
+              expr.fields().stream().map(this::toProto).collect(Collectors.toList());
+          bldr.setStruct(Expression.Nested.Struct.newBuilder().addAllFields(values));
```
benbellick: Are we losing nullability here?

core/src/main/java/io/substrait/relation/ExpressionCopyOnWriteVisitor.java
```diff
@@ -203,6 +203,15 @@ public Optional<Expression> visit(Expression.StructLiteral expr, EmptyVisitation
     return visitLiteral(expr);
   }
 
+  @Override
+  public Optional<Expression> visit(Expression.StructNested expr, EmptyVisitationContext context)
+      throws E {
+    Optional<List<Expression>> expressions = visitExprList(expr.fields(), context);
+    return expressions.map(
+        expressionList ->
+            Expression.StructNested.builder().from(expr).fields(expressionList).build());
```
benbellick: Again, are we losing nullability here?

core/src/main/java/io/substrait/relation/ProtoRelConverter.java
```diff
@@ -581,18 +581,26 @@ protected FileOrFiles newFileOrFiles(ReadRel.LocalFiles.FileOrFiles file) {
...
-              .fields(
-                  struct.getFieldsList().stream()
-                      .map(converter::from)
-                      .collect(java.util.stream.Collectors.toList()))
-              .build());
+      expressions.add(converter.from(struct));
+    }
+
+    for (io.substrait.proto.Expression.Nested.Struct expr : virtualTable.getExpressionsList()) {
+      expressions.add(converter.from(expr));
```
benbellick, line 642: This looks great now :)

core/src/test/java/io/substrait/relation/VirtualTableScanTest.java
```diff
@@ -37,13 +77,27 @@ void check() {
...
+                        R.I64,
+                        R.FP32,
+                        R.FP64,
                         R.STRING,
                         R.struct(R.STRING, R.STRING),
                         R.list(R.struct(R.STRING)),
                         R.map(R.struct(R.STRING), R.struct(R.STRING)))))
             .addRows(
-                struct(
+                ExpressionCreator.nestedStruct(
```
benbellick, line 68: Are all of the changes to this test (besides this line) strictly necessary? It would be preferable to keep the diff of the PR as small as possible unless you are introducing a test which specifically tests new functionality.

core/src/test/java/io/substrait/type/proto/ReadRelRoundtripTest.java
benbellick, line 1: It would be great to see a file like this added but `VirtualTableRelRoundTripTest.java`.

core/src/test/java/io/substrait/relation/VirtualTableScanTest.java
```diff
@@ -59,6 +113,79 @@ void check() {
...
   }
 
+  @Test
+  void valuesAndFieldsComparisonTest() {
+    io.substrait.proto.Rel valuesStructProto = protoRelVTValues();
+    io.substrait.proto.Rel fieldsStructProto = protoRelVTFields();
+    Rel relWithValues = protoRelConverter.from(valuesStructProto);
+    Rel relWithFields = protoRelConverter.from(fieldsStructProto);
+    assertEquals(relWithValues, relWithFields);
+  }
```
benbellick, line 96: Great test!

core/src/test/java/io/substrait/relation/VirtualTableScanTest.java
```diff
@@ -59,6 +113,79 @@ void check() {
...
+
+  @Test
+  void setUsingValuesOrFieldsTest() {
+    io.substrait.proto.Rel valuesAndFieldsProto = setBothValuesAndFields();
```
benbellick: Can you actually just define this value inline of in a function right below? I think it will make the test easier to read.

core/src/test/java/io/substrait/relation/VirtualTableScanTest.java
```diff
@@ -59,6 +113,79 @@ void check() {
...
+                    .addExpressions(nestedStruct)
+                    .addValues(literalStruct)
+                    .build())
+            .setBaseSchema(schema)
+            .build();
+
+    return io.substrait.proto.Rel.newBuilder().setRead(readRel).build();
+  }
+
+  io.substrait.proto.Rel protoRelVTFields() {
```
benbellick: Perhaps we can expand VT out to VirtualTable so that it is a bit clearer to readers?

core/src/test/java/io/substrait/relation/VirtualTableScanTest.java
```diff
@@ -59,6 +113,79 @@ void check() {
...
+    io.substrait.proto.ReadRel readRel =
+        ReadRel.newBuilder()
+            .setVirtualTable(ReadRel.VirtualTable.newBuilder().addExpressions(struct).build())
+            .setBaseSchema(schema)
+            .build();
+
+    return io.substrait.proto.Rel.newBuilder().setRead(readRel).build();
+  }
+
+  io.substrait.proto.Rel protoRelVTValues() {
```
benbellick: and here as well?

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
benbellick, line 1: To be honest, I'm not as familiar with this part of the codebase. I'll take a more thorough look in the future, but could be good to just get core good first.

core/src/test/java/io/substrait/type/proto/ReadRelRoundtripTest.java
gord02: There is a test case for virtualTable in the readRelRoundtripTest file that uses the updated StructNested. Is your suggestion to move the testing logic I have in the logicalValuesTest and the VirtualTableScanTest to this new file? There is a roundtrip done in the logicalValuesTest as well.

core/src/main/java/io/substrait/expression/proto/ProtoExpressionConverter.java
```diff
@@ -502,6 +502,18 @@ public Expression.Literal from(io.substrait.proto.Expression.Literal literal) {
     }
   }
 
+  public Expression.StructNested from(io.substrait.proto.Expression.Literal.Struct struct) {
+    return Expression.StructNested.builder()
+        .fields(struct.getFieldsList().stream().map(this::from).collect(Collectors.toList()))
```
benbellick, line 507: Is this a placewhere we need to pass in nullable-ness too?

core/src/main/java/io/substrait/expression/proto/ProtoExpressionConverter.java
```diff
@@ -502,6 +502,18 @@ public Expression.Literal from(io.substrait.proto.Expression.Literal literal) {
...
+        .build();
+  }
+
+  public Expression.StructNested from(io.substrait.proto.Expression.Nested.Struct struct) {
+    return Expression.StructNested.builder()
```
benbellick: and here?

core/src/main/java/io/substrait/expression/proto/ExpressionProtoConverter.java
```diff
@@ -75,6 +75,13 @@ private Expression lit(Consumer<Expression.Literal.Builder> consumer) {
     return Expression.newBuilder().setLiteral(builder).build();
   }
 
+  private Expression nested(Consumer<Expression.Nested.Builder> consumer) {
+    Expression.Nested.Builder builder = Expression.Nested.newBuilder();
+    builder.setNullable(builder.getNullable());
```
benbellick: Since builder was just newly constructed in the line above, this line doesn't do anything accept reset to value to just again be the default
```suggestion
```

core/src/main/java/io/substrait/relation/ExpressionCopyOnWriteVisitor.java
```diff
@@ -203,6 +203,19 @@ public Optional<Expression> visit(Expression.StructLiteral expr, EmptyVisitation
...
+  public Optional<Expression> visit(Expression.StructNested expr, EmptyVisitationContext context)
+      throws E {
+    Optional<List<Expression>> expressions = visitExprList(expr.fields(), context);
+    return expressions.map(
+        expressionList ->
+            Expression.StructNested.builder()
+                .from(expr)
+                .nullable(expr.nullable())
+                .fields(expressionList)
+                .build());
```
benbellick: I'm not sure here, is it necessary to actually have `.from(expr)`? Is it just being overwritten by the next two lines?

core/src/main/java/io/substrait/relation/RelProtoConverter.java
```diff
@@ -646,17 +647,35 @@ public Rel visit(Cross cross, EmptyVisitationContext context) throws RuntimeExce
...
+        Expression.StructLiteral expression = (Expression.StructLiteral) row;
+        structs.add(
+            io.substrait.proto.Expression.Nested.Struct.newBuilder()
+                .addAllFields(
+                    expression.fields().stream()
+                        .map(this::toProto)
+                        .collect(java.util.stream.Collectors.toList()))
+                .build());
+      }
+    }
```
benbellick, line 762: Is all of this logic above conditioned on type (`StructNested` vs. `StructLiteral`) actually necessary? `VirtualTableScan`'s rows always have type `StructNested` as seen [here](https://github.com/substrait-io/substrait-java/pull/579/files#diff-3de3348ee4a6b041dde42198bd9749e541025e474b62be6a5e307b80df813da9R14).

Considering this, I think you can just do the logic assuming that everything is a `StructNested`.

core/src/test/java/io/substrait/relation/VirtualTableScanTest.java
```diff
@@ -59,6 +85,76 @@ void check() {
...
+    io.substrait.proto.Expression.Literal.Struct literalStruct =
+        io.substrait.proto.Expression.Literal.Struct.newBuilder()
+            .addFields(literal)
+            .addFields(literal)
+            .build();
+    io.substrait.proto.Expression.Nested.Struct nestedStruct =
+        io.substrait.proto.Expression.Nested.Struct.newBuilder()
+            .addFields(expression)
+            .addFields(expression)
+            .build();
```
benbellick:
```suggestion
        io.substrait.proto.Expression.Literal.Struct.newBuilder()
            .addFields(literal)
            .build();
    io.substrait.proto.Expression.Nested.Struct nestedStruct =
        io.substrait.proto.Expression.Nested.Struct.newBuilder()
            .addFields(expression)
            .build();
```

nit: unless adding them twice tests something specific, I suggest dropping those two so that the intention of the test is a bit clearer.

core/src/test/java/io/substrait/relation/VirtualTableScanTest.java
```diff
@@ -59,6 +85,76 @@ void check() {
...
+
+    return io.substrait.proto.Rel.newBuilder().setRead(readRel).build();
+  }
+
+  io.substrait.proto.Rel protoRelVirtualTableValues() {
+    io.substrait.proto.Expression.Literal.Struct struct =
+        io.substrait.proto.Expression.Literal.Struct.newBuilder()
+            .addFields(literal)
+            .addFields(literal)
+            .build();
```
benbellick: Similar story in `protoRelVirtualTableFields` and `protoRelVirtualTableValues`, where I'm unsure if the double `addFields` is necessary.

benbellick: Okay I tried altering this test by only having one call each of the `addFields` and the test fails. Is this expected behavior? If so, could you explain in a comment why this is the case? If not, then might be something in the implementation that needs to be fixed.

core/src/test/java/io/substrait/type/proto/ReadRelRoundtripTest.java
benbellick: ~~Ah good catch, there is indeed a test in `ReadRelRoundtripTest.java`. I personally think that test should be moved to a file called `VirtualTableRelRoundtripTest.java` since it doesn't actually have to do with `ReadRel` and is instead its own thing. I realize you didn't put it there, but may be a good thing to do in this PR anyways if you don't mind.~~

**EDIT: I misspoke above. Forgot that `VirtualTable` is encoded as just a read_type in a `ReadRel`, so I do actually think `ReadRelRoundtripTest.java` is the right place for it.** 

Furthermore, considering that we have been having a lot of discussion about appropriately passing null values around, could we introduce another roundtrip test for virtual tables which also ensure that nullability is preserved? The current test only tests the nullability-is-false case. What about when nullability is true?

benbellick (changes requested):
Okay got the chance to do another pass. I think I'm just going to review the core section for now until I think its in a good place, and then I'll continue onwards with the review in isthmus / spark if that works for you! Thanks again for your work 🚀

core/src/test/java/io/substrait/relation/VirtualTableScanTest.java
gord02: The issue was likely caused by the schema not being updated along with the virtualTable. The defined schema needs to match the entries in the table( number of columns in both the schema and in each row in the table itself)

core/src/main/java/io/substrait/relation/ExpressionCopyOnWriteVisitor.java
gord02: It is the .fields(expressionList) which is redundant. Good catch!

core/src/test/java/io/substrait/relation/VirtualTableScanTest.java
benbellick: Ah yeah true 🤦

core/src/main/java/io/substrait/expression/proto/ProtoExpressionConverter.java
benbellick: I misspoke! This is fine. Nullable-ness is a property of `io.substrait.proto.Expression.Literal`, not of `io.substrait.proto.Expression.Literal.Struct`.

core/src/main/java/io/substrait/expression/proto/ProtoExpressionConverter.java
benbellick: Same here, my bad!

core/src/main/java/io/substrait/expression/Expression.java
```diff
@@ -662,6 +669,30 @@ public <R, C extends VisitationContext, E extends Throwable> R accept(
     }
   }
 
+  @Value.Immutable
+  abstract class StructNested implements Nested {
```
benbellick: So I think this is great, and consistent with everything else in the repo. We have this class `StructNested` which is really representing a special instance of `Nested`, which is a special instance of `Expression`. So `StructNested < Nested < Expression`. We need this representation to fairly pass around `Nested` as an instance of the `Expression` proto as seen [here](https://github.com/substrait-io/substrait/blob/996ffb94af7af017701873e0820cddda83a4531d/proto/substrait/algebra.proto#L974)

On the other hand, we want to use a version of `StructNested` in virtual tables which is _not_ an instance of expression. As seen [here](https://github.com/substrait-io/substrait/blob/benbellick%2Flambdas/proto/substrait/algebra.proto#L155), the usage of `Expression.Nested.Struct` here is (perhaps confusingly) _not_ a usage of an expression, but is rather direct access to this raw proto which is itself a list of expressions. 


Thus, this class as you have implemented is excellent and captures the one-to-one nature with the protos. But on the other hand, we need a version of Struct Nested which is actually just direct access to the proto, rather than as a special case of an expression. I'm not sure yet what the best way to represent this is.

benbellick: Okay now I think I follow what the problem here is. Yes, I see that as a consequence of using this particular representation, we have to deal with the consequence of treating this thing as an `Expression` when in reality we don't want that. I think that this will be an okay approach, but then I have a comment below which I'll link here.

core/src/main/java/io/substrait/relation/VirtualTableScan.java
```diff
@@ -28,11 +29,11 @@ protected void check() {
...
         == NamedFieldCountingTypeVisitor.countNames(this.getInitialSchema().struct());
-    List<Expression.StructLiteral> rows = getRows();
+    List<Expression.StructNested> rows = getRows();
 
-    assert rows.size() > 0
-        && names.stream().noneMatch(s -> s == null)
-        && rows.stream().noneMatch(r -> r == null)
+    assert !rows.isEmpty()
+        && names.stream().noneMatch(Objects::isNull)
+        && rows.stream().noneMatch(Objects::isNull)
```
benbellick, line 50: Understanding now the issue with the use of `StructNested`, it would be great if right here you add an assertion making sure all of the rows are `!row.nullable()`. And then add a comment and explain this extra context, saying something like: 
```
In the codebase, we use `StructNested` to represent two subtly distinct cases: 
- an instance of the Expression proto with the particular case of `rex_type` set to `Nested` (and then `Struct`)
- access to the raw proto of `Expression.Struct.Nested` itself

Here we are using the second case, and thus we are carrying around a nullable field as a consequence of the first option, but we enforce it to be false to ensure a user doesn't accidentally pass around meaningless additional context. 
```

core/src/main/java/io/substrait/expression/Expression.java
benbellick: https://github.com/substrait-io/substrait-java/pull/579/files#r2528246579

core/src/main/java/io/substrait/relation/VirtualTableScan.java
benbellick: And also adding a small test to demonstrate this would be great.

core/src/main/java/io/substrait/relation/ExpressionCopyOnWriteVisitor.java
benbellick: Okay I think really the cleanest thing here then is to just have `.from` and `.fields`. I understand now that `.from` says "give me all fields from before" and then  `.fields` overwrites it. Can't remember before if I told you to add nullability before 😅 

So I think just the following is sufficient: 
```java
            Expression.StructNested.builder()
                .from(expr)
                .fields(expressionList)
                .build());
```

But honestly it doesn't matter 🤷

core/src/main/java/io/substrait/expression/proto/ProtoExpressionConverter.java
```diff
@@ -502,6 +502,18 @@ public Expression.Literal from(io.substrait.proto.Expression.Literal literal) {
     }
   }
 
+  public Expression.StructNested from(io.substrait.proto.Expression.Literal.Struct struct) {
```
benbellick:
```suggestion
  public Expression.StructLiteral from(io.substrait.proto.Expression.Literal.Struct struct) {
```

So we aren't actually deprecating the usage of `Expression.StructLiteral` in general, but just in the particular case of `VirtualTable`s. Thus, we should preserve the function which converts `io.substrait.proto.Expression.Literal.Struct` to `Expression.StructNested`.

However, we were relying on this implementation [here](https://github.com/substrait-io/substrait-java/pull/579/files#diff-dd48c750d05998223daf061d1d409a5691a8e3c22e93cf68467f393a062b785cR597) to go from `io.substrait.proto.Expression.Literal.Struct` to `Expression.StructNested`. I propose that you add another function somewhere (if it doesn't already exist) which takes converts a `Expression.StructLiteral` to a `Expression.StructNested`. Then in the linked place [here](https://github.com/substrait-io/substrait-java/pull/579/files#diff-dd48c750d05998223daf061d1d409a5691a8e3c22e93cf68467f393a062b785cR597) we would just call that new function before adding it to the list.

core/src/main/java/io/substrait/relation/VirtualTableScan.java
```diff
@@ -28,11 +29,24 @@ protected void check() {
...
-
-    assert rows.size() > 0
-        && names.stream().noneMatch(s -> s == null)
-        && rows.stream().noneMatch(r -> r == null)
+    List<Expression.StructNested> rows = getRows();
+
+    /**
+     * In the codebase, we use `StructNested` to represent two subtly distinct cases: - an instance
+     * of the Expression proto with the particular case of `rex_type` set to `Nested` (and then
+     * `Struct`) - access to the raw proto of `Expression.Struct.Nested` itself
```
benbellick: The formatting seems a bit strange here, can we have each bullet on its own line?

core/src/main/java/io/substrait/expression/ExpressionCreator.java
```diff
@@ -286,6 +287,32 @@ public static Expression.StructLiteral struct(
...
+  }
+
+  public static Expression.StructNested nestedStruct(boolean nullable, Expression... fields) {
+    return Expression.StructNested.builder()
+        .nullable(nullable)
+        .addAllFields(Arrays.asList(fields))
+        .build();
+  }
+
+  public static List<Expression.StructNested> nestedStruct(
```
benbellick: Can we throw a comment in here to explain what this function does? I know none of the other functions in here have comments, but this function is actually a bit distinct. The others are direct wrappers around the builder, whereas this one is actually a converter from one type to the other. I would explicitly make note of the fact that it is used to convert a deprecated field for virtual tables to the proper expression-based representation or something like that.

core/src/main/java/io/substrait/relation/ProtoRelConverter.java
```diff
@@ -581,18 +582,24 @@ protected FileOrFiles newFileOrFiles(ReadRel.LocalFiles.FileOrFiles file) {
...
+
     for (io.substrait.proto.Expression.Literal.Struct struct : virtualTable.getValuesList()) {
-      structLiterals.add(
-          Expression.StructLiteral.builder()
-              .fields(
-                  struct.getFieldsList().stream()
-                      .map(converter::from)
-                      .collect(java.util.stream.Collectors.toList()))
-              .build());
+      expressions.addAll(ExpressionCreator.nestedStruct(false, converter.from(struct)));
```
benbellick: nit: add a comment here clarifying why this value is set to false for future readers.

core/src/test/java/io/substrait/relation/VirtualTableScanTest.java
```diff
@@ -59,6 +84,64 @@ void check() {
...
+  void valuesAndFieldsComparisonTest() {
+    io.substrait.proto.Rel valuesStructProto = protoRelVirtualTableValues();
+    io.substrait.proto.Rel fieldsStructProto = protoRelVirtualTableFields();
+    Rel relWithValues = protoRelConverter.from(valuesStructProto);
+    Rel relWithFields = protoRelConverter.from(fieldsStructProto);
+    assertEquals(relWithValues, relWithFields);
+  }
+
+  @Test
+  void setUsingValuesOrFieldsTest() {
```
benbellick: nit
```suggestion
  void setUsingOnlyValuesOrFieldsTest() {
```

core/src/test/java/io/substrait/type/proto/ReadRelRoundtripTest.java
```diff
@@ -58,4 +64,35 @@ void virtualTable() {
...
+                    .addFields(ExpressionCreator.i64(false, 1))
+                    .nullable(false)
+                    .build())
+            .build();
+    virtTable = VirtualTableScan.builder().from(virtTable).build();
+    verifyRoundTrip(virtTable);
+  }
+
+  @Test
+  void notNullableVirtualTable() {
```
benbellick: I think testing for this behavior belongs in the `VirtualTableScan.java` file, not in this roundtrip test. This file should be reserved for checking that round trip is consistent with the initial value. 

Furthermore, it would be great if you could add a test here where some columns is in fact nullable. So specifically a nullable column and some values should be concrete and some values should have null. It should additionally have a non-nullable column. This way, we can be pretty confident no information is being lost.

core/src/main/java/io/substrait/expression/ExpressionCreator.java
```diff
@@ -286,6 +287,33 @@ public static Expression.StructLiteral struct(
...
+
+  public static Expression.StructNested nestedStruct(boolean nullable, Expression... fields) {
+    return Expression.StructNested.builder()
+        .nullable(nullable)
+        .addAllFields(Arrays.asList(fields))
+        .build();
+  }
+
+  //  This function is meant to convert the deprecated StructLiteral rows in a VirtualTable to the
+  // new StructNested type
```
benbellick:
```suggestion
  /**
   * Converts StructLiteral instances to StructNested for VirtualTableScan.
   * This is a convenience method for migrating from the legacy StructLiteral-based
   * VirtualTable API to the new StructNested-based API.
   *
   * @param nullable whether the resulting StructNested instances should be nullable
   * @param structs the StructLiteral instances to convert
   * @return a list of StructNested instances with the same field structure
   */
```

core/src/main/java/io/substrait/expression/ExpressionCreator.java
benbellick: Commented above a more expanded version of the comment.

core/src/main/java/io/substrait/relation/VirtualTableScan.java
benbellick: Sorry, you had this one marked as resolved but it looks like it still isn't correct, so I unresolved it. Thanks!

core/src/test/java/io/substrait/type/proto/ReadRelRoundtripTest.java
```diff
@@ -64,35 +61,4 @@ void virtualTable() {
...
-                NamedStruct.of(Stream.of("column1").collect(Collectors.toList()), R.struct(R.I64)))
-            .addRows(
-                Expression.StructNested.builder()
-                    .addFields(ExpressionCreator.i64(false, 1))
-                    .nullable(false)
-                    .build())
-            .build();
-    virtTable = VirtualTableScan.builder().from(virtTable).build();
-    verifyRoundTrip(virtTable);
-  }
```
benbellick: Why did we lose this test in the last commit? Looked like you moved the one below but not this one.

benbellick: Actually, this one shouldn't even be moved. This is an appropriate place for it because it is a legitimate round trip test.

core/src/test/java/io/substrait/type/proto/ReadRelRoundtripTest.java
benbellick, line 1: I would still like to see a RoundTrip test which features a virtual table that contains a mix of nullable and non-nullable columns, and the nullable columns contains a mix of null and non-null values. Let me know if I can help at all!

core/src/main/java/io/substrait/expression/ExpressionCreator.java
benbellick: Commenting just to make this part of my review

core/src/main/java/io/substrait/expression/ExpressionCreator.java
benbellick: Commenting just to make this part of my review

core/src/test/java/io/substrait/type/proto/ReadRelRoundtripTest.java
gord02: Other than the .nullable(false), this function is identical to the virtualTable() test. So I wasn't sure it was worth having anymore considering structNested cannot be null. Should it be put back just for completeness of the tests? Testing for nullable=false which is fine, and also nullable=true which throws error?

core/src/main/java/io/substrait/expression/Expression.java
vbarua: Is there a reason to call this `StructNested` instead of just `NestedStruct`? In conversation, I would use "nested struct" to describe this. In the future as well, we will need to have `NestedMap` and `NestedList`, so having `Nested-` as the prefix would help group these together.

core/src/main/java/io/substrait/expression/ExpressionCreator.java
```diff
@@ -286,6 +287,33 @@ public static Expression.StructLiteral struct(
...
 
+  public static Expression.StructNested nestedStruct(
+      boolean nullable, Iterable<Expression> fields) {
+    return Expression.StructNested.builder().nullable(nullable).addAllFields(fields).build();
+  }
+
+  public static Expression.StructNested nestedStruct(boolean nullable, Expression... fields) {
+    return Expression.StructNested.builder()
+        .nullable(nullable)
+        .addAllFields(Arrays.asList(fields))
```
vbarua: minor: You can just use the following
```suggestion
        .addFields(fields)
```

to avoid having to construct an ArrayList.

core/src/main/java/io/substrait/relation/ProtoRelConverter.java
```diff
@@ -581,18 +582,26 @@ protected FileOrFiles newFileOrFiles(ReadRel.LocalFiles.FileOrFiles file) {
 
   protected VirtualTableScan newVirtualTable(ReadRel rel) {
     ReadRel.VirtualTable virtualTable = rel.getVirtualTable();
+    // If both values and expressions are set, raise an error
+    if (virtualTable.getValuesCount() > 0 && virtualTable.getExpressionsCount() > 0) {
+      throw new IllegalArgumentException(
+          "Virtual table cannot have both values and expressions set");
```
vbarua: minor
```suggestion
          "VirtualTable cannot have both values and expressions set");
```

To more closely link this to the VirtualTable message.

core/src/main/java/io/substrait/expression/ExpressionCreator.java
vbarua: Do we even need to expose this publicly? It looks like it's only used in one place. Would it make sense to move this to ProtoRelConverter and just have it as a private utility method?

core/src/main/java/io/substrait/relation/RelProtoConverter.java
```diff
@@ -646,17 +647,22 @@ public Rel visit(Cross cross, EmptyVisitationContext context) throws RuntimeExce
...
             .setCommon(common(virtualTableScan))
-            .setVirtualTable(
-                ReadRel.VirtualTable.newBuilder()
-                    .addAllValues(
-                        virtualTableScan.getRows().stream()
-                            .map(this::toProto)
-                            .map(t -> t.getLiteral().getStruct())
-                            .collect(Collectors.toList()))
-                    .build())
+            .setVirtualTable(ReadRel.VirtualTable.newBuilder().addAllExpressions(structs).build())
```
vbarua, line 767: Thinking about this from a migration perspective. Not all downstreams may have support for the `expressions` field. My default suggestion would be to set _both_ `expressions` and `values` but it looks like a number of the libraries don't support that, which actually makes this a bit tricky.

What I might suggest here is to use the least powerful field that we need for this. What I mean by that is, if every individual expression in the VirtualTable is a literal, we can just encode it using `values`. However, if any of them are complex expressions, we need to use `expressions`. This means that for anyone consuming plans from Substrait that are only using literals in VirtualTables, they will still get the same plans once we make these changes. Eventually, we'll want to fully remove the usage of `values`.

It would also be valid to force consumers of substrait-java plans to upgrade their handling to deal with `expressions`.

core/src/main/java/io/substrait/relation/VirtualTableScan.java
```diff
@@ -28,11 +29,24 @@ protected void check() {
...
+     *
+     * <p>Here we are using the second case, and thus we are carrying around a nullable field as a
```
vbarua: minor: the `<p>` tag doesn't do anything here as this is just a regular comment and not part of the Javadoc for this method.

core/src/main/java/io/substrait/relation/VirtualTableScan.java
```diff
@@ -28,11 +29,24 @@ protected void check() {
...
+     * consequence of the first option, but we enforce it to be false to ensure a user doesn't
+     * accidentally pass around meaningless additional context.
+     */
+    for (Expression.StructNested row : rows) {
+      assert !row.nullable();
```
vbarua, line 45: As you point out:
> In the codebase, we use `StructNested` to represent two subtly distinct cases: - an instance
     * of the Expression proto with the particular case of `rex_type` set to `Nested` (and then
     * `Struct`) - access to the raw proto of `Expression.Struct.Nested` itself

We actually can't distinguish between these two cases here, and we shouldn't because we might get provided the second case from an external plan. It's a valid construction IMO, so we shouldn't restrict it here.

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
```diff
@@ -629,24 +629,53 @@ public RelNode visit(VirtualTableScan virtualTableScan, Context context) {
         typeFactory.createStructType(fieldTypes, correctFieldNames);
 
     final List<ImmutableList<RexLiteral>> tuples = new ArrayList<>();
-    for (final Expression.StructLiteral row : virtualTableScan.getRows()) {
+    final List<RexNode> projectExpressions = new ArrayList<>();
+
+    for (final Expression rowExpr : virtualTableScan.getRows()) {
```
vbarua: You can use a tighter type bound here.
```suggestion
    for (final Expression.StructNested rowExpr : virtualTableScan.getRows()) {
```

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
```diff
@@ -629,24 +629,53 @@ public RelNode visit(VirtualTableScan virtualTableScan, Context context) {
...
-      for (final Expression.Literal literal : row.fields()) {
-        final RexNode rexNode = literal.accept(expressionRexConverter, context);
-        if (rexNode instanceof RexLiteral) {
-          final RexLiteral rexLiteral = (RexLiteral) rexNode;
-          rexRow.add(rexLiteral);
-        } else {
-          throw new UnsupportedOperationException(
-              "VirtualTableScan only supports literal values, found: "
-                  + rexNode.getClass().getName());
+      if (rowExpr instanceof Expression.StructLiteral) {
```
vbarua: The tighter type bound above helps to show that this branch can never be reached.

isthmus/src/test/java/io/substrait/isthmus/LogicalValuesTest.java
```diff
@@ -0,0 +1,71 @@
+package io.substrait.isthmus;
+
+import io.substrait.dsl.SubstraitBuilder;
+import io.substrait.expression.Expression;
+import io.substrait.extension.DefaultExtensionCatalog;
+// import io.substrait.proto.Expression;
```
vbarua: minor: remove this

isthmus/src/test/java/io/substrait/isthmus/LogicalValuesTest.java
```diff
@@ -0,0 +1,71 @@
...
+
+public class LogicalValuesTest extends PlanTestBase {
+
+  final SubstraitBuilder b = new SubstraitBuilder(extensions);
+  SubstraitToCalcite substraitToCalcite = new SubstraitToCalcite(extensions, typeFactory);
+  ExtensionCollector functionCollector = new ExtensionCollector();
+  RelProtoConverter relProtoConverter = new RelProtoConverter(functionCollector);
+
+  @Test
+  void testStructNested() throws Exception {
```
vbarua: minor: Nothing in this test throw an Exception

isthmus/src/test/java/io/substrait/isthmus/LogicalValuesTest.java
```diff
@@ -0,0 +1,71 @@
...
+  }
+
+  VirtualTableScan createVirtualTableScan() {
+    Expression.ScalarFunctionInvocation scalarExpr =
+        b.scalarFn(
+            DefaultExtensionCatalog.FUNCTIONS_ARITHMETIC,
+            "add:i32_i32",
+            R.I32,
+            b.i32(7),
+            b.i32(42));
```
vbarua: There is a `b.add(Expression left, Expression right)` method which you can use here.

isthmus/src/test/java/io/substrait/isthmus/LogicalValuesTest.java
```diff
@@ -0,0 +1,71 @@
...
+  @Test
+  void testStructNested() throws Exception {
+    VirtualTableScan virtualTableScan = createVirtualTableScan();
+    RelNode relNode = substraitToCalcite.convert(virtualTableScan); //    substrait rel to calcite
+
+    LogicalValues logicalValues = (LogicalValues) relNode.getInput(0);
+    Assert.equals(1, logicalValues.tuples.size()); // one row
+    Assert.equals(2, logicalValues.tuples.get(0).size()); // 2 literal expressions
+    LogicalProject logicalProject = (LogicalProject) relNode;
+    Assert.equals(1, logicalProject.getProjects().size()); // one non-literal expression
```
vbarua: If you inspect the rowType of the LogicalProject you've created, you'll notice that it only outputs a single column. We should have two columns, because the VirtualTable we're converting has two columns. The issues is that Calcite, unlike Substrait, only outputs the values of project expressions.

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
```diff
@@ -629,24 +629,53 @@ public RelNode visit(VirtualTableScan virtualTableScan, Context context) {
...
+    LogicalValues logicalValues =
+        LogicalValues.create(
+            relBuilder.getCluster(), rowTypeWithNames, ImmutableList.copyOf(tuples));
+    if (projectExpressions.isEmpty()) {
+      return logicalValues;
     }
 
-    return LogicalValues.create(
-        relBuilder.getCluster(), rowTypeWithNames, ImmutableList.copyOf(tuples));
+    return relBuilder.push(logicalValues).project(projectExpressions).build();
```
vbarua: This doesn't actually output the correct number of columns when you use it. I left a comment in LogicalValueTest about it.

core/src/main/java/io/substrait/relation/RelProtoConverter.java
benbellick: I made a comment about it also [here](https://github.com/substrait-io/substrait-java/pull/579#discussion_r2479475164). I agree that this isn't the ideal migration strategy, but I had given it the go-ahead only because it seems to be the exact behavior of the `substrait-go` library, which also enforces only one of literals vs.  expressions is present on consumed protos and only expressions are present on produced protos.

core/src/main/java/io/substrait/expression/ExpressionCreator.java
benbellick: That's a good point, I agree with making it a private utility 👍 (though I would still like there to be a comment on it, but then it doesn't have to be quite as nicely formatted).

core/src/test/java/io/substrait/type/proto/ReadRelRoundtripTest.java
benbellick: Ah in that case, yeah no worries. That test sounds redundant after all.

core/src/main/java/io/substrait/expression/Expression.java
benbellick: All of the other things in here are named `specific` then `general`, e.g. `BoolLiteral` which is associated with proto: 
```proto
  message Literal {
    oneof literal_type {
      bool boolean = 1;
      ...
    }
  ...
  }
```

In a similar pattern, `Struct` is the specific version of the general `Nested` proto: 
```proto
  message Nested {
    ...
    oneof nested_type {
      Struct struct = 3;
      List list = 4;
      Map map = 5;
    }
    message Struct {
      // Zero or more possibly heterogeneously-typed list of expressions that
      // form the struct fields.
      repeated Expression fields = 1;
    }
    ...
  }
```

So from my viewpoint, the name `StructNested` is consistent with the other names.

core/src/main/java/io/substrait/relation/VirtualTableScan.java
benbellick: I disagree here. The point trying to be made is that `StructNested` POJO-class represents the following whole proto: 
```proto
  // Expression to dynamically construct nested types.
  message Nested {
    // Whether the returned nested type is nullable.
    bool nullable = 1;
    ...
    oneof nested_type {
      ...
      Struct struct = 3; // <- the particular case where this one is set
    }
    ...
    message Struct {
      // Zero or more possibly heterogeneously-typed list of expressions that
      // form the struct fields.
      repeated Expression fields = 1;
    }
    ...
  }
```

i.e. it is a special case of the entire `Nested` message, and thus has a `nullable` field. 


OTOH, the virtual table proto contains the `Struct` proto directly (not wrapped in Nested): 
```proto
  message VirtualTable {
    repeated Expression.Literal.Struct values = 1 [deprecated = true];
    repeated Expression.Nested.Struct expressions = 2;
  }
```

So in other words, the `VirtualTable` _definitionally_ doesn't have access to the `nullable` field above, as that is a field of the `Nested` proto. Thus, I think it makes sense to assert that it is set to false on construction, as having nullable set to true on construction indicates a logical error on the users part. And we would never receive a proto where that is set to true because protos coming in simply won't have that field. 

On the other hand, I can imagine instead preferring to have two different classes that represent `StructNested`, one which inherits from `Literal` and indicates that we are constructing a special case of a `Literal`, and another one which **doesn't** extend from `Literal` but instead just from `Nested` to indicate that it is just a special case of `Nested`. 

What do you think?

isthmus/src/test/java/io/substrait/isthmus/LogicalValuesTest.java
gord02: The way I wrote the code was just to use the logicalProject to store expressions that could not fit in the logicalValues table, which are all non-literal expressions. Should this be changed such that all the expressions are reflected in the logicalProject?

core/src/test/java/io/substrait/type/proto/ReadRelRoundtripTest.java
benbellick: Here is an example test you could put in this file: 
```java
  @Test
  void virtualTableWithNullable() {
    io.substrait.relation.ImmutableVirtualTableScan virtTable =
        VirtualTableScan.builder()
            .initialSchema(
                NamedStruct.of(
                    Stream.of("nullable_col", "non_nullable_col").collect(Collectors.toList()),
                    R.struct(N.I64, R.I64)))
            .addRows(
                Expression.StructNested.builder()
                    .addFields(ExpressionCreator.typedNull(N.I64))
                    .addFields(ExpressionCreator.i64(false, 1))
                    .build())
            .addRows(
                Expression.StructNested.builder()
                    .addFields(ExpressionCreator.i64(true, 2))
                    .addFields(ExpressionCreator.i64(false, 3))
                    .build())
            .build();
    verifyRoundTrip(virtTable);
  }
```

Note that in order for it to work, I added this line to `TestBase.java`: 
```java
public abstract class TestBase {

  protected static final SimpleExtension.ExtensionCollection defaultExtensionCollection =
      DefaultExtensionCatalog.DEFAULT_COLLECTION;

  protected TypeCreator R = TypeCreator.REQUIRED;
  protected TypeCreator N = TypeCreator.NULLABLE; // <- NEW LINE
  ...
}
```

core/src/main/java/io/substrait/expression/ExpressionCreator.java
gord02: Is ProtoRelConverter a good place considering it converts from pojo to pojo and not proto?

core/src/main/java/io/substrait/relation/VirtualTableScan.java
vbarua: Ah! I see your point.
> We actually can't distinguish between these two cases here

I was wrong about this, because this check is specific to VirtualTables where we can make the assumption you're suggesting.

core/src/main/java/io/substrait/expression/ExpressionCreator.java
vbarua: > Is ProtoRelConverter a good place

That would be where I would put it, close to where it used,

core/src/main/java/io/substrait/expression/Expression.java
vbarua: > All of the other things in here are named specific then general, e.g. BoolLiteral which is associated with proto

We actually do both already 😅. For example, we have `ScalarSubquery` as opposed to `SubqueryScalar`. I would optimize for general readability over mapping to the protobuf structure exactly.

The literal ones are a bit weird, because I think both "Boolean Literal" or "Literal Boolean" roll off the tongue. But when talking about "Struct Nested" or "Nested Struct", you never really say the former.

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelVisitor.java
```diff
@@ -172,8 +174,46 @@ public Rel visit(org.apache.calcite.rel.core.Calc calc) {
...
+        structs.add(
+            Expression.NestedStruct.builder()
+                .addAllFields(
+                    projectRow.getProjects().stream()
+                        .map(this::toExpression)
+                        .collect(Collectors.toUnmodifiableList()))
+                .build());
+      }
+      return VirtualTableScan.builder().initialSchema(type).addAllRows(structs).build();
+    }
```
vbarua: I'm a little wary of this un-conversion. The isLogicalValue check has a lot of reflection stuff going on, and would also need to check for a bit more stuff like each `Values` relation having a single empty row to make sure we don't apply this for cases it doesn't apply.

I would recommend that we file a future issue for implementing something like this and tackle it as it's own thing if we need it. I would want to see a lot more testing for something like this specifically.

PR #584, opened 2025-10-29
fix: namespace -> urn

### LINKED ISSUE #583: Fix places that incorrectly use old `namespace` in place of `urn`
See [here](https://github.com/substrait-io/substrait-java/blob/ab5943e53f73ed73f816f2c3f59502378b3bf15c/core/src/main/java/io/substrait/dsl/SubstraitBuilder.java). 

E.g.
```java
public AggregateFunctionInvocation aggregateFn(
      String namespace, String key, Type outputType, Expression... args) {
...
}
```

benbellick:
Change usages of `namespace` to correctly be `urn`.

Closes #583


nielspardon (approved):
LGTM

PR #593, opened 2025-11-03
fix(isthmus): support TPC-DS queries 1, 30, 81

bestbeforetoday:
TPC-DS queries 1, 30 and 81 can already be converted from SQL to a Substrait plan, but these Substrait plans cannot be successfully converted back to SQL.

This change updates the Calcite version to 1.41.0, which contains fixes that allow successful conversion of TPC-DS queries from Substrait back to SQL.

In order to work around a regression introduced in Calcite 1.41.0 ([CALCITE-7276](https://issues.apache.org/jira/browse/CALCITE-7276)), this change adds the `withIdentifierExpansion` SQL validator configuration option, and fixes Rex to Substrait conversion of SQL Update where the table modify has a project as its input.

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -1,6 +1,6 @@
 [versions]
 antlr = "4.13.2"
-calcite = "1.40.0"
+calcite = "1.41.0"
 classgraph = "4.8.184"
 commons-lang3 = "[3.18.0,)"
 graal = "25.0.0"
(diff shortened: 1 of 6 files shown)


isthmus/src/test/java/io/substrait/isthmus/DmlRoundtripTest.java
```diff
@@ -21,6 +22,7 @@ void testDelete() throws SqlParseException {
   }
 
   @Test
+  @Disabled("Fails with Calcite >1.40.0")
```
bestbeforetoday: @ZorinAnton any idea why this test fails when moving from Calcite 1.40.0 to 1.41.0?

bestbeforetoday: It appears to be a regression in Calcite 1.41.0:

https://issues.apache.org/jira/browse/CALCITE-7276

PR #595, opened 2025-11-06
fix: handle valid properties of Simple Extension types array

### LINKED ISSUE #594: yaml extension parser rejects valid yaml (e.g. with description field)
E.g.
```yaml
%YAML 1.2
---
urn: extension:test:custom_extensions
types:
  - name: "customType1"
    description: >-
      some description
```
Fails parsing:
```
    java.lang.IllegalStateException: com.fasterxml.jackson.databind.exc.UnrecognizedPropertyException: Unrecognized field "description" (class io.substrait.extension.ImmutableSimpleExtension$Type$Json), not marked as ignorable (3 known properties: "urn", "name", "structure"])
     at [Source: (StringReader); line: 8, column: 1] (through reference chain: io.substrait.extension.ImmutableSimpleExtension$ExtensionSignatures$Json["types"]->java.util.ArrayList[0]->io.substrait.extension.ImmutableSimpleExtension$Type$Json["description"])
```

There may be more valid things that aren't accounted for.

benbellick:
Closes #594

### DIFF
--- a/core/src/main/java/io/substrait/extension/SimpleExtension.java
+++ b/core/src/main/java/io/substrait/extension/SimpleExtension.java
@@ -535,11 +535,18 @@ public class SimpleExtension {
 
     public abstract String name();
 
+    public abstract Optional<String> description();
+
     @JacksonInject(SimpleExtension.URN_LOCATOR_KEY)
     public abstract String urn();
 
     // TODO: Handle conversion of structure object to Named Struct representation
     protected abstract Optional<Object> structure();
+
+    // TODO: Properly handle parameters
+    protected abstract Optional<List<Object>> parameters();
+
+    protected abstract Optional<Boolean> variadic();
 
     public TypeAnchor getAnchor() {
       return anchorSupplier.get();
--- a/core/src/test/resources/extensions/custom_extensions.yaml
+++ b/core/src/test/resources/extensions/custom_extensions.yaml
@@ -3,6 +3,10 @@
 urn: extension:test:custom_extensions
 types:
   - name: "customType1"
+    description: >-
+      some description
+    variadic: true
+    parameters: []
   - name: "customType2"
     structure:
       field1: i32


core/src/main/java/io/substrait/extension/SimpleExtension.java
```diff
@@ -535,12 +535,19 @@ public abstract static class Type {
...
     public abstract String urn();
 
     // TODO: Handle conversion of structure object to Named Struct representation
     protected abstract Optional<Object> structure();
 
+    // TODO: Properly handle parameters
+    protected abstract Optional<List<Object>> parameters();
+
+    protected abstract Optional<Boolean> variadic();
+
```
vbarua, line 550: Looks reasonable to me. All of these fields are declared in 
https://github.com/substrait-io/substrait/blob/cbf7ed74eb710bbc83b81dff6147816e7afded4a/text/simple_extensions_schema.yaml#L27-L37

core/src/main/java/io/substrait/extension/SimpleExtension.java
```diff
@@ -535,12 +535,19 @@ public abstract static class Type {
...
+    public abstract Optional<String> description();
+
     @JacksonInject(SimpleExtension.URN_LOCATOR_KEY)
     public abstract String urn();
 
     // TODO: Handle conversion of structure object to Named Struct representation
     protected abstract Optional<Object> structure();
 
+    // TODO: Properly handle parameters
+    protected abstract Optional<List<Object>> parameters();
```
vbarua, line 547: Could you make an issue to track this as well.

vbarua (approved):
Overall looks good. Would be good to have a follow-up issue for the TODOs.

core/src/main/java/io/substrait/extension/SimpleExtension.java
benbellick: https://github.com/substrait-io/substrait-java/issues/599

PR #597, opened 2025-11-10
feat(isthmus): true, false, distinct comparator scalar fn mappings

asuriy:
- Added `is_true`, `is_false`, `is_not_true`, `is_not_false` and `is_distinct_from` functions in `FunctionMappings.java`
- Added tests into new file: `ComparisonFunctionsTest.java`

### DIFF
--- a/isthmus/src/test/java/io/substrait/isthmus/ComparisonFunctionsTest.java
+++ b/isthmus/src/test/java/io/substrait/isthmus/ComparisonFunctionsTest.java
@@ -1,0 +1,49 @@
+package io.substrait.isthmus;
+
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.CsvSource;
+import org.junit.jupiter.params.provider.ValueSource;
+
+public class ComparisonFunctionsTest extends PlanTestBase {
+  static String CREATES =
+      "CREATE TABLE numbers (int_a INT, int_b INT, double_a DOUBLE, double_b DOUBLE)";
+
+  @Test
+  void is_true() throws Exception {
+    String query = "SELECT ((int_a > int_b) IS TRUE) FROM numbers";
+    assertSqlSubstraitRelRoundTrip(query, CREATES);
+  }
+
+  @Test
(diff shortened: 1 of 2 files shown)


andrew-coleman (approved):

isthmus/src/test/java/io/substrait/isthmus/ComparisonFunctionsTest.java
```diff
@@ -0,0 +1,49 @@
...
+  void is_not_false() throws Exception {
+    String query = "SELECT ((int_a > int_b) IS NOT FALSE) FROM numbers";
+    assertSqlSubstraitRelRoundTrip(query, CREATES);
+  }
+
+  @ParameterizedTest
+  @CsvSource({"int_a, int_b", "int_b, int_a", "double_a, double_b", "double_b, double_a"})
+  void is_distinct_from(String left, String right) throws Exception {
+    String query = String.format("SELECT (%s IS DISTINCT FROM %s) FROM numbers", left, right);
+    assertSqlSubstraitRelRoundTrip(query, CREATES);
```
vbarua, line 40: Minor comment looking at this from the future. This doesn't actually test the IS DISTINCT FROM mapping because a query like
```sql
SELECT (int_a IS DISTINCT FROM int_b) FROM numbers
```
gets processed into the following Calcite representation
```
LogicalProject(EXPR$0=[AND(OR(IS NOT NULL($0), IS NOT NULL($1)), IS NOT TRUE(=($0, $1)))])
  LogicalTableScan(table=[[NUMBERS]])
```

I only noticed this because I was poking around updating some tests. The same thing happens with IS NOT DISTINCT FROM.

asuriy: Thanks for bringing this to my attention @vbarua! I'll take a look and rectify this tomorrow morning.

asuriy: @vbarua I believe I've addressed the logical rewrite in this pull request: https://github.com/substrait-io/substrait-java/pull/628 - let me know if this works

Issue #600, opened 2025-11-11
`VirtualTable` builders should ensure that schema and rows are consistent


benbellick:

nielspardon:
Implemented by #601, which added row-type-vs-schema validation to `VirtualTableScan.check()`, and extended by #684, which made the Calcite → Substrait direction take literal nullability from the schema. The PR did not use a closing keyword, so this stayed open.

Filed #1047 as a follow-up: the validation is written with `assert`, so it only fires under `-ea` (which Gradle's test tasks enable) and does not enforce anything for consumers of the library.

Closing as completed.

PR #601, opened 2025-11-11
feat: enforce row type consistency with schema in virtual tables

benbellick:
- ensure that virtual table builders enforce consistency of types across schema and rows (#600)
- allow nullability to be explicitly set in conversion from spark literal to substrait literal

### DIFF
--- a/core/src/test/java/io/substrait/relation/VirtualTableScanTest.java
+++ b/core/src/test/java/io/substrait/relation/VirtualTableScanTest.java
@@ -6,6 +6,7 @@ import static io.substrait.expression.ExpressionCreator.list;
 import static io.substrait.expression.ExpressionCreator.map;
 import static io.substrait.expression.ExpressionCreator.string;
 import static io.substrait.expression.ExpressionCreator.struct;
+import static io.substrait.expression.ExpressionCreator.typedNull;
 import static org.junit.jupiter.api.Assertions.assertDoesNotThrow;
 import static org.junit.jupiter.api.Assertions.assertThrows;
 
@@ -70,10 +71,10 @@ class VirtualTableScanTest extends TestBase {
         ImmutableVirtualTableScan.builder()
             .initialSchema(
                 NamedStruct.of(
-                    Arrays.asList("id", "name", "age"), R.struct(R.I64, R.STRING, R.I32)))
+                    Arrays.asList("id", "name", "age"), R.struct(R.I64, N.STRING, R.I32)))
             .addRows(
-                struct(false, i64(false, 1L), string(false, "Alice"), i32(false, 30)),
-                struct(false, i64(false, 2L), string(false, "Bob"), i32(false, 25)))
+                struct(false, i64(false, 1L), string(true, "Alice"), i32(false, 30)),
+                struct(false, i64(false, 2L), typedNull(N.STRING), i32(false, 25)))
             .build();
     assertDoesNotThrow(virtualTableScan::check);
   }
--- a/core/src/test/java/io/substrait/TestBase.java
+++ b/core/src/test/java/io/substrait/TestBase.java
@@ -17,6 +17,7 @@ public abstract class TestBase {
       DefaultExtensionCatalog.DEFAULT_COLLECTION;
 
   protected TypeCreator R = TypeCreator.REQUIRED;
+  protected TypeCreator N = TypeCreator.NULLABLE;
 
   protected SubstraitBuilder b = new SubstraitBuilder(defaultExtensionCollection);
   protected ExtensionCollector functionCollector = new ExtensionCollector();


spark/src/main/scala/io/substrait/spark/expression/ToSubstraitLiteral.scala
```diff
@@ -31,103 +31,88 @@ import scala.collection.JavaConverters
...
+      nullable: Boolean): SExpression.Literal = {
+    struct(
+      nullable,
+      JavaConverters.asJavaIterable(fields.zip(structData.values).map {
+        case (field, any) => apply(Literal(any, field.dataType), Some(field.nullable))
+      }))
   }
 
-  private def convertWithValue(literal: Literal): Option[SExpression.Literal] = {
+  private def convertWithValue(literal: Literal, nullable: Boolean): Option[SExpression.Literal] = {
```
benbellick, line 87: The diff in this function is scarier looking than it actually is. Basically, we want to be able to explicitly pass down the `nullable` value instead of assuming it to always be `null`.

spark/src/main/scala/io/substrait/spark/expression/ToSubstraitLiteral.scala
```diff
@@ -31,103 +31,88 @@ import scala.collection.JavaConverters
...
-      val seconds = (ms / Util.MICROS_PER_SECOND % Util.SECONDS_PER_DAY).toInt
-      val micros = ms % Util.MICROS_PER_SECOND
-      intervalDay(false, days, seconds, micros, Util.MICROSECOND_PRECISION)
-    }
-    val _intervalYear: Int => SExpression.Literal = (m: Int) => intervalYear(false, m / 12, m % 12)
-    val _string: String => SExpression.Literal = string(false, _)
-    val _binary: Array[Byte] => SExpression.Literal = binary(false, _)
-    val _array: (ArrayData, DataType, Boolean) => SExpression.Literal = sparkArray2Substrait
-    val _map: (MapData, DataType, DataType, Boolean) => SExpression.Literal = sparkMap2Substrait
-    val _struct: (GenericInternalRow, Array[StructField]) => SExpression.Literal =
```
benbellick, line 103: Since we are no longer assuming non-nullability, there is no need for all of these helpers. Instead, they are inlined in the code.

spark/src/main/scala/io/substrait/spark/expression/ToSubstraitLiteral.scala
```diff
@@ -139,15 +124,45 @@ class ToSubstraitLiteral {
...
+            .convert(literal.dataType, nullable = true)
+            .map(typedNull)
+            .getOrElse(throw new UnsupportedOperationException(
+              s"Unable to convert the type ${literal.dataType.typeName}"))
+        } else {
+          convertWithValue(literal, n)
+            .getOrElse(
+              throw new UnsupportedOperationException(
+                s"Unable to convert the type ${literal.dataType.typeName}"))
+        }
```
andrew-coleman: Purely a suggestion, but I wonder if the code would be clearer if the nested `if` blocks were pulled up into separate `case` conditions:
```suggestion
      case Some(false) if (literal.value == null) =>
        throw new IllegalArgumentException("Cannot create a non-nullable type for a null value")
      case Some(true) if (literal.value == null) =>
        ToSubstraitType
          .convert(literal.dataType, nullable = true)
          .map(typedNull)
          .getOrElse(throw new UnsupportedOperationException(
            s"Unable to convert the type ${literal.dataType.typeName}"))
      case Some(n) =>
        // Use explicit nullability
        convertWithValue(literal, n)
          .getOrElse(
            throw new UnsupportedOperationException(
              s"Unable to convert the type ${literal.dataType.typeName}"))
```

core/src/test/java/io/substrait/relation/VirtualTableScanTest.java
```diff
@@ -59,6 +63,113 @@ void check() {
...
+  void checkInvalidRowWrongFieldCount() {
+    // Row has wrong number of fields
+    assertThrows(
+        AssertionError.class,
+        () ->
+            ImmutableVirtualTableScan.builder()
+                .initialSchema(
+                    NamedStruct.of(
+                        Arrays.asList("id", "name", "age"), R.struct(R.I64, R.STRING, R.I32)))
+                .addRows(struct(false, i64(false, 1L), string(false, "Alice"))) // Missing age field
```
gord02, line 111: This case is covered [here](https://github.com/substrait-io/substrait-java/blob/848bfcf83f4235574a686a4d48407a8f8ba5d99e/core/src/main/java/io/substrait/relation/VirtualTableScan.java#L33)

spark/src/main/scala/io/substrait/spark/expression/ToSubstraitLiteral.scala
benbellick: I like this suggestion, thanks!

core/src/test/java/io/substrait/relation/VirtualTableScanTest.java
benbellick: Agreed! Just introducing a test to ensure that it is tested. Was there a test for that case already that I was missing?

gord02: Not explicitly no. Good point!

PR #602, opened 2025-11-12
feat: add ExchangeRel support to core

cht42:
Implement ExchangeRel in core. the spec https://github.com/substrait-io/substrait/blob/main/proto/substrait/algebra.proto#L470

### DIFF
--- a/core/src/test/java/io/substrait/type/proto/ExchangeRelRoundtripTest.java
+++ b/core/src/test/java/io/substrait/type/proto/ExchangeRelRoundtripTest.java
@@ -14,7 +14,7 @@ import java.util.Collections;
 import java.util.List;
 import org.junit.jupiter.api.Test;
 
-public class ExchangeRelRoundtripTest extends TestBase {
+class ExchangeRelRoundtripTest extends TestBase {
 
   final Rel baseTable =
       b.namedScan(


isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
```diff
@@ -591,6 +596,31 @@ public RelNode visit(NamedUpdate update, Context context) {
...
+  @Override
+  public RelNode visit(RoundRobinExchange exchange, Context context) throws RuntimeException {
+    return visitFallback(exchange, context);
+  }
+
+  @Override
+  public RelNode visit(BroadcastExchange exchange, Context context) throws RuntimeException {
+    return visitFallback(exchange, context);
+  }
+
```
nielspardon, line 623: I guess these could be mapped to Calcite's `LogicalExchange`:
https://calcite.apache.org/javadocAggregate/org/apache/calcite/rel/logical/LogicalExchange.html

nielspardon: Is this something you were planning to look into in a follow-up PR? To map to Spark/Calcite equivalents where it makes sense?

cht42: Hey, yes exactly, I figured it would be easier to break down the work into different PRs

cht42:
@nielspardon do you have any other comments ? or is the PR good to merge ?

core/src/main/java/io/substrait/relation/ProtoRelConverter.java
```diff
@@ -977,6 +994,163 @@ protected ConsistentPartitionWindow newConsistentPartitionWindow(
...
+    Rel input = from(rel.getInput());
+    List<AbstractExchangeRel.ExchangeTarget> targets =
+        rel.getTargetsList().stream().map(this::newExchangeTarget).collect(Collectors.toList());
+
+    ProtoExpressionConverter converter =
+        new ProtoExpressionConverter(lookup, extensions, input.getRecordType(), this);
+    List<FieldReference> fieldReferences =
+        rel.getScatterByFields().getFieldsList().stream()
+            .map(converter::from)
+            .collect(Collectors.toList());
```
nielspardon: I do realize most of the other usages of `ProtoExpressionConverter` in this class also just use `converter` as the variable name. Given that we have so many different converters in the code base it might make sense to use a more specific variable name to make it clear which one it is. We could use the approach from the `newAggregate()` method in this class which calls the variable `protoExprConverter`.

```suggestion
    ProtoExpressionConverter protoExprConverter =
        new ProtoExpressionConverter(lookup, extensions, input.getRecordType(), this);
    List<FieldReference> fieldReferences =
        rel.getScatterByFields().getFieldsList().stream()
            .map(protoExprConverter::from)
            .collect(Collectors.toList());
```

core/src/main/java/io/substrait/relation/ProtoRelConverter.java
```diff
@@ -977,6 +994,163 @@ protected ConsistentPartitionWindow newConsistentPartitionWindow(
...
+        rel.getTargetsList().stream().map(this::newExchangeTarget).collect(Collectors.toList());
+    ProtoExpressionConverter converter =
+        new ProtoExpressionConverter(lookup, extensions, input.getRecordType(), this);
+
+    ImmutableSingleBucketExchange.Builder builder =
+        SingleBucketExchange.builder()
+            .input(input)
+            .partitionCount(rel.getPartitionCount())
+            .targets(targets)
+            .expression(converter.from(rel.getSingleTarget().getExpression()));
```
nielspardon: same here, we could use a more specific variable name
```suggestion
    ProtoExpressionConverter protoExprConverter =
        new ProtoExpressionConverter(lookup, extensions, input.getRecordType(), this);

    ImmutableSingleBucketExchange.Builder builder =
        SingleBucketExchange.builder()
            .input(input)
            .partitionCount(rel.getPartitionCount())
            .targets(targets)
            .expression(protoExprConverter.from(rel.getSingleTarget().getExpression()));
```

core/src/main/java/io/substrait/relation/ProtoRelConverter.java
```diff
@@ -977,6 +994,163 @@ protected ConsistentPartitionWindow newConsistentPartitionWindow(
...
+    ProtoExpressionConverter converter =
+        new ProtoExpressionConverter(lookup, extensions, input.getRecordType(), this);
+
+    ImmutableMultiBucketExchange.Builder builder =
+        MultiBucketExchange.builder()
+            .input(input)
+            .partitionCount(rel.getPartitionCount())
+            .targets(targets)
+            .expression(converter.from(rel.getMultiTarget().getExpression()))
+            .constrainedToCount(rel.getMultiTarget().getConstrainedToCount());
```
nielspardon: same here, we could use a more specific variable name
```suggestion
    ProtoExpressionConverter protoExprConverter =
        new ProtoExpressionConverter(lookup, extensions, input.getRecordType(), this);

    ImmutableMultiBucketExchange.Builder builder =
        MultiBucketExchange.builder()
            .input(input)
            .partitionCount(rel.getPartitionCount())
            .targets(targets)
            .expression(protoExprConverter.from(rel.getMultiTarget().getExpression()))
            .constrainedToCount(rel.getMultiTarget().getConstrainedToCount());
```

core/src/main/java/io/substrait/relation/RelProtoConverter.java
```diff
@@ -522,6 +530,111 @@ public Rel visit(NamedUpdate update, EmptyVisitationContext context) throws Runt
...
+
+  private ExchangeRel.ExchangeTarget toProto(AbstractExchangeRel.ExchangeTarget target) {
+    ExchangeRel.ExchangeTarget.Builder builder =
+        ExchangeRel.ExchangeTarget.newBuilder().addAllPartitionId(target.getPartitionIds());
+    if (target.getType() instanceof TargetType.Uri) {
+      builder.setUri(((TargetType.Uri) target.getType()).getUri());
+    }
+    if (target.getType() instanceof TargetType.Extended) {
+      builder.setExtended(((TargetType.Extended) target.getType()).getExtended());
+    }
```
nielspardon, line 624: given that `TargetType.Uri` and `TargetType.Extended` are classes `target.getType()` can not be both so I would write this with an `else if`:
```suggestion
    if (target.getType() instanceof TargetType.Uri) {
      builder.setUri(((TargetType.Uri) target.getType()).getUri());
    } else if (target.getType() instanceof TargetType.Extended) {
      builder.setExtended(((TargetType.Extended) target.getType()).getExtended());
    }
```

spark/src/main/scala/io/substrait/debug/RelToVerboseString.scala
```diff
@@ -212,6 +213,71 @@ class RelToVerboseString(addSuffix: Boolean) extends DefaultRelVisitor[String] {
...
+  }
+
+  override def visit(exchange: BroadcastExchange, context: EmptyVisitationContext): String = {
+    withBuilder(exchange, 10)(
+      builder => {
+        builder
+          .append("=partitionCount")
+          .append(exchange.getPartitionCount)
+          .append("=targets")
+          .append(exchange.getTargets)
```
nielspardon: shouldn't the equals sign always be the suffix of these string literals?

nielspardon (commented):
the changes look good. I only have some minor remarks

nielspardon:
Thanks for addressing the feedback. the one thing I didn't notice yesterday which would be great if you can take care of is to also add some unit tests for the changes in the core. We have relation roundtrip test here which you can mirror: https://github.com/substrait-io/substrait-java/tree/main/core/src/test/java/io/substrait/type/proto

core/src/test/java/io/substrait/type/proto/ExchangeRelRoundtripTest.java
```diff
@@ -0,0 +1,108 @@
...
+import io.substrait.relation.physical.RoundRobinExchange;
+import io.substrait.relation.physical.ScatterExchange;
+import io.substrait.relation.physical.SingleBucketExchange;
+import io.substrait.relation.physical.TargetType;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
+import org.junit.jupiter.api.Test;
+
+public class ExchangeRelRoundtripTest extends TestBase {
```
nielspardon: minor: this can be package private for JUnit 5
```suggestion
class ExchangeRelRoundtripTest extends TestBase {
```

PR #606, opened 2025-11-13
feat(isthmus): mapping of string positional scalar fns from substrait

asuriy:
- Added `contains`, `starts_with`, `ends_with` and `strpos` to `FunctionMappings.java`
- Added BIG_QUERY and STANDARD SqlLibraries to `SubstraitOperatorTable.java`
- Added passing tests to `StringFunctionTest.java`


bestbeforetoday:
A few suggested changes:

1. Also test with the variants of `STARTS_WITH` and `ENDS_WITH` supported by Calcite: `STARTSWITH` and `ENDSWITH` respectively. They _should_ work without any code changes but I think worth testing explicitly.
2. It might be worth testing with string literals too. For example, `SELECT POSITION('text' IN vc) FROM strings`. Including the single quote (`'`) character can be a bit tricky with a JUnit `@CsvSource` since, by default, it uses single quote as its own quote character for parsing. You would need to change the quote character, something like:
    ```java
    @CsvSource(quoteCharacter = '`', value = {"'text', vc"})
    ````

PR #609, opened 2025-11-17
Benbellick/handle structured udt

### LINKED ISSUE #603: UserDefined literals only supported `Any` representation, not `Struct` representation
This is the same issue present in substrait-go (https://github.com/substrait-io/substrait-go/issues/170)

benbellick:
Closes #603

### DIFF
--- a/isthmus/src/main/java/io/substrait/isthmus/expression/ExpressionRexConverter.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/expression/ExpressionRexConverter.java
@@ -121,18 +121,19 @@ public class ExpressionRexConverter
 
   @Override
   public RexNode visit(Expression.UserDefinedStruct expr, Context context) throws RuntimeException {
+    return toUserDefinedStructLiteral(expr, context);
+  }
+
+  private RexLiteral toUserDefinedStructLiteral(Expression.UserDefinedStruct expr, Context context) {
     java.util.List<org.apache.calcite.rel.type.RelDataType> fieldTypes =
         new java.util.ArrayList<>(expr.fields().size());
     java.util.List<RexLiteral> fieldLiterals = new java.util.ArrayList<>(expr.fields().size());
 
     for (Expression.Literal field : expr.fields()) {
       fieldTypes.add(toStructFieldType(field));
-      fieldLiterals.add(toStructFieldLiteral(field, context));
+      fieldLiterals.add(literalToRexLiteral(field, context));
     }
 
-    // Generate dummy field names (f0, f1, f2, etc.) to satisfy Calcite's ROW type requirements.
-    // Substrait UserDefinedStruct doesn't have field names - just ordered field values.
-    // These synthetic names are discarded during conversion back to Substrait.
     java.util.List<String> fieldNames =
         java.util.stream.IntStream.range(0, expr.fields().size())
             .mapToObj(i -> "f" + i)
@@ -147,7 +148,7 @@ public class ExpressionRexConverter
             fieldTypes,
             fieldNames);
 
-    return rexBuilder.makeLiteral(fieldLiterals, customType, false);
+    return (RexLiteral) rexBuilder.makeLiteral(fieldLiterals, customType, false);
   }
 
   private org.apache.calcite.rel.type.RelDataType toStructFieldType(Expression.Literal field) {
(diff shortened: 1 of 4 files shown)


isthmus/src/test/java/io/substrait/isthmus/utils/UserTypeFactory.java
```diff
@@ -35,11 +36,48 @@ public Type createSubstrait(boolean nullable) {
     return TypeCreator.of(nullable).userDefined(urn, name);
   }
 
+  /**
+   * Test-specific variant of the core implementation that treats Calcite-copied types as
+   * equivalent, even when they are not the same Java instance. Calcite often clones UDTs during
```
asolimando, line 41: I don't think this is accurate, in general you use [canonize](https://github.com/apache/calcite/blob/51965cfbcba638ed9335e5d3cc7b560003d5f85c/core/src/main/java/org/apache/calcite/rel/type/RelDataTypeFactoryImpl.java#L446) to "create" data types, which internally uses a cache and the same object is returned if already seen before.

Lots of places in Calcite compare them by instance (Java's `==`), so if that was true it would break in a few places.

RelDataTypeFactory is pretty explicit on this in its [javadoc](https://github.com/apache/calcite/blob/main/core/src/main/java/org/apache/calcite/rel/type/RelDataTypeFactory.java#L45-L48).

```
public RelDataType createUserDefinedType(
      String urn, String name,
      List<Type.Parameter> params, boolean nullable) {
    SubstraitUserDefinedType type =
        new SubstraitUserDefinedType(urn, name, params, nullable);
    return canonize(type); <-- this is what guarantees uniqueness for your types
  }
```

Make sure you never create types manually, at that point you should be able to drop methods like `matchesCalciteAlias` that are just working around the issue, possibly also `matchesSubstraitType`

isthmus/src/main/java/io/substrait/isthmus/type/SubstraitUserDefinedType.java
```diff
@@ -0,0 +1,251 @@
...
+        .urn(urn)
+        .name(name)
+        .typeParameters(typeParameters)
+        .nullable(nullable)
+        .build();
+  }
+
+  /** Creates a SubstraitUserDefinedType from a Substrait Type.UserDefined. */
+  public static SubstraitUserDefinedType from(io.substrait.type.Type.UserDefined type) {
+    return new SubstraitUserDefinedType(
```
asolimando, line 84: Here you must canonize, and for doing that you need a handle to a Calcite type factory

isthmus/src/main/java/io/substrait/isthmus/type/SubstraitUserDefinedType.java
```diff
@@ -0,0 +1,251 @@
...
+        String urn,
+        String name,
+        List<io.substrait.proto.Type.Parameter> typeParameters,
+        boolean nullable) {
+      super(urn, name, typeParameters, nullable);
+    }
+
+    /** Creates a SubstraitUserDefinedAnyType from a Substrait Type.UserDefined. */
+    public static SubstraitUserDefinedAnyType from(io.substrait.type.Type.UserDefined type) {
+      return new SubstraitUserDefinedAnyType(
```
asolimando, line 145: Same, canonization is needed here

isthmus/src/main/java/io/substrait/isthmus/type/SubstraitUserDefinedType.java
```diff
@@ -0,0 +1,251 @@
...
+
+    /**
+     * Creates a SubstraitUserDefinedStructType from a Substrait Type.UserDefined and field
+     * information.
+     */
+    public static SubstraitUserDefinedStructType from(
+        io.substrait.type.Type.UserDefined type,
+        List<RelDataType> fieldTypes,
+        List<String> fieldNames) {
+      return new SubstraitUserDefinedStructType(
```
asolimando, line 247: Canonization needed

isthmus/src/main/java/io/substrait/isthmus/type/SubstraitUserDefinedType.java
```diff
@@ -0,0 +1,251 @@
...
+        List<io.substrait.proto.Type.Parameter> typeParameters,
+        boolean nullable,
+        List<RelDataType> fieldTypes,
+        List<String> fieldNames) {
+      super(urn, name, typeParameters, nullable);
+      if (fieldTypes.size() != fieldNames.size()) {
+        throw new IllegalArgumentException("Field types and names must have same length");
+      }
+      this.fieldTypes = fieldTypes;
+      this.fieldNames = fieldNames;
```
asolimando, line 183: Here you must call again `computeDigest()` after setting `fieldTypes` and `fieldNames` (true for any field in general), because `RelDataTypeImpl` identity is based on the digest field. This is true for all subclasses adding or changing the field.

Similarly, I expect to see `equals()` and `hashCode()` overrides when you have extra fields, I think you should add them.

isthmus/src/main/java/io/substrait/isthmus/type/SubstraitUserDefinedType.java
```diff
@@ -0,0 +1,251 @@
...
+
+    @Override
+    public List<String> getFieldNames() {
+      return fieldNames;
+    }
+
+    @Override
+    public SqlTypeName getSqlTypeName() {
+      // Can be considered as ROW since it has structure
+      return SqlTypeName.ROW;
```
asolimando, line 223: I am not 100% sure, but I am afraid you would then be matching built-in `ROW` types, I'd rather play it safe and use `SqlTypeName.OTHER` (like the base class) and use `isStruct()` override to indicate structure, something like:

```
@Override
public boolean isStruct() {
  return true;
}
```

I am starting from the assumption that, if any structured type would make sense as `ROW`, we would not have this boolean method, but I don't have any evidence for saying you shouldn't do it.

isthmus/src/main/java/io/substrait/isthmus/expression/ExpressionRexConverter.java
```diff
@@ -109,12 +109,110 @@ public RexNode visit(Expression.NullLiteral expr, Context context) throws Runtim
...
+
+    for (Expression.Literal field : expr.fields()) {
+      fieldTypes.add(toStructFieldType(field));
+      fieldLiterals.add(literalToRexLiteral(field, context));
+    }
+
+    java.util.List<String> fieldNames =
+        java.util.stream.IntStream.range(0, expr.fields().size())
+            .mapToObj(i -> "f" + i)
+            .collect(java.util.stream.Collectors.toList());
```
asolimando, line 140: From https://substrait.io/types/type_classes/#communicating-user-defined-types I was under the impression that UDTs were coming with their own field names, `expr.fields()` might have them set?

If that's not necessarily the case, we should at least preserve them whenever they are set.

isthmus/src/main/java/io/substrait/isthmus/type/SubstraitUserDefinedType.java
```diff
@@ -0,0 +1,251 @@
...
+      return fields;
+    }
+
+    @Override
+    public int getFieldCount() {
+      return fieldTypes.size();
+    }
+
+    @Override
+    public RelDataTypeField getField(String fieldName, boolean caseSensitive, boolean elideRecord) {
```
asolimando, line 201: The `elideRecord` parameter is not used here, IIRC it's used for nested field access. You should check how `RelDataTypeImpl.getField()` behaves and possibly:
- remove the override if not needed,
- call super.getField() at the top and add your extra logic after,
- add the nested type handling and honor the `elideRecord` param

isthmus/src/main/java/io/substrait/isthmus/type/SubstraitUserDefinedType.java
asolimando: `RelDataTypeImpl` is `Serializable`, are we making sure the extra fields are serializable too?

isthmus/src/test/java/io/substrait/isthmus/type/SubstraitUserDefinedTypeTest.java
```diff
@@ -0,0 +1,42 @@
...
+  void sameParametersRemainEqual() {
+    Type.Parameter integerParam = Type.Parameter.newBuilder().setInteger(7).build();
+    SubstraitUserDefinedType left =
+        new SubstraitUserDefinedType.SubstraitUserDefinedAnyType(
+            URN, NAME, List.of(integerParam), true);
+    SubstraitUserDefinedType right =
+        new SubstraitUserDefinedType.SubstraitUserDefinedAnyType(
+            URN, NAME, List.of(integerParam), true);
+
+    assertEquals(left, right);
```
asolimando, line 40: you should be able to assert `left == right` after fixing type canonization

isthmus/src/test/java/io/substrait/isthmus/type/SubstraitUserDefinedTypeTest.java
asolimando, line 1: Can we add tests for nested UDT field access? (to exercise the `elideRecord` argument in getField) and, if supported, a UDT within a struct within UDT (if such a mixup is valid)?

It might be interesting to test also two UDTs with same URN/name but different type parameters like `Point<i32>` and `Point<i64>`, I expect them to be different, as `i32` != `i64`

asolimando (commented):
I have some initial suggestions, there are a couple of places where I am not 100% sure like switching from `SqlTypeName.ROW` to `SqlTypeName.OTHER` and I have explicitly mentioned that, the rest should be safe to be implemented, feel free to ping me when it's reviewable again :)

PR #613, opened 2025-11-21
feat(core): handle struct-based UDT literals

### LINKED ISSUE #611: Handle struct-based UDT literals in core
UDT (=User Defined Type) literals in plans can be encoded via struct-based representation [according to the substrait spec](https://substrait.io/types/type_classes/#communicating-user-defined-types). This issue is to introduce handling for this UDT representation in the `core` part of `substrait-java`.

benbellick:
Support both opaque (google.protobuf.Any) and
structured (Literal.Struct) encodings for user-defined type literals per [Substrait spec](https://substrait.io/types/type_classes/#communicating-user-defined-types).

- Split UserDefinedLiteral into UserDefinedAny and UserDefinedStruct
- Move type parameters to interface level for parameterized types
- Add first-class POJO representation for type parameters
- Test coverage including roundtrip tests
- Throw exception on unhandled struct-based representation in isthmus

Closes #611

### DIFF
--- a/core/src/main/java/io/substrait/expression/Expression.java
+++ b/core/src/main/java/io/substrait/expression/Expression.java
@@ -712,60 +712,9 @@ public interface Expression extends FunctionArg {
   }
 
   /**
-   * Represents a user-defined value associated with a {@link UserDefinedAnyLiteral}.
+   * User-defined literal with value encoded as {@link com.google.protobuf.Any}.
    *
-   * <p>This interface abstracts the underlying binary representation of user-defined literal
-   * values. Users should implement this interface and provide protobuf conversion logic for their
-   * custom types.
-   */
-  interface UserDefinedAnyValue {
-    /**
-     * Converts this user-defined value to a protobuf Any message for serialization.
-     *
-     * @return the protobuf Any representation of this value
-     */
-    com.google.protobuf.Any toProto();
-
-    /**
-     * Creates a UserDefinedAnyValue from a protobuf Any message.
-     *
-     * <p>This factory method provides a default implementation that wraps the protobuf Any. Users
-     * can create custom implementations for their specific types.
-     *
-     * @param any the protobuf Any to wrap
-     * @return a UserDefinedAnyValue wrapping the protobuf Any
-     */
-    static UserDefinedAnyValue of(com.google.protobuf.Any any) {
-      return DefaultUserDefinedAnyValue.of(any);
-    }
-  }
-
-  /**
-   * Default implementation of {@link UserDefinedAnyValue} that wraps a protobuf Any message.
-   *
-   * <p>This implementation is used when deserializing from protobuf. Users can provide custom
-   * implementations for their specific types.
-   */
-  @Value.Immutable
-  abstract class DefaultUserDefinedAnyValue implements UserDefinedAnyValue {
-    abstract com.google.protobuf.Any any();
-
-    @Override
-    public com.google.protobuf.Any toProto() {
-      return any();
-    }
-
-    static DefaultUserDefinedAnyValue of(com.google.protobuf.Any any) {
-      return ImmutableExpression.DefaultUserDefinedAnyValue.builder().any(any).build();
-    }
-  }
-
-  /**
-   * User-defined literal with value encoded as binary data.
-   *
-   * <p>This encoding allows for arbitrary binary data to be stored in the literal value. Users
-   * should implement {@link UserDefinedAnyValue} to provide their custom type representation and
-   * handle protobuf conversion logic.
+   * <p>This encoding allows for arbitrary binary data to be stored in the literal value.
    */
   @Value.Immutable
   abstract class UserDefinedAnyLiteral implements UserDefinedLiteral {
@@ -778,7 +727,7 @@ public interface Expression extends FunctionArg {
     @Override
     public abstract List<io.substrait.type.Type.Parameter> typeParameters();
 
-    public abstract UserDefinedAnyValue value();
+    public abstract com.google.protobuf.Any value();
 
     @Override
     public Type.UserDefined getType() {
--- a/core/src/main/java/io/substrait/expression/ExpressionCreator.java
+++ b/core/src/main/java/io/substrait/expression/ExpressionCreator.java
@@ -307,31 +307,7 @@ public class ExpressionCreator {
   }
 
   /**
-   * Create a UserDefinedAnyLiteral with a custom UserDefinedAnyValue implementation.
-   *
-   * @param nullable whether the literal is nullable
-   * @param urn the URN of the user-defined type
-   * @param name the name of the user-defined type
-   * @param typeParameters the type parameters for the user-defined type (can be empty list)
-   * @param value the value implementing UserDefinedAnyValue interface
-   */
-  public static Expression.UserDefinedAnyLiteral userDefinedLiteralAny(
-      boolean nullable,
-      String urn,
-      String name,
-      java.util.List<io.substrait.type.Type.Parameter> typeParameters,
-      Expression.UserDefinedAnyValue value) {
-    return Expression.UserDefinedAnyLiteral.builder()
-        .nullable(nullable)
-        .urn(urn)
-        .name(name)
-        .addAllTypeParameters(typeParameters)
-        .value(value)
-        .build();
-  }
-
-  /**
-   * Create a UserDefinedAnyLiteral from a protobuf Any message.
+   * Create a UserDefinedAnyLiteral with google.protobuf.Any representation.
    *
    * @param nullable whether the literal is nullable
    * @param urn the URN of the user-defined type
@@ -345,8 +321,13 @@ public class ExpressionCreator {
       String name,
       java.util.List<io.substrait.type.Type.Parameter> typeParameters,
       Any value) {
-    return userDefinedLiteralAny(
-        nullable, urn, name, typeParameters, Expression.UserDefinedAnyValue.of(value));
+    return Expression.UserDefinedAnyLiteral.builder()
+        .nullable(nullable)
+        .urn(urn)
+        .name(name)
+        .addAllTypeParameters(typeParameters)
+        .value(value)
+        .build();
   }
 
   /**
--- a/core/src/main/java/io/substrait/expression/proto/ExpressionProtoConverter.java
+++ b/core/src/main/java/io/substrait/expression/proto/ExpressionProtoConverter.java
@@ -388,7 +388,7 @@ public class ExpressionProtoConverter
                       expr.typeParameters().stream()
                           .map(typeProtoConverter::toProto)
                           .collect(java.util.stream.Collectors.toList()))
-                  .setValue(expr.value().toProto());
+                  .setValue(expr.value());
 
           bldr.setNullable(expr.nullable()).setUserDefined(userDefinedBuilder).build();
         });
--- a/isthmus/src/main/java/io/substrait/isthmus/expression/ExpressionRexConverter.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/expression/ExpressionRexConverter.java
@@ -112,7 +112,7 @@ public class ExpressionRexConverter
   public RexNode visit(Expression.UserDefinedAnyLiteral expr, Context context)
       throws RuntimeException {
     RexLiteral binaryLiteral =
-        rexBuilder.makeBinaryLiteral(new ByteString(expr.value().toProto().toByteArray()));
+        rexBuilder.makeBinaryLiteral(new ByteString(expr.value().toByteArray()));
     RelDataType type = typeConverter.toCalcite(typeFactory, expr.getType());
     return rexBuilder.makeReinterpretCast(type, binaryLiteral, rexBuilder.makeLiteral(false));
   }
--- a/isthmus/src/main/java/io/substrait/isthmus/expression/CallConverters.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/expression/CallConverters.java
@@ -81,7 +81,7 @@ public class CallConverters {
                     .urn(t.urn())
                     .name(t.name())
                     .addAllTypeParameters(t.typeParameters())
-                    .value(Expression.UserDefinedAnyValue.of(anyValue))
+                    .value(anyValue)
                     .build();
               } catch (com.google.protobuf.InvalidProtocolBufferException e) {
                 throw new IllegalStateException("Failed to parse UserDefinedAnyLiteral value", e);


core/src/main/java/io/substrait/extension/DefaultExtensionCatalog.java
```diff
@@ -44,6 +45,8 @@ private static SimpleExtension.ExtensionCollection loadDefaultCollection() {
             .map(c -> String.format("/functions_%s.yaml", c))
             .collect(Collectors.toList());
 
+    defaultFiles.add("/extension_types.yaml");
```
benbellick, line 48: I only realized these were missing because I wanted to use the `point` definition in here for testing struct-based representation of UDT literals!

core/src/main/java/io/substrait/expression/proto/ExpressionProtoConverter.java
```diff
@@ -359,21 +357,45 @@ public Expression visit(
...
   @Override
   public Expression visit(
-      io.substrait.expression.Expression.UserDefinedLiteral expr, EmptyVisitationContext context) {
+      io.substrait.expression.Expression.UserDefinedAnyLiteral expr,
+      EmptyVisitationContext context) {
     int typeReference =
         extensionCollector.getTypeReference(SimpleExtension.TypeAnchor.of(expr.urn(), expr.name()));
     return lit(
         bldr -> {
-          try {
```
benbellick, line 385: This exception doesn't happen anymore because we don't parse the `Any` here. Instead, we have a reference to the pre-parsed proto directly.

core/src/main/java/io/substrait/type/proto/ParameterizedProtoConverter.java
```diff
@@ -262,6 +262,13 @@ public ParameterizedType userDefined(int ref) {
           "User defined types are not supported in Parameterized Types for now");
     }
 
+    @Override
+    public ParameterizedType userDefined(
+        int ref, java.util.List<io.substrait.type.Type.Parameter> typeParameters) {
+      throw new UnsupportedOperationException(
+          "User defined types are not supported in Parameterized Types for now");
```
benbellick, line 269: This is consistent with the above, where we don't yet support ParamerizedType conversion.

core/src/main/java/io/substrait/type/proto/TypeExpressionProtoVisitor.java
```diff
@@ -355,6 +355,13 @@ public DerivationExpression userDefined(int ref) {
           "User defined types are not supported in Derivation Expressions for now");
     }
 
+    @Override
+    public DerivationExpression userDefined(
+        int ref, java.util.List<io.substrait.type.Type.Parameter> typeParameters) {
+      throw new UnsupportedOperationException(
+          "User defined types are not supported in Derivation Expressions for now");
```
benbellick, line 362: This is also consistent with the above.

core/src/main/java/io/substrait/type/proto/TypeProtoConverter.java
```diff
@@ -199,4 +213,38 @@ protected Integer i(final int integerValue) {
...
+      io.substrait.type.Type.Parameter parameter) {
+    if (parameter instanceof io.substrait.type.Type.ParameterNull) {
+      return Type.Parameter.newBuilder()
+          .setNull(com.google.protobuf.Empty.getDefaultInstance())
+          .build();
+    } else if (parameter instanceof io.substrait.type.Type.ParameterDataType) {
+      io.substrait.type.Type.ParameterDataType dataType =
+          (io.substrait.type.Type.ParameterDataType) parameter;
+      TypeProtoConverter converter =
+          new TypeProtoConverter(new io.substrait.extension.ExtensionCollector());
```
benbellick: hmm... wonder if I really need to construct a new `TypeProtoConverter` here?

core/src/test/java/io/substrait/type/proto/LiteralRoundtripTest.java
```diff
@@ -3,23 +3,288 @@
...
+          + "        type: dataType\n"
+          + "      - name: mode\n"
+          + "        type: enum\n"
+          + "    structure:\n"
+          + "      value: T\n";
+
+  private static final SimpleExtension.ExtensionCollection NESTED_TYPES_EXTENSIONS =
+      SimpleExtension.load("nested_types.yaml", NESTED_TYPES_YAML);
+
+  private static final ExtensionCollector NESTED_TYPES_FUNCTION_COLLECTOR =
```
benbellick, line 68: Don't love all of these created.

isthmus/src/main/java/io/substrait/isthmus/expression/ExpressionRexConverter.java
```diff
@@ -109,14 +109,21 @@ public RexNode visit(Expression.NullLiteral expr, Context context) throws Runtim
...
+        rexBuilder.makeBinaryLiteral(new ByteString(expr.value().toByteString().toByteArray()));
     RelDataType type = typeConverter.toCalcite(typeFactory, expr.getType());
     return rexBuilder.makeReinterpretCast(type, binaryLiteral, rexBuilder.makeLiteral(false));
   }
 
+  @Override
+  public RexNode visit(Expression.UserDefinedStructLiteral expr, Context context)
+      throws RuntimeException {
+    throw new UnsupportedOperationException(
+        "UserDefinedStructLiteral representation is not yet supported in Isthmus");
```
benbellick, line 124: This is https://github.com/substrait-io/substrait-java/issues/612

core/src/main/java/io/substrait/expression/Expression.java
```diff
@@ -662,21 +662,96 @@ public <R, C extends VisitationContext, E extends Throwable> R accept(
...
+   *
+   * <p>User-defined literals can be encoded in one of two ways as per the Substrait spec:
+   *
+   * <ul>
+   *   <li>As {@code google.protobuf.Any} - see {@link UserDefinedAnyLiteral}
+   *   <li>As {@code Literal.Struct} - see {@link UserDefinedStructLiteral}
+   * </ul>
+   *
+   * @see UserDefinedAnyLiteral
+   * @see UserDefinedStructLiteral
```
vbarua: super minor: both of the `@see` feel redundant because you reference them above.

core/src/main/java/io/substrait/expression/Expression.java
```diff
@@ -662,21 +662,96 @@ public <R, C extends VisitationContext, E extends Throwable> R accept(
...
 
+    @Override
     public abstract String name();
 
     @Override
-    public Type getType() {
-      return Type.withNullability(nullable()).userDefined(urn(), name());
+    public abstract List<io.substrait.type.Type.Parameter> typeParameters();
+
+    public abstract List<Literal> fields();
```
vbarua, line 770: Is there a reason to use a `List<Literal>` here instead of just a `Expression.StructLiteral`, which would map directly to the protobuf?

core/src/main/java/io/substrait/expression/Expression.java
```diff
@@ -662,21 +662,96 @@ public <R, C extends VisitationContext, E extends Throwable> R accept(
...
+        ExpressionVisitor<R, C, E> visitor, C context) throws E {
+      return visitor.visit(this, context);
+    }
+  }
+
+  /**
+   * User-defined literal with value encoded as {@code Literal.Struct}.
+   *
+   * <p>This encoding uses a structured list of fields to represent the literal value.
+   */
```
vbarua, line 758: The docstring feels a little inconsistent. You have
> literal with value encoded as {@code Literal.Struct}

but in the class the values are encoded as a `List<Literal>`, and the second part of the docstring is consistent with that.

core/src/main/java/io/substrait/expression/Expression.java
```diff
@@ -662,21 +662,96 @@ public <R, C extends VisitationContext, E extends Throwable> R accept(
...
+    @Override
+    public abstract String urn();
+
+    @Override
+    public abstract String name();
+
+    @Override
+    public abstract List<io.substrait.type.Type.Parameter> typeParameters();
+
+    public abstract com.google.protobuf.Any value();
```
vbarua: Release Notes
Capturing the value as an Any instead of a ByteString does feel nicer ✨

core/src/main/java/io/substrait/expression/Expression.java
```diff
@@ -662,21 +662,96 @@ public <R, C extends VisitationContext, E extends Throwable> R accept(
...
+   *
+   * <ul>
+   *   <li>As {@code google.protobuf.Any} - see {@link UserDefinedAnyLiteral}
+   *   <li>As {@code Literal.Struct} - see {@link UserDefinedStructLiteral}
+   * </ul>
+   *
+   * @see UserDefinedAnyLiteral
+   * @see UserDefinedStructLiteral
+   */
+  interface UserDefinedLiteral extends Literal {
```
vbarua, line 706: Release Notes
We should call out that we don't construct UserDefinedLiterals anymore.

core/src/main/java/io/substrait/expression/proto/ProtoExpressionConverter.java
```diff
@@ -492,10 +492,40 @@ public Expression.Literal from(io.substrait.proto.Expression.Literal literal) {
...
+                      .collect(Collectors.toList()),
+                  userDefinedLiteral.getValue());
+            case STRUCT:
+              return ExpressionCreator.userDefinedLiteralStruct(
+                  literal.getNullable(),
+                  urn,
+                  name,
+                  userDefinedLiteral.getTypeParametersList().stream()
+                      .map(protoTypeConverter::from)
+                      .collect(Collectors.toList()),
```
vbarua: minor: The following type parameter conversion code
```java
                  userDefinedLiteral.getTypeParametersList().stream()
                      .map(protoTypeConverter::from)
                      .collect(Collectors.toList())
```
is shared between both branches and could be extracted out.

core/src/main/java/io/substrait/type/Type.java
```diff
@@ -393,6 +393,30 @@ abstract class UserDefined implements Type {
...
     public abstract String name();
 
+    /**
+     * Returns the type parameters for this user-defined type.
+     *
+     * <p>Type parameters are used to represent parameterized/generic types, such as {@code
+     * vector<i32>} or custom types like {@code FixedArray<100>}. Each parameter in the list can be
+     * either a type (like {@code i32}) or a value (like the integer {@code 100}).
+     *
+     * <p>Unlike built-in parameterized types ({@link Map}, {@link ListType}, {@link Decimal}),
```
vbarua: minor: The types linked in this line don't actually resolve correctly. They need the `Type.` prefix (ie. Type.Map) in order to resolve.

core/src/main/java/io/substrait/type/Type.java
```diff
@@ -393,6 +393,30 @@ abstract class UserDefined implements Type {
 
     public abstract String name();
 
+    /**
+     * Returns the type parameters for this user-defined type.
+     *
+     * <p>Type parameters are used to represent parameterized/generic types, such as {@code
+     * vector<i32>} or custom types like {@code FixedArray<100>}. Each parameter in the list can be
```
vbarua: My reading of this line is that `FixedArray` as a custom type but `vector` is not a custom type. However `vector` is _also_ a custom type, assuming you're referencing https://substrait.io/types/type_classes/#compound-user-defined-types

core/src/main/java/io/substrait/type/Type.java
```diff
@@ -402,4 +426,50 @@ public <R, E extends Throwable> R accept(TypeVisitor<R, E> typeVisitor) throws E
...
   }
+
+  /**
+   * Represents a type parameter for user-defined types.
+   *
+   * <p>Type parameters can be data types (like {@code i32} in {@code List<i32>}), or value
+   * parameters (like the {@code 10} in {@code VARCHAR<10>}). This interface provides a type-safe
+   * representation of all possible parameter kinds.
+   */
+  interface Parameter {}
```
vbarua, line 436: Is the function of https://github.com/substrait-io/substrait-java/blob/main/core/src/main/java/io/substrait/function/ParameterizedType.java applicable here? It feels vaguely related.

core/src/main/java/io/substrait/type/Type.java
```diff
@@ -402,4 +426,50 @@ public <R, E extends Throwable> R accept(TypeVisitor<R, E> typeVisitor) throws E
...
+  @Value.Immutable
+  abstract class ParameterDataType implements Parameter {
+    public abstract Type type();
+  }
+
+  /** A boolean value parameter. */
+  @Value.Immutable
+  abstract class ParameterBooleanValue implements Parameter {
+    public abstract boolean value();
+  }
```
vbarua, line 448: Is there a usecase for having boolean literal type parameters? I can't imagine a usecase where something like `MySpecialType<false>` is something you would need.

core/src/main/java/io/substrait/type/Type.java
```diff
@@ -402,4 +426,50 @@ public <R, E extends Throwable> R accept(TypeVisitor<R, E> typeVisitor) throws E
...
+  @Value.Immutable
+  abstract class ParameterIntegerValue implements Parameter {
+    public abstract long value();
+  }
+
+  /** An enum value parameter (represented as a string). */
+  @Value.Immutable
+  abstract class ParameterEnumValue implements Parameter {
+    public abstract String value();
+  }
```
vbarua, line 460: What would this type be used for?

core/src/main/java/io/substrait/type/Type.java
```diff
@@ -402,4 +426,50 @@ public <R, E extends Throwable> R accept(TypeVisitor<R, E> typeVisitor) throws E
...
+
+  /** A string value parameter. */
+  @Value.Immutable
+  abstract class ParameterStringValue implements Parameter {
+    public abstract String value();
+  }
```
vbarua, line 466: What would this type be used for?

core/src/test/java/io/substrait/TestBase.java
```diff
@@ -25,9 +29,22 @@ public abstract class TestBase {
   protected ProtoRelConverter protoRelConverter =
       new ProtoRelConverter(functionCollector, defaultExtensionCollection);
 
+  protected ExpressionProtoConverter expressionProtoConverter =
+      new ExpressionProtoConverter(functionCollector, relProtoConverter);
```
vbarua: I would recommend the following
```java
  protected ExpressionProtoConverter expressionProtoConverter =
      relProtoConverter.getExpressionProtoConverter();
```

Effectively, when we instantiate the RelProtoConverter, that creates its own associated ExpressionProtoConverter and links that back to the RelProtoConverter. The constructor handles sharing any necessary state. Pulling the ExpressionProtoConverter out of the RelProtoConverter guarantees that the state is shared correctly.

vbarua (commented):
Started taking a look at this and left some comments. Mostly looks reasonable. Want to come back and look at your tests with fresh eyes, and also think about parameterized types with a fresh 🧠.

core/src/main/java/io/substrait/type/Type.java
benbellick: Hmm I will fix that but that seems like the kind of thing that should be caught by CI/CD

core/src/main/java/io/substrait/expression/Expression.java
benbellick: Sharing [this comment](https://github.com/substrait-io/substrait-java/pull/579#discussion_r2528204873) because I believe it is relevant. 

------------

Basically it is the same issue, which is that the POJO class called `StructLiteral` is representing a special case of the `Literal` class.
```proto
  message Literal { // Both StructLiteral and UserDefinedStructLiteral are representing this _whole_ message
    oneof literal_type {
      ...
      Struct struct = 25;
      ...
      UserDefined user_defined = 33;
    }
    ...
    message Struct {
      // A possibly heterogeneously typed list of literals
      repeated Literal fields = 1;
    }

    message UserDefined {
      oneof type_anchor_type {
        // points to a type_anchor defined in this plan
        uint32 type_reference = 1;

        // points to a type_alias_anchor defined in this plan.
        uint32 type_alias_reference = 5;
      }

      // The parameters to be bound to the type class, if the type class is
      // parameterizable.
      repeated Type.Parameter type_parameters = 3;

      // a user-defined literal can be encoded in one of two ways
      oneof val {
        // the value of the literal, serialized using some type-specific protobuf message
        google.protobuf.Any value = 2;
        // the value of the literal, serialized using the structure definition in its declaration
        Literal.Struct struct = 4;
      }
    }
  }
```

Back to your comment, switching this member variable to be `Expression.StructLiteral` would amount to embedding one literal proto inside of another. For example, the proto `Struct` doesn't actually have a nullable field. But the POJO `Expression.StructLiteral` does have a nullable field because it inherits from `Literal`. 

This doesn't mean that we couldn't replace the member variable as you suggest, but if we did do that, we would be carrying around extra meaningless data, which I think is more confusing ultimately.

core/src/main/java/io/substrait/expression/Expression.java
benbellick: I guess that this could be made less confusing. [The intention was to mean the proto `Literal.Struct`](https://github.com/substrait-io/substrait/blob/84816cd57e23b99b9e1816e1fffc740f121dbb22/proto/substrait/algebra.proto#L1143), no some java class. What do you think about just altering the message to say 
> User-defined literal with value encoded via the proto message {@code Literal.Struct}.

?

benbellick: I altered it to instead link to the actual proto. Let me know what you think! I also did the same for the any case.

core/src/main/java/io/substrait/type/Type.java
benbellick: Interesting, in my IDE all three resolve totally fine. When I open in intellij, then I get some errors. Nonetheless, I am adding the resolution so that it works everywhere.

core/src/main/java/io/substrait/type/Type.java
benbellick: As discussed offline, these were inspired by [this portion](https://github.com/substrait-io/substrait/blob/84816cd57e23b99b9e1816e1fffc740f121dbb22/text/simple_extensions_schema.yaml#L86-L90) of the simple extension schema:
```yaml
  type_param_defs: # an array of compound type parameter definitions
    type: array
    items:
      ...
      properties:
        ...
        type: # expected metatype for the parameter
          type: string
          enum:
            - dataType
            - boolean
            - integer
            - enumeration
            - string
```

So while I don't understand the usage of it, I thought it was best to include all of them for consistency with the spec.

core/src/main/java/io/substrait/type/Type.java
benbellick: https://github.com/substrait-io/substrait-java/pull/613#discussion_r2566074545

core/src/main/java/io/substrait/type/Type.java
benbellick: https://github.com/substrait-io/substrait-java/pull/613#discussion_r2566074545

🤷 just wanted to be consistent with the spec

core/src/main/java/io/substrait/type/Type.java
benbellick: I ended up just simplifying this doc a bunch by removing the middle paragraph. It is more confusing IMO, and I don't want to run the risk of making a claim about something which is not thoroughly specified in the spec itself.

core/src/main/java/io/substrait/type/Type.java
benbellick: 🤔 yes that is an interesting point. Looking into it!

benbellick: So as I understand it, `ParameterizedType.java` is used for representing abstract types with parameters in yaml files. Where as the Parameter above being introduced is actually a concrete argument passed into the type. 

So for example, `List<any1>` could be a ParameterizedType, whereas `List<int32>` is a type with parameters `[int32]`.

core/src/main/java/io/substrait/expression/ExpressionCreator.java
```diff
@@ -306,13 +306,51 @@ public static Expression.NestedStruct nestedStruct(boolean nullable, Expression.
...
+        .build();
+  }
+
+  /**
+   * Create a UserDefinedStructLiteral with Struct representation.
+   *
+   * @param nullable whether the literal is nullable
+   * @param urn the URN of the user-defined type
+   * @param name the name of the user-defined type
+   * @param typeParameters the type parameters for the user-defined type (can be empty list)
```
nielspardon: I guess one would add the article here:

```suggestion
   * @param typeParameters the type parameters for the user-defined type (can be an empty list)
```

core/src/main/java/io/substrait/expression/ExpressionCreator.java
```diff
@@ -306,13 +306,51 @@ public static Expression.NestedStruct nestedStruct(boolean nullable, Expression.
...
-  public static Expression.UserDefinedLiteral userDefinedLiteral(
-      boolean nullable, String urn, String name, Any value) {
-    return Expression.UserDefinedLiteral.builder()
+  /**
+   * Create a UserDefinedAnyLiteral with google.protobuf.Any representation.
+   *
+   * @param nullable whether the literal is nullable
+   * @param urn the URN of the user-defined type
+   * @param name the name of the user-defined type
+   * @param typeParameters the type parameters for the user-defined type (can be empty list)
```
nielspardon: I guess one would add the article here:
```suggestion
   * @param typeParameters the type parameters for the user-defined type (can be an empty list)
```

core/src/main/java/io/substrait/expression/Expression.java
```diff
@@ -693,21 +693,94 @@ public <R, C extends VisitationContext, E extends Throwable> R accept(
...
+    @Override
+    public abstract String urn();
+
+    @Override
+    public abstract String name();
 
+    @Override
+    public abstract List<io.substrait.type.Type.Parameter> typeParameters();
+
+    public abstract com.google.protobuf.Any value();
```
nielspardon: One observation/remark I would have here is that in other places we have been trying to not expose protobuf objects in the Substrait core POJO API like e.g. in `AdvancedExtension` we provide the empty `Optimization` and `Enhancement` interfaces and ask users of the Substrait Java SDK to implement protobuf conversion logic for those instead. At first glance it feels like we would be a little inconsistent if for user defined any literals we expose the protobuf any value directly.

examples/substrait-spark/src/main/java/io/substrait/examples/util/SubstraitStringify.java
```diff
@@ -96,7 +96,7 @@ public static List<String> explain(io.substrait.plan.Plan plan) {
   /**
    * Explains the Sustrait relation
    *
-   * @param plan Subsrait relation
+   * @param rel Subsrait relation
```
nielspardon, line 99: side note: this fix is also in https://github.com/substrait-io/substrait-java/pull/642

core/src/main/java/io/substrait/expression/Expression.java
benbellick: Hey Niels, this was a great point. I just pushed a commit addressing this by introducing the interface `UserDefinedAnyValue`. I'll leave the PR open for a little while to give you the opportunity to provide feedback on that component if you'd like. Otherwise, I'll merge it in later today. Thanks for the review!

benbellick: Actually, after implementing the abstraction pattern similar to `AdvancedExtension`, I think we should actually defer the change you are suggesting. The abstraction adds an extra layer (`UserDefinedAnyValue`) without clear benefits. Users still need to work with `com.google.protobuf.Any` directly to construct values, and the indirection makes the API harder to understand. Since you mentioned this could be revisited later, let's keep it simple for now and use `Any` directly as the spec defines.

nielspardon: Yes, on that point I also wasn't sure how much additional value the additional layer of abstraction would bring.

Issue #622, opened 2025-11-26
parameterConsistency field not deserializable from YAML


benbellick:
YAML extension files containing `parameterConsistency` in variadic function definitions fail to load with the error:

```
Unrecognized field "parameterConsistency" (class ImmutableSimpleExtension$VariadicBehavior$Json),
not marked as ignorable (2 known properties: "min", "max"])
```

As a minimal test case to demonstrate the failure: 
```java
@Test
public void testParameterConsistencyLoading() {
  String yamlContent =
      "urn: extension:test:example\n"
          + "scalar_functions:\n"
          + "  - name: test_func\n"
          + "    impls:\n"
          + "      - args:\n"
          + "          - name: arg1\n"
          + "            value: string\n"
          + "        variadic:\n"
          + "          min: 1\n"
          + "          parameterConsistency: CONSISTENT\n"
          + "        return: string\n";

  SimpleExtension.ExtensionCollection collection =
      SimpleExtension.load("test://example", yamlContent);

  assertEquals(
      SimpleExtension.VariadicBehavior.ParameterConsistency.CONSISTENT,
      collection.scalarFunctions().get(0).variadic().get().parameterConsistency());
}
```

This can be fixed by adding the `@Value.Default` annotation to the  `VariadicBehavior.parameterConsistency()` method. Without it, the Immutables library cannot recognize it as a deserializable field.

This will need one extra step, which is to actually thread the parameterConsistency info through to the function validation to ensure that the parameterConsistency is actually enforced.

PR #624, opened 2025-11-27
feat: handle parameterConsistency option in YAML extensions

Adam-Alani:
This pull request introduces improved handling and validation of variadic parameter consistency for function signatures, particularly distinguishing between consistent and inconsistent parameter types in variadic arguments. The changes ensure that function argument matching logic respects the `parameterConsistency` setting and are thoroughly tested with new unit tests.

**Enhancements to variadic parameter consistency:**

* Updated the function argument matching logic in `FunctionConverter` so that, when `parameterConsistency` is set to `CONSISTENT`, all variadic arguments must have the same type (ignoring nullability), and when set to `INCONSISTENT`, variadic arguments can differ in type. This logic is now explicitly enforced in the `inputTypesMatchDefinedArguments` method. [[1]](diffhunk://#diff-78351c0f66175cf526eabe57fddf55c6a7e80e7a29fdaee6a69d59b318f45c1cL247-R249) [[2]](diffhunk://#diff-78351c0f66175cf526eabe57fddf55c6a7e80e7a29fdaee6a69d59b318f45c1cL259-R274) [[3]](diffhunk://#diff-78351c0f66175cf526eabe57fddf55c6a7e80e7a29fdaee6a69d59b318f45c1cL295-R332)
* Added a default value for `parameterConsistency` in the `SimpleExtension.VariadicBehavior` interface, defaulting to `CONSISTENT` if not specified.

FIxes: https://github.com/substrait-io/substrait-java/issues/622


isthmus/src/test/java/io/substrait/isthmus/expression/VariadicParameterConsistencyTest.java
```diff
@@ -0,0 +1,250 @@
...
+import io.substrait.function.ParameterizedType;
+import io.substrait.isthmus.PlanTestBase;
+import io.substrait.isthmus.TypeConverter;
+import io.substrait.type.Type;
+import io.substrait.type.TypeCreator;
+import java.util.List;
+import org.junit.jupiter.api.Test;
+
+/** Tests for variadic parameter consistency validation in FunctionConverter. */
+class VariadicParameterConsistencyTest extends PlanTestBase {
```
vbarua: I would suggest renaming this to `VariadicBehaviourTest` for generality. You've already got some limited testing for `min` behaviour present, and this would be a good place to add more tests for this kind of stuff.

core/src/test/java/io/substrait/extension/TypeExtensionTest.java
```diff
@@ -97,4 +97,27 @@ void roundtripNumberedAnyTypes() {
     Plan planReturned = protoPlanConverter.from(protoPlan);
     assertEquals(plan, planReturned);
   }
+
+  @Test
+  public void testParameterConsistencyLoading() {
```
vbarua: I think this is a reasonable test, but I don't think it belongs in this test class, which from the comments is more specific to custom type handling in extensions.

A separate `VariadicBehaviourTest` class (or somesuch name) would be a better fit.

isthmus/src/main/java/io/substrait/isthmus/expression/FunctionConverter.java
```diff
@@ -292,9 +300,39 @@ private boolean inputTypesMatchDefinedArguments(
       }
 
       // If all the types match, check if the wildcard types are compatible.
-      // TODO: Determine if non-enumerated wildcard types (i.e. `any` as opposed to `any1`) need to
-      //   have the same type.
-      return wildcardToType.values().stream().allMatch(s -> s.size() == 1);
+      // When parameterConsistency is INCONSISTENT, wildcard types can differ.
+      // When parameterConsistency is CONSISTENT (or not variadic), wildcard types must be the same.
```
vbarua: As I understand it, the `parameterConsistency` behaviour _only_ applies to the variadic arguments. It's not related to how we typecheck the enumerated wildcard types (i.e `any1`, `any2`, etc). From the (limited) [documentation](https://substrait.io/expressions/scalar_functions/#type-parameter-resolution-in-variadic-functions) on it:
> When the last argument of a function is variadic and declares a type parameter e.g. fn(A, B, C...), the C parameter can be marked as either consistent or inconsistent. If marked as consistent, the function can only be bound to arguments where all the C types are the same concrete type. If marked as inconsistent, each unique C can be bound to a different type within the constraints of what T allows.

CONSISTENT means that the types of all the variadic arguments have to be the same. INCONSISTENT means... I'm not sure tbh because in
> each unique C can be bound to a different type within the constraints of what T allows.

I don't know what T is. This is something I can follow up with the community about.

core/src/main/java/io/substrait/extension/SimpleExtension.java
```diff
@@ -235,6 +235,7 @@ enum ParameterConsistency {
       INCONSISTENT
     }
 
+    @Value.Default
     default ParameterConsistency parameterConsistency() {
       return ParameterConsistency.CONSISTENT;
```
vbarua, line 240: I think this is a reasonble default, because I think it's what most people expect in practice and it's also the first value in the [enumeration](https://github.com/substrait-io/substrait/blob/84816cd57e23b99b9e1816e1fffc740f121dbb22/text/simple_extensions_schema.yaml#L166-L168).

We can formalize this in the spec more concretely.

vbarua (commented):
Left some comments. The big question I have is actually around what INCONSISTENT is supposed to mean. It's not super clear to me based on what's currently in the specification, and I'm not sure if the checks you've added in the FunctionConverter are correct. We can follow up with the upstream to clarify the intended behaviour for the Isthmus code. 

The code you've added to the core to parse the parameterConsistency looks good to me, and stops substrait-java from blowing up if it encounters an extension with parameter consistency set as Ben pointed out in https://github.com/substrait-io/substrait-java/issues/622. If you make the core parsing changes their own PR, I'd be happy to merge those in.

isthmus/src/main/java/io/substrait/isthmus/expression/FunctionConverter.java
benbellick: Ah I realize that this is attempting to handle the second part of the ticket I linked (https://github.com/substrait-io/substrait-java/issues/622). 

However, I don't believe this is the right place to handle it. Considering that `parameterConsistency` has meaning regardless of calcite, those checks should really be inside of `core/` IMO.

core/src/main/java/io/substrait/expression/VariadicParameterConsistencyValidator.java
```diff
@@ -0,0 +1,74 @@
...
+      SimpleExtension.Function func, List<FunctionArg> arguments) {
+    java.util.Optional<SimpleExtension.VariadicBehavior> variadic = func.variadic();
+    if (!variadic.isPresent()) {
+      return;
+    }
+
+    SimpleExtension.VariadicBehavior variadicBehavior = variadic.get();
+    if (variadicBehavior.parameterConsistency()
+        != SimpleExtension.VariadicBehavior.ParameterConsistency.CONSISTENT) {
+      // INCONSISTENT allows different types, so validation passes
```
benbellick, line 34: So unfortunately, this is going to be a bit more complicated than it initially seemed. [From the docs](https://github.com/substrait-io/substrait/blob/5f031b69ed211e1ec307be3db7989d64c65d33a2/proto/substrait/function.proto#L35-L36): 
>       // Each argument can be any possible concrete type afforded by the bounds
>       // of any parameter defined in the arguments specification.

Let me show you a (made-up) example:
```yaml
urn: "urn:example:extension"
scalar_functions:
  - name: "inconsistent_sum"
    impls:
      - args:
          - value: "decimal<P,S>"
            variadic:
              min: 1
              parameterConsistency: INCONSISTENT
        return: "decimal<38,S>"

```

This means that the following are valid: 
- `inconsistent_sum(decimal<10, 2>, decimal<10, 2>, decimal<10, 2>)`
- `inconsistent_sum(decimal<10, 2>, decimal<10, 2>)`
- `inconsistent_sum(decimal<15, 8>, decimal<11, 8>, decimal<119, 8>)`


On the other hand, the following are all invalid: 
- `inconsistent_sum(decimal<10, 2>, decimal<10, 3>, decimal<10, 4>)`
- `inconsistent_sum(decimal<10, 2>, i32)`

The above example is showing that there is the implicit constraint across the variadic parameters that the scale `S` must be the same as the output (and thus must all be the same across the variadic parameter). On the other hand, the precision `P` is free to be anything. But all of the parameters do have to be `decimal`. 

Not sure if that was instructive or not 😅

core/src/main/java/io/substrait/expression/VariadicParameterConsistencyValidator.java
```diff
@@ -0,0 +1,74 @@
...
+    int fixedArgCount = func.args().size();
+    if (argumentTypes.size() <= fixedArgCount) {
+      // No variadic arguments, validation passes
+      return;
+    }
+
+    // For CONSISTENT, all variadic arguments must have the same type (ignoring nullability)
+    int firstVariadicArgIdx = Math.max(variadicBehavior.getMin() - 1, 0);
+    for (int i = firstVariadicArgIdx; i < argumentTypes.size() - 1; i++) {
+      Type currentType = argumentTypes.get(i);
```
benbellick, line 86: nit: could we just save the firstVariadicArg one time and then compare all of the rest of them to that single one (instead of comparing across the list pair wise)? I think this amounts to a more informative error message.

core/src/main/java/io/substrait/expression/VariadicParameterConsistencyValidator.java
benbellick: For readability, it may make sense to have two private methods that implement each behavior, then have the public method just be a switch on the parameter consistency options. Just an idea!

core/src/main/java/io/substrait/expression/VariadicParameterConsistencyValidator.java
```diff
@@ -0,0 +1,74 @@
...
+      Type nextType = argumentTypes.get(i + 1);
+      // Normalize both types to nullable for comparison (ignoring nullability)
+      if (!io.substrait.type.TypeCreator.asNullable(currentType)
+          .equals(io.substrait.type.TypeCreator.asNullable(nextType))) {
```
benbellick: nit: IMO this is useful enough to warrant its own method. Could imagine something in `core/src/main/java/io/substrait/type/Type.java` like: 
```java
  /**
   * Compares this type with another type, ignoring nullability differences.
   *
   * @param other the type to compare with
   * @return true if the types are equal when both are treated as nullable
   */
  default boolean equalsIgnoringNullability(Type other) {
    return TypeCreator.asNullable(this).equals(TypeCreator.asNullable(other));
  }
```

Thoughts?

core/src/test/java/io/substrait/expression/VariadicParameterConsistencyTest.java
```diff
@@ -0,0 +1,289 @@
...
+            .urn("extension:test:variadic")
+            .name("test_func")
+            .args(args)
+            .variadic(
+                variadic != null ? java.util.Optional.of(variadic) : java.util.Optional.empty())
+            .returnType(R.I64)
+            .options(java.util.Collections.emptyMap())
+            .build();
+
+    return Expression.ScalarFunctionInvocation.builder()
```
benbellick: nit: [can we use `ExpressionCreator.scalarFunction` here instead?](https://github.com/substrait-io/substrait-java/blob/8845bbff6954c2ba5105469401a94bf54c686f49/core/src/main/java/io/substrait/expression/ExpressionCreator.java#L371)

core/src/test/java/io/substrait/expression/VariadicParameterConsistencyTest.java
```diff
@@ -0,0 +1,289 @@
...
+        .declaration(declaration)
+        .arguments(arguments)
+        .outputType(R.I64)
+        .options(java.util.Collections.emptyList())
+        .build();
+  }
+
+  @Test
+  void testConsistentVariadicWithSameTypes() {
```
benbellick, line 42: Basically every place here where you are using the builders directly, I think it would be better to use the `ExpressionCreator` helpers. E.g. 
`Expression.I64Literal.builder().value(1).build()` -> `ExpressionCreator.i8(false, 1)`

core/src/test/java/io/substrait/extension/VariadicBehaviorTest.java
```diff
@@ -0,0 +1,33 @@
...
+            + "          min: 1\n"
+            + "          parameterConsistency: CONSISTENT\n"
+            + "        return: string\n";
+
+    SimpleExtension.ExtensionCollection collection =
+        SimpleExtension.load("test://example", yamlContent);
+
+    assertEquals(
+        SimpleExtension.VariadicBehavior.ParameterConsistency.CONSISTENT,
+        collection.scalarFunctions().get(0).variadic().get().parameterConsistency());
```
benbellick, line 30: Lets be super thorough and also add a test to make sure that `parameterConsistency: INCONSISTENT` is also passed through.

core/src/test/java/io/substrait/expression/VariadicParameterConsistencyTest.java
```diff
@@ -0,0 +1,289 @@
...
+                    Expression.I64Literal.builder().value(1).build(),
+                    Expression.I64Literal.builder().value(2).nullable(true).build(),
+                    Expression.I64Literal.builder().value(3).nullable(true).build(),
+                    Expression.I64Literal.builder().value(4).build())),
+        "Consistent variadic with same types but different nullability should pass");
+  }
+
+  @Test
+  void testConsistentVariadicWithDifferentTypes() {
+    // Function: test_func(i64, i64...) with CONSISTENT parameterConsistency
```
benbellick, line 79: All of these comments which explicitly show a text representation of the function being tested are good 👍

core/src/test/java/io/substrait/expression/VariadicParameterConsistencyTest.java
```diff
@@ -0,0 +1,289 @@
...
+    List<SimpleExtension.Argument> args =
+        List.of(SimpleExtension.ValueArgument.builder().value(R.I64).name("arg1").build());
+
+    SimpleExtension.VariadicBehavior variadic =
+        ImmutableSimpleExtension.VariadicBehavior.builder()
+            .min(1)
+            .parameterConsistency(SimpleExtension.VariadicBehavior.ParameterConsistency.CONSISTENT)
+            .build();
+
+    // Variadic arguments have different types - should fail
```
benbellick: But let's drop all of these comments which repeat the message in the assert.

benbellick (changes requested):
A lot of my comments are about relatively minor things. But the most important point is about the meaning of INCONSISTENT. I do think that the upstream could benefit from a bit more clarity there. Thanks for your work and let me know if anything I have said is unclear!

core/src/main/java/io/substrait/expression/VariadicParameterConsistencyValidator.java
Adam-Alani: I'll open a different issue for this and a follow up PR, and we can add a big TODO comment

Adam-Alani: https://github.com/substrait-io/substrait-java/issues/633

core/src/main/java/io/substrait/expression/VariadicParameterConsistencyValidator.java
```diff
@@ -0,0 +1,99 @@
...
+import java.util.List;
+import java.util.Optional;
+import java.util.stream.Collectors;
+
+/**
+ * Helper class for validating variadic parameter consistency in function invocations. Validates
+ * that when parameterConsistency is CONSISTENT, all variadic arguments have the same type (ignoring
+ * nullability).
+ */
+public class VariadicParameterConsistencyValidator {
```
benbellick:
```suggestion
class VariadicParameterConsistencyValidator {
```
[This way we keep it package private](https://docs.oracle.com/javase/tutorial/java/javaOO/accesscontrol.html).

core/src/main/java/io/substrait/expression/VariadicParameterConsistencyValidator.java
benbellick: You might as well put the ticket number in the comment as well.

core/src/main/java/io/substrait/expression/AggregateFunctionInvocation.java
```diff
@@ -25,6 +25,16 @@ public Type getType() {
 
   public abstract Expression.AggregationInvocation invocation();
 
+  /**
+   * Validates that variadic arguments satisfy the parameter consistency requirement. When
+   * CONSISTENT, all variadic arguments must have the same type (ignoring nullability). When
+   * INCONSISTENT, arguments can have different types.
```
benbellick, line 31: Lets clarify this to say something other than "have different types". Maybe, "have any type satisfying the type constraints" or something?

core/src/main/java/io/substrait/expression/VariadicParameterConsistencyValidator.java
```diff
@@ -0,0 +1,99 @@
...
+            .filter(arg -> arg instanceof Expression || arg instanceof Type)
+            .map(
+                arg -> {
+                  if (arg instanceof Expression) {
+                    return ((Expression) arg).getType();
+                  } else {
+                    return (Type) arg;
+                  }
+                })
+            .collect(Collectors.toList());
```
benbellick, line 62: I wonder if all of this filtering logic could be simplified if we just introduced an `EnumArg` type?

benbellick: I think its fine to leave this for now, but just making note of it.

core/src/main/java/io/substrait/expression/VariadicParameterConsistencyValidator.java
```diff
@@ -0,0 +1,99 @@
...
+
+    // Count how many Expression/Type arguments are in the fixed arguments (before variadic)
+    // Note: func.args() includes all argument types (Expression, Type, EnumArg), but we only
+    // care about Expression/Type arguments for type consistency checking
+    int fixedTypeArgCount = 0;
```
benbellick: I think a better name may be something like `nonVariadicArgCount`? It would be nice to if you had an example e.g. sum(i32...) has a `nonVariadicArgCount` of 0. Or is it 1?

The reason why I am suggested an example is just that it isn't entirely clear to me whether or not the first non-variadic example is included in this list or not without closely reading the code.

core/src/main/java/io/substrait/expression/VariadicParameterConsistencyValidator.java
```diff
@@ -0,0 +1,99 @@
...
+    for (int i = 0; i < func.args().size() && i < arguments.size(); i++) {
+      FunctionArg arg = arguments.get(i);
+      if (arg instanceof Expression || arg instanceof Type) {
+        fixedTypeArgCount++;
+      }
+    }
```
benbellick: Take it or leave it, but since you are already doing it functional style already above. But yeah truly feel free to ignore if these seems more confusing. 

```suggestion
    int fixedTypeArgCount =
        (int)
            arguments.stream()
                .limit(func.args().size())
                .filter(arg -> arg instanceof Expression || arg instanceof Type)
                .count();

```

core/src/test/java/io/substrait/expression/VariadicParameterConsistencyTest.java
benbellick: Sorry unresolving this one, did you maybe miss it?

core/src/test/java/io/substrait/expression/VariadicParameterConsistencyTest.java
benbellick: Also unresolving this one, did you maybe miss it?

core/src/test/java/io/substrait/extension/VariadicBehaviorTest.java
benbellick: Also unresolving this one, think its also missed.

core/src/main/java/io/substrait/expression/AggregateFunctionInvocation.java
benbellick: Unresolving this as I think the comment is still the same as before.

core/src/main/java/io/substrait/expression/Expression.java
```diff
@@ -799,6 +799,16 @@ public Type getType() {
       return outputType();
     }
 
+    /**
+     * Validates that variadic arguments satisfy the parameter consistency requirement. When
+     * CONSISTENT, all variadic arguments must have the same type (ignoring nullability). When
+     * INCONSISTENT, arguments can have different types.
```
benbellick, line 836: This comment should (like above) clarify what different types means.

core/src/main/java/io/substrait/expression/AggregateFunctionInvocation.java
benbellick: How about something like this? 
```suggestion
   * INCONSISTENT, arguments can have different types, provided that they satisfy the 
   * implied type constraints. 
```

core/src/main/java/io/substrait/expression/Expression.java
```diff
@@ -840,6 +850,16 @@ public Type getType() {
 
     public abstract AggregationInvocation invocation();
 
+    /**
+     * Validates that variadic arguments satisfy the parameter consistency requirement. When
+     * CONSISTENT, all variadic arguments must have the same type (ignoring nullability). When
+     * INCONSISTENT, arguments can have different types.
```
benbellick, line 887: and here, too

core/src/main/java/io/substrait/expression/Expression.java
```diff
@@ -799,6 +799,16 @@ public Type getType() {
...
+     */
+    @Value.Check
+    protected void check() {
+      VariadicParameterConsistencyValidator.validate(declaration(), arguments());
+    }
+
```
benbellick, line 842: Looks like you covered both variadic `Scalar` functions and variadic `Window` functions. Any reason to not also cover variadic `Aggregate` functions?

core/src/main/java/io/substrait/expression/AggregateFunctionInvocation.java
benbellick: Ah I remember now, this is a bit thornier than I thought because we are going to implement this in a separate PR. I think it's fine to leave this as is, but I'm going to make a comment on the other ticket about it.

benbellick: https://github.com/substrait-io/substrait-java/issues/633#issuecomment-3656678208

core/src/main/java/io/substrait/expression/Expression.java
benbellick: Similarly to above, you can ignore this and save it for the subsequent ticket: https://github.com/substrait-io/substrait-java/issues/633#issuecomment-3656678208

core/src/main/java/io/substrait/expression/Expression.java
benbellick: Again, ignore this comment, and save for the subsequent ticket: https://github.com/substrait-io/substrait-java/issues/633#issuecomment-3656678208

core/src/main/java/io/substrait/expression/Expression.java
Adam-Alani: I might be missing something, but I did add the check() in `AggregateFunctionInvocation` class, should it be somewhere else too?

benbellick: Ah I misunderstood, I see now that `AggregateFunctionInvocation` is in a separate file. Thanks!

benbellick (approved):
LGTM!

Issue #375, opened 2025-04-07
support for Nested expressions


vbarua:
We should be able to convert Nested expressions from protobuf to Calcite and back.

https://github.com/substrait-io/substrait/blob/4a472447392d9beba0401987b4d845079241c30d/proto/substrait/algebra.proto#L1128-L1167

PR #627, opened 2025-12-01
feat: support Nested Lists

gord02:
This PR is to add support for the Expression Nested List type. 

[Issue](https://github.com/substrait-io/substrait-java/issues/375)

Key Changes:
- Substrait pojo to proto and back conversion 
- Substrait pojo to calcite and back conversion 
- Added new type in Isthmus to support Nested functions list during the conversion to Calcite from Substrait 
- Added code to not do a project remapping on an empty table 

Testing:
- added tests for roundtrip between substrait to proto 
- added tests for roundtrip between substrait and calcite 

./gradlew test --tests io.substrait.isthmus.NestedExpressionsTest --debug-jvm

### DIFF
--- a/core/src/test/java/io/substrait/type/proto/ProjectRelRoundtripTest.java
+++ b/core/src/test/java/io/substrait/type/proto/ProjectRelRoundtripTest.java
@@ -16,8 +16,6 @@ public class ProjectRelRoundtripTest extends TestBase {
           Arrays.asList("col_a", "col_b", "col_c", "col_d"),
           Arrays.asList(R.I64, R.FP64, R.STRING, R.I32));
 
-  final Rel emptyTable = b.emptyScan();
-
   @Test
   void simpleProjection() {
     // Project single field


core/src/main/java/io/substrait/expression/Expression.java
```diff
@@ -922,6 +929,26 @@ public <R, C extends VisitationContext, E extends Throwable> R accept(
     }
   }
 
+  @Value.Immutable
+  abstract class NestedList implements Nested {
+    public abstract List<Expression> values();
+
+    @Override
+    public Type getType() {
+      return Type.withNullability(nullable()).list(values().get(0).getType());
```
benbellick, line 952: This is totally fine but just a note that we should then actually make sure somewhere that all of these values have the same type.

core/src/main/java/io/substrait/expression/proto/ExpressionProtoConverter.java
```diff
@@ -496,6 +496,26 @@ public Expression visit(
         .build();
   }
 
+  private Expression nested(Consumer<Expression.Nested.Builder> consumer) {
+    Expression.Nested.Builder builder = Expression.Nested.newBuilder();
+    builder.setNullable(builder.getNullable());
+    consumer.accept(builder);
+    return Expression.newBuilder().setNested(builder).build();
+  }
```
benbellick: nit: this is personal taste but I think that having this private method is more confusing than just deleting it and having the builder inline in the function below. 

I realize that this is how `lit` is handled above, but I also think there that the amount of code saved by introducing that helper is not worth the confusion. Though the `lit` helper can be better justified by the fact that there are 30+ places where it is called. Here, however, it is just once. And once your work is done, it will be at most three (for `Struct`, `List`, and `Map`).

core/src/main/java/io/substrait/expression/proto/ProtoExpressionConverter.java
```diff
@@ -361,6 +363,15 @@ private WindowBound toWindowBound(io.substrait.proto.Expression.WindowFunction.B
     }
   }
 
+  public Expression.Nested from(io.substrait.proto.Expression.Nested nested) {
+    if (nested.getNestedTypeCase() == io.substrait.proto.Expression.Nested.NestedTypeCase.LIST) {
```
benbellick: nit (style): lets use a switch expression like the below `from` implementation to be consistent (we will later want to handle the other nested types).

core/src/test/java/io/substrait/type/proto/NestedListExpressionTest.java
```diff
@@ -0,0 +1,52 @@
...
+            .build();
+    expressionList.add(literalNestedList);
+
+    io.substrait.relation.Project project =
+        io.substrait.relation.Project.builder()
+            .expressions(expressionList)
+            .input(b.emptyScan())
+            .build();
+
+    verifyRoundTrip(project);
```
benbellick, line 57: Totally minor thing, but in a PR I am working on, [I was going to add a `verifyRoundTrip` overload for just expressions themselves](https://github.com/substrait-io/substrait-java/pull/613/files#diff-4d5ac1f0e1e87e6b8d11b570e48cadaed58d5151ed8d208dacedfeef5ab8b26fR45) so that you don't have to wrap everything in a relation. Could be useful here as well, but feel free to ignore.

core/src/test/java/io/substrait/type/proto/NestedListExpressionTest.java
```diff
@@ -0,0 +1,52 @@
...
+            .addValues(nonLiteralExpression)
+            .addValues(nonLiteralExpression)
+            .build();
+    expressionList.add(nonLiteralNestedList);
+
+    io.substrait.relation.Project project =
+        io.substrait.relation.Project.builder()
+            .expressions(expressionList)
+            .input(b.emptyScan())
+            .build();
```
benbellick: same nit:

```suggestion
  void nonLiteralNestedListTest() {
    Expression.NestedList nonLiteralNestedList =
        Expression.NestedList.builder()
            .addValues(nonLiteralExpression)
            .addValues(nonLiteralExpression)
            .build();

    io.substrait.relation.Project project =
        io.substrait.relation.Project.builder()
            .addExpressions(nonLiteralNestedList)
            .input(b.emptyScan())
            .build();
```

core/src/test/java/io/substrait/type/proto/NestedListExpressionTest.java
```diff
@@ -0,0 +1,52 @@
...
+            .addValues(literalExpression)
+            .addValues(literalExpression)
+            .build();
+    expressionList.add(literalNestedList);
+
+    io.substrait.relation.Project project =
+        io.substrait.relation.Project.builder()
+            .expressions(expressionList)
+            .input(b.emptyScan())
+            .build();
```
benbellick: nit (style):

```suggestion
    Expression.NestedList literalNestedList =
        Expression.NestedList.builder()
            .addValues(literalExpression)
            .addValues(literalExpression)
            .build();

    io.substrait.relation.Project project =
        io.substrait.relation.Project.builder()
            .addExpressions(literalNestedList)
            .input(b.emptyScan())
            .build();
```

core/src/main/java/io/substrait/expression/Expression.java
benbellick: Also, what happens when you have an empty list?

benbellick: Yeah I think we are missing a check to make sure all of the types are the same. Something like: 
```java
    @Value.Check
    protected void check() {
      if (!values().isEmpty()) {
        Type firstType = values().get(0).getType();
        assert values().stream().allMatch(v -> v.getType().equals(firstType))
            : "All values in NestedList must have the same type";
      }
    }
```

Or even something like this if you wanted to play code golf 😅:
```java
    @Value.Check
    protected void check() {
      assert values().stream().map(Expression::getType).distinct().count() <= 1
          : "All values in NestedList must have the same type";
    }
```

Lets definitely make sure to include a test that shows that we cannot construct with mismatched types somewhere.

core/src/main/java/io/substrait/expression/Expression.java
```diff
@@ -922,6 +929,34 @@ public <R, C extends VisitationContext, E extends Throwable> R accept(
...
+  abstract class NestedList implements Nested {
+    public abstract List<Expression> values();
+
+    @Value.Check
+    protected void check() {
+      assert !values().isEmpty() : "To specify an empty list, use Literal.empty_list";
+
+      assert values().stream().map(Expression::getType).distinct().count() <= 1
+          : "All values in NestedList must have the same type";
+    }
```
benbellick, line 948:
```suggestion
  /**
   * A nested list expression with one or more elements.
   *
   * <p>Note: This class cannot be used to construct an empty list. To create an empty list, use
   * {@link ExpressionCreator#emptyList(boolean, Type)} which returns an {@link EmptyListLiteral}.
   */
  @Value.Immutable
  abstract class NestedList implements Nested {
    public abstract List<Expression> values();

    @Value.Check
    protected void check() {
      assert !values().isEmpty() : "To specify an empty list, use ExpressionCreator.emptyList()";

      assert values().stream().map(Expression::getType).distinct().count() <= 1
          : "All values in NestedList must have the same type";
    }
```

Would be great to add a similarly nice javadoc to the ExpressionCreator.java method which constructs a `NestedList`. Thanks!

core/src/main/java/io/substrait/expression/proto/ProtoExpressionConverter.java
```diff
@@ -361,6 +363,16 @@ private WindowBound toWindowBound(io.substrait.proto.Expression.WindowFunction.B
...
   }
 
+  public Expression.Nested from(io.substrait.proto.Expression.Nested nested) {
+    switch (nested.getNestedTypeCase()) {
+      case LIST:
+        List<Expression> list =
+            nested.getList().getValuesList().stream().map(this::from).collect(Collectors.toList());
+        return ExpressionCreator.nestedList(nested.getNullable(), list);
+    }
+    return null;
```
benbellick: Lets throw an exception here instead of returning `null`. (this is how the below `from` function handles it.)

core/src/test/java/io/substrait/type/proto/NestedListExpressionTest.java
```diff
@@ -0,0 +1,94 @@
...
+    verifyRoundTrip(project);
+  }
+
+  @Test
+  void EmptyNestedListTest() {
+    ImmutableExpression.NestedList.Builder builder = Expression.NestedList.builder();
+    assertThrows(AssertionError.class, builder::build);
+  }
```
benbellick, line 41: 🚀 good to have this tested!

core/src/test/java/io/substrait/type/proto/NestedListExpressionTest.java
benbellick, line 1: all in all great test set 😎

benbellick (changes requested):
Just left a few more comments but the `core/` part is looking great!

isthmus/src/main/java/io/substrait/isthmus/NestedFunctions.java
```diff
@@ -0,0 +1,19 @@
...
+import org.apache.calcite.sql.type.ReturnTypes;
+
+/** Substrait-specific extension function for the Expression Nested List type */
+public class NestedFunctions {
+
+  public static final SqlFunction NESTED_LIST =
+      new SqlFunction(
+          "nested_list",
+          SqlKind.OTHER_FUNCTION,
+          ReturnTypes.BOOLEAN,
```
benbellick: So again, I don't really know the calcite stuff that well, but at the very least this seems unusual. What do nested list types have to do with `SqlFunction`? And why would the return type be `BOOLEAN`? 

Perhaps this is some kind of workaround to get around an inherent limitation of calcite / isthmus. But if that is the case, then there should be some thorough paragraph somewhere explaining why this would be necessary.

gord02: Good callout! That code was not safe and has been reworked. We do need a new type to represent the nested list for round trip tests, but the underlying type is still the same as ARRAY_VALUE_CONSTRUCTOR. So I refactored it accordingly to more safely handle conversion. Thanks!

core/src/main/java/io/substrait/expression/proto/ProtoExpressionConverter.java
```diff
@@ -361,6 +363,17 @@ private WindowBound toWindowBound(io.substrait.proto.Expression.WindowFunction.B
...
   }
 
+  public Expression.Nested from(io.substrait.proto.Expression.Nested nested) {
+    switch (nested.getNestedTypeCase()) {
+      case LIST:
+        List<Expression> list =
+            nested.getList().getValuesList().stream().map(this::from).collect(Collectors.toList());
+        return ExpressionCreator.nestedList(nested.getNullable(), list);
+      default:
+        throw new IllegalStateException("Unimplemented nested type: " + nested.getNestedTypeCase());
```
vbarua: minor: We can use `UnsupportedOperationException` here, which better matches your error message well 🙂

core/src/main/java/io/substrait/expression/ExpressionCreator.java
```diff
@@ -281,6 +281,17 @@ public static Expression.StructLiteral struct(boolean nullable, Expression.Liter
     return Expression.StructLiteral.builder().nullable(nullable).addFields(values).build();
   }
 
+  /**
+   * Creator a nested list expression with one or more elements.
```
vbarua:
```suggestion
   * Creates a nested list expression with one or more elements.
```

core/src/test/java/io/substrait/type/proto/NestedListExpressionTest.java
```diff
@@ -0,0 +1,94 @@
...
+import io.substrait.expression.ImmutableExpression;
+import org.junit.jupiter.api.Test;
+
+class NestedListExpressionTest extends TestBase {
+  io.substrait.expression.Expression literalExpression =
+      Expression.BoolLiteral.builder().value(true).build();
+  Expression.ScalarFunctionInvocation nonLiteralExpression = b.add(b.i32(7), b.i32(42));
+
+  @Test
+  void DifferentTypedLiteralsNestedListTest() {
```
vbarua:
```suggestion
  void rejectNestedListWithElementsOfDifferentTypes() {
```

minor suggestion: if we include the expected behaviour in the test name, it's easier to see at at glance what its testing. Similar suggestions for test below.

core/src/test/java/io/substrait/type/proto/NestedListExpressionTest.java
```diff
@@ -0,0 +1,94 @@
...
+    io.substrait.relation.Project project =
+        io.substrait.relation.Project.builder()
+            .addExpressions(builder.build())
+            .input(b.emptyScan())
+            .build();
+    verifyRoundTrip(project);
+  }
+
+  @Test
+  void EmptyNestedListTest() {
```
vbarua:
```suggestion
  void rejectEmptyNestedListTest() {
```

core/src/test/java/io/substrait/type/proto/NestedListExpressionTest.java
```diff
@@ -0,0 +1,94 @@
...
+
+  @Test
+  void DifferentTypedLiteralsNestedListTest() {
+    ImmutableExpression.NestedList.Builder builder =
+        Expression.NestedList.builder().addValues(literalExpression).addValues(b.i32(12));
+    assertThrows(AssertionError.class, builder::build);
+  }
+
+  @Test
+  void SameTypedLiteralsNestedListTest() {
```
vbarua:
```suggestion
  void acceptNestedListWithElementsOfSameType() {
```

Beyond the name suggestion for this test, it is a bit redundant because it doesn't really verify anything differently than your normal tests below do.

isthmus/src/main/java/io/substrait/isthmus/expression/NestedExpressionConverter.java
```diff
@@ -0,0 +1,33 @@
...
+import io.substrait.expression.Expression;
+import io.substrait.isthmus.CallConverter;
+import java.util.List;
+import java.util.Optional;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+import org.apache.calcite.rex.RexCall;
+import org.apache.calcite.rex.RexNode;
+
+public class NestedExpressionConverter implements CallConverter {
```
vbarua: I think its a reasonable choice not to have this be a nested class in CallConverters. That said, could you rename this to `NestedExpressionCallConverter` to make differentiate from the other things we call converters.

isthmus/src/main/java/io/substrait/isthmus/expression/NestedExpressionConverter.java
```diff
@@ -0,0 +1,33 @@
...
+
+  public NestedExpressionConverter() {}
+
+  @Override
+  public Optional<Expression> convert(
+      RexCall call, Function<RexNode, Expression> topLevelConverter) {
+
+    if (!call.getOperator().getName().equals("NESTEDLIST")) {
```
asolimando:
```suggestion
    if (!(call.getOperator() instanceof NestedListConstructor)) {
```

string-comparison is brittle and should be avoided

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelVisitor.java
```diff
@@ -179,6 +179,10 @@ public Rel visit(org.apache.calcite.rel.core.Project project) {
             .map(this::toExpression)
             .collect(java.util.stream.Collectors.toList());
 
+    //    if there is no input fields, don’t put a remapping on it
```
asolimando:
```suggestion
    // if there are no input fields, don’t put a remapping on it
```
nit

isthmus/src/main/java/io/substrait/isthmus/expression/NestedExpressionConverter.java
vbarua: minor suggestion: The "NESTEDLIST" name string here has to be the same as the name you pass in the NestedListConstructor.

If you add a field to that class like
```java
  public static final String NAME = "NESTEDLIST"; 
```

we could replace all usages of that string with that variable, and make this implicit link explicit in the code.

isthmus/src/test/java/io/substrait/isthmus/NestedExpressionsTest.java
```diff
@@ -0,0 +1,76 @@
...
+      Expression.BoolLiteral.builder().value(true).build();
+  Expression.ScalarFunctionInvocation nonLiteralExpression = b.add(b.i32(7), b.i32(42));
+  Expression.ScalarFunctionInvocation nonLiteralExpression2 = b.add(b.i32(3), b.i32(4));
+
+  final List<Type> tableType = List.of(R.I32, R.FP32, N.STRING, N.BOOLEAN);
+  final Rel commonTable = b.namedScan(List.of("example"), List.of("a", "b", "c", "d"), tableType);
+  final Rel emptyTable = b.emptyScan();
+
+  @Test
+  void NestedListWithJustLiteralsTest() {
```
asolimando:
```suggestion
  void NestedListWithLiteralsTest() {
```
nit: but reads better IMO

isthmus/src/main/java/io/substrait/isthmus/NestedListConstructor.java
```diff
@@ -0,0 +1,34 @@
...
+import org.apache.calcite.sql.SqlKind;
+import org.apache.calcite.sql.SqlOperatorBinding;
+import org.apache.calcite.sql.fun.SqlMultisetValueConstructor;
+import org.apache.calcite.sql.type.SqlTypeUtil;
+import org.apache.calcite.sql.validate.SqlValidatorUtil;
+
+/**
+ * Substrait-specific constructor to map back to the Expression NestedList type in Substrait. This
+ * constructor creates a special type of SqlKind.ARRAY_VALUE_CONSTRUCTOR for lists that store
+ * non-literal expressions.
```
asolimando: Conceptually and by looking at the tests (`NestedListWithJustLiteralsTest`), it seems this can indeed handle literal expressions, can you double check?

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelVisitor.java
```diff
@@ -179,6 +179,10 @@ public Rel visit(org.apache.calcite.rel.core.Project project) {
...
+    if (project.getInput().getRowType().getFieldCount() == 0) {
+      return Project.builder().expressions(expressions).input(apply(project.getInput())).build();
+    }
```
vbarua, line 185: This change looks good, but is a bit out of place in this PR. From experimenting locally it seems that one of your tests fails locally with it because a pointless remap is introduced.

I do think it's a good change, and we should keep it in this PR, but could you add an explicit test for this features along the lines of `avoidProjectRemapOnEmptyInput()` to ProjectRelRoundtripTest.

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelVisitor.java
vbarua: minor/pedantic: small rewording + avoiding extra whitespace
```suggestion
    // if there are no input fields, no remap is necessary
```

isthmus/src/test/java/io/substrait/isthmus/NestedExpressionsTest.java
vbarua: General Java style point. Tests are methods, so their names should be in `camelCase`. Only constructor functions that share the name of the class are capitalized

```suggestion
  void nestedListWithJustLiteralsTest() {
```

isthmus/src/test/java/io/substrait/isthmus/NestedExpressionsTest.java
```diff
@@ -0,0 +1,76 @@
...
+            .input(emptyTable)
+            .build();
+
+    RelNode relNode = substraitToCalcite.convert(project); //    substrait rel to calcite
+    Rel project2 = SubstraitRelVisitor.convert(relNode, extensions); // calcite to substrait
+    assertEquals(project, project2); // pojo -> calcite -> pojo
+  }
+
+  @Test
+  void NestedListWithNonLiteralsTest() {
```
vbarua:
```suggestion
  void nestedListWithNonLiteralsTest() {
```

isthmus/src/test/java/io/substrait/isthmus/NestedExpressionsTest.java
```diff
@@ -0,0 +1,76 @@
...
+
+    io.substrait.relation.Project project =
+        io.substrait.relation.Project.builder()
+            .expressions(expressionList)
+            .input(emptyTable)
+            .build();
+
+    RelNode relNode = substraitToCalcite.convert(project); //    substrait rel to calcite
+    Rel project2 = SubstraitRelVisitor.convert(relNode, extensions); // calcite to substrait
+    assertEquals(project, project2); // pojo -> calcite -> pojo
```
vbarua: I would replace these 3 lines with just
```java
assertFullRoundTrip(project);
```

which is a utility helper we have that checks more conversion than what you are checking now. Same for the test below.

isthmus/src/test/java/io/substrait/isthmus/NestedExpressionsTest.java
```diff
@@ -0,0 +1,76 @@
...
+        io.substrait.relation.Project.builder()
+            .expressions(expressionList)
+            .input(commonTable)
+            .remap(Rel.Remap.of(Collections.singleton(4)))
+            .build();
+
+    RelNode relNode = substraitToCalcite.convert(project); //    substrait rel to calcite
+    Rel project2 = SubstraitRelVisitor.convert(relNode, extensions); // calcite to substrait
+    assertEquals(project, project2); // pojo -> calcite -> pojo
+  }
```
vbarua, line 97: One more test that might be good to include in this file is a NestedList that includes as an expression a field from the input using a FieldReference.

isthmus/src/main/java/io/substrait/isthmus/NestedListConstructor.java
```diff
@@ -0,0 +1,34 @@
...
+ */
+public class NestedListConstructor extends SqlMultisetValueConstructor {
```
asolimando: I guess you picked `SqlMultisetValueConstructor` instead of the more natural `SqlArrayValueConstructor` because [CallConverters.java#L144](https://github.com/substrait-io/substrait-java/blob/c5b1536cc346706afa3dd36e568455f2ef843e5d/isthmus/src/main/java/io/substrait/isthmus/expression/CallConverters.java#L144) would then match on your `NestedList` and treat it as the regular `SqlArrayValueConstructor` (then invoke [LiteralConstructorConverter.java#L32](https://github.com/substrait-io/substrait-java/blob/c5b1536cc346706afa3dd36e568455f2ef843e5d/isthmus/src/main/java/io/substrait/isthmus/expression/LiteralConstructorConverter.java#L32) which is not what we want here).

`SqlMultisetValueConstructor` is conceptually wrong as a multiset is radically different from an array/list in standard SQL, on top of that I imagine that tomorrow you might want to support something similar and hit the same issue you are trying to avoid here by using this class in the first place.

The impedance mismatch is that in SQL (and Calcite), arrays and lists are technically the same entity, while IIRC in Substrait they are treated as different entities (@benbellick can you confirm this?).

By looking at `LiteralConstructorConverter`, there is an implicit assumption that arrays store only literals, we go down that route without checking if elements in the array are really literals ([LiteralConstructorConverter.java#L62](https://github.com/substrait-io/substrait-java/blob/c5b1536cc346706afa3dd36e568455f2ef843e5d/isthmus/src/main/java/io/substrait/isthmus/expression/LiteralConstructorConverter.java#L62)).

It's probably enough to change `LiteralConstructorConverter::toNonEmptyListLiteral` to something like this (haven't tested it):
```
private Optional<Expression> toNonEmptyListLiteral(
      RexCall call, Function<RexNode, Expression> topLevelConverter) {
    List<Expression> expressions = call.operands.stream()
        .map(topLevelConverter)

    // Check if all operands are actually literals
    if (expressions.stream().allMatch(e -> e instanceof Expression.Literal)) {
      return Optional.of(ExpressionCreator.list(
          call.getType().isNullable(),
          expressions.stream()
              .map(e -> (Expression.Literal) e)
              .collect(Collectors.toList())));
    }

    return Optional.empty();
  }
```

I suggest to extend `SqlArrayValueConstructor` (which, I know, extends `SqlMultisetValueConstructor`, but still), then fix `LiteralConstructorConverter` as suggested, so that we can continue with `NestedExpressionConverter` which comes just after, and we should be good

isthmus/src/test/java/io/substrait/isthmus/NestedExpressionsTest.java
```diff
@@ -0,0 +1,76 @@
...
+import io.substrait.type.Type;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import org.apache.calcite.rel.RelNode;
+import org.junit.jupiter.api.Test;
+
+class NestedExpressionsTest extends PlanTestBase {
+
+  protected static final SimpleExtension.ExtensionCollection defaultExtensionCollection =
```
asolimando, line 22: I think we could add a few more tests.

1. mixed literal and non-literal expressions (IIRC this is the key scenario that justifies having both ListLiteral and NestedList), so here I'd test an "array" containing both literals and runtime expressions (like function calls) in the same list
2. field references: not sure it's legal in this context, but if it is, it would be nice to add a test for arrays containing references to input table columns (common use case for building dynamic arrays from row data)
3. ListLiteral preservation: due to what discussed [here](https://github.com/substrait-io/substrait-java/pull/627#discussion_r2607829240), I'd rather make sure creating a `ListLiteral` (all literals) round-trip it through Calcite, and assert that it remains a `ListLiteral` and was not converted to `NestedList` (especially if you accept my suggestion from the aforementioned comment)

Additional tests could be:
- single-element lists,
- nullable round-trip preservation through Calcite
- different data types beyond integers and booleans
- type coercion/type unification

isthmus/src/main/java/io/substrait/isthmus/NestedListConstructor.java
gord02: It can handle both literals and non-literals. I'll update the comment to avoid the confusion

isthmus/src/main/java/io/substrait/isthmus/NestedListConstructor.java
gord02: Just to make sure I am understanding it correctly, let me know your thoughts on the following scenarios: We want to ensure that the roundtrip of both a nested list with literals and non-literals are both returned to a nested list. If the literalConstructorConverter is run first on a list of just literals, then it would pass and then wouldn't be mapped back to a nested list. In the other case, where the nestedExpressionConverter is run first, the literal lists that were originally not a NestedList would become a nested list. Does the above account for this or is the difference not important? 

Also, is there a way to meaningfully extend the SqlArrayValueConstructor class? Its definition is bare with just a constructor to its parent type.

benbellick: There is no distinction between lists and arrays in substrait. Only `list` is a legitimate type. Array is informally used in some places in the docs, but that doesn't actually exist as a distinct type.

isthmus/src/test/java/io/substrait/isthmus/NestedExpressionsTest.java
gord02: The proto algebra for Nested List describes it as "homogeneously-typed list of one or more expressions". This covers the mixed literal and non-literal expressions and type coercion/type unification cases which are in the NestedListExpressionTest file(which aren't allowed).

isthmus/src/main/java/io/substrait/isthmus/NestedListConstructor.java
vbarua: To your point about
> We want to ensure that the roundtrip of both a nested list with literals and non-literals are both returned to a nested list.

The design you have works to solve the problem of "NestedList in is always NestedList out". Thinking about what Alessandro is saying though, I actually don't think it's worth preserving that. There is a little bit of redundancy on the Substrait side in that list containing only literal values can be expressed equivalently as a Literal.List or a Nested.List.

In Calcite, these can both be mapped a SqlArrayValueConstructor. You introduced your NestedListConstructor to be able to distinguish between the two incoming cases so as to roundtrip them, but what if instead of doing that we just said:
1. A `SqlArrayValueConstructor` call with all literals gets turned into a Substrait Literal.List
2. A `SqlArrayValueConstructor` call with _any_ non-literals gets turned into Substrat Nested.List

This isn't as nice from a roundtrip perspective, but from a pure Calcite perspective we're mapping to the most specific Substrait construct that we can use, which is better in my opinion.

One thing I would push for that's different from what Alessandro is suggesting is to have a single call converter `SqlArrayValueConstructorCallConverter` that combines the array converting code in LiteralConstructorConverter and your code in NestedExpressionCallConverter. Effectively, let's not split the code that handles arrays with only literals, and arrays with non-literals, and just put it all in one class so that it's easy to see how the behaviour works and you don't have to look at 2 different converters for 1 Calcite construct.

asolimando: Thanks @benbellick and @vbarua for the clarification, the fact that only `List` exists in Substrait actually makes things easier, as we can map it to `Array` in standard SQL (and in Calcite, where it's informally called `List` too sometimes).

In light of this clarification, my previous suggestion is an unneeded over complication, and I fully agree with Victor's proposal.

isthmus/src/test/java/io/substrait/isthmus/NestedExpressionsTest.java
```diff
@@ -66,11 +65,97 @@ void NestedListWithNonLiteralsTest() {
...
+    assertFullRoundTrip(project);
+  }
+
+  @Test
+  void nestedListWithStringLiteralsTest() {
+    Expression.NestedList nestedList =
+        Expression.NestedList.builder().addValues(b.str("xzy")).addValues(b.str("abc")).build();
+
+    Rel project =
+        io.substrait.relation.Project.builder()
```
asolimando: Nit: I am not very familiar with coding style conventions in substrait-java, but I generally prefer to use imports over fully qualified names for Java classes (in case you accept to change it, I saw this pattern in quite a few places, it should be fixed consistently)

isthmus/src/test/java/io/substrait/isthmus/NestedExpressionsTest.java
asolimando: (I don't have sufficient permissions to resolve the conversation but I am happy with the change)

isthmus/src/test/java/io/substrait/isthmus/NestedExpressionsTest.java
vbarua: This is a pattern we use sometimes in substrait-java, _specifically_ when we have to deal with both Calcite and Substrait relations in the same file. An example of this is the SubstraitRelVisitor, which uses both
```
io.substrait.relation.Project;
org.apache.calcite.rel.core.Project
```
so in the code we reference the Substrait Project using `Project` and the Calcite Project using `org.apache.calcite.rel.core.Project` to help disambiguate them.

In this code though, because we don't have a conflict the general guidance would be to import
```
import io.substrait.relation.Project;
```
and use that.

isthmus/src/main/java/io/substrait/isthmus/NestedListConstructor.java
gord02: I created a new file SqlArrayValueConstructorCallConverter that has the code from the LiteralConstructorConverter and the nestedListConstructor and therefore removed both files. I also updated the tests to check for the listLiteral in the event of converting from a nestedList of literals, and compared the resulting list of expressions to the beginning list.

core/src/test/java/io/substrait/type/proto/ProjectRelRoundtripTest.java
```diff
@@ -146,4 +148,11 @@ void emptyProjection() {
 
     verifyRoundTrip(projection);
   }
+
+  @Test
+  void avoidProjectRemapOnEmptyInput() {
+    Rel projection =
+        Project.builder().input(emptyTable).addExpressions(b.add(b.i32(1), b.i32(2))).build();
+    verifyRoundTrip(projection);
```
vbarua: This doesn't actually test the code you added to avoid remaps on empty inputs. If I comment out that code, this test still passes. The `verifyRoundTrip` method only checks the conversion to protobuf and back, it doesn't pass through Calcite which is where you've added your code.

We need to move this test into the Isthmus tests. `FetchTest` is an example file with tests for Fetch relation. You could add a `ProjectTest` file in that directory, and then use the `assertFullRoundTrip`, which should run your code.

isthmus/src/main/java/io/substrait/isthmus/expression/CallConverters.java
```diff
@@ -141,7 +141,7 @@ public static List<CallConverter> defaults(TypeConverter typeConverter) {
         CallConverters.CASE,
         CallConverters.CAST.apply(typeConverter),
         CallConverters.REINTERPRET.apply(typeConverter),
-        new LiteralConstructorConverter(typeConverter));
```
vbarua, line 144: I think we still need the LiteralConstructorConverter to handle `SqlMapValueConstructor`. Nothing breaks if you remove this?

isthmus/src/main/java/io/substrait/isthmus/expression/ExpressionRexConverter.java
```diff
@@ -320,6 +321,24 @@ public RexNode visit(Expression.ListLiteral expr, Context context) throws Runtim
...
 
+  @Override
+  public RexNode visit(Expression.NestedList expr, Context context) {
+    List<RexNode> args =
+        expr.values().stream().map(e -> e.accept(this, context)).collect(Collectors.toList());
+
+    // to preserve NestedList nullability
+    RelDataType elementType;
+    if (args.isEmpty()) {
+      elementType = typeFactory.createSqlType(SqlTypeName.ANY);
```
vbarua: This should be impossible right? We have a check to prevent the creation of NestedLists with no elements? It would be reasonable to throw an `IllegalStateException("NestedList must have at least 1 element").

isthmus/src/main/java/io/substrait/isthmus/expression/CallConverters.java
gord02: I moved the logic from LiteralConstructorConverter to the SqlArrayValueConstructorCallConverter including the SqlMapValueConstructor logic. Should this map logic be made into its own file now?

isthmus/src/main/java/io/substrait/isthmus/expression/ExpressionRexConverter.java
gord02: Good catch, yes!

isthmus/src/main/java/io/substrait/isthmus/expression/CallConverters.java
vbarua: Ah, that's why it didn't blow up. `SqlArrayValueConstructorCallConverter` is not a good place for the Map handling code, I didn't even think to look for the code there 😅

What do you think about adding a `SqlMapValueConstructorCallConverter` and having the map code live there.

PR #629, opened 2025-12-03
style: add final modifier where possible

nielspardon:
This PR makes parameters and local variables final if possible and adds corresponding PMD checks so any new parameters or local variables will be declared final if they could be final.

### DIFF
--- a/build-logic/src/main/resources/substrait-pmd.xml
+++ b/build-logic/src/main/resources/substrait-pmd.xml
@@ -12,6 +12,9 @@
 
   <rule ref="category/java/codestyle.xml/UnnecessaryModifier" />
   <rule ref="category/java/codestyle.xml/FieldDeclarationsShouldBeAtStartOfClass" />
+  <rule ref="category/java/codestyle.xml/FinalParameterInAbstractMethod" />
+  <rule ref="category/java/codestyle.xml/LocalVariableCouldBeFinal" />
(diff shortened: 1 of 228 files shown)


build-logic/src/main/resources/substrait-pmd.xml
```diff
@@ -12,6 +12,9 @@
 
   <rule ref="category/java/codestyle.xml/UnnecessaryModifier" />
   <rule ref="category/java/codestyle.xml/FieldDeclarationsShouldBeAtStartOfClass" />
+  <rule ref="category/java/codestyle.xml/FinalParameterInAbstractMethod" />
+  <rule ref="category/java/codestyle.xml/LocalVariableCouldBeFinal" />
+  <rule ref="category/java/codestyle.xml/MethodArgumentCouldBeFinal" />
```
benbellick, line 17: I am definitely in favor of the first two `FinalParameterInAbstractMethod` and `LocalVariableCouldBeFinal`, but I am less convinced about introducing `MethodArgumentCouldBeFinal`. It looks like a lot noise is introduced in this PR by bringing in that rule. Considering that usage of `final` doesn't actually prevent mutation of the object, just reassignment to a new object, it doesn't seem that useful. 

We also already have the immutables library to protect against mutation for the classes for which that matters. 

Happy to be convinced though if there is a compelling reason to switch to using `final`.

benbellick (changes requested):

PR #642, opened 2025-12-17
fix: improve the javadoc generation

mbwhite:
Make all components put javadoc into a consistent root. This achieves the following

- means the javadoc for the generated protobuf can be kept quiet - as we can't adjust the javadoc here
- there is a single place for javadoc that could be published
- the isthmus and core and protobuf projects can be interlinked
- isthmus can be linked with the calcite javadoc
- overview pages can be added
- all under a versioned directory for easy maintence

To date there isn't (AFAIK) a stightforward maintainable way to merge javadoc form multi-component projects

### DIFF
--- a/core/build.gradle.kts
+++ b/core/build.gradle.kts
@@ -367,8 +367,7 @@ tasks.named<Javadoc>("javadoc") {
   // Keep normal behavior for main javadoc (warnings allowed to show/fail if you want)
   (options as StandardJavadocDocletOptions).apply {
     encoding = "UTF-8"
-    destinationDir = rootProject.layout.buildDirectory.dir("docs/${version}/core").get().asFile
-
+    setDestinationDir(rootProject.layout.buildDirectory.dir("docs/${version}/core").get().asFile)
     addStringOption("overview", "${rootProject.projectDir}/core/src/main/javadoc/overview.html")
     links("../core-proto/")
   }
--- a/isthmus/build.gradle.kts
+++ b/isthmus/build.gradle.kts
@@ -155,8 +155,7 @@ tasks.named<Javadoc>("javadoc") {
   // Keep normal behavior for main javadoc (warnings allowed to show/fail if you want)
   (options as StandardJavadocDocletOptions).apply {
     encoding = "UTF-8"
-    destinationDir = rootProject.layout.buildDirectory.dir("docs/${version}/isthmus").get().asFile
-
+    setDestinationDir(rootProject.layout.buildDirectory.dir("docs/${version}/isthmus").get().asFile)
     addStringOption("overview", "${rootProject.projectDir}/isthmus/src/main/javadoc/overview.html")
     links("../core-proto/")
     links("../core/")
@@ -166,12 +165,12 @@ tasks.named<Javadoc>("javadoc") {
 
 // Bundle both passes into the Javadoc JAR used for publishing.
 tasks.named<Jar>("javadocJar") {
-  val shared = rootProject.layout.buildDirectory.dir("docs/${version}").get().asFile
-  if (!shared.exists()) {
-    println("Creating a dir for javadoc ${rootProject.buildDir}/${version}")
-    shared.mkdirs()
-  }
+  // auto creates the directories if needed
+  val docsDir = rootProject.layout.buildDirectory.dir("docs/${version}")
+  destinationDirectory.set(docsDir)
 
+  from(tasks.named("javadoc"))
   // Ensure javadoc tasks have produced output
-  dependsOn(tasks.named("javadoc"))
+  // Handle duplicate files (e.g., allclasses-index.html) from multiple javadoc tasks
+  duplicatesStrategy = DuplicatesStrategy.EXCLUDE
 }


examples/substrait-spark/src/main/java/io/substrait/examples/util/SubstraitStringify.java
```diff
@@ -74,7 +74,6 @@ public SubstraitStringify() {
   /**
    * Explains the Sustrait plan
    *
-   * @param plan Subsrait plan
```
nielspardon, line 76: I do still see a `plan` parameter in the method declaration below so should the javadoc tag for it really be removed?

examples/substrait-spark/src/main/java/io/substrait/examples/util/SubstraitStringify.java
```diff
@@ -96,7 +95,7 @@ public static List<String> explain(io.substrait.plan.Plan plan) {
   /**
    * Explains the Sustrait relation
    *
-   * @param plan Subsrait relation
+   * @param rel Subsrait relation
```
nielspardon: this change feels a bit unrelated to the other changes but while you are at it maybe fix the other typo as well?
```suggestion
   * @param rel Substrait relation
```

mbwhite: done

examples/substrait-spark/src/main/java/io/substrait/examples/util/SubstraitStringify.java
mbwhite: a mistake yes

overview.html
```diff
@@ -0,0 +1,22 @@
...
+  <ul>
+    <li><a href="./index.html">core</a> Core classes and interfaces</li>
+    <li>
+      <a href="../core-proto/index.html">core-protobuf</a> The ProtoBuf wrapper
+      classes
+    </li>
+    <li>
+      <a href="../isthmus/index.html">Isthmus API</a> Utility module that uses
+      Calcite to convert SQL to/from Substrait
+    </li>
```
andrew-coleman: Suggest adding some extra punctuation between the link title and the description.
```suggestion
    <li><a href="./index.html">core</a> - Core classes and interfaces</li>
    <li>
      <a href="../core-proto/index.html">core-protobuf</a> - The Protocol Buffers wrapper
      classes
    </li>
    <li>
      <a href="../isthmus/index.html">Isthmus API</a> - Utility module that uses
      Calcite to convert SQL to/from Substrait
    </li>
```

core/src/main/javadoc/overview.html
```diff
@@ -0,0 +1,22 @@
+<body>
```
nielspardon, line 1: having the overview.html in the repository root feels a bit odd to me and my mind is trying to figure out whether it should be in a subfolder and which one

benbellick: Maybe `src/main/doc`, `src/main/javadoc`, `src/doc`? I don't have experience at all with this file here. But just throwing some ideas out. I agree that this shouldn't be at the root level.

overview.html
mbwhite: done

core/src/main/javadoc/overview.html
mbwhite: I've moved it to `core/src/mian/javadoc` (cf https://maven.apache.org/plugins/maven-javadoc-plugin/examples/javadoc-resources.html#:~:text=For%20more%20information%2C%20see%20javadoc,%2Fmain%2Fjavadoc%2Foverview.)

mbwhite:
@nielspardon @andrew-coleman @benbellick @bestbeforetoday @vbarua 

I've updated the merge conflicts; if there aren't any more urgent comments on this one... would it be ok to merge please?  Help ro reduce any conflicts with other JavaDoc changes. Also this one helps to reduce the overall 'noise' level from the generated protobuf's javadoc.

Thanks

isthmus/build.gradle.kts
```diff
@@ -147,3 +147,29 @@ tasks {
 sourceSets { test { proto.srcDirs("src/test/resources/extensions") } }
 
 protobuf { protoc { artifact = "com.google.protobuf:protoc:" + libs.protoc.get().getVersion() } }
+
+tasks.named<Javadoc>("javadoc") {
+  description = "Generate Javadoc for main sources (excludes protobuf-generated sources)."
```
nielspardon: I think for `isthmus` this is not excluding protobuf generated sources since does not have to
```suggestion
  description = "Generate Javadoc for main sources."
```

isthmus/build.gradle.kts
```diff
@@ -147,3 +147,29 @@ tasks {
...
+
+  // Keep normal behavior for main javadoc (warnings allowed to show/fail if you want)
+  (options as StandardJavadocDocletOptions).apply {
+    encoding = "UTF-8"
+    destinationDir = rootProject.layout.buildDirectory.dir("docs/${version}/isthmus").get().asFile
+
+    addStringOption("link", "../core-proto")
+    addStringOption("link", "../core")
+    addStringOption("link", "https://calcite.apache.org/javadocAggregate/")
```
nielspardon: I checked out your branch and ran it locally. I would have expected that these lines lead to classes from core / core-proto being linked to either. Right now it seems that only the linking for the Calcite classes works, e.g.:
image

Is that the expected behavior?

core/build.gradle.kts
```diff
@@ -273,3 +273,67 @@ protobuf {
...
+  // Exclude the protobuf-generated directory from the main pass
+  val protoDirFile = protoJavaDir.get().asFile
+  exclude { spec -> spec.file.toPath().startsWith(protoDirFile.toPath()) }
+
+  // Keep normal behavior for main javadoc (warnings allowed to show/fail if you want)
+  (options as StandardJavadocDocletOptions).apply {
+    encoding = "UTF-8"
+    destinationDir = rootProject.layout.buildDirectory.dir("docs/${version}/core").get().asFile
+
+    addStringOption("overview", "${rootProject.projectDir}/core/src/main/javadoc/overview.html")
```
nielspardon, line 371: the overview.html customization is useful for navigation across the 3 javadoc directories though I'm wondering whether it should be included in all 3 javadoc directories?

when you click on `core-protobuf` coming from `core` here:
image



then you end up on `core-protobuf` with no way to get back to `core` easily:

image

core/build.gradle.kts
```diff
@@ -273,3 +273,67 @@ protobuf {
...
+}
+
+// Second pass: Javadoc for main code, excluding the generated protobuf sources.
+tasks.named<Javadoc>("javadoc") {
+  mustRunAfter("javadocProto")
+  description = "Generate Javadoc for main sources (excludes protobuf-generated sources)."
+
+  // Exclude the protobuf-generated directory from the main pass
+  val protoDirFile = protoJavaDir.get().asFile
+  exclude { spec -> spec.file.toPath().startsWith(protoDirFile.toPath()) }
```
nielspardon, line 364: I noticed the auto-generated classes by the immutables annotation processor are missing from the javadoc. Is there another config we need to set to make them appear?

core/build.gradle.kts
nielspardon: I'm also wondering whether if we had Java modules (if we weren't required to stay at Java 8 compatibility) would allow us to render a single Javadoc across all 3 components. But then we could also rely on services like [javadoc.io](https://javadoc.io/doc/io.substrait) for rendering javadoc.

isthmus/build.gradle.kts
mbwhite: have updated this

isthmus/build.gradle.kts
mbwhite: no;  configuring the linking is a little confusing to configure; there are several ways to do it and evidence would suggest they are mutually exclusive.  Certainly the paths need to be relative paths.

They should be linked together now.

core/build.gradle.kts
mbwhite: Have added overview.html for all core/core-proto/isthmus.

they should all be linked together

core/build.gradle.kts
mbwhite: I've added them in; we may need to address any javadoc issues similarly to the protobufs. 
But until the warnings are reduced it's hard to really see

vbarua: > I've added them in; we may need to address any javadoc issues similarly to the protobufs.

I haven't been following this work too closely, but it seems like there are some difficulties associated with docs for the generated code. I wanted to mentioned that I've started building up the scaffolding to extract code generation from the language libraries that was discussed in https://github.com/substrait-io/substrait/issues/925. I'm focusing on substrait-python for now, but substrait-java is also a target for this, and once the generated code is gone hopefully these issues will be easier to deal with.

core/build.gradle.kts
```diff
@@ -273,3 +273,113 @@ protobuf {
...
+    addStringOption("overview", "${rootProject.projectDir}/core/src/main/javadoc/overview.html")
+    links("../core-proto/")
+  }
+}
+
+// Bundle both passes into the Javadoc JAR used for publishing.
+tasks.named<Jar>("javadocJar") {
+  val shared = rootProject.layout.buildDirectory.dir("docs/${version}").get().asFile
+  if (!shared.exists()) {
+    println("Creating a dir for javadoc ${rootProject.buildDir}/docs/${version}")
```
nielspardon: I've seen a deprecation warning for the use`buildDir` when I ran this locally. seems one should use `layout.buildDirectory.get()` instead.

isthmus/build.gradle.kts
```diff
@@ -147,3 +147,31 @@ tasks {
...
+    links("../core/")
+    links("https://calcite.apache.org/javadocAggregate/")
+  }
+}
+
+// Bundle both passes into the Javadoc JAR used for publishing.
+tasks.named<Jar>("javadocJar") {
+  val shared = rootProject.layout.buildDirectory.dir("docs/${version}").get().asFile
+  if (!shared.exists()) {
+    println("Creating a dir for javadoc ${rootProject.buildDir}/${version}")
```
nielspardon: I've seen a deprecation warning for the use`buildDir` when I ran this locally. seems one should use `layout.buildDirectory.get()` instead.

core/build.gradle.kts
```diff
@@ -273,3 +273,113 @@ protobuf {
...
+// First pass: Javadoc for generated protobufs — ignore warnings.
+tasks.register<Javadoc>("javadocProto") {
+  dependsOn("generateProto", "compileJava")
+
+  group = JavaBasePlugin.DOCUMENTATION_GROUP
+  description = "Generate Javadoc for protobuf-generated sources (warnings suppressed)."
+
+  // Only the generated proto sources
+  setSource(fileTree(protoJavaDir) { include("**/*.java") })
+  // source(fileTree(immuteableJavaDir) { include("**/*.java") })
```
nielspardon: I guess this line is not really needed
```suggestion
```

core/build.gradle.kts
```diff
@@ -273,3 +273,113 @@ protobuf {
...
+
+tasks.register<Javadoc>("javadocImmutable") {
+  dependsOn("compileJava", "javadoc")
+
+  group = JavaBasePlugin.DOCUMENTATION_GROUP
+  description = "Generate Javadoc for immutable-generated sources (warnings suppressed)."
+
+  // Only the generated proto sources
+  setSource(fileTree(immuteableJavaDir) { include("**/*.java") })
+  // source(fileTree(immuteableJavaDir) { include("**/*.java") })
```
nielspardon:
```suggestion
```

core/build.gradle.kts
nielspardon: nice, I can see the immutables classes now

core/build.gradle.kts
nielspardon: nice, thanks

isthmus/build.gradle.kts
nielspardon: great, that you figured this out. things are linked nicely now.

PR #643, opened 2025-12-17
feat(isthmus): support month()-style datetime operators

mbwhite:
The Substrait Function for Month style mappings doesn't work as the calcite only has 2 arguments, but substrait has three.

Isthmus (in either direction) assumes that there is a 1:1 mapping between a substrait function and a calcite function. With the same number of arguments in each. (although types can be coerced if needed)

In this case, the 3 arg Substrait function needs to be mapped to a 2 arg calcite one. In effect this is working by swapping out the function to a different one and then letting the rest of the logic work as before.

A concern is the increased complexity - and the code as a whole wasn't built with this in mind.

### DIFF
--- a/core/src/main/java/io/substrait/expression/EnumArg.java
+++ b/core/src/main/java/io/substrait/expression/EnumArg.java
@@ -4,8 +4,6 @@ import io.substrait.extension.SimpleExtension;
 import io.substrait.util.VisitationContext;
 import java.util.Optional;
 import org.immutables.value.Value;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 
 /**
  * Captures the {@link SimpleExtension.EnumArgument} option value in a {@link
@@ -16,9 +14,6 @@ import org.slf4j.LoggerFactory;
  */
 @Value.Immutable
 public interface EnumArg extends FunctionArg {
-
-  Logger LOGGER = LoggerFactory.getLogger(EnumArg.class);
-
   EnumArg UNSPECIFIED_ENUM_ARG = builder().value(Optional.empty()).build();
 
   Optional<String> value();


core/src/main/java/io/substrait/expression/EnumArg.java
```diff
@@ -26,7 +31,12 @@ default <R, C extends VisitationContext, E extends Throwable> R accept(
   }
 
   static EnumArg of(SimpleExtension.EnumArgument enumArg, String option) {
-    assert (enumArg.options().contains(option));
+    if (!enumArg.options().contains(option)) {
+      LOGGER.warn(
+          String.format(
+              "Function does not accept those arguments %s %s", option, enumArg.options()));
+      return null;
+    }
```
bestbeforetoday, line 32: This risks being a breaking change to this class API since null can now be returned whereas that didn't previously happen. Unexpected null returns run the risk of breaking callers and causing unexpected delayed failures.

It might be better to either leave it as-is or actively generate an exception here where the problem occurs. Something like:

```suggestion
    if (!enumArg.options().contains(option)) {
      throw new IllegalArgumentException(
          "EnumArg value " + option + " not valid for options: " + enumArg.options());
    }
```

isthmus/src/main/java/io/substrait/isthmus/expression/ExtractDateFunctionMapper.java
```diff
@@ -0,0 +1,103 @@
...
+import org.apache.calcite.rex.RexNode;
+import org.apache.calcite.sql.fun.SqlStdOperatorTable;
+import org.apache.calcite.sql.type.BasicSqlType;
+import org.apache.calcite.sql.type.SqlTypeName;
+
+/**
+ * Custom mapping for the Calcite MONTH/DAY/QUARTER functions.
+ *
+ * <p>These come from Calcite as 2 argument functions (like YEAR) but in Substrait these functions
+ * are 3 agumentes; the additional being if this is a 0 or 1 based value. Calcite is a 1 based value
```
bestbeforetoday:
```suggestion
 * are 3 arguments; the additional being if this is a 0 or 1 based value. Calcite is a 1 based value
```

isthmus/src/main/java/io/substrait/isthmus/expression/ExtractDateFunctionMapper.java
```diff
@@ -0,0 +1,103 @@
...
+    }
+
+    final RelDataType dataType = call.operands.get(1).getType();
+    if (!(dataType instanceof BasicSqlType)) {
+      return Optional.empty();
+    }
+
+    if (!((BasicSqlType) dataType).getSqlTypeName().equals(SqlTypeName.DATE)) {
+      return Optional.empty();
+    }
```
bestbeforetoday: I don't think the check for BasicSqlType and subsequent cast is necessary. Perhaps this can be simplified and use the same style as the check immediately above:


```suggestion
    RexNode dataType = call.operands.get(1);
    if (!dataType.getType().getSqlTypeName().equals(SqlTypeName.DATE)) {
      return Optional.empty();
    }
```

isthmus/src/main/java/io/substrait/isthmus/expression/ExtractDateFunctionMapper.java
```diff
@@ -0,0 +1,103 @@
...
+  }
+
+  @Override
+  public Optional<List<FunctionArg>> getExpressionArguments(
+      final Expression.ScalarFunctionInvocation expression) {
+    String name = expression.declaration().toString();
+
+    if ("extract:req_req_date".equals(name)) {
+      final List<FunctionArg> newArgs = new ArrayList<>(expression.arguments());
+      newArgs.remove(1);
```
bestbeforetoday: Minor nit since we're dealing with very small lists but removing elements from anywhere other than the end of an ArrayList generally performs badly since it needs to move all subsequent elements. A LinkedList might be a better choice.

isthmus/src/main/java/io/substrait/isthmus/expression/LiteralConverter.java
```diff
@@ -132,6 +133,10 @@ public Expression.Literal convert(RexLiteral literal) {
             return ExpressionCreator.string(n, ((NlsString) value).getValue());
           } else if (value instanceof Enum) {
             Enum<?> v = (Enum<?>) value;
+            if (value instanceof INDEXING) {
+              return ExpressionCreator.string(n, v.name());
+            }
+
```
bestbeforetoday: This does not seem to be necessary. Tests pass without this change.

isthmus/src/main/java/io/substrait/isthmus/expression/ScalarFunctionConverter.java
```diff
@@ -24,13 +24,17 @@ public class ScalarFunctionConverter
...
   /**
    * Function mappers provide a hook point for any custom mapping to Substrait functions and
    * arguments.
    */
   private final List<ScalarFunctionMapper> mappers;
 
+  public enum INDEXING {
+    ONE,
+    ZERO
+  }
```
bestbeforetoday: Enum names are generally upper camel case, like class names.

bestbeforetoday: Since this enum is used across multiple classes, I am not sure being a nested class within ScalarFunctionConverter is really the right place for it. This couples several other classes to ScalarFunctionConvert that previously did not reference it. Perhaps it should be declared in its own file. Maybe it is `ExtractIndexing` since I don't see it used anywhere else?

bestbeforetoday (changes requested):
Generally looks good to me with just a few minor tweaks desirable.

examples/isthmus-api/.gitignore
```diff
@@ -2,3 +2,4 @@ _apps
 _data
 **/*/bin
 build
+substrait.plan
```
vbarua, line 5: Why do we need to ignore these?

isthmus/src/main/java/io/substrait/isthmus/expression/EnumConverter.java
```diff
@@ -84,6 +91,10 @@ private static Optional<Enum<?>> constructValue(
       return option.get().map(SqlTrimFunction.Flag::valueOf);
     }
 
+    if (cls.isAssignableFrom(INDEXING.class)) {
+      return option.get().map(INDEXING::valueOf);
+    }
```
vbarua: What does this code actually do? If I delete it locally, none of the tests fail?

isthmus/src/main/java/io/substrait/isthmus/expression/EnumConverter.java
```diff
@@ -72,6 +73,12 @@ public class EnumConverter {
     calciteEnumMap.put(
         argAnchor(DefaultExtensionCatalog.FUNCTIONS_STRING, "rtrim:str_str", 0),
         SqlTrimFunction.Flag.class);
+    calciteEnumMap.put(
+        argAnchor(DefaultExtensionCatalog.FUNCTIONS_DATETIME, "extract:req_req_date", 0),
+        TimeUnitRange.class);
+    calciteEnumMap.put(
+        argAnchor(DefaultExtensionCatalog.FUNCTIONS_DATETIME, "extract:req_req_date", 1),
+        INDEXING.class);
```
vbarua: If I remove this mapping, none of the tests fail.

core/src/main/java/io/substrait/expression/EnumArg.java
mbwhite: Yes I would agree with you - but the thinking was along these lines.  (should I have added this as a comment!)

Let's assuming a case where the option is not valid;  previously this invalid state would have been completely ignore, and code continued.  **unless** the Classloader assertions where enabled - in that case the `assert` would have thrown an exception.   External behaviour would in most cases have been any option would be accepted.   Only in cases such as in unit tests would an exception be propagated to the end user.

Hence to keep the same behaviour in the majority of the cases, hence null. 

However - I would completely agree that the IllegalArugmentException would be better - but *technically* is a breaking change as a whole... but would given the preV1 state of Substrait would be one a strong argument could be made for.

isthmus/src/main/java/io/substrait/isthmus/expression/ExtractDateFunctionMapper.java
mbwhite: Note to self: re-enable code spell checker..

isthmus/src/main/java/io/substrait/isthmus/expression/ExtractDateFunctionMapper.java
mbwhite: Ah - hadn't noticed that `getSqlTypeName` was defined on the `RexDataType`

isthmus/src/main/java/io/substrait/isthmus/expression/ExtractDateFunctionMapper.java
mbwhite: As I'm 'in the area' I can alter that.

Did look at the implementation of both of them - so the `ArrayList` does a `System.arraycopy` but the `LinkedList` does do a linear scan of the internal node list to find the element to remove (doing equality at ever step).  Wonder what the relative performance is like.. experiment for another day.

isthmus/src/main/java/io/substrait/isthmus/expression/ScalarFunctionConverter.java
mbwhite: I wasn't completely sure where best to put this - as it's only (at the moment anyway) used briefly;

I'll move it to a separate file

examples/isthmus-api/.gitignore
mbwhite: It's created for the examples, and I was used them as a test and kept adding it to the commit.

isthmus/src/main/java/io/substrait/isthmus/expression/LiteralConverter.java
mbwhite: Correct - this dates I think from my initial revisions...

isthmus/src/main/java/io/substrait/isthmus/expression/EnumConverter.java
mbwhite: It was bypassed by some of the other special case code (which ins't in keeping with the overall code). so it is required.

```
Substrait2SqlTest > dateFunctions() FAILED
    java.lang.UnsupportedOperationException: Unable to handle symbol: ONE
```

isthmus/src/main/java/io/substrait/isthmus/expression/EnumConverter.java
mbwhite: This was a case of duplicating the code that existed for the other functions.. 

As Calcite doesn't have the concept, this isn't needed.. I've added a comment for future so it doesn't follow the pattern.

isthmus/src/main/java/io/substrait/isthmus/expression/ScalarFunctionConverter.java
mbwhite: In new file..

isthmus/src/main/java/io/substrait/isthmus/expression/EnumConverter.java
bestbeforetoday: Trying to remind myself of where this fits in... this function is used when converting Substrait function arguments that are enums back to Calcite Rex nodes. However, before it reaches this code for the extract date function, the function arguments have already been manipulated by the `getExpressionArguments` function in ExtractDateFunctionMapper, which removes the `indexing` enum from the argument list (since it isn't relevant to Calcite). The net result is that you are correct, this piece of code is never used and should be removed.

core/src/main/java/io/substrait/expression/EnumArg.java
bestbeforetoday: Returning null is not keeping the same behaviour and is also a breaking change, just like explicitly throwing an exception. It's just a less helpful breaking change since it masks (or defers) the underlying problem.

mbwhite:
@vbarua @bestbeforetoday hopefully I've been able to address all your points.  Thanks

core/src/main/java/io/substrait/expression/EnumArg.java
```diff
@@ -14,6 +16,9 @@
  */
 @Value.Immutable
 public interface EnumArg extends FunctionArg {
+
+  Logger LOGGER = LoggerFactory.getLogger(EnumArg.class);
```
bestbeforetoday: This is not used.

bestbeforetoday: This is such a minor issue I pushed an additional commit to fix this so the PR doesn't need to be held out until you change it.

bestbeforetoday (approved):

PR #645, opened 2025-12-17
fix: javadoc for the isthmus library

mbwhite:


isthmus/src/main/java/io/substrait/isthmus/OuterReferenceResolver.java
```diff
@@ -15,32 +15,71 @@
...
   }
 
+  /**
+   * Returns the number of "steps out" (nesting depth) for a given {@link RexFieldAccess}.
+   *
+   * @param fieldAccess the field access referencing a {@link RexCorrelVariable}
+   * @return the number of outer scopes between the access and its correlation source, or {@code
+   *     null} if not tracked
+   */
   public int getStepsOut(RexFieldAccess fieldAccess) {
```
nielspardon: that makes me wonder if we would need to change the method return type given that it's declared as a primitive `int` and not an `Integer` meaning it can not be null

nielspardon: looks like this method is not called anywhere in the current code base

mbwhite: yes that's a good point... it does seem redundant

nielspardon: https://github.com/substrait-io/substrait-java/pull/665

isthmus/src/main/java/io/substrait/isthmus/PreCalciteAggregateValidator.java
```diff
@@ -58,32 +57,19 @@ private static boolean isValidCalciteMeasure(Aggregate.Measure measure) {
...
     }
 
-    // Calcite stores grouping fields in an ImmutableBitSet and does not track the order of the
-    // grouping fields. The output record shape that Calcite generates ALWAYS has the groupings in
-    // ascending field order. This causes issues with Substrait in cases where the grouping fields
-    // in Substrait are not defined in ascending order.
-
-    // For example, if a grouping is defined as (0, 2, 1) in Substrait, Calcite will output it as
-    // (0, 1, 2), which means that the Calcite output will no longer line up with the expectations
-    // of the Substrait plan.
```
nielspardon, line 83: I'm not sure how I feel about removing this much information. I'm wondering whether we should better preserve this somehow.

isthmus/src/main/java/io/substrait/isthmus/RelNodeVisitor.java
```diff
@@ -18,80 +18,215 @@
...
+   * Fallback method for visiting any {@link RelNode} type not explicitly handled.
+   *
+   * @param other the relational node
+   * @return the result of visiting this node
+   * @throws EXCEPTION if an error occurs during processing
+   */
   public abstract OUTPUT visitOther(RelNode other) throws EXCEPTION;
 
+  /** Protected constructor to enforce subclassing. */
   protected RelNodeVisitor() {}
```
nielspardon, line 219: probably should add some checkstyle rules to catch such cases where the constructor is defined in the middle of the class

nielspardon: I created an issue for this: https://github.com/substrait-io/substrait-java/issues/667

isthmus/src/main/java/io/substrait/isthmus/PreCalciteAggregateValidator.java
mbwhite: I've reinstated the comment

isthmus/src/main/java/io/substrait/isthmus/SqlToSubstrait.java
```diff
@@ -21,14 +21,31 @@
 public class SqlToSubstrait extends SqlConverterBase {
   private final SqlOperatorTable operatorTable;
 
+  /**
+   * Creates a SQL-to-Substrait converter using the default extension catalog and no feature
+   * overrides.
```
nielspardon: If we remove the negation on the second part I guess it comes down to this? Would that be clearer?

```suggestion
   * Creates a SQL-to-Substrait converter using the default extension catalog and default features.
```

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
```diff
@@ -95,23 +95,50 @@
...
     this(extensions, typeFactory, relBuilder, ImmutableFeatureBoard.builder().build());
   }
 
+  /**
+   * Constructs a converter using the given extensions and feature board.
+   *
+   * @param extensions Substrait extension collection (scalar/aggregate/window function variants)
+   * @param typeFactory Calcite type factory
+   * @param relBuilder Calcite {@link RelBuilder}
+   * @param featureBoard feature flags controlling behavior (e.g., dynamic UDFs)
```
nielspardon: I would not call out specific features at this point since it would be hard to keep up-to-date. I would expect the features to be documented in the FeatureBoard class anyway.

```suggestion
   * @param featureBoard feature flags controlling the conversion behavior
```

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
```diff
@@ -217,6 +280,21 @@ public static RelNode convert(
...
+   *
+   * <p>This method initializes a {@link RelBuilder} with the given cluster and catalog, then
+   * constructs a {@link SubstraitRelNodeConverter} to perform the conversion.
+   *
+   * @param relRoot the root Substrait relation to convert
+   * @param relOptCluster the Calcite cluster providing optimization context
+   * @param catalogReader the Calcite catalog reader for schema resolution
+   * @param parserConfig the SQL parser configuration
+   * @param extensions the Substrait extension collection (scalar, aggregate, window functions)
+   * @param featureBoard feature flags controlling conversion behavior (e.g., dynamic UDF support)
```
nielspardon:
```suggestion
   * @param featureBoard feature flags controlling conversion behavior
```

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelVisitor.java
```diff
@@ -68,24 +68,52 @@
...
+  /** Feature flags governing conversion. */
   protected final FeatureBoard featureBoard;
+
   private Map<RexFieldAccess, Integer> fieldAccessDepthMap;
 
+  /**
+   * Creates a visitor with default features.
+   *
+   * @param typeFactory Calcite type factory.
+   * @param extensions Substrait extension collection.
```
nielspardon: I'm not sure how we want to handle punctuation. I learned that you use a full-stop if you try to end a complete sentence. In these cases we do not seem to have a complete sentence since we don't have a verb. What do you think?

nielspardon: somehow this pattern is very present in this file but I can't recall seeing it in the previous files in this PR

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelVisitor.java
```diff
@@ -182,17 +244,35 @@ public Rel visit(org.apache.calcite.rel.core.Values values) {
...
   public Rel visit(org.apache.calcite.rel.core.Filter filter) {
     Expression condition = toExpression(filter.getCondition());
     return Filter.builder().condition(condition).input(apply(filter.getInput())).build();
   }
 
+  /**
+   * Converts a Calcite {@link org.apache.calcite.rel.core.Calc}.
+   *
+   * @param calc Calcite calc relation.
+   * @return Converted relation or {@code super.visit(calc)}.
```
nielspardon: I would not mention `or {@code super.visit(calc)}`. Feels like we are repeating the implementation. Doesn't calling `super.visit(calc)` effectively mean that it's not implemented yet and we are falling back? Should we call this fact out instead?

nielspardon: I guess we could also easily remove a method that just calls its super implementation

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelVisitor.java
```diff
@@ -248,6 +334,12 @@ private Join.JoinType asJoinType(org.apache.calcite.rel.core.Join join) {
     throw new UnsupportedOperationException("Unsupported join type: " + join.getJoinType());
   }
 
+  /**
+   * Converts a Calcite {@link org.apache.calcite.rel.core.Correlate}.
+   *
+   * @param correlate Calcite correlate relation.
+   * @return Converted relation or {@code super.visit(correlate)}.
```
nielspardon: same here. not sure mentioning `or {@code super.visit(correlate)}` is a useful information for the reader

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelVisitor.java
```diff
@@ -379,11 +496,23 @@ Aggregate.Measure fromAggCall(RelNode input, Type.Struct inputType, AggregateCal
     return builder.build();
   }
 
+  /**
+   * Converts a Calcite {@link org.apache.calcite.rel.core.Match}.
+   *
+   * @param match Calcite match relation.
+   * @return Converted relation or {@code super.visit(match)}.
```
nielspardon: same here. not sure having `or {@code super.visit(match)}` is a useful information

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelVisitor.java
```diff
@@ -453,11 +589,24 @@ private static Expression.SortDirection asSortDirection(RelFieldCollation collat
     throw new IllegalArgumentException("Unsupported collation direction: " + direction);
   }
 
+  /**
+   * Converts a Calcite {@link org.apache.calcite.rel.core.Exchange}.
+   *
+   * @param exchange Calcite exchange relation.
+   * @return Converted relation or {@code super.visit(exchange)}.
```
nielspardon: one more. not sure `or {@code super.visit(exchange)}` is a useful information

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelVisitor.java
```diff
@@ -566,6 +715,12 @@ private NamedStruct getSchema(final RelNode queryRelRoot) {
     return typeConverter.toNamedStruct(rowType);
   }
 
+  /**
+   * Handles Calcite {@link CreateTable} as Substrait CTAS.
```
nielspardon: should we spell out CTAS once here? or do you think it is commonly known by our readers what it stands for?

isthmus/src/main/java/io/substrait/isthmus/TypeConverter.java
```diff
@@ -20,11 +20,29 @@
 import org.apache.calcite.sql.type.SqlTypeName;
 import org.jspecify.annotations.Nullable;
 
+/**
+ * Utility for converting between Calcite {@link org.apache.calcite.rel.type.RelDataType} and
+ * Substrait {@link io.substrait.type.Type}.
+ *
+ * <p>Supports primitive, complex, and user *
+ *
+ * <p>Supports primitive, complex, and user-defined types in both directions.
```
nielspardon: I guess this can be collapsed

```suggestion
 * <p>Supports primitive, complex, and user-defined types in both directions.
```

isthmus/src/main/java/io/substrait/isthmus/calcite/SubstraitOperatorTable.java
```diff
@@ -69,6 +75,19 @@ public class SubstraitOperatorTable implements SqlOperatorTable {
 
   private SubstraitOperatorTable() {}
```
nielspardon, line 77: you may also want to add a javadoc comment for this private constructor like for the previous ones mainly for consistency

isthmus/src/main/java/io/substrait/isthmus/calcite/SubstraitTable.java
```diff
@@ -4,21 +4,45 @@
...
+   */
   public SubstraitTable(String tableName, RelDataType rowType) {
     this.tableName = tableName;
     this.rowType = rowType;
   }
 
+  /**
+   * Returns the name of the table.
+   *
+   * @return The table name as a {@link String}.
```
nielspardon: It feels obvious to me that the table name is a string.

```suggestion
   * @return the table name
```

isthmus/src/main/java/io/substrait/isthmus/calcite/rel/DdlSqlToRelConverter.java
```diff
@@ -46,10 +69,23 @@ public RelRoot visit(SqlCall sqlCall) {
...
   protected RelRoot handleNonDdl(final SqlNode sqlNode) {
     return converter.convertQuery(sqlNode, true, true);
   }
 
+  /**
+   * Handles {@code CREATE TABLE AS SELECT} statements.
+   *
+   * @param sqlCreateTable the CREATE TABLE node
+   * @return a {@link RelRoot} wrapping a synthetic {@code CreateTable} relational node
+   * @throws IllegalArgumentException if the statement is not CTAS
```
nielspardon: I guess there should be an article

```suggestion
   * @throws IllegalArgumentException if the statement is not a CTAS
```

isthmus/src/main/java/io/substrait/isthmus/expression/AggregateFunctionConverter.java
```diff
@@ -75,6 +109,15 @@ protected AggregateFunctionInvocation generateBinding(
         arguments);
   }
 
+  /**
+   * Attempts to convert a Calcite aggregate call to a Substrait invocation.
+   *
+   * @param input input relational node
+   * @param inputType Substrait input struct type
+   * @param call Calcite aggregate call
+   * @param topLevelConverter converter for RexNodes to Expressions
```
nielspardon, line 118: In other instances you added an `{@link }` for mentions of classes. I guess we want to be consistent on this style.
```suggestion
   * @param topLevelConverter converter for {@link RexNode}s to {@link Expression}s
```

isthmus/src/main/java/io/substrait/isthmus/expression/AggregateFunctionConverter.java
```diff
@@ -122,15 +180,30 @@ private WrappedAggregateCall(
       this.inputType = inputType;
     }
 
+    /**
+     * Returns operands as input references over the argument list.
+     *
+     * @return stream of RexNode operands
```
nielspardon, line 186:
```suggestion
     * @return stream of {@link RexNode} operands
```

isthmus/src/main/java/io/substrait/isthmus/expression/FunctionConverter.java
```diff
@@ -49,60 +48,63 @@
...
- * <ul>
- *   <li><b>Calcite → Substrait:</b> Subclasses implement {@code convert()} methods to convert
- *       Calcite calls to Substrait function invocations
- *   <li><b>Substrait → Calcite:</b> {@link #getSqlOperatorFromSubstraitFunc} converts Substrait
- *       function keys to Calcite {@link SqlOperator}s
- * </ul>
- *
- * <p>When multiple functions with the same name and signature are passed into the constructor, a
- * <b>last-wins precedence strategy</b> is used for resolution. The last function in the input list
- * takes precedence during Calcite to Substrait conversion.
```
nielspardon, line 63: I think this bit is important information and should be retained.

isthmus/src/main/java/io/substrait/isthmus/expression/FunctionConverter.java
```diff
@@ -155,18 +157,14 @@ public FunctionConverter(
   }
 
   /**
-   * Converts a Substrait function to a Calcite {@link SqlOperator} (Substrait → Calcite direction).
-   *
-   * <p>Given a Substrait function key (e.g., "concat:str_str") and output type, this method finds
-   * the corresponding Calcite {@link SqlOperator}. When multiple operators match, the output type
-   * is used to disambiguate.
```
nielspardon, line 162: I think the disambiguation part gets lost with this rewrite.

nielspardon: I see it's still somewhat there in the parameter description though I'm wondering whether this is more of a behavioral description which should also be mentioned in the method description.

isthmus/src/main/java/io/substrait/isthmus/expression/FunctionConverter.java
```diff
@@ -212,6 +228,13 @@ protected class FunctionFinder {
     private final Optional<SingularArgumentMatcher<F>> singularInputType;
     private final Util.IntRange argRange;
 
+    /**
+     * Creates a function finder for a Substrait name/operator over given variants.
+     *
+     * @param substraitName canonical Substrait function name
```
nielspardon, line 238: this makes me wonder whether we should have an example here to help disambiguate between the [function signature](https://substrait.io/extensions/#function-signature) `concat:str_str` and the function name `concat`

isthmus/src/main/java/io/substrait/isthmus/expression/FunctionMappings.java
```diff
@@ -9,10 +9,17 @@
...
 
+/**
+ * Defines static mappings between Calcite {@link SqlOperator} signatures and Substrait base
+ * function names, including scalar, aggregate, and window function signatures.
+ *
+ * <p>Also provides type-based resolvers to disambiguate operators by output type.
+ */
 public class FunctionMappings {
   // Static list of signature mapping between Calcite SQL operators and Substrait base function
   // names.
```
nielspardon: I guess this is redundant now?
```suggestion
```

isthmus/src/main/java/io/substrait/isthmus/expression/FunctionMappings.java
```diff
@@ -146,54 +156,120 @@ public class FunctionMappings {
...
           resolver(SqlStdOperatorTable.BIT_LEFT_SHIFT, Set.of("i8", "i16", "i32", "i64")));
 
+  /**
+   * Prints all scalar signatures (for quick inspection).
+   *
+   * @param args CLI arguments (unused)
+   */
   public static void main(String[] args) {
     SCALAR_SIGS.forEach(System.out::println);
   }
```
nielspardon: it feels to me like we should drop this method

nielspardon: https://github.com/substrait-io/substrait-java/pull/672

isthmus/src/main/java/io/substrait/isthmus/expression/FunctionMappings.java
```diff
@@ -146,54 +156,120 @@ public class FunctionMappings {
...
 
+  /**
+   * Creates a signature mapping entry.
+   *
+   * @param operator the Calcite operator
+   * @param substraitName the Substrait base function name
```
nielspardon: here you call this the `Substrait base function name`, earlier I saw `Substrait canonical function name`. might be a good idea to be consistent here and maybe also give examples to disambiguate with the function signature.

isthmus/src/main/java/io/substrait/isthmus/expression/FunctionMappings.java
nielspardon: If you rebase then this method has been removed.

isthmus/src/main/java/io/substrait/isthmus/expression/ListSqlOperatorFunctions.java
```diff
@@ -10,8 +10,29 @@
 import org.apache.calcite.sql.fun.SqlMultisetSetOperator;
 import org.apache.calcite.sql.fun.SqlStdOperatorTable;
 
+/**
```
nielspardon: If you rebase then this file has been removed.

isthmus/src/main/java/io/substrait/isthmus/sql/SubstraitSqlDialect.java
```diff
@@ -28,10 +37,20 @@ public static SqlString toSql(RelNode relNode) {
...
+   * @param context The {@link SqlDialect.Context} providing configuration for SQL generation.
+   */
   public SubstraitSqlDialect(Context context) {
     super(context);
   }
 
+  /**
+   * Indicates whether this dialect supports approximate COUNT(DISTINCT) operations.
+   *
+   * @return {@code true}, as Substrait SQL dialect supports approximate count distinct.
```
nielspardon: we could add an "always" to make it even more obvious
```suggestion
   * @return always {@code true}, as Substrait SQL dialect supports approximate count distinct.
```

isthmus/src/main/java/io/substrait/isthmus/expression/IgnoreNullableAndParameters.java
```diff
@@ -4,232 +4,528 @@
...
 
+  /**
+   * Compares {@link Type.UserDefined} for exact equality (URI and name).
+   *
+   * @param type user-defined type
+   * @return {@code true} if {@code typeToMatch} equals {@code type}
+   * @throws RuntimeException if comparison cannot be performed
+   */
   @Override
   public Boolean visit(Type.UserDefined type) throws RuntimeException {
```
nielspardon: interesting, this seems to be the only method with a throws declaration

bestbeforetoday: Having the Javadoc for the thrown RuntimeException is good (if it really is thrown — I haven't checked). The code itself really should not declare unchecked exceptions though.

isthmus/src/main/java/io/substrait/isthmus/expression/FunctionConverter.java
mbwhite: Comment restored

substrait
nielspardon, line 1: are you maybe reverting the Git submodule update from main?

mbwhite: not on purpose... looks like regular `git rebase` doesn't update the submodules.

nielspardon: can you check that it is on tag v0.79.0 as in main? looks like it still pulls in something new

isthmus/src/main/java/io/substrait/isthmus/expression/ListSqlOperatorFunctions.java
nielspardon, line 1: I had deleted this file in https://github.com/substrait-io/substrait-java/pull/672

I guess you would also delete it from your branch so you're not bringing it back.

isthmus/src/main/java/io/substrait/isthmus/expression/FunctionMappings.java
```diff
@@ -146,50 +153,120 @@ public class FunctionMappings {
...
 
+  /**
+   * Prints all scalar signatures (for quick inspection).
+   *
+   * @param args CLI arguments (unused)
+   */
+  public static void main(String[] args) {
+    SCALAR_SIGS.forEach(System.out::println);
+  }
+
```
nielspardon: I deleted this one in: https://github.com/substrait-io/substrait-java/pull/672
```suggestion
```

build.gradle.kts
```diff
@@ -62,9 +62,9 @@ allprojects {
       kotlinGradle { ktfmt("0.61").googleStyle() }
       java {
         target("src/*/java/**/*.java")
-        // since kfmt also brings in google-java-format we need to sync versions
-        // https://github.com/facebook/ktfmt/blob/v0.61/gradle/libs.versions.toml#L7
-        googleJavaFormat("1.23.0")
```
benbellick, line 67: https://github.com/substrait-io/substrait-java/pull/743 introduced this change (just providing context, I don't know if this is a necessary change or not)

mbwhite: Yes;  this was being a bit of a pain yesterday and I was talking to @nielspardon on this; we were seeing almost random behaviour.

I'll rebase and retry here.

mbwhite: so that's gone now and looks like the build was happy.

Issue #640, opened 2025-12-15
CI: loosen commitlint restrictions


benbellick:
CI enforces that all commits in a PR conform to a commitlint standard. However, we just squash merge all PRs anyways, so all of those commit messages are lost.

Let's drop the requirement and just make sure that we follow commitlint standards in the PR title and description (which can then be used as the commit title and message by default).

nielspardon:
On this issue do you think we should remove the pre-commit hook and only run against PR title + description during CI?

nielspardon:
Reopening since we should discuss actually loosening the restrictions. #854 at least aligned the config for pre-commit and CI with the other repositories.

bestbeforetoday:
If we don't use the subject lines from the individual commits then it seems pointless to enforce them with commitlint. We might as well only check the PR title that will be used as the resulting commit into main from the PR.

PR #646, opened 2025-12-17
ci: enforce commitlint only on PR titles and descriptions

vbarua:
As we squash merge in this repository, checking that individual commit message
follow the Conventional Commits spec is not useful.

Instead, enforce that the PR title and description do follow the specification,
as that does end up in the final commit.

### DIFF
--- a/.github/workflows/pr_title.yaml
+++ b/.github/workflows/pr_title.yaml
@@ -1,0 +1,63 @@
+name: PR Title Check
+
+on:
+  pull_request_target:
+    types: [opened, edited, synchronize, reopened]
+jobs:
+  commitlint:
+    name: PR title / description conforms to semantic-release
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/setup-node@v6
+        with:
+          node-version: "20"
+      - run: npm install @commitlint/config-conventional
+      - run: >
+          echo 'module.exports = {
+            // Workaround for https://github.com/dependabot/dependabot-core/issues/5923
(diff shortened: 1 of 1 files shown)


.github/workflows/pr_title.yaml
```diff
@@ -0,0 +1,63 @@
+name: PR Title Check
```
vbarua, line 1: This was copied wholesale from https://github.com/substrait-io/substrait/blob/d1045ad1c080ecc06ea0515a8396644023bda8aa/.github/workflows/pr_breaking.yml

vbarua:
Addresses https://github.com/substrait-io/substrait-java/issues/640

.github/workflows/pr_title.yaml
```diff
@@ -0,0 +1,63 @@
...
+              }
+            }
+            // Post the comment about Conventional Commits
+            github.rest.issues.createComment({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              issue_number: context.issue.number,
+              body: message
+            })
+            core.setFailed(message)
```
benbellick, line 63: I wonder if we could make this action simpler by **not** commenting on the PR? And instead just have the CI/CD check fail? I'm not so particular, this just seems simpler and easier to maintain that way 🤷

benbellick (approved):

.github/workflows/pr_title.yaml
vbarua: I'm going to leave the comment for now, as it includes links to the conventional commit standard, which helps users understand what the requirements are explicitly.

PR #647, opened 2025-12-17
feat(isthmus): configurable fallback to dynamic function mapping

ZorinAnton:
This PR introduces a new `autoFallbackToDynamicFunctionMapping` feature flag that enables automatic support for all unmapped extension functions (scalar, aggregate, and window functions) in SQL↔Substrait conversions.


vbarua (commented):
I'm working on a PR (https://github.com/substrait-io/substrait-java/pull/649) to try and make wiring in what you're doing here easier. We're starting to run into the limits of the existing design of the function conversion system, and I'm wary of adding more complexity before cleaning it up.

My big issue is with how we're using the feature flags for this, because we're embedding a ton of conditional configuration logic into constructors all over the codebase, which makes it very difficult to reason about an already difficult part of the system. What I would like to do is separate out the configuration from the constructors. Effectively, can we construct all the function converters we need _outside_ of the SubstraitRelNode and SubstraitRelVisitor, and then inject them in.

PR #648, opened 2025-12-18
fix: add javadoc to core (part i)

mbwhite:


core/src/main/java/io/substrait/extension/DefaultExtensionCatalog.java
```diff
@@ -4,28 +4,64 @@
...
   public static final String FUNCTIONS_ROUNDING = "extension:io.substrait:functions_rounding";
+
+  /** Extension identifier for decimal rounding functions. */
   public static final String FUNCTIONS_ROUNDING_DECIMAL =
       "extension:io.substrait:functions_rounding_decimal";
+
+  /** Extension identifier for set functions. */
   public static final String FUNCTIONS_SET = "extension:io.substrait:functions_set";
+
+  /** Extension identifier for string functions. */
```
benbellick, line 53: IMO, these comments are a bit excessive and I would drop them. That's just my opinion though :)

mbwhite: I'm certainly sympathetic to that; with JavaDoc requiring all public members to be commented you can end up with this pattern. 

It's a pity JavaDoc hasn't evolved more fine grained control over years

PR #649, opened 2025-12-19
feat: introduce ConverterProvider to control conversion behaviour

vbarua:
Introduces the ConverterProvider class as a single source of configuration for
all conversion related classes between SQL, Calcite and Substrait. All
conversion classes now have constructors consuming a ConverterProvider.

The no-arg ConverterProvider constructor provides reasonable defaults for
conversions. Other constructors can override some behaviours, and full
customization can be achieved by extending the ConverterProvider, as is done in 
DynamicConverterProvider

BREAKING CHANGE: removed FeatureBoard class
BREAKING CHANGE: all constructors consume a FeatureBoard have been removed
BREAKING CHANGE: a number of methods/constructors have been replaced with ConverterProvider equivalents

### DIFF
--- a/isthmus/src/main/java/io/substrait/isthmus/DynamicConverterProvider.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/DynamicConverterProvider.java
@@ -25,7 +25,7 @@ public class DynamicConverterProvider extends ConverterProvider {
   public DynamicConverterProvider(
       SimpleExtension.ExtensionCollection extensions, RelDataTypeFactory typeFactory) {
     super(extensions, typeFactory);
-    this.scalarFunctionConverter = createScalarFunctionConverter(extensions, typeFactory);
+    this.scalarFunctionConverter = createScalarFunctionConverter();
   }
 
   @Override
@@ -42,10 +42,7 @@ public class DynamicConverterProvider extends ConverterProvider {
             .collect(Collectors.toList());
     callConverters.add(
         new ScalarFunctionConverter(
-            extensions.scalarFunctions(),
-            additionalSignatures,
-            typeFactory,
-            TypeConverter.DEFAULT));
+            extensions.scalarFunctions(), additionalSignatures, typeFactory, typeConverter));
     return callConverters;
   }
 
@@ -63,9 +60,7 @@ public class DynamicConverterProvider extends ConverterProvider {
     return SqlOperatorTables.chain(operatorTable, SqlOperatorTables.of(generatedDynamicOperators));
   }
 
-  private static ScalarFunctionConverter createScalarFunctionConverter(
-      SimpleExtension.ExtensionCollection extensions, RelDataTypeFactory typeFactory) {
-
+  private ScalarFunctionConverter createScalarFunctionConverter() {
     List<FunctionMappings.Sig> additionalSignatures = Collections.emptyList();
 
     java.util.Set<String> knownFunctionNames =
@@ -92,6 +87,6 @@ public class DynamicConverterProvider extends ConverterProvider {
     }
 
     return new ScalarFunctionConverter(
-        extensions.scalarFunctions(), additionalSignatures, typeFactory, TypeConverter.DEFAULT);
+        extensions.scalarFunctions(), additionalSignatures, typeFactory, typeConverter);
   }
 }


isthmus/src/main/java/io/substrait/isthmus/ConverterProvider.java
```diff
@@ -0,0 +1,77 @@
...
+import io.substrait.isthmus.expression.ScalarFunctionConverter;
+import io.substrait.isthmus.expression.SqlArrayValueConstructorCallConverter;
+import io.substrait.isthmus.expression.SqlMapValueConstructorCallConverter;
+import io.substrait.isthmus.expression.WindowFunctionConverter;
+import java.util.ArrayList;
+import java.util.List;
+import org.apache.calcite.rel.type.RelDataTypeFactory;
+import org.apache.calcite.rex.RexBuilder;
+
+public class ConverterProvider {
```
vbarua, line 47: We wire in the same 4 converters (3 function converter + 1 type converter) in a number of places (SubstraitRelVisitor, SubstraitRelNodeConverter, ...). Instead of doing that, we can centralized the configuration of converters into one provider*. This would also allow us to provide extended providers, like the DynamicConverterProvider, that could implement the dynamic fallback behaviour in a single location rather that sprinkling it throughout the various parts of the codebase, which makes it very difficult to reason about.

\* I'm not set on the name

isthmus/src/main/java/io/substrait/isthmus/SqlToSubstrait.java
```diff
@@ -2,70 +2,39 @@
...
-            SqlOperatorTables.chain(
-                SubstraitOperatorTable.INSTANCE, SqlOperatorTables.of(generatedDynamicOperators));
-        return;
-      }
-    }
-    this.operatorTable = SubstraitOperatorTable.INSTANCE;
+  /** Use {@link SqlToSubstrait#SqlToSubstrait(ConverterProvider)} instead */
+  @Deprecated
+  public SqlToSubstrait(SimpleExtension.ExtensionCollection extensions) {
+    this(new ConverterProvider(extensions));
```
bestbeforetoday: It seems strange to define a new constructor that is already deprecated. I wonder if it should not exist at all.

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelVisitor.java
```diff
@@ -59,88 +57,58 @@
...
-    ArrayList<CallConverter> converters = new ArrayList<CallConverter>();
-    converters.addAll(CallConverters.defaults(typeConverter));
-    converters.add(scalarFunctionConverter);
-    converters.add(CallConverters.CREATE_SEARCH_CONV.apply(new RexBuilder(typeFactory)));
-    this.aggregateFunctionConverter = aggregateFunctionConverter;
-    this.rexExpressionConverter =
-        new RexExpressionConverter(this, converters, windowFunctionConverter, typeConverter);
-    this.typeConverter = typeConverter;
-    this.featureBoard = features;
+      TypeConverter typeConverter) {
```
bestbeforetoday: The constructor signature has changed anyway so this is effectively a new constructor that replaces one we no longer want people to use. Since any calling code needs to be changed, perhaps it would be better not to introduce a new (deprecated) constructor and just have them make the desired change first time.

isthmus/src/main/java/io/substrait/isthmus/DynamicConverterProvider.java
```diff
@@ -0,0 +1,100 @@
...
+        ExtensionUtils.getDynamicExtensions(extensions);
+    if (!dynamicExtensionCollection.scalarFunctions().isEmpty()
+        || !dynamicExtensionCollection.aggregateFunctions().isEmpty()) {
+      List<SqlOperator> generatedDynamicOperators =
+          SimpleExtensionToSqlOperator.from(dynamicExtensionCollection, typeFactory);
+      return SqlOperatorTables.chain(
+          SubstraitOperatorTable.INSTANCE, SqlOperatorTables.of(generatedDynamicOperators));
+    }
+
+    return SubstraitOperatorTable.INSTANCE;
```
bestbeforetoday: Maybe this should get the default value from the superclass method.

isthmus/src/main/java/io/substrait/isthmus/DynamicConverterProvider.java
```diff
@@ -0,0 +1,100 @@
...
+  public SqlOperatorTable getSqlOperatorTable() {
+    SimpleExtension.ExtensionCollection dynamicExtensionCollection =
+        ExtensionUtils.getDynamicExtensions(extensions);
+    if (!dynamicExtensionCollection.scalarFunctions().isEmpty()
+        || !dynamicExtensionCollection.aggregateFunctions().isEmpty()) {
+      List<SqlOperator> generatedDynamicOperators =
+          SimpleExtensionToSqlOperator.from(dynamicExtensionCollection, typeFactory);
+      return SqlOperatorTables.chain(
+          SubstraitOperatorTable.INSTANCE, SqlOperatorTables.of(generatedDynamicOperators));
+    }
```
bestbeforetoday: **Very** minor style _suggestion_: testing if both are empty gives the very simple short-circuit case where the default value is returned, which means less indented code and might improve readability.

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelVisitor.java
vbarua: Ah, the original constructor consumed a FeatureBoard which I've fully removed in this PR. The original constructor doesn't exists anymore.

I'm just going to fully remove this constructor.

isthmus/src/main/java/io/substrait/isthmus/SqlToSubstrait.java
vbarua: You're right, this doesn't replace an existing one because I removed the FeatureBoard. I'll remove it.

isthmus/src/main/java/io/substrait/isthmus/DynamicConverterProvider.java
```diff
@@ -0,0 +1,98 @@
...
+import io.substrait.isthmus.expression.ScalarFunctionConverter;
+import java.util.Collections;
+import java.util.List;
+import java.util.stream.Collectors;
+import org.apache.calcite.rel.type.RelDataTypeFactory;
+import org.apache.calcite.sql.SqlOperator;
+import org.apache.calcite.sql.SqlOperatorTable;
+import org.apache.calcite.sql.util.SqlOperatorTables;
+
+public class DynamicConverterProvider extends ConverterProvider {
```
vbarua, line 15: This class represents the ultimate goal of this PR. Previously, all of the code for handling dynamic conversions was smeared over various parts of the codebase. Now, it's all in one place, which make it easier to reason about how the dynamic conversion behaviour is applied.

isthmus/src/main/java/io/substrait/isthmus/SqlToSubstrait.java
```diff
@@ -1,71 +1,30 @@
...
   }
 
-  public SqlToSubstrait(FeatureBoard features) {
-    this(DefaultExtensionCatalog.DEFAULT_COLLECTION, features);
-  }
-
-  public SqlToSubstrait(SimpleExtension.ExtensionCollection extensions, FeatureBoard features) {
-    super(features, extensions);
-
-    if (featureBoard.allowDynamicUdfs()) {
```
vbarua, line 35: The behaviour below is now encapsulated in the DynamicConverterProvider

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
```diff
@@ -104,138 +100,37 @@ public class SubstraitRelNodeConverter
...
-  }
-
-  private static ScalarFunctionConverter createScalarFunctionConverter(
-      SimpleExtension.ExtensionCollection extensions,
-      RelDataTypeFactory typeFactory,
-      boolean allowDynamicUdfs) {
-
-    List<FunctionMappings.Sig> additionalSignatures;
-
-    if (allowDynamicUdfs) {
```
vbarua, line 171: The dynamic function handling code is now encapsulated in the DynamicConverterProvider

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelVisitor.java
```diff
@@ -58,88 +54,41 @@
...
-  public SubstraitRelVisitor(
-      RelDataTypeFactory typeFactory,
-      SimpleExtension.ExtensionCollection extensions,
-      FeatureBoard features) {
-
-    this.typeConverter = TypeConverter.DEFAULT;
-    ArrayList<CallConverter> converters = new ArrayList<>();
-    converters.addAll(CallConverters.defaults(typeConverter));
-
-    if (features.allowDynamicUdfs()) {
```
vbarua, line 97: This code is now encapuslated in the DynamicConverterProvider

isthmus/src/test/java/io/substrait/isthmus/CustomFunctionTest.java
```diff
@@ -95,22 +96,7 @@ public RelDataType toCalcite(Type.UserDefined type) {
...
-          FunctionMappings.s(customScalarAnyToAnyFn),
-          FunctionMappings.s(customScalarAny1Any1ToAny1Fn),
-          FunctionMappings.s(customScalarAny1Any2ToAny2Fn),
-          FunctionMappings.s(customScalarListAnyFn),
-          FunctionMappings.s(customScalarListAnyAndAnyFn),
-          FunctionMappings.s(customScalarListStringFn),
-          FunctionMappings.s(customScalarListStringAndAnyFn),
-          FunctionMappings.s(customScalarListStringAndAnyVariadic0Fn),
-          FunctionMappings.s(customScalarListStringAndAnyVariadic1Fn),
-          FunctionMappings.s(toBType));
```
vbarua, line 112: Moved below so that it can be declared `static`

isthmus-cli/src/main/java/io/substrait/isthmus/cli/IsthmusEntryPoint.java
```diff
@@ -57,11 +52,6 @@ enum OutputFormat {
     BINARY, // protobuf BINARY format
   }
 
-  @Option(
-      names = {"--unquotedcasing"},
-      description = "Calcite's casing policy for unquoted identifiers: ${COMPLETION-CANDIDATES}")
-  private Casing unquotedCasing = Casing.TO_UPPER;
```
bestbeforetoday, line 63: Removing this option will break the use-case described in PR #351. I guess it should be possible to continue support by having IsthmusEntryPoint explicitly use its own subclass of ConverterProvider that overrides `getSqlParserConfig()` to set the desired unquotedCasing value.

isthmus/src/main/java/io/substrait/isthmus/DynamicConverterProvider.java
```diff
@@ -0,0 +1,93 @@
...
+
+  private final SimpleExtension.ExtensionCollection extensions;
+
+  public DynamicConverterProvider(
+      SimpleExtension.ExtensionCollection extensions, RelDataTypeFactory typeFactory) {
```
bestbeforetoday, line 26: Does it make sense for this class to have a constructor that takes only the `extensions` parameter and uses the default `SubstraitTypeSystem.TYPE_FACTORY` type factory, similar to the constructors of ConverterProvider? Or is the explicit type factory preferred since it is used to build the dynamic operators so is something the caller should be considering?

isthmus/src/main/java/io/substrait/isthmus/DynamicConverterProvider.java
```diff
@@ -0,0 +1,93 @@
...
+
+  @Override
+  public SqlOperatorTable getSqlOperatorTable() {
+    SqlOperatorTable operatorTable = super.getSqlOperatorTable();
+    SimpleExtension.ExtensionCollection dynamicExtensionCollection =
+        ExtensionUtils.getDynamicExtensions(extensions);
+    if (dynamicExtensionCollection.scalarFunctions().isEmpty()
+        && dynamicExtensionCollection.aggregateFunctions().isEmpty()) {
+      return operatorTable;
+    }
```
bestbeforetoday, line 57: Looking a bit harder at the code, I am not sure whether behaviour changes if this guard is omitted. It makes me nervous that this check has implicit knowledge of the implementation of the SimpleExtensionToSqlOperator. I wonder if it could just be omitted and we accept there is an additional chained operator table doing nothing in some cases. If this condition really needs to be guarded against, perhaps it could be done either in SimpleExtensionToSqlOperator, or by checking if the `generatedDynamicOperators` list below is empty to avoid the need for implementation knowledge of SimpleExtensionToSqlOperator?

isthmus/src/main/java/io/substrait/isthmus/DynamicConverterProvider.java
```diff
@@ -0,0 +1,93 @@
...
+            .collect(Collectors.toSet());
+
+    List<SimpleExtension.ScalarFunctionVariant> dynamicFunctions =
+        extensions.scalarFunctions().stream()
+            .filter(f -> !knownFunctionNames.contains(f.name().toLowerCase()))
+            .collect(Collectors.toList());
+
+    if (dynamicFunctions.isEmpty()) {
+      additionalSignatures = Collections.emptyList();
+    } else {
```
bestbeforetoday: Will the code result in an empty `additionalSignatures` list even without this guard? If so, perhaps it is better just to omit the guard.

isthmus-cli/src/main/java/io/substrait/isthmus/cli/IsthmusEntryPoint.java
vbarua: That is intentional, and I've talked with the person who introduced that change (he also works at Datadog). SQL parsing is not really an Isthmus / Substrait concern IMO, as it's purely a Calcite capability. The CLI parser was primarily intended for quick testing purposes, and we shouldn't commit to maintaining a specific parser configuration beyond what we care about for testing in this project.

On that note though, a thing I could do is split this PR into 2 pieces. One change to remove the FeatureBoard as it's own thing, and then another to introduce the ConverterProvider. It would potentially make the review easier. Let me know if you'd like me to do that.

isthmus/src/main/java/io/substrait/isthmus/DynamicConverterProvider.java
vbarua: That seems pretty reasonable, added some extra constructors.

isthmus/src/main/java/io/substrait/isthmus/DynamicConverterProvider.java
vbarua: My aim with this PR was to retain the existing behaviour/functionality of the dynamic function implementation. I agree that it could use another pass, and in fact my initial merge of that work was based on making it opt-int so we could iterate on it https://github.com/substrait-io/substrait-java/pull/457#pullrequestreview-3542723982.

I want to avoid modifications of that logic in this PR, as I think that should be it's own work. Does that sound reasonable? Now that all the logic is one place as well, hopefully it will be easier to to reason about.

isthmus/src/main/java/io/substrait/isthmus/DynamicConverterProvider.java
vbarua: This is an easy enough simplification. Done.

isthmus-cli/src/main/java/io/substrait/isthmus/cli/IsthmusEntryPoint.java
bestbeforetoday: I think it is fine to both remove the FeatureBoard and introduce the ConverterProvider in this PR. My concern was that the `isthmus` CLI command is losing the `--unquotedcasing` option in the process.

If nobody needs that option anymore, it could be removed. It seems like a good idea to do that in a different PR since it is unrelated to this change.

If we want to keep that option, it should be easy to implement using ConverterProvider. It even acts as a demonstration that isthmus configuration can easily be customised by calling code.

```java
  @Option(
      names = {"--unquotedcasing"},
      description = "Calcite's casing policy for unquoted identifiers: ${COMPLETION-CANDIDATES}")
  private Casing unquotedCasing = Casing.TO_UPPER;

  private ConverterProvider converterProvider() {
    return new ConverterProvider() {
      @Override
      public SqlParser.Config getSqlParserConfig() {
        return super.getSqlParserConfig().withUnquotedCasing(unquotedCasing);
      }
    };
  }
```

Then pass `converterProvider()` to the constructors of `SqlExpressionToSubstrait` and `SqlToSubstrait`.

isthmus/src/main/java/io/substrait/isthmus/DynamicConverterProvider.java
bestbeforetoday: I see what you mean. Even if the code is not optimal, it is what was there before your refactor. No need to hold up this PR to tidy up existing code.

isthmus-cli/src/main/java/io/substrait/isthmus/cli/IsthmusEntryPoint.java
vbarua: Oooh, I really like that idea. Implemented.

PR #652, opened 2025-12-23
test: avoid duplicating functionality from base test classes

vbarua:
* Reuse base R and N type creators
* Standardize on sb from SubstraitBuilder util
* Remove defaultExtensionCollection from tests
* Introduce reusable substraitToCalcite in PlanTestBase

### DIFF
--- a/core/src/test/java/io/substrait/TestBase.java
+++ b/core/src/test/java/io/substrait/TestBase.java
@@ -15,10 +15,11 @@ import io.substrait.type.TypeCreator;
 import java.io.IOException;
 
 public abstract class TestBase {
-  protected final SimpleExtension.ExtensionCollection extensions;
 
-  protected TypeCreator R = TypeCreator.REQUIRED;
-  protected TypeCreator N = TypeCreator.NULLABLE;
+  protected static final TypeCreator R = TypeCreator.REQUIRED;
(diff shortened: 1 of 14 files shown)


core/src/test/java/io/substrait/TestBase.java
```diff
@@ -10,24 +12,38 @@
...
-  protected TypeCreator N = TypeCreator.NULLABLE;
+  protected final SimpleExtension.ExtensionCollection extensions;
 
-  protected SubstraitBuilder b = new SubstraitBuilder(defaultExtensionCollection);
   protected ExtensionCollector functionCollector = new ExtensionCollector();
   protected RelProtoConverter relProtoConverter = new RelProtoConverter(functionCollector);
-  protected ProtoRelConverter protoRelConverter =
-      new ProtoRelConverter(functionCollector, defaultExtensionCollection);
+
+  protected SubstraitBuilder sb;
```
vbarua, line 27: I opted to use `sb` for SubstraitBuilder, which felt slightly more descriptive than `b`. A short-name is useful because when it's used, it's called frequently to build a plan component. A descriptive name like `substraitBuilder` ends up being noisy.

andrew-coleman (approved):

PR #661, opened 2026-01-06
build: enable Gradle caching / remove setup-gradle

nielspardon:
I've been seeing build failures recently where Maven Central responds with a 403 while fetching dependencies. This could be due to some Github Actions workers causing excessive traffic for Maven Central and being blocked: https://central.sonatype.org/faq/403-error-central/

This PR enables Gradle caching on Github Actions to limit traffic to Maven Central: https://github.com/actions/setup-java?tab=readme-ov-file#caching-gradle-dependencies

Further since we are using the Gradle wrapper I don't think we need the `setup-gradle` Github Actions task so I'm removing it which should also speed up the builds a little.

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -39,8 +39,7 @@ jobs:
         with:
           java-version: '17'
           distribution: 'temurin'
-      - name: Setup Gradle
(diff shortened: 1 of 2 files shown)


nielspardon:
closing this for now. will try to understand the setup-gradle task better.

PR #665, opened 2026-01-08
refactor(isthmus): simplify OuterReferenceResolver

nielspardon:
This simplifies the `OuterReferenceResolver` as identified in https://github.com/substrait-io/substrait-java/pull/645/files#r2672534015 

The way the `OuterReferenceResolver` is used in the code today does not encapsulate the `fieldAccessDepthMap` within the resolver but it gets used directly in `SubstraitRelVisitor` meaning that the `getStepsOut` method is never used. We could either try to encapsulate the map better within `OuterReferenceResolver` or we simply return the map as a result of the `OuterReferenceResolver` which this PR does:

- removes `getStepsOut` and `getFieldAccessDepthMap`
- returns `fieldAccessDepthMap` in `apply` method


benbellick:
I haven't looked into this yet but wondering if there is value in adding any tests that ensure the behavior is still as we expect?

bestbeforetoday (approved):

PR #676, opened 2026-01-15
fix(isthmus): correct strpos parameter order

### LINKED ISSUE #662: STRPOS and POSITION
The subtrait function strpos was mapped to calcite's POSITION function in this [PR](https://github.com/substrait-io/substrait-java/pull/610/files).

I believe the two are not one to one mappings of one another as the semantics agruments are swapped. Substrait describes [strpos](https://github.com/substrait-io/substrait/blob/92d2e757a330f9c973bb566817dc92afd1badcb2/extensions/functions_string.yaml#L609), as the second arugment being the substring, while [calcite's position](https://github.com/apache/calcite/blob/35768a76e71eb2b8c1150494b7cd53af77c40574/core/src/main/java/net/hydromatic/optiq/runtime/SqlFunctions.java#L958), has the first argument being the substring.

bestbeforetoday:
Both SQL STRPOS and POSITION functions are modelled by Calcite as Position. Parameters are substring followed by the input to be searched. This maps to the Substrait strpos function, which has the parameters of input to be searched followed by substring. The default mapping retains the argument order, which results in SQL being translated to incorrect Substrait, and correct Substrait translated to incorrect SQL. Round-trip between Substrait and SQL still works since the parameter order is reversed in both directions, resulting in the correct order again.

This change adds a custom function mapper to reverse the order of the first two parameters when converting between Calcite Position and Substrait strpos functions.

Closes #662

### DIFF
--- a/isthmus/src/main/java/io/substrait/isthmus/expression/PositionFunctionMapper.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/expression/PositionFunctionMapper.java
@@ -1,0 +1,59 @@
+package io.substrait.isthmus.expression;
+
+import io.substrait.expression.Expression;
+import io.substrait.expression.FunctionArg;
+import io.substrait.extension.SimpleExtension.ScalarFunctionVariant;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import java.util.Optional;
+import java.util.stream.Collectors;
+import org.apache.calcite.rex.RexCall;
+import org.apache.calcite.rex.RexNode;
+import org.apache.calcite.sql.fun.SqlStdOperatorTable;
+
+/**
+ * Custom mapping for the Calcite {@code POSITION} function to the Substrait {@code strpos}
+ * function. Calcite also represents the SQL {@code STRPOS} function as {@code POSITION}.
+ *
+ * <p>Calcite {@code POSITION} has <em>substring</em> followed by <em>input</em> parameters, while
+ * Substrait {@code strpos} has <em>input</em> followed by <em>substring</em>. When mapping between
+ * Calcite and Substrait, the parameters need to be reversed
+ *
+ * <p>{@code POSITION(substring IN input)} maps to {@code strpos(input, substring)}.
+ */
+final class PositionFunctionMapper implements ScalarFunctionMapper {
+  private static final String strposFunctionName = "strpos";
+  private final List<ScalarFunctionVariant> strposFunctions;
+
+  public PositionFunctionMapper(List<ScalarFunctionVariant> functions) {
+    strposFunctions =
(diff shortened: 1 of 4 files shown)


isthmus/src/main/java/io/substrait/isthmus/expression/PositionFunctionMapper.java
```diff
@@ -0,0 +1,57 @@
...
+  }
+
+  @Override
+  public Optional<SubstraitFunctionMapping> toSubstrait(final RexCall call) {
+    if (!SqlStdOperatorTable.POSITION.equals(call.op)) {
+      return Optional.empty();
+    }
+
+    List<RexNode> operands = new ArrayList<>(call.getOperands());
+    Collections.swap(operands, 0, 1);
```
benbellick, line 43: Probably not a bad idea to put a comment here just because we have some naked numbers used to reference parameters.

isthmus/src/test/java/io/substrait/isthmus/StringFunctionTest.java
```diff
@@ -227,87 +225,95 @@ void testContains(String left, String right) throws Exception {
...
+      value = {"vc, 'substring'", "'string', vc"},
       quoteCharacter = '`')
-  void testStrposWithLiteral(String left, String right) throws Exception {
+  void testStrposWithLiteral(String input, String substring) throws Exception {
+    String query = String.format("SELECT STRPOS(%s, %s) FROM strings", input, substring);
+    assertSqlRoundTrip(query);
+  }
 
-    String query = String.format("SELECT STRPOS(%s, %s) > 0 FROM strings", left, right);
+  @Test
```
benbellick, line 264:
```suggestion
  @Test
  // Verifies POSITION→strpos parameter swap (#662).
  // SQL POSITION(substring IN input) must become Substrait strpos(input, substring).
```

Just so someone has a reference for the future :)

isthmus/src/main/java/io/substrait/isthmus/expression/PositionFunctionMapper.java
bestbeforetoday: Is the class Javadoc not sufficient?

benbellick: It's probably fine, just my opinion if we wanted to make it extra clear. Feel free to ignore if you disagree.

benbellick:
@bestbeforetoday sorry, I'm not sure I am able to see those comment changes? 

Also, very small thing, but since we squash merge in this repo anyways, I think it is better to always do changes in new commits rather than force-pushing. This way, I can view the diff of what changed just from when I last viewed. Thanks! 🙂

PR #677, opened 2026-01-15
feat(isthmus): support Substrait struct literal conversions

### LINKED ISSUE #612: Handle struct-based UDT literals in isthmus

benbellick:
Enables the conversion of Substrait:
* Expression.StructLiteral
* Expression.UserDefinedStructLiteral

to Calcite and back

Closes #612

### DIFF
--- a/isthmus/src/main/java/io/substrait/isthmus/expression/CallConverters.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/expression/CallConverters.java
@@ -4,6 +4,7 @@ import com.google.common.collect.ImmutableList;
 import io.substrait.expression.Expression;
 import io.substrait.expression.ExpressionCreator;
 import io.substrait.isthmus.CallConverter;
+import io.substrait.isthmus.SubstraitRelNodeConverter;
 import io.substrait.isthmus.TypeConverter;
 import io.substrait.type.Type;
 import java.util.ArrayList;
@@ -68,8 +69,7 @@ public class CallConverters {
             Expression operand = visitor.apply(call.getOperands().get(0));
             Type type = typeConverter.toSubstrait(call.getType());
 
-            // For now, we only support handling of SqlKind.REINTEPRETET for the case of stored
-            // user-defined literals
+            // Calcite encoded Expression.UserDefinedAnyLiteral
             if (operand instanceof Expression.FixedBinaryLiteral
                 && type instanceof Type.UserDefined) {
               Expression.FixedBinaryLiteral literal = (Expression.FixedBinaryLiteral) operand;
@@ -90,7 +90,9 @@ public class CallConverters {
               } catch (com.google.protobuf.InvalidProtocolBufferException e) {
                 throw new IllegalStateException("Failed to parse UserDefinedAnyLiteral value", e);
               }
-            } else if (operand instanceof Expression.StructLiteral
+            }
+            // Calcite encoded Expression.UserDefinedStructLiteral
+            else if (operand instanceof Expression.StructLiteral
                 && type instanceof Type.UserDefined) {
               Expression.StructLiteral structLiteral = (Expression.StructLiteral) operand;
               Type.UserDefined t = (Type.UserDefined) type;
@@ -106,18 +108,8 @@ public class CallConverters {
             return null;
           };
 
-  //  public static SimpleCallConverter OrAnd(FunctionConverter c) {
-  //      return (call, visitor) -> {
-  //        if (call.getKind() != SqlKind.AND && call.getKind() != SqlKind.OR) {
-  //          return null;
-  //        }
-  //
-  //
-  //        return null;
-  //      };
-  //  }
   /**
-   * Converts Calcite ROW constructors into Substrait struct literals.
+   * Converts Calcite ROW constructors into Substrait {@link Expression.StructLiteral}s.
    *
    * <p>ROW values are always concrete (never null themselves) - if a value is actually null, use
    * NullLiteral instead of StructLiteral. Therefore, the resulting StructLiteral always has


isthmus/src/test/java/io/substrait/isthmus/utils/UserTypeFactory.java
```diff
@@ -19,34 +21,65 @@ public class UserTypeFactory {
...
 
   private static class InnerType extends RelDataTypeImpl {
     private final boolean nullable;
+    private final String urn;
     private final String name;
+    private final List<Type.Parameter> typeParameters;
 
-    private InnerType(boolean nullable, String name) {
-      computeDigest();
-      this.nullable = nullable;
```
benbellick, line 48: Was this a bug before? I thought that `computeDigest()` had to be the last thing called.

benbellick:
I'll be honest and say that I used a lot of agentic coding to figure this one out due to my lack of familiarity with `Calcite`. Hoping a `Calcite` expert (e.g. @asolimando) could help offer a review :)

isthmus/src/main/java/io/substrait/isthmus/expression/CallConverters.java
```diff
@@ -100,6 +118,51 @@ public class CallConverters {
...
+          return null;
+        }
+
+        List<Expression> operands =
+            call.getOperands().stream().map(visitor).collect(java.util.stream.Collectors.toList());
+        if (!operands.stream().allMatch(expr -> expr instanceof Expression.Literal)) {
+          throw new IllegalArgumentException("ROW operands must be literals.");
+        }
+
+        java.util.List<org.apache.calcite.rel.type.RelDataTypeField> fieldTypes =
```
asolimando: Nit: we could use imports over FQN here, at least for `RelDataTypeField`? (I assume there might be clashes for `List` and `Collectors`).

isthmus/src/main/java/io/substrait/isthmus/expression/CallConverters.java
```diff
@@ -86,8 +88,24 @@ public class CallConverters {
...
+              return Expression.UserDefinedStructLiteral.builder()
+                  .nullable(t.nullable())
+                  .urn(t.urn())
+                  .name(t.name())
+                  .addAllTypeParameters(t.typeParameters())
+                  .addAllFields(structLiteral.fields())
+                  .build();
             }
-            return null;
+            throw new IllegalStateException(
```
asolimando: I guess this might be a breaking change, in my understanding the whole Converters design in substrait relies on them returning `null` to signal they "gave up" and to allow further converters to try.

This change could break users relying on later converters to handle custom cases, I'd rather keep returning `null` here or add a comment explaining why in this specific cases an exception must be thrown.

isthmus/src/main/java/io/substrait/isthmus/expression/CallConverters.java
```diff
@@ -100,6 +118,51 @@ public class CallConverters {
...
+            call.getType().getFieldList();
+        List<Expression.Literal> literals = new java.util.ArrayList<>();
+
+        for (int i = 0; i < operands.size(); i++) {
+          Expression.Literal lit = (Expression.Literal) operands.get(i);
+          boolean fieldIsNullable = fieldTypes.get(i).getType().isNullable();
+
+          // ROW types are never nullable (struct literals are always concrete values).
+          // Field nullability comes from individual field types.
+          if (fieldIsNullable && !lit.nullable()) {
```
asolimando: I wonder if we can end up with a reverse case (non nullable field type but a `null` literal).

Can we rely on this case being validated by any previous validation phase or should be enforce this here?

If the answer is true, I'd add a comment explaining that, if not we should probably add a test.

isthmus/src/test/java/io/substrait/isthmus/utils/UserTypeFactory.java
asolimando: IIRC we discussed that on another PR, and I confirm that `computeDigest()` must be placed after all the fields modifications (for at least all fields appearing composing the digest, which is usually all of them).

So, yes, this was most probably a bug.

isthmus/src/main/java/io/substrait/isthmus/expression/CallConverters.java
benbellick: Yup, that definitely seems correct:
https://github.com/substrait-io/substrait-java/blob/be6f4ae3bb1f73133c38c06b8df48ff58e9c2cab/isthmus/src/main/java/io/substrait/isthmus/expression/RexExpressionConverter.java#L75-L82

Thanks for identifying that!

isthmus/src/main/java/io/substrait/isthmus/expression/CallConverters.java
benbellick: (In trying to answer this question, I found an opportunity to improve the ergonomics of null literals in #686)

AFAICT, `lit` here will always but non-nullable because the nullability is lost on conversion to calcite. It is actually not necessary to do the check for `lit.nullable()` and should be dropped. Instead, I'll just do 
```java
  lit = lit.withNullable(fieldIsNullable);
```

core/src/main/java/io/substrait/expression/Expression.java
```diff
@@ -30,6 +30,13 @@ interface Literal extends Expression {
     default boolean nullable() {
       return false;
     }
+
+    /**
+     * Returns a copy of this literal with the specified nullability.
+     *
+     * <p>This method is implemented by all concrete Literal classes via Immutables code generation.
+     */
+    Literal withNullable(boolean nullable);
```
vbarua, line 39: Oooh, that's handy.

I think we can use this to replace the https://github.com/substrait-io/substrait-java/blob/be6f4ae3bb1f73133c38c06b8df48ff58e9c2cab/core/src/main/java/io/substrait/type/TypeCreator.java#L127 which uses a whole visitor just to change the nullability 😅

Would be worth quick follow-up after this PR.

isthmus/src/main/java/io/substrait/isthmus/expression/CallConverters.java
```diff
@@ -142,21 +137,19 @@ public class CallConverters {
...
+        List<RelDataTypeField> fieldTypes = call.getType().getFieldList();
+        List<Expression.Literal> literals =
+            java.util.stream.IntStream.range(0, operands.size())
+                .mapToObj(
+                    i -> {
+                      Expression.Literal lit = (Expression.Literal) operands.get(i);
+                      boolean fieldIsNullable = fieldTypes.get(i).getType().isNullable();
+                      return lit.withNullable(fieldIsNullable);
+                    })
+                .collect(java.util.stream.Collectors.toList());
```
asolimando: Embedded FQN strikes again :)

benbellick: Good catch :)

benbellick: I should figure out how to add this to auto linting.

benbellick: I went ahead and fixed the other FQNs for Collectors in that file.

PR #682, opened 2026-01-20
feat: allow the creation VirtualTableScans with no rows 

### LINKED ISSUE #673: Drop EmptyScan in favor of allowing empty VirtualTableScan
VirtualTableScan currently requires at least one row:
 https://github.com/substrait-io/substrait-java/blob/13309df36b01b4fb3f4b8edc5c8ad8a8dffe2e92/core/src/main/java/io/substrait/relation/VirtualTableScan.java#L48 

This made sense when the schema was derived from the first row ([714eb77](https://github.com/substrait-io/substrait-java/commit/714eb774e328ee234e1673a48270916a80442205)), which required EmptyScan as a separate class to represent empty tables. However, since #272, the schema is explicitly provided via .initialSchema() and is independent of the rows. This makes the non-empty constraint outdated and EmptyScan unnecessary. 

We should allow empty VirtualTableScan and deprecate EmptyScan.

gord02:
This PR removes the `io.substrait.relation.EmptyScan` class. It instead allows `io.substrait.relation.VirtualTableScan` to be empty. 

Closes #673 

BREAKING CHANGE: The `io.substrait.relation.EmptyScan` class has been removed.

### DIFF
--- a/core/src/main/java/io/substrait/relation/VirtualTableScan.java
+++ b/core/src/main/java/io/substrait/relation/VirtualTableScan.java
@@ -17,6 +17,7 @@ public abstract class VirtualTableScan extends AbstractReadRel {
    * Checks the following invariants when construction a VirtualTableScan
    *
    * <ul>
+   *   <li>no null field names
    *   <li>no null rows
    *   <li>row shape must match field-list
    *   <li>row field types must match schema types


core/src/main/java/io/substrait/dsl/SubstraitBuilder.java
```diff
@@ -315,8 +315,8 @@ private NamedScan namedScan(
     return NamedScan.builder().names(tableName).initialSchema(namedStruct).remap(remap).build();
   }
 
-  public EmptyScan emptyScan() {
-    return EmptyScan.builder()
+  public VirtualTableScan emptyVirtualTableScan() {
+    return VirtualTableScan.builder()
         .initialSchema(NamedStruct.of(Collections.emptyList(), R.struct()))
         .build();
   }
```
benbellick, line 322:
```suggestion
```
I'm almost inclined to just delete this entirely now. Is this also schemaless?

benbellick: Yeah I think that if removing it isn't such a large diff, then I am in support of getting rid of it. If it ends up being a bit of a pain though, we can save it for later.

core/src/test/java/io/substrait/type/proto/ReadRelRoundtripTest.java
```diff
@@ -32,12 +32,6 @@ void namedScan() {
     verifyRoundTrip(namedScan);
   }
 
-  @Test
-  void emptyScan() {
-    io.substrait.relation.EmptyScan emptyScan = sb.emptyScan();
-    verifyRoundTrip(emptyScan);
-  }
```
benbellick, line 39: I think we might as well just keep this test but make it a roundtrip test on the empty virtual table.

core/src/main/java/io/substrait/dsl/SubstraitBuilder.java
benbellick: Okay I see now its used in a bunch of tests, so no worries about keeping it for now.

benbellick (commented):
all looks reasonable to me! Only thing really is just putting that roundtrip test back in. 

Also, maybe we will want to mark this as a breaking change since we are technically modifying the API here by dropping a public class. @vbarua

core/src/main/java/io/substrait/relation/ProtoRelConverter.java
```diff
@@ -186,21 +186,14 @@ public Rel from(io.substrait.proto.Rel rel) {
...
-    } else if (rel.hasNamedTable()) {
+    if (rel.hasNamedTable()) {
       return newNamedScan(rel);
     } else if (rel.hasLocalFiles()) {
       return newLocalFiles(rel);
     } else if (rel.hasExtensionTable()) {
       return newExtensionTable(rel);
     } else {
-      return newEmptyScan(rel);
+      return newVirtualTable(rel);
```
benbellick: The previous implementation of this function was using `newEmptyScan` as a fallback (but was explicitly checking if a table was a virtual table before constructing one). 

The new implementation now just falls back to a virtual table every time. I think a more accurate implementation would go back to checking if the rel is a virtual table. Then as a fallback, it should probably just throw an exception like:
```java
          "ReadRel must have one of: namedTable, localFiles, extensionTable, or virtualTable");
```

This way, we don't silently treat potential future table types as a virtual table.

benbellick (commented):
Just the one comment here and then I think it's great!
https://github.com/substrait-io/substrait-java/pull/682/files#r2712955699

isthmus/src/test/java/io/substrait/isthmus/VirtualTableScanTest.java
```diff
@@ -59,6 +61,12 @@ void expressionContainingVirtualTable() {
         explain(relNode));
   }
 
+  @Test
+  void emptyVirtualTableScan() {
```
asolimando, line 66: This test is good but very minimal (empty schema, empty rows), can you add tests with a non-empty schema and zero rows?

Something along the line of:
```
 NamedStruct schema = NamedStruct.of(List.of("col1"), R.struct(R.I32));
 assertDoesNotThrow(() -> createVirtualTableScan(schema));
```

asolimando (approved):
I have checked only the isthum/Calcite part and LGTM, asking only for one extra test

core/src/main/java/io/substrait/relation/VirtualTableScan.java
```diff
@@ -17,8 +17,6 @@ public abstract class VirtualTableScan extends AbstractReadRel {
    * Checks the following invariants when construction a VirtualTableScan
    *
    * <ul>
-   *   <li>non-empty rowset
-   *   <li>non-null field-names
```
benbellick, line 21: I think this line should be added back in right? We have only deleted the "non-empty rowset" check. The "non-null field names" check comes from `names.stream().noneMatch(Objects::isNull)` which is still there.

benbellick: I just pushed a fix myself! Hope you don't mind.

Issue #631, opened 2025-12-04
Flatten Expanded Calcite Representation of Virtual Tables w/ Complex Expressions when Converting to Substrait


gord02:
Nested Struct is implemented in this [PR](https://github.com/substrait-io/substrait-java/pull/579). 

The motivation for this PR was to implement the changes in the proto representation for VirtualTable to change from rows represented as struct literals to Nested Structs to allow for storing non-literal expressions.  This replaces the need for the Project rel above the VirtualTable storing the non-literal expressions. 

This PR currently converts the substrait VirtualTable into the following format:
```// When a VirtualTable contains non-literal expressions, they cannot be put directly into a
      // LogicalValues relation. Instead, we create a LogicalProject for each row to compute its
      // values, and combine them together using a LogicalUnion. For example the following:
      
       VirtualTable
           (e1, e2)
           (e3, e4)
      
      Becomes:
        LogicalProject(exprs=[0, 1])
          LogicalUnion(all=[true])
            LogicalProject(exprs=[e1, e2])
              <Empty Row>
            LogicalProject(exprs=[e3, e4])
              <Empty Row>

      // where each e1-e4 can represent both literal and non-literal expression
```

This implementation does not allow for round trip testing as the above format is the final output. Ideally this should be changed such that the original VirtualTable using Nested Struct is returned so that the roundtrip tests now work. 

The desired output after implementing should allow for the creation of a VirtualTableScan and have the same VirtualTable returned after Conversion to, and back from Calcite as follows:
```
Creation of VirtualTableScan using NestedStruct:
 VirtualTableScan:
    names=[col1, col2]
    NestedStruct [ fields: [(1+1,) (3)]
    NestedStruct: [ fields: [(4), (4+5)]

to Calcite:
 LogicalUnion(all=[true])
          LogicalProject(exprs=[1+1, 3])
            <Empty Row>
         LogicalProject(exprs=[4, 4+5])
            <Empty Row>

Back to Substrait:
 VirtualTableScan:
    names=[col1, col2]
    NestedStruct [ fields: [(1+1,) (3)]
    NestedStruct: [ fields: [(4), (4+5)]
```

alexandrefimov:
Taking this one.

The round trip is the part with no test behind it, so here is where it stands. A `VirtualTableScan` whose rows contain a call comes back as a different relation:

```
VirtualTableScan(col1, col2)
  (2, 4.4 + 4.5)
  (6 * 2, 8.8)

[...]

PR #684, opened 2026-01-21
fix(isthmus): preserve nullability in VirtualTableScan Calcite roundtrip

### LINKED ISSUE #683: Lost Nullability in Calcite Roundtrip
When a nullable field type is defined in the schema of a virtualTableScan, the fields are enforced to also be nullable. Calcite's logicalValues does not enforce this match of the schema and actual field implementation. The following failing test case exposes this bug. 

```java
@Test
  void lostNullabilityInVirtualTableOnRoundTrip() {
      NamedStruct schema = NamedStruct.of(List.of("col1", "col2"), R.struct(N.I32, R.FP64));
      Expression nullableI32 = Expression.I32Literal.builder().value(6).nullable(true).build();
      VirtualTableScan virtualTableScan = createVirtualTableScan(schema, List.of(nullableI32, sb.fp64(8)));
      assertFullRoundTrip(virtualTableScan);
  }
```

A workaround should either be implemented in substrait-java(potentially loosening the schema matching), or a fix in Calcite should be implemented(logicalValues should update the field list type definition when making its fields non-null).

asolimando:
Uses schema field nullability when converting Calcite `LogicalValues` literals back to Substrait, instead of deriving nullability from the `RexLiteral` type.

Calcite's `LogicalValues` may have literals with non-nullable types even when the schema field is nullable (Calcite infers the tightest possible type for literals, by design).

Previously this caused a mismatch during roundtrip conversion, as `VirtualTableScan` requires row field types to exactly match schema field types.

Fixes #683

### DIFF
--- a/isthmus/src/test/java/io/substrait/isthmus/VirtualTableScanTest.java
+++ b/isthmus/src/test/java/io/substrait/isthmus/VirtualTableScanTest.java
@@ -5,6 +5,7 @@ import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertThrows;
 
 import io.substrait.expression.Expression;
+import io.substrait.expression.ExpressionCreator;
 import io.substrait.relation.VirtualTableScan;
 import io.substrait.type.NamedStruct;
 import java.io.PrintWriter;
@@ -84,7 +85,7 @@ class VirtualTableScanTest extends PlanTestBase {
   @Test
   void nullableFieldRoundTrip() {
     NamedStruct schema = NamedStruct.of(List.of("col1", "col2"), R.struct(N.I32, R.FP64));
-    Expression nullableI32 = Expression.I32Literal.builder().value(6).nullable(true).build();
+    Expression nullableI32 = ExpressionCreator.i32(true, 6);
     VirtualTableScan virtualTableScan =
         createVirtualTableScan(schema, List.of(nullableI32, sb.fp64(8)));
     assertFullRoundTrip(virtualTableScan);
@@ -93,8 +94,8 @@ class VirtualTableScanTest extends PlanTestBase {
   @Test
   void nullLiteralRoundTrip() {
     NamedStruct schema = NamedStruct.of(List.of("col1", "col2"), R.struct(N.I32, N.FP64));
-    Expression nullI32 = Expression.NullLiteral.builder().type(N.I32).build();
-    Expression nullFp64 = Expression.NullLiteral.builder().type(N.FP64).build();
+    Expression nullI32 = ExpressionCreator.typedNull(N.I32);
+    Expression nullFp64 = ExpressionCreator.typedNull(N.FP64);
     VirtualTableScan virtualTableScan = createVirtualTableScan(schema, List.of(nullI32, nullFp64));
     assertFullRoundTrip(virtualTableScan);
   }
@@ -103,8 +104,8 @@ class VirtualTableScanTest extends PlanTestBase {
   void mixedNullabilityRoundTrip() {
     NamedStruct schema =
         NamedStruct.of(List.of("col1", "col2", "col3"), R.struct(N.I32, R.FP64, N.STRING));
-    Expression nullI32 = Expression.NullLiteral.builder().type(N.I32).build();
-    Expression nullString = Expression.NullLiteral.builder().type(N.STRING).build();
+    Expression nullI32 = ExpressionCreator.typedNull(N.I32);
+    Expression nullString = ExpressionCreator.typedNull(N.STRING);
     VirtualTableScan virtualTableScan =
         createVirtualTableScan(schema, List.of(nullI32, sb.fp64(8), nullString));
     assertFullRoundTrip(virtualTableScan);


isthmus/src/main/java/io/substrait/isthmus/SubstraitRelVisitor.java
```diff
@@ -168,13 +168,20 @@ public Rel visit(org.apache.calcite.rel.core.Values values) {
...
+                  // LogicalValues may have literals with non-nullable types even when the
+                  // schema field is nullable
                   List<Expression> fields =
-                      list.stream()
-                          .map(l -> literalConverter.convert(l))
+                      IntStream.range(0, list.size())
+                          .mapToObj(
+                              i ->
+                                  literalConverter.convert(
+                                      list.get(i), schemaFieldTypes.get(i).nullable()))
```
benbellick, line 180: This is fine but it is moments like this that I really wish Java had nice FP primitives like [zip](https://en.wikipedia.org/wiki/Zipping_(computer_science)) 😅

isthmus/src/main/java/io/substrait/isthmus/expression/LiteralConverter.java
```diff
@@ -69,10 +70,28 @@ private static BigDecimal bd(RexLiteral literal) {
   public Expression.Literal convert(RexLiteral literal) {
     // convert type first to guarantee we can handle the value.
     final Type type = typeConverter.toSubstrait(literal.getType());
-    final boolean n = type.nullable();
+    return convert(literal, type.nullable());
```
benbellick, line 73: I'm now wondering if it's the case that `type.nullable()` will always be false. If nullability of literals is always lost in calcite values, won't this always just be the default?

isthmus/src/main/java/io/substrait/isthmus/expression/LiteralConverter.java
```diff
@@ -69,10 +70,28 @@ private static BigDecimal bd(RexLiteral literal) {
...
+   * <p>This overload is useful when the target nullability should come from the schema rather than
+   * the literal's own type. For example, Calcite's LogicalValues may have literals with
+   * non-nullable types even when the schema field is nullable.
+   *
+   * @param literal the RexLiteral to convert
+   * @param nullable the nullability to use for the resulting Substrait literal
+   * @return the converted Substrait Literal
+   */
+  public Expression.Literal convert(RexLiteral literal, boolean nullable) {
+    final boolean n = nullable;
```
benbellick: small nit, but I think it makes more sense to just use `nullable` everywhere in this method instead of assigning to `n`. Unless there is a value of marking it as final here?

isthmus/src/test/java/io/substrait/isthmus/VirtualTableScanTest.java
```diff
@@ -59,6 +59,16 @@ void expressionContainingVirtualTable() {
...
 
+  @Test
+  void nullableFieldRoundTrip() {
+    // Test that nullable fields in the schema are preserved during Calcite roundtrip
+    NamedStruct schema = NamedStruct.of(List.of("col1", "col2"), R.struct(N.I32, R.FP64));
+    Expression nullableI32 = Expression.I32Literal.builder().value(6).nullable(true).build();
+    VirtualTableScan virtualTableScan =
+        createVirtualTableScan(schema, List.of(nullableI32, sb.fp64(8)));
+    assertFullRoundTrip(virtualTableScan);
+  }
```
benbellick, line 92: This is a great test. I wonder if it is worth creating an **evil** test that also checks nested nullability? I.e. 
```
nstruct<
  col1:list<i32?>,
  col2:nstruct<
    f1:i32?,
    f2:i32
  >
>
```

benbellick: I actually tried to look into this a bit locally... seems like the conversion to calcite struggles a bit anyways to handle these nested types (nullable or not). That may just be another ticket worth opening up in the future

benbellick: I think this may be relevant: https://github.com/substrait-io/substrait-java/issues/631

isthmus/src/main/java/io/substrait/isthmus/expression/LiteralConverter.java
asolimando: No, `null` literals have nullable types in Calcite, so it's not always non-nullable.
I have added the `nullLiteralRoundTrip` test to show this.

isthmus/src/main/java/io/substrait/isthmus/expression/LiteralConverter.java
asolimando: No, no added value, it's a leftover from a previous attempt I did, I will use `nullable` as you suggest

isthmus/src/test/java/io/substrait/isthmus/VirtualTableScanTest.java
asolimando: I agree that this relates to #631, we can probably defer to it and scope it out of the current PR, wdyt?

isthmus/src/main/java/io/substrait/isthmus/expression/LiteralConverter.java
```diff
@@ -69,72 +70,89 @@ private static BigDecimal bd(RexLiteral literal) {
...
+   * @param nullable the nullability to use for the resulting Substrait literal
+   * @return the converted Substrait Literal
+   */
+  public Expression.Literal convert(RexLiteral literal, boolean nullable) {
     if (literal.isNull()) {
-      return ExpressionCreator.typedNull(type);
+      final Type type = typeConverter.toSubstrait(literal.getType());
+      final Type typeWithNullability =
+          nullable ? TypeCreator.asNullable(type) : TypeCreator.asNotNullable(type);
+      return ExpressionCreator.typedNull(typeWithNullability);
```
benbellick, line 92: I think that as of #686, it is the case that `typedNull` will actually throw an exception if it is passed a non-null type. Just commenting so you are aware that if the `TypeCreator.asNotNullable` case is hit here, an exception will be thrown.

isthmus/src/test/java/io/substrait/isthmus/VirtualTableScanTest.java
```diff
@@ -59,6 +59,35 @@ void expressionContainingVirtualTable() {
...
+
+  @Test
+  void nullLiteralRoundTrip() {
+    NamedStruct schema = NamedStruct.of(List.of("col1", "col2"), R.struct(N.I32, N.FP64));
+    Expression nullI32 = Expression.NullLiteral.builder().type(N.I32).build();
+    Expression nullFp64 = Expression.NullLiteral.builder().type(N.FP64).build();
```
benbellick: very minor nit: 
```suggestion
    Expression nullI32 = ExpressionCreator.typedNull(N.I32);
    Expression nullFp64 = ExpressionCreator.typedNull(N.FP64);
```

Arguably not even necessary, just slightly more concise.

isthmus/src/test/java/io/substrait/isthmus/VirtualTableScanTest.java
```diff
@@ -59,6 +59,35 @@ void expressionContainingVirtualTable() {
...
+    VirtualTableScan virtualTableScan = createVirtualTableScan(schema, List.of(nullI32, nullFp64));
+    assertFullRoundTrip(virtualTableScan);
+  }
+
+  @Test
+  void mixedNullabilityRoundTrip() {
+    NamedStruct schema =
+        NamedStruct.of(List.of("col1", "col2", "col3"), R.struct(N.I32, R.FP64, N.STRING));
+    Expression nullI32 = Expression.NullLiteral.builder().type(N.I32).build();
+    Expression nullString = Expression.NullLiteral.builder().type(N.STRING).build();
```
benbellick: and here as well
```suggestion
    Expression nullI32 = ExpressionCreator.typedNull(N.I32);
    Expression nullString = ExpressionCreator.typedNull(N.STRING);
```

isthmus/src/test/java/io/substrait/isthmus/VirtualTableScanTest.java
```diff
@@ -59,6 +59,35 @@ void expressionContainingVirtualTable() {
         explain(relNode));
   }
 
+  @Test
+  void nullableFieldRoundTrip() {
+    NamedStruct schema = NamedStruct.of(List.of("col1", "col2"), R.struct(N.I32, R.FP64));
+    Expression nullableI32 = Expression.I32Literal.builder().value(6).nullable(true).build();
```
benbellick: you can even do it for the non-null ones!
```suggestion
    Expression nullableI32 = ExpressionCreator.i32(true, 6);
```

isthmus/src/main/java/io/substrait/isthmus/expression/LiteralConverter.java
asolimando: Thanks for sharing, I gave it a bit more thinking and I think it's fine and the thrown error will be clear enough, no need to be defensive here nor adjust the error message (which is what you are suggesting, IIUC)

isthmus/src/test/java/io/substrait/isthmus/VirtualTableScanTest.java
asolimando: This is actually pretty cool, it was a bit too verbose before, thanks for the tip!

PR #691, opened 2026-01-23
feat(core): expose metadata from YAML extension files

### LINKED ISSUE #690: Handle metadata in YAML extension files
https://github.com/substrait-io/substrait/pull/922 introduced support for metadata inside of YAML extension files. We should expose the metadata in these file during runtime so that users can extend substrait via metadata.

benbellick:
Closes #690


core/src/main/java/io/substrait/extension/SimpleExtension.java
```diff
@@ -276,6 +278,8 @@ public String description() {
 
     public abstract Map<String, Option> options();
 
+    public abstract Optional<Map<String, Object>> metadata();
```
nielspardon, line 281: I'm wondering whether it would be useful to have an interface instead of `Object` for e.g. creating custom YAML deserializers / serializers and whether we should allow to add them to the `ObjectMapper` in `SimpleExtension.load()`.

Otherwise I'm fine with the changes.

benbellick: That is a useful idea, but I wonder if we can leave that for future work? I think this simple approach works for now, and it satisfies the goal of being able to expose arbitrary data to users from YAML files.

nielspardon: Well, if we switch to an interface instead of `Object` at a later point this would be a breaking change for consumers which we probably want to avoid.

nielspardon: I can see that using `Object` is simpler for default types like `String | Integer | Boolean`

PR #695, opened 2026-01-29
feat(isthmus): add strptime_* function mappings

nielspardon:
This adds Calcite mappings for the following 3 Substrait functions:

- `strptime_time` -> `PARSE_TIME`
- `strptime_timestamp` -> `PARSE_TIMESTAMP`
- `strptime_date` -> `PARSE_DATE`

I added the `ScalarFunctionConverter` into the `RexExpressionConverter` instance in `FunctionConversionTest` in order to round-trip test Substrait <-> Calcite expression mappings which also made the `subtractDateIDay` test case work.

The Calcite functions are Big Query specific variants which happen to be defined closest to the Substrait variants of this functions vs. other alternatives `TO_DATE` and `TO_TIMESTAMP`.

### DIFF
--- a/isthmus/src/test/java/io/substrait/isthmus/FunctionConversionTest.java
+++ b/isthmus/src/test/java/io/substrait/isthmus/FunctionConversionTest.java
@@ -301,8 +301,7 @@ class FunctionConversionTest extends PlanTestBase {
     assertEquals(SqlKind.OTHER_FUNCTION, calciteExpr.getKind());
     assertInstanceOf(RexCall.class, calciteExpr);
 
-    RexCall extract = (RexCall) calciteExpr;
-    assertEquals("PARSE_TIME('%H:%M:%S':VARCHAR, '12:34:56':VARCHAR)", extract.toString());
+    assertEquals("PARSE_TIME('%H:%M:%S':VARCHAR, '12:34:56':VARCHAR)", calciteExpr.toString());
 
     // tests the reverse Calcite -> Substrait
     Expression reverse = calciteExpr.accept(rexExpressionConverter);
@@ -329,10 +328,9 @@ class FunctionConversionTest extends PlanTestBase {
     assertEquals(SqlKind.OTHER_FUNCTION, calciteExpr.getKind());
     assertInstanceOf(RexCall.class, calciteExpr);
 
-    RexCall extract = (RexCall) calciteExpr;
     assertEquals(
         "PARSE_TIMESTAMP('%Y:%m:%dT%H:%M:%S':VARCHAR, '2026-01-29T12:34:56':VARCHAR)",
-        extract.toString());
+        calciteExpr.toString());
 
     // tests the reverse Calcite -> Substrait
     Expression reverse = calciteExpr.accept(rexExpressionConverter);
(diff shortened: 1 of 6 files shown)


isthmus/src/test/java/io/substrait/isthmus/FunctionConversionTest.java
```diff
@@ -278,4 +283,84 @@ void concatCharAndVarchar() throws Exception {
...
+            timeString,
+            formatString);
+
+    // tests Substrait -> Calcite
+    RexNode calciteExpr = strptimeTimeFn.accept(expressionRexConverter, Context.newContext());
+    assertEquals(SqlKind.OTHER_FUNCTION, calciteExpr.getKind());
+    assertInstanceOf(RexCall.class, calciteExpr);
+
+    RexCall extract = (RexCall) calciteExpr;
+    assertEquals("PARSE_TIME('%H:%M:%S':VARCHAR, '12:34:56':VARCHAR)", extract.toString());
```
bestbeforetoday: I don't think the cast is necessary to allow `toString` to be called.

```suggestion
    assertEquals("PARSE_TIME('%H:%M:%S':VARCHAR, '12:34:56':VARCHAR)", calciteExpr.toString());
```

isthmus/src/test/java/io/substrait/isthmus/FunctionConversionTest.java
```diff
@@ -278,4 +283,84 @@ void concatCharAndVarchar() throws Exception {
...
+            dateString,
+            formatString);
+
+    // tests Substrait -> Calcite
+    RexNode calciteExpr = strptimeDateFn.accept(expressionRexConverter, Context.newContext());
+    assertEquals(SqlKind.OTHER_FUNCTION, calciteExpr.getKind());
+    assertInstanceOf(RexCall.class, calciteExpr);
+
+    RexCall extract = (RexCall) calciteExpr;
+    assertEquals("PARSE_DATE('%Y:%m:%d':VARCHAR, '2026-01-29':VARCHAR)", extract.toString());
```
bestbeforetoday: I don't think the cast is necessary to allow `toString` to be called.

```suggestion
    assertEquals("PARSE_DATE('%Y:%m:%d':VARCHAR, '2026-01-29':VARCHAR)", calciteExpr.toString());
```

isthmus/src/test/java/io/substrait/isthmus/FunctionConversionTest.java
```diff
@@ -278,4 +283,84 @@ void concatCharAndVarchar() throws Exception {
...
+
+    // tests Substrait -> Calcite
+    RexNode calciteExpr = strptimeTimestampFn.accept(expressionRexConverter, Context.newContext());
+    assertEquals(SqlKind.OTHER_FUNCTION, calciteExpr.getKind());
+    assertInstanceOf(RexCall.class, calciteExpr);
+
+    RexCall extract = (RexCall) calciteExpr;
+    assertEquals(
+        "PARSE_TIMESTAMP('%Y:%m:%dT%H:%M:%S':VARCHAR, '2026-01-29T12:34:56':VARCHAR)",
+        extract.toString());
```
bestbeforetoday: I don't think the cast is necessary to allow `toString` to be called.

```suggestion
    assertEquals(
        "PARSE_TIMESTAMP('%Y:%m:%dT%H:%M:%S':VARCHAR, '2026-01-29T12:34:56':VARCHAR)",
        calciteExpr.toString());
```

bestbeforetoday (changes requested):
There is quite a bit of duplication between the three Strptime function mapping and test implementations that might benefit from being factored out, perhaps into a StrptimeMapper superclass and using a parameterized test. It is fine as-is if you prefer to defer any refactoring to a future PR though.

Just a few unnecessary casts noted with inline comments that are worth removing to avoid redundant code.

isthmus/src/test/java/io/substrait/isthmus/FunctionConversionTest.java
nielspardon: makes sense. this is a copy & paste artifact I didn't question :-)

bestbeforetoday (approved):

PR #700, opened 2026-02-04
feat(isthmus): add PostgreSQL TPC-H integration testing

nielspardon:
This PR adds integration testing for isthmus against a PostgreSQL database.

- it adds TPC-H data generated with scale factor 0.01 aka tiny `*.tbl` files
- it uses [testcontainers](https://java.testcontainers.org/modules/databases/postgres/) to automatically spin up a PostgreSQL container during testing
- populates the PostgreSQL DB by running the DDL statements and loading the data from the `*.tbl` files
- runs a reference SQL and a SQL generated from the Substrait plan against the PostgreSQL database and checks whether both SQLs produce the same records
- I had to create copies of the DDL statements and the TPC-H queries for PostgreSQL since there are smaller syntax differences vs the copies we currently have in the `queries` subfolder e.g. 
  - PostgreSQL expects `interval '1 day'` instead of `interval '1' day(3)`.
  - identifiers in Substrait plans are essentially handled as case-sensitive so I had to uppercase and quote identifiers in the reference SQLs and the DDL statements
- the good news all but TPC-H query 21 are producing the same results based on the testing method I'm using. just need to figure out what's going wrong with this query

This is currently a draft since I want to get some early feedback. What I'm still planning to do:

- add javadocs for the new code
- Q: should we always run the integration tests or should we run them in a separate Gradle task?
- I also want to enable this for the TPC-DS in a follow-up PR and then see whether we can extend this more databases supported by testcontainers


bestbeforetoday:
I have recently stumbled on a few cases where, despite passing our unit tests that round-trip SQL/Substrait, specific TPC-H queries originating from Substrait plans do not execute correctly on a target database. Either the SQL is not valid for the database (I happened to be using DuckDB locally) or — more worryingly — the SQL executes but produces different results from the reference SQL. I think this is something we should be checking (and catching) here in the substrait-java repository. Some variant of what is posted in this PR seems like a good approach to me and I would like to move this forward.

What I am imagining is:

1. Converting TPC-H queries to Substrait plans (using Isthmus).
2. Converting the generated Substrait plans to Calcite.
3. Executing the Substrait-derived Calcite plan against the test data (either directly using Calcite or by converting to SQL).
4. Comparing the Substrait-derived results to the results of executing the TPC-H reference queries directly.

This is similar to some of the testing that is done in the Spark subproject. This testing should demonstrate the failure cases I mentioned above and provide confidence that fixes resolve them.

@benbellick We can use a tool like [tpchgen](https://github.com/clflushopt/tpchgen-rs) to efficiently generate TPC-H datasets on demand and avoid checking in lots of test data. Would this satisfy your concern?

benbellick:
@bestbeforetoday I think that sounds like a great plan moving forwards. I do see the need for introducing some tests like this, and I appreciated you investigating ways to keep the diff smaller. I won't have time to properly review today but hopefully I will this week. Thanks!

bestbeforetoday:
Thanks for the feedback @benbellick. It will take me a little time to evolve to a state where a review is required so no hurry there.

nielspardon:
replaced by #855

PR #701, opened 2026-02-10
refactor(isthmus): strptime function mapping

bestbeforetoday:
Avoid code duplication. Also test operator name and operands explicitly instead of relying on Calcite's toString implementation, which might change over time.

### DIFF
--- a/isthmus/src/test/java/io/substrait/isthmus/FunctionConversionTest.java
+++ b/isthmus/src/test/java/io/substrait/isthmus/FunctionConversionTest.java
@@ -17,10 +17,8 @@ import io.substrait.isthmus.expression.ScalarFunctionConverter;
 import io.substrait.isthmus.expression.WindowFunctionConverter;
 import io.substrait.type.Type;
 import io.substrait.type.TypeCreator;
-import java.util.List;
 import java.util.stream.Stream;
 import org.apache.calcite.rex.RexCall;
-import org.apache.calcite.rex.RexLiteral;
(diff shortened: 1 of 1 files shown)


isthmus/src/main/java/io/substrait/isthmus/expression/AbstractStrptimeFunctionMapper.java
```diff
@@ -0,0 +1,71 @@
...
+  private final List<ScalarFunctionVariant> strptimeFunctions;
+
+  /**
+   * Constructs an abstract strptime function mapper.
+   *
+   * @param substraitFunctionName the name of the Substrait function (e.g., "strptime_date")
+   * @param calciteOperator the Calcite operator to map from (e.g., SqlLibraryOperators.PARSE_DATE)
+   * @param functions the list of all available scalar function variants
+   */
+  protected AbstractStrptimeFunctionMapper(
```
nielspardon, line 35: if you want to have even more reuse then `PositionFunctionMapper` also just swaps the position of the first two arguments of its function call

bestbeforetoday: This refactor consolidates a family of closely related date/time parsing functions that require the same argument mapping and differ only by the type they return. POSITION deals with string searching so, while it also requires the same argument mapping, the relationship is not obvious and could be confusing. I would prefer to keep this PR focused on removing duplication in the closely related date/time parsing functions.

PR #706, opened 2026-02-17
build(isthmus-cli): fix javadoc warnings

bestbeforetoday:
Change the isthmus-cli build to fail on warnings if they are introduced in the future.


nielspardon (approved):
LGTM, added the other active committers as reviewers for FYI

Issue #687, opened 2026-01-22
Support lambdas


benbellick:
https://github.com/substrait-io/substrait/pull/889 introduced lambda support in substrait. Let's introduce lambdas into `substrait-java`.

It is broken down below into two distinct issues so that they can be merged in two independent PRs to keep the diff small.

PR #710, opened 2026-02-24
feat: add lambda expression support

limameml:
This PR is to add support for the Lambda Expression

## Summary
  - Add lambda expression support to substrait-java core and isthmus
  - Implement Type.Func for function types
  - Add Expression.Lambda and lambda parameter references
  - Full proto round-trip support
  - Calcite (isthmus) integration (single-level lambdas only)

  ## Test plan
  - Added LambdaExpressionRoundtripTest in core
  - Added LambdaExpressionTest in isthmus

closes https://github.com/substrait-io/substrait-java/issues/687

### DIFF
--- a/core/src/test/resources/expressions/lambda/invalid/param_index.json
+++ b/core/src/test/resources/expressions/lambda/invalid/param_index.json
@@ -1,0 +1,21 @@
+{
+  "lambda": {
+    "parameters": {
+      "types": [
+        { "i32": { "nullability": "NULLABILITY_REQUIRED" } }
+      ]
+    },
+    "body": {
+      "selection": {
+        "directReference": {
+          "structField": {
+            "field": 5
+          }
+        },
+        "lambdaParameterReference": {
+          "stepsOut": 0
+        }
+      }
+    }
+  }
+}


core/src/main/java/io/substrait/expression/Expression.java
```diff
@@ -707,6 +707,31 @@ public <R, C extends VisitationContext, E extends Throwable> R accept(
...
+    public abstract Type.Struct parameters();
+
+    public abstract Expression body();
+
+    @Override
+    public Type getType() {
+      List<Type> paramTypes = parameters().fields();
+      Type returnType = body().getType();
+
+      return Type.withNullability(false).func(paramTypes, returnType);
```
benbellick, line 775: Is setting nullability false correct here? I am pretty sure `func` can be nullable, e.g. `func?<i32 -> i32>`. 

Also, probably will be good to make a test captures this then.

core/src/main/java/io/substrait/expression/Expression.java
```diff
@@ -902,6 +927,28 @@ public <R, C extends VisitationContext, E extends Throwable> R accept(
     }
   }
 
+  @Value.Immutable
```
benbellick: This is okay if you feel confident in your `LambdaInvocation` work, but considering the size of this PR, it will make it easier both on you and the reviewer to separate out the `Lambda` work and `LambdaInvocation` work into two separate PRs.

core/src/main/java/io/substrait/expression/proto/ProtoExpressionConverter.java
```diff
@@ -37,6 +37,7 @@ public class ProtoExpressionConverter {
   private final Type.Struct rootType;
   private final ProtoTypeConverter protoTypeConverter;
   private final ProtoRelConverter protoRelConverter;
+  private final List<Type.Struct> lambdaParameterStack = new ArrayList<>();
```
benbellick: Can we use a stack here?

core/src/main/java/io/substrait/expression/proto/ProtoExpressionConverter.java
```diff
@@ -260,6 +281,44 @@ public Type visit(Type.Struct type) throws RuntimeException {
...
+        {
+          io.substrait.proto.Expression.Lambda protoLambda = expr.getLambda();
+          Type.Struct parameters =
+              (Type.Struct)
+                  protoTypeConverter.from(
+                      io.substrait.proto.Type.newBuilder()
+                          .setStruct(protoLambda.getParameters())
+                          .build());
+
+          lambdaParameterStack.add(parameters);
```
benbellick: then we can use the standard stack `push` here

core/src/main/java/io/substrait/expression/proto/ProtoExpressionConverter.java
```diff
@@ -260,6 +281,44 @@ public Type visit(Type.Struct type) throws RuntimeException {
...
+
+          Expression body;
+          try {
+            body = from(protoLambda.getBody());
+          } finally {
+            lambdaParameterStack.remove(lambdaParameterStack.size() - 1);
```
benbellick: and the standard stack `pop` here

core/src/main/java/io/substrait/expression/Expression.java
limameml: I followed the substrait-go implementation: the GetType function for lambda returns Nullability:    types.NullabilityRequired
```
func (l *Lambda) GetType() types.Type {
	return &types.FuncType{
		Nullability:    types.NullabilityRequired,
		ParameterTypes: l.Parameters.Types,
		ReturnType:     l.Body.GetType(),
	}
}
```

core/src/main/java/io/substrait/expression/Expression.java
limameml: I'll handle `LambdaInvocation` in a separate pr

core/src/main/java/io/substrait/expression/proto/ProtoExpressionConverter.java
limameml: I need to access `lambdaIndex = lambdaParameterStack.size() - 1 - stepsOut` here:` lambdaParameters = lambdaParameterStack.get(lambdaIndex);` when building the Lambda parametre ref that's why I didn't use a stack

benbellick: That's fair. Java's stack does extend `Vector` though so you still get the `get` method. I'm not so particular on this one. 

The only thing I will say is I do think that the whole stack parameter stuff can be confusing for people without a little bit of PL experience, which is why I pushed @slimsammylim to encapsulate some of that logic and put it in docstring comments. I could see a case for encapsulating this logic in a local private class, though it probably makes more sense to do this only if this logic comes up again.

I expected to see something like this logic elsewhere, as you need to do it to do validation when building lambdas. Though I didn't find anything like this. Does the builder for lambdas in this PR actually validate that the lambda is semantically correct?

core/src/main/java/io/substrait/expression/Expression.java
benbellick: hmm I am thinking about this right now but that may have been wrong in `substrait-go` 🤔

benbellick: I think this is a spec error. I raised an issue in upstream [here](https://github.com/substrait-io/substrait/issues/976). 

Mind leaving a comment there referencing this issue as a TODO so we remember to fix it once it gets resolved?

core/src/main/java/io/substrait/expression/proto/ProtoExpressionConverter.java
bestbeforetoday: I agree that it would be nice for clarity to encapsulate this quite specific behaviour in a class ([single responsibility principle](https://en.wikipedia.org/wiki/SOLID)). It doesn't need to extend one of the full collection interfaces, just provide the methods you need:
- `void push(Type.Struct value)`
- `Type.Struct pop()` (throw runtime NoSuchElementException if empty)
- `Type.Struct get(int stepsOut)` (throw runtime IndexOutOfBoundsException on invalid stepsOut)

Or make it generic if you will reuse it for other types. The caller doesn't need to worry about clever indexing logic or what collection implementation is used internally; just make simple method calls that clearly describe the intent.

core/src/main/java/io/substrait/expression/proto/ProtoExpressionConverter.java
```diff
@@ -75,6 +76,26 @@ public FieldReference from(io.substrait.proto.Expression.FieldReference referenc
...
+              reference.getLambdaParameterReference();
+
+          int stepsOut = lambdaParamRef.getStepsOut();
+          int lambdaIndex = lambdaParameterStack.size() - 1 - stepsOut;
+          if (lambdaIndex < 0 || lambdaIndex >= lambdaParameterStack.size()) {
+            throw new IllegalArgumentException(
+                String.format(
+                    "Lambda parameter reference with stepsOut=%d is invalid (current depth: %d)",
+                    stepsOut, lambdaParameterStack.size()));
+          }
```
bestbeforetoday: This logic could sit within a simple class to encapsulate the lambda parameter stack behaviour (as described in other comments) instead of adding extra complexity to this method.

core/src/main/java/io/substrait/expression/proto/ProtoExpressionConverter.java
```diff
@@ -75,6 +76,26 @@ public FieldReference from(io.substrait.proto.Expression.FieldReference referenc
...
+
+          Type.Struct lambdaParameters = lambdaParameterStack.get(lambdaIndex);
+          return FieldReference.newLambdaParameterReference(
+              reference.getDirectReference().getStructField().getField(),
```
Slimsammylim: Something I'm curious about is why we use `reference.getDirectReference().getStructField().getField()` here instead of `getDirectReferenceSegments(reference.getDirectReference())` like the `ROOT_REFERENCE` case does. 

I'm unfamiliar with substrait-java, but does this current implementation support nested field access through lambda parameters? 

*This may be an unimportant issue especially if it's not common to access nested fields in lambdas

limameml: because the `ofRoot` method takes the struct itself while the `newLambdaParameterReference` takes the index, like the `newRootStructOuterReference` methode

core/src/main/java/io/substrait/expression/proto/ProtoExpressionConverter.java
limameml: Hey @benbellick, for your last point I don't quite understand what validation I should add, doesn't the validation happen when building the LAMBDA_PARAMETER_REFERENCE?

core/src/main/java/io/substrait/expression/proto/ProtoExpressionConverter.java
```diff
@@ -574,4 +609,30 @@ public Expression.SortField fromSortField(SortField s) {
   public static FunctionOption fromFunctionOption(io.substrait.proto.FunctionOption o) {
     return FunctionOption.builder().name(o.getName()).addAllValues(o.getPreferenceList()).build();
   }
+
+  private static class LambdaParameterStack {
```
benbellick: I think this is much better. Thanks for encapsulating this into its own class. 

Would you mind also providing some simple docstring-style method comments to explain the logic here? So that that future readers of the code can understand what is going on at a first glance.

core/src/main/java/io/substrait/expression/proto/ProtoExpressionConverter.java
Slimsammylim: Ah, I see. I am still confused as to whether this supports nested field access through lambda parameters. I messed around w some tests through claude, and I believe it is unsupported. However, this is a nit and I'm not sure it will come up often in practice, so we can come back to this later if it ends up being an issue!

vbarua:
I made a PR into this PR as part of understanding how it works, and fixing a small gap when loading Substrait plans with lambdas: https://github.com/limameml/substrait-java/pull/1

core/src/main/java/io/substrait/expression/Expression.java
```diff
@@ -758,6 +758,33 @@ public <R, C extends VisitationContext, E extends Throwable> R accept(
...
+
+    public abstract Expression body();
+
+    @Override
+    public Type getType() {
+      List<Type> paramTypes = parameters().fields();
+      Type returnType = body().getType();
+
+      // TO DO: fix Lambda return type once this issue
+      // https://github.com/substrait-io/substrait/issues/976 is resolved
```
vbarua:
```suggestion
      // TODO: Type.Func nullability is hardcoded to false here because the spec does not allow for
      //   declaring otherwise.
      // See: https://github.com/substrait-io/substrait/issues/976
```

Being a bit more precise about what the issue is, and what the implications are in the code.

core/src/main/java/io/substrait/relation/VirtualTableScan.java
```diff
@@ -242,5 +242,11 @@ public Integer visit(Type.Map type) throws RuntimeException {
     public Integer visit(Type.UserDefined type) throws RuntimeException {
       return 0;
     }
+
+    @Override
+    public Integer visit(Type.Func type) throws RuntimeException {
+      return type.parameterTypes().stream().mapToInt(p -> p.accept(this)).sum()
+          + type.returnType().accept(this);
```
vbarua: This visitor is used to determine how many _names_ are needed for a ReadRel based on the schema.

For example, a table with row type ```struct<i8, i16, i32>``` needs 3 names. A table like ```struct<i8, struct<i16, i32>, i64>``` needs 5 names, because in Substrait we name every struct field. For full details see https://substrait.io/types/named_structs/#determining-names

Now, given something like ```struct<i8, func<...>, i32>``` we don't need to look inside the func for names. The type of the column is just `func`, and none of the parameters to the `func` can be reference by name.

All of this is to say that this can just be 0.

```suggestion
      return 0;
```

examples/substrait-spark/src/main/java/io/substrait/examples/util/ExpressionStringify.java
```diff
@@ -202,6 +202,12 @@ public String visit(Expression.NestedStruct expr, EmptyVisitationContext context
     return "<NestedStruct >";
   }
 
+  @Override
+  public String visit(Expression.Lambda expr, EmptyVisitationContext context)
+      throws RuntimeException {
+    return "<Lambda>";
```
vbarua: minor: extract ` ` for consistency with the other methods
```suggestion
    return "<Lambda >";
```

core/src/main/java/io/substrait/expression/proto/ProtoExpressionConverter.java
vbarua: > I am still confused as to whether this supports nested field access through lambda parameters.

If by nested access you mean accessing nested fields of a lambda parameter inside the body, for example
```($0) -> $0.a.[0].hello``` then I think that is broken here actually. That might be fine for now and we can treat it is a follow-up item, but we should detect if we have a nested field lookup happening by checking to to content of `reference.getDirectReference().getStructField().getChild();` If it's empty, there is only a direct reference to the lambda param. If it's not empty we can blow up.

To @Slimsammylim's we point, to handle this correctly we would need to use `getDirectReferenceSegments(reference.getDirectReference())` to create the list of reference segments and then re-use some of the other constructors.

Comment (limameml) [reply core/src/main/java/io/substrait/expression/proto/ProtoExpressionConverter.java]: ok I will add a check and go back to it in a follow up pr.

File: core/src/test/java/io/substrait/type/proto/LambdaExpressionRoundtripTest.java [hunk 12]
```diff
@@ -0,0 +1,361 @@
...
+import java.util.List;
+import org.junit.jupiter.api.Test;
+
+/**
+ * Tests for Lambda expression round-trip conversion through protobuf. Based on equivalent tests
+ * from substrait-go.
+ */
+class LambdaExpressionRoundtripTest extends TestBase {
+
+  /** Test that lambdas with no parameters are valid. Building: () -> i32(42) : func<() -> i32> */
```
benbellick: nit: The parenthesis make it look like a cast to me 🤷 
```suggestion
  /** Test that lambdas with no parameters are valid. Building: () -> 42 : func<() -> i32> */
```

core/src/test/java/io/substrait/type/proto/LambdaExpressionRoundtripTest.java
```diff
@@ -0,0 +1,361 @@
...
+   */
+  @Test
+  void invalidOuterRef_stepsOutTooHigh() {
+    Type.Struct params = Type.Struct.builder().nullable(false).addFields(R.I32).build();
+
+    // Create a parameter reference with stepsOut=1 but no outer lambda exists
+    FieldReference invalidRef = FieldReference.newLambdaParameterReference(0, params, 1);
+
+    Expression.Lambda lambda =
+        Expression.Lambda.builder().parameters(params).body(invalidRef).build();
```
benbellick: Shouldn't this throw an exception? Perhaps I am misunderstanding what the builder does, but don't we expect the builder to validate that we built a correct lambda? Why are we able to build it, and convert it to a proto, but not convert it back?

core/src/test/java/io/substrait/type/proto/LambdaExpressionRoundtripTest.java
```diff
@@ -0,0 +1,361 @@
...
+    FieldReference invalidRef = FieldReference.newLambdaParameterReference(0, outerParams, 2);
+
+    Expression.Lambda innerLambda =
+        Expression.Lambda.builder().parameters(innerParams).body(invalidRef).build();
+
+    Expression.Lambda outerLambda =
+        Expression.Lambda.builder().parameters(outerParams).body(innerLambda).build();
+
+    // Convert to proto
+    io.substrait.proto.Expression protoExpression = expressionProtoConverter.toProto(outerLambda);
```
benbellick: Same issue here as above I think.

core/src/test/java/io/substrait/type/proto/LambdaExpressionRoundtripTest.java
benbellick: IMO this issue will need to be resolved before merging. Sorry to be a blocker! It doesn't make sense to me that we would be able to make an invalid proto and only validate on the way back, unless I am misunderstanding something.

benbellick: https://github.com/substrait-io/substrait-go/blob/48bbdb43eeccc61c7adfecc3549a1b004fd52fd8/expr/lambda_test.go#L148-L165

^ You can see an example here of how in `substrait-go` the builder throws an error

vbarua: Good point Ben. It would make sense for these to fail, especially given that we're already claiming them as validation errors that should fail.

core/src/main/java/io/substrait/expression/FieldReference.java
```diff
@@ -38,13 +40,18 @@ public <R, C extends VisitationContext, E extends Throwable> R accept(
   public boolean isSimpleRootReference() {
     return segments().size() == 1
         && !inputExpression().isPresent()
-        && !outerReferenceStepsOut().isPresent();
+        && !outerReferenceStepsOut().isPresent()
+        && !lambdaParameterReferenceStepsOut().isPresent();
```
benbellick, line 52: Thinking out loud here, haven't checked yet, but do we actually validate that only one of `outerReferenceStepsOut` or `lambdaParamterReferenceStepsOut` is set?

core/src/main/java/io/substrait/expression/FieldReference.java
```diff
@@ -20,6 +20,8 @@ public abstract class FieldReference implements Expression {
 
   public abstract Optional<Integer> outerReferenceStepsOut();
 
+  public abstract Optional<Integer> lambdaParameterReferenceStepsOut();
```
benbellick, line 23: Also doesn't have to be for this PR, but I wonder if in the future, it makes sense to separate out outer references and lambda references to both extend the base class `FieldReference`.

core/src/main/java/io/substrait/expression/LambdaBuilder.java
```diff
@@ -0,0 +1,150 @@
...
+    }
+
+    /**
+     * Creates a validated reference to a parameter of this lambda.
+     *
+     * @param paramIndex index of the parameter within this lambda's parameter struct
+     * @return a {@link FieldReference} pointing to the specified parameter
+     * @throws IndexOutOfBoundsException if paramIndex is out of bounds
+     */
+    public FieldReference ref(int paramIndex) {
```
benbellick:
```suggestion
    public FieldReference ref(int paramIndex) throws IndexOutOfBoundsException {
```
or whatever the syntax is

core/src/main/java/io/substrait/expression/FieldReference.java
```diff
@@ -134,6 +141,15 @@ public static FieldReference newInputRelReference(int index, List<Rel> rels) {
             index, currentOffset));
   }
 
+  public static FieldReference newLambdaParameterReference(
+      int paramIndex, Type.Struct lambdaParamsType, int stepsOut) {
```
vbarua: minor: could we switch the parameter ordering here to be

```
int stepsOut, int paramIndex, Type.Struct lambdaParamsType
```

Because the value of the lambdaParamsType can be a long invocation it's more readable if we put it at the end.

I was looking at the usage in some of the lambda builder tests, and IMO
```java
                FieldReference.newLambdaParameterReference(
                    0, Type.Struct.builder().nullable(false).addFields(R.I32).build(), 1))
```

is harder to read than
```java
                FieldReference.newLambdaParameterReference(
                    1, 0, Type.Struct.builder().nullable(false).addFields(R.I32).build()))
```

`stepsOut` is first because I think about creating the reference as "step out x levels and grab the yth field from the type z"

vbarua (approved):
Thanks for you additions @benbellick, they make sense to me.

I did identify one minor minor API improvement based on tests you added, but otherwise this is looking good to me.

core/src/main/java/io/substrait/type/StringTypeVisitor.java
```diff
@@ -150,4 +150,13 @@ public String visit(Type.Map type) throws RuntimeException {
   public String visit(Type.UserDefined type) throws RuntimeException {
     return String.format("u!%s%s", type.name(), n(type));
   }
+
+  @Override
+  public String visit(Type.Func type) throws RuntimeException {
+    return String.format(
+        "func%s<%s -> %s>",
+        n(type),
+        type.parameterTypes().stream().map(t -> t.accept(this)).collect(Collectors.joining(", ")),
```
benbellick, line 159: I think that for one param, e.g. `T`, the short name is `func<T -> U>`. But for multiple, it is `func<(T1,T2) -> U>`. This looks like it will just do `func<T1,T2->U>`

benbellick: hmm, I think that I am wrong actually 🤷 I see tests down below that have `(T1,T2)` in them.

core/src/main/java/io/substrait/type/parser/ParseToPojo.java
```diff
@@ -411,19 +412,61 @@ public TypeExpression visitNStruct(final SubstraitTypeParser.NStructContext ctx)
 
     @Override
     public TypeExpression visitFunc(final SubstraitTypeParser.FuncContext ctx) {
-      throw new UnsupportedOperationException();
```
benbellick, line 414: This part was a bit tough for me to follow, but it seems consistent with the above implementations and I think this complexity is a bit inherent because of the grammar, so I think this is fine 👍

core/src/test/java/io/substrait/expression/LambdaBuilderTest.java
```diff
@@ -0,0 +1,107 @@
...
+  @Test
+  void invalidFieldIndex_outOfBounds() {
+    assertThrows(
+        IndexOutOfBoundsException.class, () -> lb.lambda(List.of(R.I32), params -> params.ref(5)));
+  }
+
+  // (x: i32)@p -> p[-1] — negative index
+  @Test
+  void negativeFieldIndex() {
+    assertThrows(Exception.class, () -> lb.lambda(List.of(R.I32), params -> params.ref(-1)));
```
benbellick: hmm wonder if we can check something better than Exception.class here?

core/src/main/java/io/substrait/expression/FieldReference.java
```diff
@@ -38,13 +40,18 @@ public <R, C extends VisitationContext, E extends Throwable> R accept(
...
-        && !outerReferenceStepsOut().isPresent();
+        && !outerReferenceStepsOut().isPresent()
+        && !lambdaParameterReferenceStepsOut().isPresent();
   }
 
   public boolean isOuterReference() {
     return outerReferenceStepsOut().orElse(0) > 0;
   }
 
+  public boolean isLambdaParameterReference() {
```
benbellick, line 60: nit: lets give this a docstring

isthmus/src/main/java/io/substrait/isthmus/expression/FunctionMappings.java
benbellick, line 1: I'm fairly certain that SQL -> Substrait conversion from `isthmus` doesn't work yet. 

I tried adding the following test and it failed: 
```java
package io.substrait.isthmus;

class ListFunctionTest extends PlanTestBase {

  @Test
  void transformIdentity() throws Exception {
    assertFullRoundTrip(
        "SELECT transform(col, x -> x) FROM (VALUES (ARRAY[1, 2, 3])) AS t(col)");
  }
}
```

IMO, this is no problem. This PR is already quite large, so I'd rather leave that out. However, can we open up a ticket to track that this work still needs to be done just so we don't forget?

core/src/main/java/io/substrait/expression/FieldReference.java
benbellick: To be explicit, yeah don't do this is this PR, it is fine as is :)

core/src/test/java/io/substrait/expression/LambdaBuilderTest.java
limameml: let's go with an IndexOutOfBoundsException

core/src/main/java/io/substrait/expression/FieldReference.java
limameml: There is no check I will add it

isthmus/src/main/java/io/substrait/isthmus/expression/FunctionMappings.java
limameml: I see the problem lies in the definition of the list functions in calcite, since the list functions don't have a built in definition the way I defined the works for substrait to sql translation but they need a more complexe definition I think it will be something like what @vbarua did with AggregateFunctions.java

limameml: I'll open a ticket for that

limameml: https://github.com/substrait-io/substrait-java/issues/757

core/src/main/java/io/substrait/expression/FieldReference.java
benbellick: @limameml I think this one might still need to be addressed?

benbellick: I went ahead and took care of it myself, so no worries!

PR #721, opened 2026-03-03
feat(isthmus): migrate to PrecisionTime / PrecisionTimestamp

nielspardon:
This PR migrates isthmus to use PrecisionTime / PrecisionTimestamp instead of the deprecated Time / Timestamp types.

### DIFF
--- a/core/src/main/java/io/substrait/expression/ExpressionCreator.java
+++ b/core/src/main/java/io/substrait/expression/ExpressionCreator.java
@@ -10,6 +10,7 @@ import java.math.BigDecimal;
 import java.nio.ByteBuffer;
 import java.time.Instant;
 import java.time.LocalDateTime;
+import java.time.LocalTime;
 import java.time.ZoneOffset;
 import java.util.Arrays;
 import java.util.List;
@@ -90,6 +91,23 @@ public class ExpressionCreator {
   }
 
   /**
+   * Creates a precision time literal from a LocalTime value.
+   *
+   * <p>This method converts a Java {@link LocalTime} to a Substrait precision time literal with
+   * nanosecond precision (precision = 9). The time value is represented as nanoseconds since
+   * midnight (00:00:00).
+   *
+   * @param nullable whether the literal can be null
+   * @param value the LocalTime value to convert
+   * @return a PrecisionTimeLiteral with nanosecond precision representing the given time
+   * @see #precisionTime(boolean, long, int) for creating precision time with custom precision
+   */
+  public static Expression.PrecisionTimeLiteral precisionTime(boolean nullable, LocalTime value) {
+    long epochNano = value.toNanoOfDay();
+    return precisionTime(nullable, epochNano, 9);
+  }
+
+  /**
    * @deprecated Timestamp is deprecated in favor of PrecisionTimestamp
    */
   @Deprecated
@@ -153,6 +171,27 @@ public class ExpressionCreator {
         .value(value)
         .precision(precision)
         .build();
+  }
+
+  /**
+   * Creates a precision timestamp literal from a LocalDateTime value.
+   *
+   * <p>This method converts a Java {@link LocalDateTime} to a Substrait precision timestamp literal
+   * with nanosecond precision (precision = 9). The timestamp value is represented as nanoseconds
+   * since the Unix epoch (1970-01-01 00:00:00 UTC), assuming the LocalDateTime is in UTC timezone.
+   *
+   * @param nullable whether the literal can be null
+   * @param value the LocalDateTime value to convert (interpreted as UTC)
+   * @return a PrecisionTimestampLiteral with nanosecond precision representing the given timestamp
+   * @see #precisionTimestamp(boolean, long, int) for creating precision timestamp with custom
+   *     precision
+   */
+  public static Expression.PrecisionTimestampLiteral precisionTimestamp(
+      boolean nullable, LocalDateTime value) {
+    long epochNano =
+        TimeUnit.SECONDS.toNanos(value.toEpochSecond(ZoneOffset.UTC))
+            + value.toLocalTime().getNano();
+    return precisionTimestamp(nullable, epochNano, 9);
   }
 
   public static Expression.PrecisionTimestampTZLiteral precisionTimestampTZ(
--- a/core/src/main/java/io/substrait/expression/ExpressionCreator.java
+++ b/core/src/main/java/io/substrait/expression/ExpressionCreator.java
@@ -122,7 +122,7 @@ public class ExpressionCreator {
   public static Expression.TimestampLiteral timestamp(boolean nullable, LocalDateTime value) {
     long epochMicro =
         TimeUnit.SECONDS.toMicros(value.toEpochSecond(ZoneOffset.UTC))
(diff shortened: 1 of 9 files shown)


isthmus/src/main/java/io/substrait/isthmus/expression/ExpressionRexConverter.java
```diff
@@ -212,9 +213,16 @@ public RexNode visit(Expression.TimeLiteral expr, Context context) throws Runtim
 
   @Override
   public RexNode visit(PrecisionTimeLiteral expr, Context context) throws RuntimeException {
-    return rexBuilder.makeLiteral(
-        createTimeString(expr.value(), expr.precision()),
-        typeConverter.toCalcite(typeFactory, expr.getType()));
+    int maxPrecision = typeFactory.getTypeSystem().getMaxPrecision(SqlTypeName.TIME);
+    if (expr.precision() > maxPrecision) {
+      throw new UnsupportedOperationException(
```
bestbeforetoday: `UnsupportedOperationException` is usually used when invocation of the operation itself is not supported. For example, adding or modifying a value in a read-only collection. Conversion of precision time literals to Calcite Rex nodes is a valid operation. When an invalid argument is supplied to a supported operation, I might expect to see `IllegalArgumentException`. Or, if it makes sense, possibly a new Isthmus-specific exception such as `IllegalPrecisionException`. Same observation for other instances of this precision check.

isthmus/src/test/java/io/substrait/isthmus/CalciteCallTest.java
```diff
@@ -35,7 +35,7 @@ class CalciteCallTest extends CalciteObjs {
   @Test
   void extract() {
     test(
-        "extract:req_ts",
+        "extract:req_pts",
```
bestbeforetoday, line 38: Do we still need to test `req_ts`?

core/src/main/java/io/substrait/expression/ExpressionCreator.java
```diff
@@ -89,6 +90,23 @@ public static Expression.PrecisionTimeLiteral precisionTime(
...
+   * <p>This method converts a Java {@link LocalTime} to a Substrait precision time literal with
+   * nanosecond precision (precision = 9). The time value is represented as nanoseconds since
+   * midnight (00:00:00).
+   *
+   * @param nullable whether the literal can be null
+   * @param value the LocalTime value to convert
+   * @return a PrecisionTimeLiteral with nanosecond precision representing the given time
+   * @see #precisionTime(boolean, long, int) for creating precision time with custom precision
+   */
+  public static Expression.PrecisionTimeLiteral precisionTime(boolean nullable, LocalTime value) {
```
bestbeforetoday, line 105: This doesn't look to be used anywhere in the codebase. If it is worth adding, it should probably have a test driving it.

core/src/main/java/io/substrait/expression/ExpressionCreator.java
```diff
@@ -155,6 +173,27 @@ public static Expression.PrecisionTimestampLiteral precisionTimestamp(
...
+   * with nanosecond precision (precision = 9). The timestamp value is represented as nanoseconds
+   * since the Unix epoch (1970-01-01 00:00:00 UTC), assuming the LocalDateTime is in UTC timezone.
+   *
+   * @param nullable whether the literal can be null
+   * @param value the LocalDateTime value to convert (interpreted as UTC)
+   * @return a PrecisionTimestampLiteral with nanosecond precision representing the given timestamp
+   * @see #precisionTimestamp(boolean, long, int) for creating precision timestamp with custom
+   *     precision
+   */
+  public static Expression.PrecisionTimestampLiteral precisionTimestamp(
```
bestbeforetoday, line 189: This doesn't look to be used anywhere in the codebase. If it is worth adding, it should probably have a test driving it.

bestbeforetoday:
I see the build failed because some test was expecting UnsupportedOperationException instead of IllegalArgumentException. If that is existing behaviour then it seems OK to me to preserve it, even if it also seems an unintuitive and potentially confusing exception type to me. Your choice.

nielspardon:
> Assuming I have that correct, are we implementing the most useful behaviour? Should the `precisionTime` and `precisionTimestamp` functions create expressions at a precision that can be successfully converted to Calcite, or should we be more lenient on the conversion and truncate (or round) values during conversion to Calcite instead of failing?

I think this a complex and separate topic. 
- `ExpressionCreator` is a component in `core` which means it can be used independently of `isthmus` / Calcite. - However the existing `LocalDateTime`, `LocalDate`, `LocalTime` method variants in the creator were originally created to make the interoperability with Calcite easier since you can get one of the above mentioned objects from a Calcite datetime expression.
- I thought it might in general (independently of Calcite) be useful to create Substrait datetime objects from Java datetime objects without consumers of the substrait-java core having to do the conversion to milli, micro, nano, pico seconds themselves
- How we handle compatibility between core / isthmus is yet another topic. We may even need to handle precision compatibility on the SQL dialect level. I expect the Substrait dialect to play a role here. I recently added support for specifying the maximum supported precision for datetime types to the Substrait dialect: https://github.com/substrait-io/substrait/pull/961

isthmus/src/main/java/io/substrait/isthmus/expression/ExpressionRexConverter.java
nielspardon: thanks, this should be addressed now

isthmus/src/test/java/io/substrait/isthmus/CalciteCallTest.java
nielspardon: Round tripping function calls with the deprecated Timestamp, TimestampTZ, Time was already broken in some cases before this change. We don't have tests that certify backwards compatibility of parsing Substrait plans using deprecated types / functions.

core/src/main/java/io/substrait/expression/ExpressionCreator.java
```diff
@@ -155,6 +173,27 @@ public static Expression.PrecisionTimestampLiteral precisionTimestamp(
...
+      boolean nullable, LocalDateTime value) {
+    long epochNano =
+        TimeUnit.SECONDS.toNanos(value.toEpochSecond(ZoneOffset.UTC))
+            + value.toLocalTime().getNano();
```
bestbeforetoday: Would it be sufficient to just use `value.getNano()`?

bestbeforetoday (commented):
I wonder if an `ExpressionCreator.precisionTimestamp` overload taking a Java Instant might also be useful. Instant is effectively nanoseconds since Unix epoch UTC, similar to precision timestamp. Not necessary for this PR though.

core/src/main/java/io/substrait/expression/ExpressionCreator.java
nielspardon: yes, makes sense. will change that.

isthmus/src/test/java/io/substrait/isthmus/CalciteCallTest.java
nielspardon: I did create PR https://github.com/substrait-io/substrait/pull/994 to remove the deprecated time/timestamp types

bestbeforetoday (approved):

PR #723, opened 2026-03-05
feat: add missing extension YAMLs and test for completeness

### LINKED ISSUE #722: Add test to catch missing extension YAML files in DefaultExtensionCatalog
While adding `functions_aggregate_decimal_output` to `DefaultExtensionCatalog`, we noticed there was no way to catch this kind of omission. Added a test that dynamically discovers all extension YAMLs on the classpath and verifies they're loaded, so we don't silently miss new extensions again.

Also added the other missing files found by the test: `functions_set` and `functions_geometry`. `functions_list` remains excluded due to #688.

benbellick:
Closes #722

Adds a test that dynamically discovers all extension YAMLs under `substrait/extensions/` on the classpath (via ClassGraph) and verifies they are loaded by `DefaultExtensionCatalog`. Files not yet loadable are listed in `UNSUPPORTED_FILES` with comments explaining why. The test also asserts these are NOT loaded, so the skip list can't silently go stale.

Also relocates extension YAMLs from the classpath root into `substrait/extensions/`, adds the `FUNCTIONS_AGGREGATE_DECIMAL_OUTPUT` constant, and loads `functions_set`.

### DIFF
--- a/core/src/main/java/io/substrait/extension/DefaultExtensionCatalog.java
+++ b/core/src/main/java/io/substrait/extension/DefaultExtensionCatalog.java
@@ -41,9 +41,6 @@ public class DefaultExtensionCatalog {
   /** Extension identifier for geometry functions. */
   public static final String FUNCTIONS_GEOMETRY = "extension:io.substrait:functions_geometry";
 
-  /** Extension identifier for list functions. */
-  public static final String FUNCTIONS_LIST = "extension:io.substrait:functions_list";
-
   /** Extension identifier for logarithmic functions. */
   public static final String FUNCTIONS_LOGARITHMIC = "extension:io.substrait:functions_logarithmic";
 
@@ -56,6 +53,9 @@ public class DefaultExtensionCatalog {
 
   /** Extension identifier for set functions. */
   public static final String FUNCTIONS_SET = "extension:io.substrait:functions_set";
+
+  /** Extension identifier for list functions. */
+  public static final String FUNCTIONS_LIST = "extension:io.substrait:functions_list";
 
   /** Extension identifier for string functions. */
   public static final String FUNCTIONS_STRING = "extension:io.substrait:functions_string";


core/src/test/java/io/substrait/extension/DefaultExtensionCatalogTest.java
```diff
@@ -0,0 +1,66 @@
...
+
+  private static String parseUrn(File yamlFile) throws IOException {
+    JsonNode doc = YAML_MAPPER.readTree(yamlFile);
+    JsonNode urnNode = doc.get("urn");
+    return urnNode == null ? null : urnNode.asText();
+  }
+
+  private static List<File> getExtensionYamlFiles() {
+    File extensionsDir = new File("../substrait/extensions");
+    assertTrue(extensionsDir.isDirectory(), "substrait/extensions directory not found");
```
nielspardon: I'm wondering whether this should check the extension YAML files on the classpath instead of relying on the current working directory (especially given that @vbarua is planning to remove the submodule and move packaging the extension YAMLs into a separate JAR into substrait-packaging). The YAMLs should be [already on the classpath](https://github.com/substrait-io/substrait-java/blob/663049bdaf880be4f99f20444a2db91b18c26d96/core/build.gradle.kts#L230). We're already loading [classgraph](https://github.com/classgraph/classgraph) for the isthmus-cli which can be useful to scan the classpath for any `*.yaml` files.

benbellick: Sounds good to me, I'll make that change shortly. Thanks!

nielspardon: btw. it might be better to relocate the YAML files on the classpath into a subfolder like `substrait/extensions` to be easier recognizable as default YAML files. right now they are being added into the root of the classpath

PR #740, opened 2026-03-09
feat(core)!: bump substrait to v0.85.0, drop URI support

benbellick:
Completes the URI-to-URN migration started in #522.

Substrait spec v0.85.0 removed the deprecated URI fields from the proto definitions (`SimpleExtensionURI`, `extension_uris`, `extension_uri_reference`) in favor of `SimpleExtensionURN`, `extension_urns`, and `extension_urn_reference` (see [substrait#971](https://github.com/substrait-io/substrait/pull/971)). This PR bumps the submodule and drops all URI handling from substrait-java to match.

Previously, `SimpleExtension` maintained a bidirectional URI-to-URN map so that plans could accept either format on input and emit both on output. `ExtensionCollector` and `ImmutableExtensionLookup` both dealt with URI fields alongside URN fields during serialization and deserialization. With the upstream removal, all of that is unnecessary. The extension system now works exclusively with URNs.

### DIFF
--- a/core/src/main/java/io/substrait/extension/SimpleExtension.java
+++ b/core/src/main/java/io/substrait/extension/SimpleExtension.java
@@ -614,7 +614,7 @@ public class SimpleExtension {
   }
 
   /**
-   * The catalog of function and type definitions loaded from YAML extension files. Maps URN+name
+   * The catalog of function and type definitions loaded from YAML extension files. Maps URN + name
    * pairs to full definitions (argument types, return types, etc.).
    *
    * <p>Used by {@link AbstractExtensionLookup#getScalarFunction} and similar methods to resolve a


core/src/test/resources/plan-roundtrip/complex-expected-plan.json
benbellick, line 1: I think these roundtrip tests are still valuable, even if we aren't testing the migration.

core/src/main/java/io/substrait/extension/ImmutableExtensionLookup.java
```diff
@@ -22,230 +22,72 @@ private ImmutableExtensionLookup(
   }
 
   public static Builder builder() {
-    return builder(DefaultExtensionCatalog.DEFAULT_COLLECTION);
```
benbellick, line 25: We used to need this to be able to access the URI<->URN mapping. It was a bit of a hack before but was necessary during the migration.

core/src/main/java/io/substrait/extension/ImmutableExtensionLookup.java
```diff
@@ -22,230 +22,72 @@ private ImmutableExtensionLookup(
...
 
     private Builder from(
         List<SimpleExtensionURN> simpleExtensionURNs,
-        List<io.substrait.proto.SimpleExtensionURI> simpleExtensionURIs,
         List<SimpleExtensionDeclaration> simpleExtensionDeclarations) {
       Map<Integer, String> urnMap = new HashMap<>();
-      Map<Integer, String> uriMap = new HashMap<>();
 
-      // Handle URN format
+      // TODO: consider validating URN format (e.g. must start with "extension:")
```
benbellick, line 72: I think it could be reasonable to validate at least `extension:<namespace>:<name>` here, but I figure that can be left for another PR where it can be properly discussed.

core/src/main/java/io/substrait/extension/SimpleExtension.java
```diff
@@ -592,6 +592,13 @@ public Stream<SimpleExtension.Function> resolve(String urn) {
     }
   }
 
+  /**
+   * The catalog of function and type definitions loaded from YAML extension files. Maps URN+name
+   * pairs to full definitions (argument types, return types, etc.).
+   *
+   * <p>Used by {@link AbstractExtensionLookup#getScalarFunction} and similar methods to resolve a
+   * {@link FunctionAnchor} into a complete {@link Function} with signature metadata.
+   */
```
benbellick, line 622: This wasn't strictly necessary for this PR, but I get a bit confused in this codebase on the differences between `SimpleExtension`, `ExtensionCollector`, `ExtensionCollection`, and `ImmutableExtensionLookup`. I wanted to move in the direction of clearer documentation on these differences.

core/src/test/java/io/substrait/extension/ExtensionCollectionMergeTest.java
benbellick, line 1: This test had to do with the management of the URI<->URN mapping when there was an implied discrepancy of the mapping from the YAML files. Since there is no more mapping, these tests are useless now.

core/src/test/java/io/substrait/extension/ExtensionLookupTest.java
benbellick, line 1: This is a new test but testing normally expected behavior, which is just that when we look things up in the `ImmutableExtensionLookup`, they _are_ indeed there.

core/src/test/java/io/substrait/plan/PlanRoundtripTest.java
benbellick, line 1: Yeah just added this roundtrip tests (as mentioned in another comment) because we already had those nice plans, and I think this is a good category of tests to have in general.

core/src/main/java/io/substrait/extension/ImmutableExtensionLookup.java
nielspardon: if we want to validate the URN format we may want to add a regexp pattern to the extension YAML schema in the spec:
- https://github.com/substrait-io/substrait/blob/2aaae7c0d369a23b5b469badce0644097f1769a1/text/simple_extensions_schema.yaml#L8-L9
- https://json-schema.org/understanding-json-schema/reference/regular_expressions#example

I would agree that is a separate change. Do you want to create an issue for this?

core/src/main/java/io/substrait/extension/SimpleExtension.java
```diff
@@ -592,6 +592,13 @@ public Stream<SimpleExtension.Function> resolve(String urn) {
     }
   }
 
+  /**
+   * The catalog of function and type definitions loaded from YAML extension files. Maps URN+name
```
nielspardon: not having the spaces hurts my eyes :-) 
```suggestion
   * The catalog of function and type definitions loaded from YAML extension files. Maps URN + name
```

core/src/main/java/io/substrait/extension/ImmutableExtensionLookup.java
benbellick: https://github.com/substrait-io/substrait/pull/881

I have this PR open where there is some discussion about this. It seemed to stall a bit and it was relatively low priority on my end, so I decided to prioritize other things. However, I would love to see some kind of resolution there :)

core/src/test/java/io/substrait/plan/PlanRoundtripTest.java
nielspardon: Do we have the code that produced these plans also somewhere? Given that the Substrait might evolve in breaking ways we likely will need to update the JSON files at some point in time in the future.

benbellick: No code to generate these plans unfortunately. If we evolve the substrait in breaking ways, then we should at the very least see test failures which prompt us to update the plans, though the update will be by hand for now 🥲

PR #742, opened 2026-03-10
fix(core): create precision time protobuf

bestbeforetoday:
The io.substrait.type.proto.TypeProtoConverter.Types#wrap(Object) method was missing a branch to handle PrecisionTime. This caused an UnsupportedOperationException to the be thrown when creating a PrecisionTime protobuf object.

This change adds the missing branch to the wrap method, and extends type pojo/proto conversion tests to include PrecisionTime.


nielspardon (approved):
LGTM

PR #744, opened 2026-03-10
fix(core): use new aggregate grouping behavior in RelProtoConverter

nielspardon:
changes RelProtoConverter to output the new and the old aggregate grouping proto structures

relates to https://github.com/substrait-io/substrait-java/issues/299 and https://github.com/substrait-io/substrait/pull/1002

This PR changes the `RelProtoConverter` to output the new and the old grouping behavior keeping the existing `io.substrait.relation.Aggregate` behavior the same.

### DIFF
--- a/core/src/test/java/io/substrait/relation/AggregateRelTest.java
+++ b/core/src/test/java/io/substrait/relation/AggregateRelTest.java
@@ -69,9 +69,7 @@ class AggregateRelTest extends TestBase {
    */
   private static List<List<Integer>> getExpressionReferences(AggregateRel aggregateRel) {
     return aggregateRel.getGroupingsList().stream()
-        .map(
-            grouping ->
-                grouping.getExpressionReferencesList().stream().collect(Collectors.toList()))
+        .map(AggregateRel.Grouping::getExpressionReferencesList)
         .collect(Collectors.toList());
   }
 
@@ -85,16 +83,6 @@ class AggregateRelTest extends TestBase {
     return aggregateRel.getGroupingsList().stream()
         .map(grouping -> grouping.getGroupingExpressionsList())
         .collect(Collectors.toList());
-  }
-
-  /**
-   * Helper method to extract aggregate-level grouping expressions from an AggregateRel.
-   *
-   * @param aggregateRel the AggregateRel to extract grouping expressions from
-   * @return a list of expressions at the aggregate level
-   */
-  private static List<Expression> getAggregateGroupingExpressions(AggregateRel aggregateRel) {
-    return aggregateRel.getGroupingExpressionsList();
   }
 
   @Test
@@ -140,7 +128,7 @@ class AggregateRelTest extends TestBase {
     // Verify backward compatibility: deprecated grouping_expressions field is also populated
     assertEquals(List.of(List.of(col1Ref, col2Ref)), getGroupingExpressions(roundtripAgg));
     // Verify aggregate-level grouping_expressions field is populated
-    assertEquals(List.of(col1Ref, col2Ref), getAggregateGroupingExpressions(roundtripAgg));
+    assertEquals(List.of(col1Ref, col2Ref), roundtripAgg.getGroupingExpressionsList());
   }
 
   @Test
@@ -188,7 +176,7 @@ class AggregateRelTest extends TestBase {
     // Verify backward compatibility: deprecated grouping_expressions field is also populated
     assertEquals(List.of(List.of(col1Ref, col2Ref)), getGroupingExpressions(roundtripAgg));
     // Verify aggregate-level grouping_expressions field is populated
-    assertEquals(List.of(col1Ref, col2Ref), getAggregateGroupingExpressions(roundtripAgg));
+    assertEquals(List.of(col1Ref, col2Ref), roundtripAgg.getGroupingExpressionsList());
   }
 
   @Test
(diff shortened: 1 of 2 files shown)


core/src/test/java/io/substrait/relation/AggregateRelTest.java
```diff
@@ -169,5 +206,27 @@ void testAggregateWithMultipleGroupings() {
...
+        roundtripAgg.getGroupings(0).getGroupingExpressionsCount(),
+        "grouping expressions count of grouping should be 2");
+    assertEquals(
+        2,
+        roundtripAgg.getGroupings(0).getExpressionReferencesCount(),
+        "expression reference count of grouping should be 2");
+    assertEquals(
+        1,
+        roundtripAgg.getGroupings(1).getExpressionReferencesCount(),
+        "expression reference count of grouping should be 2");
```
benbellick: The message here says it should be 2 but the assert statement says it should be 1.

benbellick:
```suggestion
    assertEquals(
        List.of(List.of(0, 1), List.of(1)), getExpressionReferences(roundtripAgg));
  }
```

I realize that this style of test is consistent with above, but I think it could be easier to read these tests if we:
1. introduced a helper `getExpressionReferences`, and
2. just checked the expected output rather than the length of each part individually. 

What do you think?

benbellick: This could apply to all of the tests added here as well.

benbellick (commented):
I think it looks good, just the one comment about making the tests a bit clearer. Thanks!

core/src/test/java/io/substrait/relation/AggregateRelTest.java
nielspardon: sure, I changed the tests accordingly and rebased on latest main

nielspardon: with the new test approach this typo doesn't exist anymore. thanks for catching it

core/src/test/java/io/substrait/relation/AggregateRelTest.java
```diff
@@ -169,5 +231,17 @@ void testAggregateWithMultipleGroupings() {
...
+
+    AggregateRel roundtripAgg = roundtripRel.getAggregate();
+    // Verify new expression_references structure
+    assertEquals(List.of(List.of(0, 1), List.of(1)), getExpressionReferences(roundtripAgg));
+    // Verify backward compatibility: deprecated grouping_expressions field is also populated
+    assertEquals(
+        List.of(List.of(col1Ref, col2Ref), List.of(col2Ref)), getGroupingExpressions(roundtripAgg));
+    // Verify aggregate-level grouping_expressions field is populated
+    assertEquals(List.of(col1Ref, col2Ref), getAggregateGroupingExpressions(roundtripAgg));
   }
```
benbellick, line 262: perhaps overkill but can't hurt to include. 

```suggestion
  }

  /**
   * Tests deduplication of non-trivial grouping expressions by equals() (not identity), and that an
   * empty grouping set is handled correctly alongside non-empty ones.
   */
  @Test
  void testGroupingExpressionDeduplicationAndEmptyGroupingSet() {
    NamedScan input = sb.namedScan(List.of("t"), List.of("a", "b"), List.of(R.I32, R.I32));

    // Two independently constructed add(a, b) expressions — equal but not same instance.
    io.substrait.expression.Expression addAB1 =
        sb.add(sb.fieldReference(input, 0), sb.fieldReference(input, 1));
    io.substrait.expression.Expression addAB2 =
        sb.add(sb.fieldReference(input, 0), sb.fieldReference(input, 1));

    Aggregate agg =
        Aggregate.builder()
            .input(input)
            .addGroupings(sb.grouping(addAB1), sb.grouping(addAB2), sb.grouping())
            .build();

    Rel roundtripRel = relProtoConverter.visit(agg, EmptyVisitationContext.INSTANCE);
    AggregateRel result = roundtripRel.getAggregate();

    assertEquals(
        List.of(expressionProtoConverter.toProto(addAB1)), result.getGroupingExpressionsList());
    assertEquals(List.of(List.of(0), List.of(0), List.of()), getExpressionReferences(result));
  }
```

core/src/test/java/io/substrait/relation/AggregateRelTest.java
```diff
@@ -57,6 +60,43 @@ public static io.substrait.proto.Expression createFieldReference(int col) {
...
+   *
+   * @param aggregateRel the AggregateRel to extract expression references from
+   * @return a list of lists, where each inner list contains the expression reference indices for a
+   *     grouping
+   */
+  private static List<List<Integer>> getExpressionReferences(AggregateRel aggregateRel) {
+    return aggregateRel.getGroupingsList().stream()
+        .map(
+            grouping ->
+                grouping.getExpressionReferencesList().stream().collect(Collectors.toList()))
```
benbellick: The _most_ minor nit: I think there is no need to copy here since this is test code, so we can simplify this a bit. 
```suggestion
        .map(AggregateRel.Grouping::getExpressionReferencesList)
```

core/src/test/java/io/substrait/relation/AggregateRelTest.java
```diff
@@ -57,6 +60,43 @@ public static io.substrait.proto.Expression createFieldReference(int col) {
...
+
+  /**
+   * Helper method to extract aggregate-level grouping expressions from an AggregateRel.
+   *
+   * @param aggregateRel the AggregateRel to extract grouping expressions from
+   * @return a list of expressions at the aggregate level
+   */
+  private static List<Expression> getAggregateGroupingExpressions(AggregateRel aggregateRel) {
+    return aggregateRel.getGroupingExpressionsList();
+  }
```
benbellick:
```suggestion
```
IMO, this helper is just wrapping an existing method, lets just inline the method call.

benbellick (approved):
LGTM, just a couple small comments. Thanks!

PR #745, opened 2026-03-11
feat(isthmus): enable precision timestamp addition/subtraction

nielspardon:
This PR adds support for precision timestamp addition/subtraction in isthmus. 

- Previously, the mapping code only supported the old timestamp types.
- Also fixes the use of type names to use the proper short names.
- Removes `time` for the `DATETIME_PLUS` mapping since that is not officially supported in the default Substrait `extension:io.substrait:functions_datetime` collection.
- Adds testcases for `date`, `precision_timestamp` and `precision_timestamp_tz` (generated by IBM Bob but I checked them)

### DIFF
--- a/isthmus/src/main/java/io/substrait/isthmus/expression/FunctionMappings.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/expression/FunctionMappings.java
@@ -140,12 +140,12 @@ public class FunctionMappings {
           resolver(
               SqlStdOperatorTable.PLUS, Set.of("i8", "i16", "i32", "i64", "fp32", "fp64", "dec")),
           SqlStdOperatorTable.DATETIME_PLUS,
-          resolver(SqlStdOperatorTable.PLUS, Set.of("date", "time", "timestamp")),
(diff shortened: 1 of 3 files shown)


isthmus/src/main/java/io/substrait/isthmus/expression/FunctionMappings.java
```diff
@@ -140,12 +140,12 @@ public class FunctionMappings {
           resolver(
               SqlStdOperatorTable.PLUS, Set.of("i8", "i16", "i32", "i64", "fp32", "fp64", "dec")),
           SqlStdOperatorTable.DATETIME_PLUS,
-          resolver(SqlStdOperatorTable.PLUS, Set.of("date", "time", "timestamp")),
+          resolver(SqlStdOperatorTable.DATETIME_PLUS, Set.of("date", "ts", "tstz", "pts", "ptstz")),
```
nielspardon, line 143: this somehow still did work with the less specific/wrong SqlOperator. now it uses the right one

andrew-coleman (approved):

PR #755, opened 2026-03-18
fix(javadoc): partial isthmus calcite

mbwhite:


isthmus/src/main/java/io/substrait/isthmus/calcite/SubstraitTable.java
```diff
@@ -4,21 +4,45 @@
 import org.apache.calcite.rel.type.RelDataTypeFactory;
 import org.apache.calcite.schema.impl.AbstractTable;
 
-/** Basic {@link AbstractTable} implementation */
+/**
+ * Basic {@link AbstractTable} implementation for Substrait.
```
bestbeforetoday, line 8: Just an observation—not necessary to change anything...

The existing JavaDoc says `AbstractTable` but I think in reality this is a `org.apache.calcite.schema.Table` implementation. The fact it extends `AbstractTable` (which in turn implements the `Table` interface is more an implementation convenience.

PR #756, opened 2026-03-18
fix(javadoc): partial isthmus calcite rel

mbwhite:


isthmus/src/main/java/io/substrait/isthmus/calcite/rel/CreateTable.java
```diff
@@ -21,17 +21,33 @@ public CreateTable(List<String> tableName, RelNode input) {
...
+   */
   @Override
   public RelWriter explainTerms(RelWriter pw) {
     return super.explainTerms(pw).item("tableName", getTableName());
   }
 
+  /**
+   * Returns the inputs to this node (single input).
+   *
+   * @return a list containing the input relation
```
bestbeforetoday: It returns a Calcite RelNode, not a list. Should this have been the doc for the `inputs` parameter? `@param` doc is missing.

bestbeforetoday (approved):

PR #761, opened 2026-03-18
docs(isthmus): javadoc for io.substrait.isthmus type converters

mbwhite:


isthmus/src/main/java/io/substrait/isthmus/expression/CallConverters.java
bestbeforetoday: This one seems to have been missed. I guess there is Javadoc there... Just not very much. 😉

isthmus/src/main/java/io/substrait/isthmus/expression/CallConverters.java
```diff
@@ -182,6 +198,9 @@ else if (operand instanceof Expression.StructLiteral
    * Expand {@link org.apache.calcite.util.Sarg} values in a calcite `SqlSearchOperator` into
    * simpler expressions. The expansion logic is encoded in {@link RexUtil#expandSearch(RexBuilder,
    * RexProgram, RexNode)}
+   *
+   * <p>Returns a factory of {@link SimpleCallConverter} that expands SEARCH calls using the
```
bestbeforetoday: I find this wording slightly confusing. The `CREATE_SEARCH_CONV` static variable is a function that returns a SimpleCallConverter. You might call the variable/function itself a factory but it does not **return** a factory.

isthmus/src/main/java/io/substrait/isthmus/expression/CallConverters.java
```diff
@@ -206,10 +231,27 @@ public static List<CallConverter> defaults(TypeConverter typeConverter) {
...
+     * if not handled.
+     *
+     * @param call the Calcite call to convert
+     * @param topLevelConverter converter for nested {@link RexNode} operands
+     * @return converted expression, or {@code null} if not applicable
+     */
     @Nullable Expression apply(RexCall call, Function<RexNode, Expression> topLevelConverter);
 
+    /**
+     * Default adapter to {@link CallConverter#convert(RexCall, Function)} returning {@link
```
bestbeforetoday: I think this is intended to refer to the `apply` function rather than to itself.

```suggestion
     * Default adapter to {@link CallConverter#apply(RexCall, Function)} returning {@link
```

PR #762, opened 2026-03-18
docs(isthmus): javadoc for io.substrait.isthmus type system classes

mbwhite:


isthmus/src/main/java/io/substrait/isthmus/SubstraitTypeSystem.java
```diff
@@ -41,6 +57,11 @@ public int getMaxPrecision(final SqlTypeName typeName) {
     return super.getMaxPrecision(typeName);
   }
 
+  /**
+   * Returns the maximum numeric scale supported by this type system.
+   *
+   * @return Maximum numeric scale (38).
+   */
```
bestbeforetoday, line 67: This description is incorrect. The parent interface Javadoc says:

```java
/**
   * Returns default precision for this type if supported, otherwise
   * {@link RelDataType#PRECISION_NOT_SPECIFIED}
   * if precision is either unsupported or must be specified explicitly.
   *
   * @return Default precision
   */
```

Also, the return value (in the current implementation) is 38 for the decimal type, but not for any other type.

isthmus/src/main/java/io/substrait/isthmus/SubstraitTypeSystem.java
```diff
@@ -51,6 +72,11 @@ public int getDefaultPrecision(final SqlTypeName typeName) {
     }
   }
 
+  /**
+   * Returns the maximum numeric precision supported by this type system.
+   *
+   * @return Maximum numeric precision (38).
+   */
```
bestbeforetoday, line 85: This description is incorrect. The parent interface Javadoc says:

```java
/**
   * Returns the maximum scale allowed for this type, or
   * {@link RelDataType#SCALE_NOT_SPECIFIED}
   * if scale is not applicable for this type.
   *
   * @return Maximum allowed scale
   */
```

The maximum scale for the decimal type (in the current implementation) is 38. For all other types it is different.

isthmus/src/main/java/io/substrait/isthmus/TypeConverter.java
```diff
@@ -20,11 +20,27 @@
...
+ *
+ * @see UserTypeMapper
+ * @see io.substrait.type.Type
+ * @see org.apache.calcite.rel.type.RelDataType
+ */
 public class TypeConverter {
 
   private final UserTypeMapper userTypeMapper;
 
   // DEFAULT TypeConverter which does not handle user-defined types
```
bestbeforetoday: The Javadoc below now provides the same information so this line can be removed.

PR #764, opened 2026-03-18
docs(isthmus): javadoc for io.substrait.isthmus.expression classes

mbwhite:


isthmus/src/main/java/io/substrait/isthmus/expression/RexExpressionConverter.java
```diff
@@ -59,19 +86,34 @@ public RexExpressionConverter(
   }
 
   /**
-   * Only used for testing. Missing `ScalarFunctionConverter`, `CallConverters.CREATE_SEARCH_CONV`
+   * Testing-only constructor that wires default converters.
+   *
+   * <p>Missing {@code ScalarFunctionConverter} and {@code CallConverters.CREATE_SEARCH_CONV}.
    */
   public RexExpressionConverter() {
```
bvolpato, line 93: Maybe we make those package-private and bring a `@VisibleForTesting` contention over?

bestbeforetoday: I agree that anything visible only for testing should ideally be default scope rather than public. Ideally the code would be testable without requiring test-only visibility but I appreciate that it is sometimes necessary / preferred.

PR #765, opened 2026-03-18
fix(javadoc): partial isthmus

mbwhite:


isthmus/src/main/java/io/substrait/isthmus/SqlExpressionToSubstrait.java
```diff
@@ -27,25 +27,47 @@
...
+  /** Converter for RexNodes to Substrait expressions. */
   protected final RexExpressionConverter rexConverter;
 
+  /** Creates a converter with default features and the default extension catalog. */
   public SqlExpressionToSubstrait() {
     this(new ConverterProvider());
   }
 
+  /**
+   * Creates a converter with the given feature board and extension collection.
```
bestbeforetoday: I think the feature board is no longer a thing. Configuration is now handled by the supplied converter provider. This text needs to be changed to reflect that.

isthmus/src/main/java/io/substrait/isthmus/SqlExpressionToSubstrait.java
```diff
@@ -27,25 +27,47 @@
...
+ *
+ * <p>Supports optional CREATE TABLE statements to provide schema and column bindings for expression
+ * validation and Rex conversion.
+ */
 public class SqlExpressionToSubstrait extends SqlConverterBase {
 
+  /** Converter for RexNodes to Substrait expressions. */
   protected final RexExpressionConverter rexConverter;
 
+  /** Creates a converter with default features and the default extension catalog. */
```
bestbeforetoday: Maybe better to keep it slightly more generic like this...?

```suggestion
  /** Creates a converter with default configuration. */
```

isthmus/src/main/java/io/substrait/isthmus/SqlToSubstrait.java
```diff
@@ -19,10 +19,18 @@
 public class SqlToSubstrait extends SqlConverterBase {
   private final SqlOperatorTable operatorTable;
 
+  /**
+   * Creates a SQL-to-Substrait converter using the default extension catalog and default features.
```
bestbeforetoday: Similar to other comment, I worry about referencing "features" since this hints at the old feature board. Maybe just refer to "default configuration"?

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
```diff
@@ -90,25 +90,51 @@
...
   /** Use {@link #SubstraitRelNodeConverter(RelBuilder, ConverterProvider)} instead */
   @Deprecated
+  /**
+   * Creates a new SubstraitRelNodeConverter with the specified extensions, type factory, and
+   * relation builder.
+   *
+   * @param extensions the Substrait extension collection
+   * @param typeFactory the Calcite type factory
+   * @param relBuilder the Calcite relation builder
+   */
```
bestbeforetoday: There are two JavaDoc blocks. There should be only one, before the `@Deprecated` annotation. The `@deprecated` JavaDoc tag can provide deprecation information.

```suggestion
  /**
   * Creates a new SubstraitRelNodeConverter with the specified extensions, type factory, and
   * relation builder.
   *
   * @param extensions the Substrait extension collection
   * @param typeFactory the Calcite type factory
   * @param relBuilder the Calcite relation builder
   * @deprecated Use {@link #SubstraitRelNodeConverter(RelBuilder, ConverterProvider)} instead.
   */
@Deprecated
```

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
```diff
@@ -119,6 +145,18 @@ public SubstraitRelNodeConverter(RelBuilder relBuilder, ConverterProvider conver
     this.expressionRexConverter = converterProvider.getExpressionRexConverter(this);
   }
 
+  /**
+   * Converts a Substrait {@link Rel} plan to a Calcite {@link RelNode} using default feature
+   * settings.
+   *
+   * <p>This method creates a {@link RelBuilder} configured with the provided cluster and catalog,
+   * then delegates to the converter with default features.
```
bestbeforetoday: This part doesn't really make sense to me. Particularly about delegating to the converter. It might be simpler just to remove this description of the implementation since the end user probably doesn't really care how the internals are implemented provided the outcome matches their expectations.

PR #766, opened 2026-03-18
fix(javadoc): partial isthmus

mbwhite:


isthmus/src/main/java/io/substrait/isthmus/SubstraitRelVisitor.java
```diff
@@ -73,28 +73,57 @@ public class SubstraitRelVisitor extends RelNodeVisitor<Rel, RuntimeException> {
...
   private Map<RexFieldAccess, Integer> fieldAccessDepthMap;
 
   /** Use {@link SubstraitRelVisitor#SubstraitRelVisitor(ConverterProvider)} */
   @Deprecated
+  /**
+   * Creates a new SubstraitRelVisitor with the specified type factory and extensions.
+   *
+   * @param typeFactory the Calcite type factory
+   * @param extensions the Substrait extension collection
+   */
```
bestbeforetoday, line 93: There are two JavaDoc blocks for the one method: one before and one after the `@Deprecated` annotation. There should be a single JavaDoc block before the annotation.

I suggest using the `@deprecated` JavaDoc tag for documentation on the deprecation.

```suggestion
  /**
   * Creates a new SubstraitRelVisitor with the specified type factory and extensions.
   *
   * @param typeFactory the Calcite type factory
   * @param extensions the Substrait extension collection
   * @deprecated Use {@link #SubstraitRelVisitor(ConverterProvider)} instead.
   */
```

PR #767, opened 2026-03-18
docs: update substrait builder javadoc and spelling

mbwhite:
Update to the JavaDoc for the Substrait Builder (and fix a spelling mistake)

This is ahead of some new functions coming in a follow on pr.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,12 @@
 Release Notes
 ---
 
+## [0.86.1](https://github.com/substrait-io/substrait-java/compare/v0.86.0...v0.86.1) (2026-03-22)
+
+### Bug Fixes
+
+* **javadoc:** partial isthmus ([#759](https://github.com/substrait-io/substrait-java/issues/759)) ([8ffa3b3](https://github.com/substrait-io/substrait-java/commit/8ffa3b32ab6d5359575a531c3e0911958eb43713))
+
 ## [0.86.0](https://github.com/substrait-io/substrait-java/compare/v0.85.0...v0.86.0) (2026-03-19)
 
 ### Features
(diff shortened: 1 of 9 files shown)


core/src/main/java/io/substrait/dsl/SubstraitBuilder.java
```diff
@@ -49,25 +49,72 @@
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
+/**
+ * A builder class for constructing Substrait query plans using a fluent API.
```
bestbeforetoday: I think the reference to a "fluent API" is misleading here. The builders used internally by this class provide a fluent API where methods are chained to define all the required properties before building. This class does not do that. It provides standalone (non-chainable) factory methods for constructing Substrait objects.

core/src/main/java/io/substrait/dsl/SubstraitBuilder.java
```diff
@@ -49,25 +49,72 @@
...
+ *
+ * <p>This class provides factory methods for creating various Substrait relation operators (e.g.,
+ * aggregate, join, filter, project) and expressions (e.g., literals, function invocations, field
+ * references). It simplifies the construction of complex query plans by providing a type-safe,
+ * builder-style interface.
```
bestbeforetoday: It uses builder style interfaces internally but does not provide a builder-style interface.

core/src/main/java/io/substrait/dsl/SubstraitBuilder.java
mbwhite: Yes perfectly correct @bestbeforetoday that is the wrong use of fluent

core/src/main/java/io/substrait/dsl/SubstraitBuilder.java
mbwhite: Likewise - the wrong usage

PR #773, opened 2026-03-23
feat: substrait builder extra api

mbwhite:
Follow on to pr #767 to fill some missing gaps in the substrait builder. We've recently used the Substrait Builder and found that there some variants and missing gaps in the APIs.  We'd subclassed it to add these, so I'm now wanting to push these to the main codebase. 

There wasn't a unit test already so I've added a basic one here - which I appreciate will is quite large.

### DIFF
--- a/core/src/main/java/io/substrait/dsl/SubstraitBuilder.java
+++ b/core/src/main/java/io/substrait/dsl/SubstraitBuilder.java
@@ -97,12 +97,9 @@ public class SubstraitBuilder {
   }
 
   /**
-   * Gets the default extension collection used by this builder.
+   * Gets the extension collection used by this builder.
    *
-   * <p>This collection includes standard Substrait functions for strings, arithmetic, comparison,
-   * datetime, and other operations from {@link DefaultExtensionCatalog#DEFAULT_COLLECTION}.
-   *
-   * @return the ExtensionCollection containing standard Substrait functions
+   * @return the ExtensionCollection used with this builder
    */
   public SimpleExtension.ExtensionCollection getExtensions() {
     return extensions;
@@ -149,25 +146,6 @@ public class SubstraitBuilder {
   }
 
   /**
-   * Creates an aggregate relation with a single grouping and output field remapping.
-   *
-   * @param groupingFn function to derive the grouping from the input relation
-   * @param measuresFn function to derive the measures from the input relation
-   * @param remap the output field remapping specification
-   * @param input the input relation to aggregate
-   * @return a new {@link Aggregate} relation
-   */
-  public Aggregate aggregate(
-      Function<Rel, Aggregate.Grouping> groupingFn,
-      Function<Rel, List<Aggregate.Measure>> measuresFn,
-      Rel.Remap remap,
-      Rel input) {
-    Function<Rel, List<Aggregate.Grouping>> groupingsFn =
-        groupingFn.andThen(g -> Stream.of(g).collect(Collectors.toList()));
-    return aggregate(groupingsFn, measuresFn, Optional.of(remap), input);
-  }
-
-  /**
    * Creates an aggregate relation that groups and aggregates data from an input relation.
    *
    * <p>This method constructs a Substrait aggregate operation by applying grouping and measure
@@ -187,12 +165,12 @@ public class SubstraitBuilder {
    * @return an Aggregate relation representing the grouping and aggregation operation
    */
   public Aggregate aggregate(
-      final Function<Rel, List<Aggregate.Grouping>> groupingsFn,
-      final Function<Rel, List<Aggregate.Measure>> measuresFn,
-      final Optional<Rel.Remap> remap,
-      final Rel input) {
-    final List<Aggregate.Grouping> groupings = groupingsFn.apply(input);
-    final List<Aggregate.Measure> measures = measuresFn.apply(input);
+      Function<Rel, List<Aggregate.Grouping>> groupingsFn,
+      Function<Rel, List<Aggregate.Measure>> measuresFn,
+      Optional<Rel.Remap> remap,
+      Rel input) {
+    List<Aggregate.Grouping> groupings = groupingsFn.apply(input);
+    List<Aggregate.Measure> measures = measuresFn.apply(input);
     return Aggregate.builder()
         .groupings(groupings)
         .measures(measures)
@@ -903,7 +881,7 @@ public class SubstraitBuilder {
    * @param value value to create
    * @return i16 instance
    */
-  public Expression.I8Literal i8(final int value) {
+  public Expression.I8Literal i8(int value) {
     return Expression.I8Literal.builder().value(value).build();
   }
 
@@ -913,7 +891,7 @@ public class SubstraitBuilder {
    * @param value value to create
    * @return i16 instance
    */
-  public Expression.I16Literal i16(final int value) {
+  public Expression.I16Literal i16(int value) {
     return Expression.I16Literal.builder().value(value).build();
   }
 
@@ -933,7 +911,7 @@ public class SubstraitBuilder {
    * @param value value to create
    * @return i64 instance
    */
-  public Expression.I64Literal i64(final long value) {
+  public Expression.I64Literal i64(long value) {
     return Expression.I64Literal.builder().value(value).build();
   }
 
@@ -943,7 +921,7 @@ public class SubstraitBuilder {
    * @param value the float value
    * @return a new {@link Expression.FP32Literal}
    */
-  public Expression.FP32Literal fp32(final float value) {
+  public Expression.FP32Literal fp32(float value) {
     return Expression.FP32Literal.builder().value(value).build();
   }
 
@@ -1532,7 +1510,7 @@ public class SubstraitBuilder {
    * @param expression the boolean expression to negate
    * @return a scalar function invocation representing the logical NOT of the input expression
    */
-  public Expression not(final Expression expression) {
+  public Expression not(Expression expression) {
     return this.scalarFn(
         DefaultExtensionCatalog.FUNCTIONS_BOOLEAN,
         "not:bool",
@@ -1554,7 +1532,7 @@ public class SubstraitBuilder {
    * @return a scalar function invocation that returns true if the expression is null, false
    *     otherwise
    */
-  public Expression isNull(final Expression expression) {
+  public Expression isNull(Expression expression) {
 
     final List<Expression> args = new ArrayList<>();
     args.add(expression);
@@ -1581,12 +1559,12 @@ public class SubstraitBuilder {
    * @return a scalar function invocation expression
    */
   public Expression scalarFn(
-      final String urn,
-      final String key,
-      final Type returnType,
-      final List<? extends FunctionArg> args,
-      final List<FunctionOption> optionsList) {
-    final SimpleExtension.ScalarFunctionVariant declaration =
+      String urn,
+      String key,
+      Type returnType,
+      List<? extends FunctionArg> args,
+      List<FunctionOption> optionsList) {
+    SimpleExtension.ScalarFunctionVariant declaration =
         extensions.getScalarFunction(SimpleExtension.FunctionAnchor.of(urn, key));
     return Expression.ScalarFunctionInvocation.builder()
         .declaration(declaration)
@@ -1626,11 +1604,8 @@ public class SubstraitBuilder {
    * @return a scalar function invocation expression
    */
   public Expression scalarFn(
-      final String urn,
-      final String key,
-      final Type returnType,
-      final List<? extends FunctionArg> args) {
-    final SimpleExtension.ScalarFunctionVariant declaration =
+      String urn, String key, Type returnType, List<? extends FunctionArg> args) {
+    SimpleExtension.ScalarFunctionVariant declaration =
         extensions.getScalarFunction(SimpleExtension.FunctionAnchor.of(urn, key));
     return Expression.ScalarFunctionInvocation.builder()
         .declaration(declaration)
--- a/core/src/main/java/io/substrait/dsl/SubstraitBuilder.java
+++ b/core/src/main/java/io/substrait/dsl/SubstraitBuilder.java
@@ -920,11 +920,11 @@ public class SubstraitBuilder {
   /**
    * Creates a 32-bit integer literal expression.
    *
-   * @param value the integer value
+   * @param v the integer value
    * @return a new {@link Expression.I32Literal}
    */
-  public Expression.I32Literal i32(final int value) {
-    return Expression.I32Literal.builder().value(value).build();
+  public Expression.I32Literal i32(int v) {
+    return Expression.I32Literal.builder().value(v).build();
   }
 
   /**
@@ -950,11 +950,11 @@ public class SubstraitBuilder {
   /**
    * Creates a 64-bit floating point literal expression.
    *
-   * @param value the double value
+   * @param v the double value
    * @return a new {@link Expression.FP64Literal}
    */
-  public Expression.FP64Literal fp64(final double value) {
-    return Expression.FP64Literal.builder().value(value).build();
+  public Expression.FP64Literal fp64(double v) {
+    return Expression.FP64Literal.builder().value(v).build();
   }
 
   /**
--- a/core/src/main/java/io/substrait/dsl/SubstraitBuilder.java
+++ b/core/src/main/java/io/substrait/dsl/SubstraitBuilder.java
@@ -883,7 +883,7 @@ public class SubstraitBuilder {
   }
 
   /**
-   * Create i16 literal.
+   * Create a 16-bit integer literal expression.
    *
    * @param v value to create
    * @return i16 instance
@@ -1067,7 +1067,7 @@ public class SubstraitBuilder {
    * options.
    *
    * @param condition the expression to match against
-   * @param options the list of possible vs to match
+   * @param options the list of possible values to match
    * @return a new {@link SingleOrList} expression
    */
   public Expression singleOrList(Expression condition, Expression... options) {
@@ -1075,10 +1075,10 @@ public class SubstraitBuilder {
   }
 
   /**
-   * Creates an IN predicate expression that checks if any needle vs exist in the haystack relation.
+   * Creates an IN predicate expression that checks if any needle values exist in the haystack relation.
    *
    * @param haystack the relation to search within
-   * @param needles the vs to search for
+   * @param needles the values to search for
    * @return a new {@link Expression.InPredicate}
    */
   public Expression.InPredicate inPredicate(Rel haystack, Expression... needles) {
@@ -1501,7 +1501,7 @@ public class SubstraitBuilder {
    * Creates a logical NOT expression that negates a boolean expression.
    *
    * <p>This is a convenience method that wraps the boolean NOT function from the Substrait standard
-   * library. The result is nullable to handle NULL input vs according to three-vd logic.
+   * library. The result is nullable to handle NULL input values according to three-valued logic.
    *
    * @param expression the boolean expression to negate
    * @return a scalar function invocation representing the logical NOT of the input expression
--- a/core/src/main/java/io/substrait/dsl/SubstraitBuilder.java
+++ b/core/src/main/java/io/substrait/dsl/SubstraitBuilder.java
@@ -12,6 +12,7 @@ import io.substrait.expression.Expression.Switch;
 import io.substrait.expression.Expression.SwitchClause;
 import io.substrait.expression.FieldReference;
 import io.substrait.expression.FunctionArg;
+import io.substrait.expression.FunctionOption;
 import io.substrait.expression.WindowBound;
 import io.substrait.extension.DefaultExtensionCatalog;
 import io.substrait.extension.SimpleExtension;
@@ -39,6 +40,7 @@ import io.substrait.relation.physical.NestedLoopJoin;
 import io.substrait.type.NamedStruct;
 import io.substrait.type.Type;
 import io.substrait.type.TypeCreator;
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.LinkedList;
@@ -74,6 +76,17 @@ public class SubstraitBuilder {
   private final SimpleExtension.ExtensionCollection extensions;
 
   /**
+   * Constructs a new SubstraitBuilder with the default extension collection.
+   *
+   * <p>The builder is initialized with {@link DefaultExtensionCatalog#DEFAULT_COLLECTION}, which
+   * includes standard Substrait functions for strings, arithmetic, comparison, datetime, and other
+   * operations.
+   */
+  public SubstraitBuilder() {
+    this(DefaultExtensionCatalog.DEFAULT_COLLECTION);
+  }
+
+  /**
    * Constructs a new SubstraitBuilder with the specified extension collection.
    *
    * @param extensions the extension collection used to resolve function declarations and other
@@ -81,6 +94,18 @@ public class SubstraitBuilder {
    */
   public SubstraitBuilder(SimpleExtension.ExtensionCollection extensions) {
     this.extensions = extensions;
+  }
+
+  /**
+   * Gets the default extension collection used by this builder.
+   *
+   * <p>This collection includes standard Substrait functions for strings, arithmetic, comparison,
+   * datetime, and other operations from {@link DefaultExtensionCatalog#DEFAULT_COLLECTION}.
+   *
+   * @return the ExtensionCollection containing standard Substrait functions
+   */
+  public SimpleExtension.ExtensionCollection getExtensions() {
+    return extensions;
   }
 
   // Relations
@@ -142,13 +167,32 @@ public class SubstraitBuilder {
     return aggregate(groupingsFn, measuresFn, Optional.of(remap), input);
   }
 
-  private Aggregate aggregate(
-      Function<Rel, List<Aggregate.Grouping>> groupingsFn,
-      Function<Rel, List<Aggregate.Measure>> measuresFn,
-      Optional<Rel.Remap> remap,
-      Rel input) {
-    List<Aggregate.Grouping> groupings = groupingsFn.apply(input);
-    List<Aggregate.Measure> measures = measuresFn.apply(input);
+  /**
+   * Creates an aggregate relation that groups and aggregates data from an input relation.
+   *
+   * <p>This method constructs a Substrait aggregate operation by applying grouping and measure
+   * functions to the input relation. The grouping function defines how rows are grouped together,
+   * while the measure function defines the aggregate computations (e.g., SUM, COUNT, AVG) to
+   * perform on each group.
+   *
+   * <p>The optional remap parameter allows reordering or filtering of output columns, which is
+   * useful for controlling the final schema of the aggregate result.
+   *
+   * @param groupingsFn a function that takes the input relation and returns a list of grouping
+   *     expressions defining how to partition the data
+   * @param measuresFn a function that takes the input relation and returns a list of aggregate
+   *     measures to compute for each group
+   * @param remap an optional remapping specification to reorder or filter output columns
+   * @param input the input relation to aggregate
+   * @return an Aggregate relation representing the grouping and aggregation operation
+   */
+  public Aggregate aggregate(
+      final Function<Rel, List<Aggregate.Grouping>> groupingsFn,
+      final Function<Rel, List<Aggregate.Measure>> measuresFn,
+      final Optional<Rel.Remap> remap,
+      final Rel input) {
+    final List<Aggregate.Grouping> groupings = groupingsFn.apply(input);
+    final List<Aggregate.Measure> measures = measuresFn.apply(input);
     return Aggregate.builder()
         .groupings(groupings)
         .measures(measures)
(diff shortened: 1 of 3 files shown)


core/src/main/java/io/substrait/dsl/SubstraitBuilder.java
```diff
@@ -853,24 +897,64 @@ public Expression.BoolLiteral bool(boolean v) {
...
   /**
    * Creates a 32-bit integer literal expression.
    *
-   * @param v the integer value
+   * @param value the integer value
    * @return a new {@link Expression.I32Literal}
    */
-  public Expression.I32Literal i32(int v) {
-    return Expression.I32Literal.builder().value(v).build();
+  public Expression.I32Literal i32(final int value) {
```
benbellick: Do we need to alter all of the signatures in this PR as final? It's adding a lot of diff to this PR and IMO it doesn't really add anything extra in terms of guarantees.

mbwhite: Adding `final` is something my editor is set to do, but happy to work with the project guidelines;  

`i32` and `fp64` are the only two existing methods I've altered. Everything else is new (or newly public)

benbellick: Sounds good. I'm on my phone so can't look at the rest of the file, but can we only use final for new methods if it's consistent with the other methods then? Do the other ones mark every method as final?

core/src/main/java/io/substrait/dsl/SubstraitBuilder.java
```diff
@@ -853,6 +897,26 @@ public Expression.BoolLiteral bool(boolean v) {
     return Expression.BoolLiteral.builder().value(v).build();
   }
 
+  /**
+   * Create i16 literal.
+   *
+   * @param value value to create
+   * @return i16 instance
+   */
+  public Expression.I8Literal i8(final int value) {
```
benbellick:
```suggestion
  public Expression.I8Literal i8(int value) {
```

And yeah looking at these files on my computer now. It does look like we consistently _don't_ use `final` in all of the other methods. Can we drop the usage of `final` in all of these places to keep our API consistent? That way we can have a separate conversation about introducing final if we want it. 

https://github.com/substrait-io/substrait-java/pull/629#discussion_r2585519436

core/src/main/java/io/substrait/dsl/SubstraitBuilder.java
```diff
@@ -142,13 +167,32 @@ public Aggregate aggregate(
...
+   *
+   * @param groupingsFn a function that takes the input relation and returns a list of grouping
+   *     expressions defining how to partition the data
+   * @param measuresFn a function that takes the input relation and returns a list of aggregate
+   *     measures to compute for each group
+   * @param remap an optional remapping specification to reorder or filter output columns
+   * @param input the input relation to aggregate
+   * @return an Aggregate relation representing the grouping and aggregation operation
+   */
+  public Aggregate aggregate(
```
benbellick, line 164: Sorry, I'm a bit confused here. Now that this method is made public, what is the difference between this one and the above one? Just that `Rel.Remap` is not required, but is instead optional? 

Can we instead just either a) have one `aggregate` builder in which the `Remap` is marked optional, or b) have two different impls, one in which the `Remap` is required, and the other in which there is no `Remap` argument? 

Otherwise, its a bit unclear to me which one users are supposed to use.

benbellick: I think my preference for now is just the one singular one where it is marked optional. This would technically be a breaking change but would be a minimal diff.

benbellick (changes requested):
Just a few comments. Thanks!

core/src/main/java/io/substrait/dsl/SubstraitBuilder.java
mbwhite: certainly will get on to that in the morning

mbwhite: @benbellick you're right - that addition isn't needed. my mistake

core/src/main/java/io/substrait/dsl/SubstraitBuilder.java
```diff
@@ -124,31 +149,31 @@ public Aggregate aggregate(
...
-      Function<Rel, List<Aggregate.Grouping>> groupingsFn,
-      Function<Rel, List<Aggregate.Measure>> measuresFn,
-      Optional<Rel.Remap> remap,
-      Rel input) {
-    List<Aggregate.Grouping> groupings = groupingsFn.apply(input);
-    List<Aggregate.Measure> measures = measuresFn.apply(input);
+      final Function<Rel, List<Aggregate.Grouping>> groupingsFn,
+      final Function<Rel, List<Aggregate.Measure>> measuresFn,
+      final Optional<Rel.Remap> remap,
+      final Rel input) {
```
benbellick: Still have `final` on these ones, can we get rid of those?

mbwhite: sorry missed those... changes on their way in

core/src/main/java/io/substrait/dsl/SubstraitBuilder.java
```diff
@@ -83,6 +96,18 @@ public SubstraitBuilder(SimpleExtension.ExtensionCollection extensions) {
     this.extensions = extensions;
   }
 
+  /**
+   * Gets the default extension collection used by this builder.
+   *
+   * <p>This collection includes standard Substrait functions for strings, arithmetic, comparison,
+   * datetime, and other operations from {@link DefaultExtensionCatalog#DEFAULT_COLLECTION}.
```
benbellick: This isn't necessarily true, right? Only if you use the parameter-less constructor do you get the default collection. If you use the builder with a parameter, then you could put whatever you want in I think.

mbwhite: you're quite correct - it's entirely what has be supplied..

nielspardon:
@benbellick since you started this review, do you have time for another look?

core/src/main/java/io/substrait/dsl/SubstraitBuilder.java
```diff
@@ -1439,6 +1501,79 @@ public Expression.ScalarFunctionInvocation or(Expression... args) {
...
+  /**
+   * Creates a logical NOT expression that negates a boolean expression.
+   *
+   * <p>This is a convenience method that wraps the boolean NOT function from the Substrait standard
+   * library. The result is nullable to handle NULL input values according to three-valued logic.
+   *
+   * @param expression the boolean expression to negate
+   * @return a scalar function invocation representing the logical NOT of the input expression
+   */
+  public Expression not(Expression expression) {
```
benbellick, line 1513: Slightly relevant issue I just found FYI https://github.com/substrait-io/substrait/issues/1041

core/src/main/java/io/substrait/dsl/SubstraitBuilder.java
```diff
@@ -1439,6 +1501,79 @@ public Expression.ScalarFunctionInvocation or(Expression... args) {
...
+   * @param key the function signature (e.g., "substring:str_i32_i32")
+   * @param returnType the return type of the function
+   * @param args the function arguments
+   * @param optionsList the function options controlling behavior
+   * @return a scalar function invocation expression
+   */
+  public Expression scalarFn(
+      String urn,
+      String key,
+      Type returnType,
```
benbellick, line 1564: Thinking out loud here, but does this builder validate that the returnType is actually possible for this function?

core/src/main/java/io/substrait/dsl/SubstraitBuilder.java
```diff
@@ -1459,6 +1594,26 @@ public Expression.ScalarFunctionInvocation scalarFn(
         .build();
   }
 
+  /**
+   * Creates a scalar function invocation with function options.
```
benbellick, line 1598: This says 
> with function options

but only the above method accepts options AFAICT via `optionsList`, and this one does not. Maybe we just get rid of this whole one to keep things simple?

core/src/main/java/io/substrait/dsl/SubstraitBuilder.java
```diff
@@ -1532,6 +1687,22 @@ public Plan plan(Plan.Root root) {
...
+   * the plan. This is the final step in building a complete Substrait plan that can be serialized
+   * and executed by a Substrait consumer.
+   *
+   * @param input the root relational expression of the query plan
+   * @param names the ordered list of output column names corresponding to the input relation's
+   *     output schema
+   * @return a new {@link Plan}
+   */
+  public Plan.Root root(final Rel input, final List<String> names) {
+    return Plan.Root.builder().input(input).names(names).build();
```
benbellick, line 1703: Also curious if this one validates that the length of names matches the schema of input.

core/src/test/java/io/substrait/dsl/SubstraitBuilderTest.java
```diff
@@ -0,0 +1,265 @@
...
+      final List<FieldReference> refs = builder.fieldReferences(scan, 0, 1);
+      assertEquals(2, refs.size());
+    }
+
+    @Test
+    void testCast() {
+      final Expression.I32Literal input = builder.i32(1);
+      final Expression cast = builder.cast(input, R.I64);
+      assertNotNull(cast);
+      assertTrue(cast instanceof Expression.Cast);
```
benbellick, line 79: Might be better to actually check that the type of this expression is now `R.I64`.

core/src/main/java/io/substrait/dsl/SubstraitBuilder.java
```diff
@@ -853,6 +875,26 @@ public Expression.BoolLiteral bool(boolean v) {
...
   }
 
+  /**
+   * Create i16 literal.
+   *
+   * @param value value to create
+   * @return i16 instance
+   */
+  public Expression.I8Literal i8(int value) {
+    return Expression.I8Literal.builder().value(value).build();
```
benbellick, line 885: Docs say this is making an `i16` but it is making an `i8`.

core/src/main/java/io/substrait/dsl/SubstraitBuilder.java
benbellick: Also `and` and `or` above return `Expression.ScalarFunctionInvocation`, but this one returns the more general `Expression`. Can we return `Expression.ScalarFunctionInvocation` to be consistent?

benbellick: Also the nullability should be assumed to be `MIRROR` rather than `DECLARED_OUTPUT` to be consistent with `and` and `or` above. So I think we should check nullability here and only mark the output as nullable if the input is.

core/src/main/java/io/substrait/dsl/SubstraitBuilder.java
```diff
@@ -1439,6 +1501,79 @@ public Expression.ScalarFunctionInvocation or(Expression... args) {
...
+  public Expression isNull(Expression expression) {
+
+    final List<Expression> args = new ArrayList<>();
+    args.add(expression);
+
+    return this.scalarFn(
+        DefaultExtensionCatalog.FUNCTIONS_COMPARISON,
+        "is_null:any",
+        TypeCreator.REQUIRED.BOOLEAN,
+        args,
```
benbellick, line 1544: nit: could just simplify this to the below and drop the above construction of an ArrayList
```suggestion
        List.of(expression),
```

core/src/main/java/io/substrait/dsl/SubstraitBuilder.java
```diff
@@ -1532,6 +1687,22 @@ public Plan plan(Plan.Root root) {
...
+   * Creates a Plan.Root, which is the top-level container for a Substrait query plan.
+   *
+   * <p>The {@link Plan} wraps a relational expression tree and associates output column names with
+   * the plan. This is the final step in building a complete Substrait plan that can be serialized
+   * and executed by a Substrait consumer.
+   *
+   * @param input the root relational expression of the query plan
+   * @param names the ordered list of output column names corresponding to the input relation's
+   *     output schema
+   * @return a new {@link Plan}
```
benbellick, line 1700:
```suggestion
   * @return a new {@link Plan.root}
```

core/src/test/java/io/substrait/dsl/SubstraitBuilderTest.java
```diff
@@ -0,0 +1,265 @@
...
+      assertNotNull(count);
+      assertNotNull(sum);
+
+      final Aggregate.Grouping grouping = builder.grouping(scan, 1);
+      assertNotNull(grouping);
+      assertEquals(1, grouping.getExpressions().size());
+
+      final AggregateFunctionInvocation afi1 =
+          builder.aggregateFn(
+              "extension:io.substrait:functions_aggregate_generic",
```
benbellick, line 201: nit:

```suggestion
              DefaultExtensionCatalog.FUNCTIONS_AGGREGATE_GENERIC,
```

core/src/test/java/io/substrait/dsl/SubstraitBuilderTest.java
```diff
@@ -0,0 +1,265 @@
...
+      assertEquals(Join.JoinType.INNER, join.getJoinType());
+    }
+
+    @Test
+    void testFetchAndLimit() {
+      final NamedScan scan = createSimpleScan();
+      Fetch limit = builder.limit(10, scan);
+      assertEquals(10, limit.getCount().getAsLong());
+      assertEquals(0, limit.getOffset());
+      limit = builder.limit(10, Remap.of(Arrays.asList(new Integer[] {0, 1})), scan);
```
benbellick, line 136: nit:
```suggestion
      limit = builder.limit(10, Remap.of(List.of(0, 1)), scan);
```

core/src/main/java/io/substrait/dsl/SubstraitBuilder.java
```diff
@@ -863,6 +905,26 @@ public Expression.I32Literal i32(int v) {
     return Expression.I32Literal.builder().value(v).build();
   }
 
+  /**
+   * Creates a 64-bit integer literal expression.
+   *
+   * @param value value to create
+   * @return i64 instance
+   */
+  public Expression.I64Literal i64(long value) {
```
benbellick, line 914: nit: all of the other existing methods use `v` for the parameter name.  Should we do the same for all of the new ones to be consistent?

benbellick (commented):
Left a few more comments. Thanks!

core/src/main/java/io/substrait/dsl/SubstraitBuilder.java
nielspardon: I don't think we are doing that today. Should we add such a validation check to `Plan.Root`?

benbellick: I added this in #812

core/src/main/java/io/substrait/dsl/SubstraitBuilder.java
mbwhite: It does some validation certainly; To be consistent with the status of the SubstraitBuilder I think the validation should be done in the lower level builder

mbwhite:
Sorry for the delay here.. other work commitments

core/src/main/java/io/substrait/dsl/SubstraitBuilder.java
```diff
@@ -846,27 +868,67 @@ private Sort sort(
   /**
    * Creates a boolean literal expression.
    *
-   * @param v the boolean value
+   * @param v the boolean v
```
benbellick: why are we changing `value` to `v` here? I agree that the _parameter_ should be called `v` to be consistent everywhere. But the docs should refer to the `value`.

benbellick: There are a few other places below where this would need to be fixed. Thanks!

core/src/main/java/io/substrait/dsl/SubstraitBuilder.java
```diff
@@ -124,25 +146,25 @@ public Aggregate aggregate(
...
+   * useful for controlling the final schema of the aggregate result.
+   *
+   * @param groupingsFn a function that takes the input relation and returns a list of grouping
+   *     expressions defining how to partition the data
+   * @param measuresFn a function that takes the input relation and returns a list of aggregate
+   *     measures to compute for each group
+   * @param remap an optional remapping specification to reorder or filter output columns
    * @param input the input relation to aggregate
-   * @return a new {@link Aggregate} relation
+   * @return an Aggregate relation representing the grouping and aggregation operation
```
benbellick: Why drop the link here? seems useful IMO

core/src/main/java/io/substrait/dsl/SubstraitBuilder.java
```diff
@@ -124,25 +146,25 @@ public Aggregate aggregate(
...
-   * @param measuresFn function to derive the measures from the input relation
-   * @param remap the output field remapping specification
+   * <p>This method constructs a Substrait aggregate operation by applying grouping and measure
+   * functions to the input relation. The grouping function defines how rows are grouped together,
+   * while the measure function defines the aggregate computations (e.g., SUM, COUNT, AVG) to
+   * perform on each group.
+   *
+   * <p>The optional remap parameter allows reordering or filtering of output columns, which is
+   * useful for controlling the final schema of the aggregate result.
+   *
```
benbellick:
```suggestion
```

Personally, I think that the `@param` comment below is clear enough. I'd rather the core docs here focus on aggregate specific things, rather than general relation things. What do you think?

core/src/main/java/io/substrait/dsl/SubstraitBuilder.java
```diff
@@ -1439,6 +1500,75 @@ public Expression.ScalarFunctionInvocation or(Expression... args) {
     return scalarFn(DefaultExtensionCatalog.FUNCTIONS_BOOLEAN, "or:bool", outputType, args);
   }
 
+  /**
+   * Creates a logical NOT expression that negates a boolean expression.
+   *
+   * <p>This is a convenience method that wraps the boolean NOT function from the Substrait standard
+   * library. The result is nullable to handle NULL input vs according to three-vd logic.
+   *
```
benbellick, line 1506:
```suggestion
```

To be consistent with the rest of the file, I don't think we need to declare that this is a convenience helper (the whole file is convenience helpers 😄)

core/src/main/java/io/substrait/dsl/SubstraitBuilder.java
```diff
@@ -1439,6 +1500,75 @@ public Expression.ScalarFunctionInvocation or(Expression... args) {
...
+        DefaultExtensionCatalog.FUNCTIONS_BOOLEAN, "not:bool", outputType, expression);
+  }
+
+  /**
+   * Creates a null-check expression that tests whether an expression is null.
+   *
+   * <p>This is a convenience method that wraps the is_null function from the Substrait comparison
+   * function library. The function evaluates the input expression and returns true if it is null,
+   * false otherwise. This is commonly used in conditional logic and filtering operations.
+   *
```
benbellick:
```suggestion
```

core/src/main/java/io/substrait/dsl/SubstraitBuilder.java
```diff
@@ -1439,6 +1500,75 @@ public Expression.ScalarFunctionInvocation or(Expression... args) {
...
+   * <p>The return type is always a required (non-nullable) boolean, as the null check itself always
+   * produces a definite true/false result.
```
benbellick, line 1521:
```suggestion
   * <p>The return type is always a required (non-nullable) boolean.
```
nit: feel free to ignore if you disagree. I think the comment is perfectly clear without the explanation.

mbwhite:
@benbellick my phone has told me there's a comment but I can't see in the PR... to answer your question - over enthaustic search and replace.

core/src/test/java/io/substrait/dsl/SubstraitBuilderTest.java
```diff
@@ -0,0 +1,258 @@
...
+
+      final List<FieldReference> refs = builder.fieldReferences(scan, 0, 1);
+      assertEquals(2, refs.size());
+    }
+
+    @Test
+    void testCast() {
+      final Expression.I32Literal input = builder.i32(1);
+      assertEquals(R.I32, input.getType());
+    }
```
benbellick, line 82: I don't think this test uses `cast` at all.

core/src/test/java/io/substrait/dsl/SubstraitBuilderTest.java
```diff
@@ -0,0 +1,258 @@
...
+    assertInstanceOf(SimpleExtension.ExtensionCollection.class, this.builder.getExtensions());
+  }
+
+  @Nested
+  @DisplayName("Literal and Expression Tests")
+  class ExpressionTests {
+
+    @Test
+    void testLiterals() {
+      assertEquals(true, builder.bool(true).value());
```
benbellick, line 53: nit
```suggestion
      assertTrue(builder.bool(true).value());
```

core/src/test/java/io/substrait/dsl/SubstraitBuilderTest.java
benbellick, line 1: I'm usually a strong advocate for tests, but many of these helper implementations are shallow enough that the current tests feel like they mostly duplicate the implementation. I'm not sure the broad `assertNotNull` coverage provides much value. I think this file might be stronger with one focused test that demonstrates how the builder APIs are used in practice and asserts the resulting plan/POJO, rather than trying to unit test every helper individually. What do you think about replacing these with something like this?


```java
@Test
void testBuildsNonTrivialPlan() {
  final List<String> tableName = List.of("orders");
  final List<String> columnNames = List.of("customer_id", "status", "amount");
  final List<Type> columnTypes = List.of(R.I32, R.STRING, R.I32);
  final Remap remap = Remap.of(List.of(0, 2));
  final List<String> outputNames = List.of("customer_id", "row_count");

  final NamedScan scan = builder.namedScan(tableName, columnNames, columnTypes);
  final Filter filtered =
      builder.filter(rel -> builder.not(builder.isNull(builder.fieldReference(rel, 2))), scan);
  final Project projected =
      builder.project(
          rel -> List.of(builder.fieldReference(rel, 0), builder.i64(1L), builder.fp32(1.5f)),
          filtered);
  final Aggregate aggregated =
      builder.aggregate(
          rel ->
              List.of(
                  builder.grouping(builder.fieldReference(rel, 0)),
                  builder.grouping(builder.fieldReference(rel, 1))),
          rel -> List.of(builder.countStar()),
          Optional.of(remap),
          projected);

  final Plan actual = builder.plan(builder.root(aggregated, outputNames));

  final NamedScan expectedScan =
      NamedScan.builder()
          .names(tableName)
          .initialSchema(
              NamedStruct.of(
                  columnNames,
                  Type.Struct.builder().addAllFields(columnTypes).nullable(false).build()))
          .build();
  final Filter expectedFiltered =
      Filter.builder()
          .input(expectedScan)
          .condition(
              builder.not(builder.isNull(builder.fieldReference(expectedScan, 2))))
          .build();
  final Project expectedProjected =
      Project.builder()
          .input(expectedFiltered)
          .addExpressions(builder.fieldReference(expectedFiltered, 0))
          .addExpressions(builder.i64(1L))
          .addExpressions(builder.fp32(1.5f))
          .build();
  final Aggregate expectedAggregated =
      Aggregate.builder()
          .input(expectedProjected)
          .addGroupings(builder.grouping(builder.fieldReference(expectedProjected, 0)))
          .addGroupings(builder.grouping(builder.fieldReference(expectedProjected, 1)))
          .addMeasures(builder.countStar())
          .remap(Optional.of(remap))
          .build();
  final Plan expected =
      Plan.builder()
          .addRoots(Plan.Root.builder().input(expectedAggregated).names(outputNames).build())
          .build();

  assertEquals(expected, actual);
}

```

It exercises a bunch of the new APIs and demonstrates how it is used in practice. Let me know what you think!

benbellick (approved):
Left a few comments but they are mostly about docs and the tests. Otherwise, I think its good. Thanks for your patience with my reviews!

benbellick:
> @benbellick my phone has told me there's a comment but I can't see in the PR... to answer your question - over enthaustic search and replace.

@mbwhite Sorry, I was trying to do a review but accidentally left that first comment by mistake.

core/src/main/java/io/substrait/dsl/SubstraitBuilder.java
mbwhite: No strong opinions either way to be honest.. .:-)

core/src/test/java/io/substrait/dsl/SubstraitBuilderTest.java
mbwhite: does now :-)

core/src/test/java/io/substrait/dsl/SubstraitBuilderTest.java
mbwhite: I do see your argument; 'has much being gained' is a valid question. My view is to be more cautious – keep the tests and certainly add an example. Testing that confirms the foundations are solid is going to become even more important, I feel, in the coming years.

mbwhite:
@benbellick @nielspardon  update with the latest comments Ben. :-)

I'll take the example suggestion and do a large example PR

core/src/main/java/io/substrait/dsl/SubstraitBuilder.java
```diff
@@ -853,6 +872,26 @@ public Expression.BoolLiteral bool(boolean v) {
     return Expression.BoolLiteral.builder().value(v).build();
   }
 
+  /**
+   * Create i8 literal.
```
nielspardon: why are we not using the same comment as for 32-bit and 64-bit variants?
```suggestion
   * Creates an 8-bit integer literal expression.
```

core/src/main/java/io/substrait/dsl/SubstraitBuilder.java
```diff
@@ -853,6 +872,26 @@ public Expression.BoolLiteral bool(boolean v) {
...
+   *
+   * @param v value to create
+   * @return i8 instance
+   */
+  public Expression.I8Literal i8(int v) {
+    return Expression.I8Literal.builder().value(v).build();
+  }
+
+  /**
+   * Create i16 literal.
```
nielspardon: why are we not using the same comment as for 32-bit and 64-bit variants?
```suggestion
   * Creates a 16-bit integer literal expression.
```

core/src/main/java/io/substrait/dsl/SubstraitBuilder.java
```diff
@@ -1008,19 +1067,18 @@ public IfClause ifClause(Expression condition, Expression then) {
    * options.
    *
    * @param condition the expression to match against
-   * @param options the list of possible values to match
+   * @param options the list of possible vs to match
```
nielspardon: this feels like a search'n'replace gone wrong

core/src/main/java/io/substrait/dsl/SubstraitBuilder.java
```diff
@@ -1008,19 +1067,18 @@ public IfClause ifClause(Expression condition, Expression then) {
...
    * @return a new {@link SingleOrList} expression
    */
   public Expression singleOrList(Expression condition, Expression... options) {
     return SingleOrList.builder().condition(condition).addOptions(options).build();
   }
 
   /**
-   * Creates an IN predicate expression that checks if any needle values exist in the haystack
-   * relation.
+   * Creates an IN predicate expression that checks if any needle vs exist in the haystack relation.
```
nielspardon: same here: this feels like a search'n'replace gone wrong

core/src/main/java/io/substrait/dsl/SubstraitBuilder.java
```diff
@@ -1008,19 +1067,18 @@ public IfClause ifClause(Expression condition, Expression then) {
...
    *
    * @param haystack the relation to search within
-   * @param needles the values to search for
+   * @param needles the vs to search for
```
nielspardon: same here: this feels like a search'n'replace gone wrong

core/src/main/java/io/substrait/dsl/SubstraitBuilder.java
```diff
@@ -1439,6 +1497,71 @@ public Expression.ScalarFunctionInvocation or(Expression... args) {
     return scalarFn(DefaultExtensionCatalog.FUNCTIONS_BOOLEAN, "or:bool", outputType, args);
   }
 
+  /**
+   * Creates a logical NOT expression that negates a boolean expression.
+   *
+   * <p>This is a convenience method that wraps the boolean NOT function from the Substrait standard
+   * library. The result is nullable to handle NULL input vs according to three-vd logic.
```
nielspardon: same here: this feels like a search'n'replace gone wrong `three-vd`

nielspardon (changes requested):

core/src/main/java/io/substrait/dsl/SubstraitBuilder.java
```diff
@@ -1543,12 +1666,28 @@ public Plan.Root root(Rel rel) {
    * Creates a plan from a plan root.
    *
    * @param root the plan root
-   * @return a new {@link Plan}
+   * @return a new {@link Plan.Root}
```
nielspardon: this feels like a wrong modification. the return type of this method is still a `Plan` and not a `Plan.Root`
```suggestion
   * @return a new {@link Plan}
```

core/src/main/java/io/substrait/dsl/SubstraitBuilder.java
nielspardon: this one should have been changed to `Plan.Root`

nielspardon (approved):
LGTM

PR #775, opened 2026-03-25
build: add workaround for VS Code detection of immutables classes

nielspardon:
If you are a user of a VS Code based IDE then you may have been noticing that the IDE started having trouble recognizing the immutables generated classes. This is a known issue with VS Code / Eclipse and one workaround is to add additional Gradle configuration using the Eclipse plugin: https://github.com/redhat-developer/vscode-java/issues/2981

This PR adds the corresponding Gradle configuration for the Gradle modules using `annotationProcessor`.

Also fixes the .gitignore for Eclipse / VS Code `bin` directory since the current pattern is too specific missing `bin` directories in subfolders.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,10 +7,11 @@ gen
 *.iml
 out/**
 *.iws
-.vscode
 .pmdCache
 
-*/bin
+# VS Code ignores
+.vscode
+bin
 .metals
 .bloop
 .project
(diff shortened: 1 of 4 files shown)


bestbeforetoday (approved):
I don't know if this is the best solution but it does seem to work for me when loading substrait-java in VS Code... After building the project at the command-line anyway, although I'm not sure if that was the trigger, restarting VS Code, or something else. 

Thank you for figuring out a smoother experience for VS Code users!

PR #777, opened 2026-03-25
docs(isthmus): javadoc for ContentProvider & subclasses

mbwhite:
ContentProvider JavaDoc

### DIFF
--- a/isthmus/src/main/java/io/substrait/isthmus/DynamicConverterProvider.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/DynamicConverterProvider.java
@@ -12,22 +12,72 @@ import org.apache.calcite.sql.SqlOperator;
 import org.apache.calcite.sql.SqlOperatorTable;
 import org.apache.calcite.sql.util.SqlOperatorTables;
 
+/**
+ * A converter provider that dynamically generates SQL operators and call converters from Substrait
+ * extension functions.
+ *
+ * <p>This class extends {@link ConverterProvider} to support dynamic function extensions that are
+ * not part of the standard Substrait function set. It automatically generates SQL operators for
+ * dynamic extensions and creates appropriate call converters to handle them during query
+ * translation.
+ *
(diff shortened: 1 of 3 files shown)


isthmus/src/main/java/io/substrait/isthmus/DynamicConverterProvider.java
```diff
@@ -12,22 +12,68 @@
...
+ *
+ * <p>This class extends {@link ConverterProvider} to support dynamic function extensions that are
+ * not part of the standard Substrait function set. It automatically generates SQL operators for
+ * dynamic extensions and creates appropriate call converters to handle them during query
+ * translation.
+ *
+ * <p>Dynamic extensions are identified as functions that are not present in the known function
+ * mappings ({@link FunctionMappings#SCALAR_SIGS}). These functions are then converted to SQL
+ * operators and integrated into the operator table and call converter list.
+ */
```
bestbeforetoday, line 31: I _think_ the primary intent for this converter provider implementation is to support user-defined functions. The description doesn't contradict that use-case but it also doesn't make it very clear — at least to me — when this would be the converter provider they should pick.

It might be good to incorporate something similar to these words (lifted from ExtensionUtils.getDynamicExtensions(), which this converter provider uses):

> A dynamic extension is a user-defined function (UDF) that is not part of the standard Substrait function catalog. These are custom functions that users define and provide at runtime, extending the built-in function set with domain-specific or application-specific operations.

mbwhite: I've used this together with a yaml definition to support functions that already exist in the engine (sparksql) but weren't in Subtrait.

User defined but not implemented :-)

Though I'm not sure know how this class relates to the 'AutomaticDynamicFunctionMappingCOnverterProvider'

mbwhite: though I've added in the paragraph as that is useful.

PR #778, opened 2026-03-26
docs(core): add javadoc, excluding Expression

bestbeforetoday:
This change adds missing Javadoc to address build warnings in all classes except for io.substrait.expression.Expression.

### DIFF
--- a/core/src/main/java/io/substrait/relation/Aggregate.java
+++ b/core/src/main/java/io/substrait/relation/Aggregate.java
@@ -12,11 +12,25 @@ import java.util.stream.Collectors;
 import java.util.stream.Stream;
 import org.immutables.value.Value;
 
+/**
+ * Represents an aggregate relation that groups input rows and computes aggregate functions.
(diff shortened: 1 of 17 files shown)


core/src/main/java/io/substrait/relation/BiRel.java
```diff
@@ -3,10 +3,21 @@
 import java.util.Arrays;
 import java.util.List;
 
+/** Abstract base class for binary relations that have exactly two input relations. */
```
mbwhite, line 6: can you have a binary relation without two inputs :-)

core/src/main/java/io/substrait/relation/Aggregate.java
```diff
@@ -71,26 +85,58 @@ public <O, C extends VisitationContext, E extends Exception> O accept(
     return visitor.visit(this, context);
   }
 
+  /** Represents a grouping set - a set of expressions to group by. */
```
andrew-coleman, line 88: ... by which to group :)

PR #779, opened 2026-03-26
docs(core): io.substrait.expression.Expression

bestbeforetoday:
Resolve Javadoc warnings for the io.substrait.expression.Expression class, excluding "no comment" warnings.

### DIFF
--- a/core/src/main/java/io/substrait/expression/Expression.java
+++ b/core/src/main/java/io/substrait/expression/Expression.java
@@ -13,9 +13,18 @@ import java.util.Map;
 import java.util.UUID;
 import org.immutables.value.Value;
 
+/**
+ * Represents a Substrait expression that can be evaluated to produce a value. Expressions include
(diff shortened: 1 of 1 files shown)


core/src/main/java/io/substrait/expression/Expression.java
```diff
@@ -1189,8 +1258,10 @@ public static ImmutableExpression.SortField.Builder builder() {
     }
   }
 
+  /** Base interface for subquery expressions. */
   interface Subquery extends Expression {}
 
+  /** Represents a set predicate subquery (EXISTS or UNIQUE). */
```
mbwhite: might in theory be something else in addition to EXITS and UNIQUE?

bestbeforetoday: Fair point. It might evolve so being more generic might be better here.

PR #780, opened 2026-03-26
fix(isthmus): std_dev, variance function mappings

### LINKED ISSUE #803: use enum arg function signatures for `std_dev` and `variance` functions instead of function option signatures

### LINKED ISSUE #807: deprecate `std_dev` and `variance` functions using function options

nielspardon:
This PR implements the bidirectional conversion between Calcite and Substrait for the
statistical aggregate functions standard deviation and variance (`STDDEV_POP`,
`STDDEV_SAMP`, `VAR_POP`, `VAR_SAMP`).

### Problem

Substrait uses a single function name for both the population and sample variants:

- `std_dev` for both `STDDEV_POP` and `STDDEV_SAMP`
- `variance` for both `VAR_POP` and `VAR_SAMP`

The population (n denominator) vs. sample (n-1 denominator) distinction is captured by a
`distribution` value (`POPULATION` or `SAMPLE`). Previously these functions were not mapped,
so the existing TPC-DS test cases using them were silently mis-mapped to `AVG` (Calcite
represents these statistical functions with `SqlAvgAggFunction`).

### Solution

This uses the **non-deprecated** signatures introduced in substrait-io/substrait#1011
(Substrait v0.87.0), which carry the distinction as an **enum function argument** (the
leading `distribution` argument, e.g. `std_dev:req_fp64`), rather than the deprecated
function-option form (substrait-io/substrait#1019). 

Resolves #803
Resolves #807 

Since the input arguments are cast to `FP64` when necessary, the integer-based signatures
proposed in substrait-io/substrait#1012 are not required.

**Calcite → Substrait:**

- Added function mappings in `FunctionMappings` for all four statistical operators.
- `AggregateFunctionConverter` synthesizes the leading `distribution` enum operand based on
  the Calcite `SqlKind`, so the generic function matcher resolves the enum-arg variant and
  builds the `EnumArg` automatically (no bespoke option plumbing).
- Statistical inputs are cast to `FP64` where necessary.

**Substrait → Calcite:**

- `FunctionConverter.getSqlOperatorFromSubstraitFunc` disambiguates the population/sample
  operator from the `distribution` enum argument.
- `SubstraitRelNodeConverter` and `PreCalciteAggregateValidator` skip the non-value enum
  argument when building Calcite aggregate operands.

**DSL & shared enum:**

- A shared `StatisticalDistribution` enum (in `:core`) is the single source of truth for the
  `SAMPLE` / `POPULATION` values used by both the DSL builder and isthmus.
- `SubstraitBuilder` gains `stddevPopulation`, `stddevSample`, `variancePopulation`, and
  `varianceSample` convenience methods.

### Non-floating-point inputs

`std_dev` / `variance` only define `fp32` / `fp64` signatures, so a statistical aggregate
over an integer (or other non-fp) column is rewritten in `SubstraitRelVisitor` to cast the
argument to `fp64` (appending a cast column so other aggregates over the same column are
unaffected) and cast the result back to the type Calcite inferred. The rewrite is
idempotent, so the converted plan is stable under further round trips.

### Testing

- `AggregationFunctionsTest` exercises full round trips (POJO ⇄ proto and Substrait ⇄
  Calcite) for all four functions, with and without grouping.
- `StatisticalFunctionTest` verifies the SQL round trip, asserts that each SQL operator
  maps to the enum-arg signature (`std_dev:req_fp64` etc.) with the correct `distribution`
  `EnumArg` and no function options, and covers non-fp (integer) inputs — including a
  column shared with a non-statistical aggregate, and with grouping.

🤖 Generated with AI

### DIFF
--- a/core/src/main/java/io/substrait/dsl/SubstraitBuilder.java
+++ b/core/src/main/java/io/substrait/dsl/SubstraitBuilder.java
@@ -1,6 +1,7 @@
 package io.substrait.dsl;
 
 import io.substrait.expression.AggregateFunctionInvocation;
+import io.substrait.expression.EnumArg;
 import io.substrait.expression.Expression;
 import io.substrait.expression.Expression.Cast;
 import io.substrait.expression.Expression.FailureBehavior;
@@ -13,6 +14,7 @@ import io.substrait.expression.Expression.SwitchClause;
 import io.substrait.expression.FieldReference;
 import io.substrait.expression.FunctionArg;
 import io.substrait.expression.FunctionOption;
+import io.substrait.expression.StatisticalDistribution;
 import io.substrait.expression.WindowBound;
 import io.substrait.extension.DefaultExtensionCatalog;
 import io.substrait.extension.SimpleExtension;
@@ -1371,7 +1373,7 @@ public class SubstraitBuilder {
    * @param input the input relation containing the field
    * @param field the zero-based index of the field to aggregate
    * @return an aggregate measure computing population standard deviation with
-   *     distribution=POPULATION option
+   *     distribution=POPULATION enum argument
    */
   public Aggregate.Measure stddevPopulation(Rel input, int field) {
     return stddevPopulation(fieldReference(input, field));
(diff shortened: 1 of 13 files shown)


mbwhite (approved):
LGTM - will be good to get these cleaned-up

nielspardon:
rebased on latest main and updated the PR to use the new enum arg function variants instead of the deprecated function option variants

andrew-coleman (approved):

PR #782, opened 2026-03-27
feat(core): add MaskExpression POJO and ReadRel projection support

flex-seongmin:
## Summary

Implement the `MaskExpression` POJO layer for `ReadRel` projection support in substrait-java.

The Substrait protobuf spec defines `Expression.MaskExpression` as a field on `ReadRel` (field 4: `projection`), but the Java POJO layer had only a TODO comment and no actual implementation. This means any `ReadRel` projection information was silently dropped during proto-to-POJO conversion, and POJO-constructed plans could not express column projection at the read level.

This change closes that gap by introducing a full POJO model for `MaskExpression` and wiring it through the existing conversion infrastructure.

## Changes

- **`MaskExpression.java`** (new) -- `@Value.Enclosing` interface with 11 `@Value.Immutable` inner types (`MaskExpr`, `Select`, `StructSelect`, `StructItem`, `ListSelect`, `ListSelectItem`, `ListElement`, `ListSlice`, `MapSelect`, `MapKey`, `MapKeyExpression`) matching the proto definition 1:1
- **`MaskExpressionProtoConverter.java`** (new) -- Dedicated bidirectional converter between `io.substrait.proto.Expression.MaskExpression` and the POJO types, keeping proto FQN references isolated from the relation converters
- **`AbstractReadRel.java`** -- Replaced TODO comment with `Optional<MaskExpression.MaskExpr> getProjection()`
- **`ProtoRelConverter.java`** -- Added `optionalMaskExpression()` helper; all four `ReadRel` subtypes (`NamedScan`, `LocalFiles`, `VirtualTableScan`, `ExtensionTable`) now populate projection during proto-to-POJO conversion
- **`RelProtoConverter.java`** -- All four `visit(ReadRel)` methods now serialize projection back to proto via `MaskExpressionProtoConverter.toProto()`

## Testing

Added 11 roundtrip tests in `ReadRelRoundtripTest` covering:

- Simple flat column projection (`StructSelect` only)
- Nested struct projection (`StructItem` with child `Select.ofStruct`)
- List element and slice selection (`ListSelect`)
- List selection with nested child (`ListSelect` + child `StructSelect`)
- Map key selection (`MapSelect` with `MapKey`)
- Map key expression / wildcard selection (`MapKeyExpression`)
- Map selection with nested child (`MapSelect` + child `StructSelect`)
- `maintainSingularStruct` flag
- Projection combined with filter
- Projection on `VirtualTableScan` and `LocalFiles`

### DIFF
--- a/core/src/main/java/io/substrait/expression/MaskExpression.java
+++ b/core/src/main/java/io/substrait/expression/MaskExpression.java
@@ -19,10 +19,10 @@ import org.immutables.value.Value;
 public interface MaskExpression {
 
   /**
-    * The top-level struct selection describing which fields to include.
-    *
-    * @return the top-level struct selection
-    */
+   * The top-level struct selection describing which fields to include.
+   *
+   * @return the top-level struct selection
+   */
   StructSelect getSelect();
 
   /**


nielspardon:
Hi @flex-seongmin, thanks for your contribution. I enabled the Github Action workflow and it looks like your PR has some formatting issues. Running `./gradlew spotlessApply` should automatically fix these for you.

core/src/main/java/io/substrait/expression/MaskExpression.java
```diff
@@ -0,0 +1,230 @@
...
+    }
+
+    static Select ofList(ListSelect listSelect) {
+      return builder().list(listSelect).build();
+    }
+
+    static Select ofMap(MapSelect mapSelect) {
+      return builder().map(mapSelect).build();
+    }
+  }
```
nielspardon, line 67: This seems to be mirroring the generated proto Java classes. What if we would turn this into an empty base interface without the `ofStruct`, `ofList`, `ofMap` methods. Then `StructSelect`, `ListSelect` and `MapSelect` can extend that interface and you can use a visitor pattern to process the different types. This would fit more with the existing code like e.g. [Expression](https://github.com/substrait-io/substrait-java/blob/370f0713b01353f95298768563a8083840707fce/core/src/main/java/io/substrait/expression/Expression.java) with its [ExpressionVisitor](https://github.com/substrait-io/substrait-java/blob/main/core/src/main/java/io/substrait/expression/ExpressionVisitor.java).

```suggestion
  interface Select {
  /**
   * Accepts a visitor to traverse this mask expression.
   *
   * @param <R> the return type of the visitor
   * @param <C> the context type
   * @param <E> the exception type that may be thrown
   * @param visitor the visitor to accept
   * @param context the visitation context
   * @return the result of the visitation
   * @throws E if an error occurs during visitation
   */
  <R, C extends VisitationContext, E extends Throwable> R accept(
      MaskExpressionVisitor<R, C, E> visitor, C context) throws E;
  }
```

```java
interface StructSelect extends Select {
...
    @Override
    public <R, C extends VisitationContext, E extends Throwable> R accept(
        MaskExpressionVisitor<R, C, E> visitor, C context) throws E {
      return visitor.visit(this, context);
    }
}

interface ListSelect extends Select {
...
    @Override
    public <R, C extends VisitationContext, E extends Throwable> R accept(
        MaskExpressionVisitor<R, C, E> visitor, C context) throws E {
      return visitor.visit(this, context);
    }
}

interface MapSelect extends Select {
...
    @Override
    public <R, C extends VisitationContext, E extends Throwable> R accept(
        MaskExpressionVisitor<R, C, E> visitor, C context) throws E {
      return visitor.visit(this, context);
    }
}
```

core/src/main/java/io/substrait/expression/MaskExpression.java
```diff
@@ -0,0 +1,230 @@
...
+     */
+    @Value.Default
+    default boolean getMaintainSingularStruct() {
+      return false;
+    }
+
+    static ImmutableMaskExpression.MaskExpr.Builder builder() {
+      return ImmutableMaskExpression.MaskExpr.builder();
+    }
+  }
```
nielspardon: You could also merged this into the top-level `MaskExpression` interface which would be similar to what we are doing in [`Expression`](https://github.com/substrait-io/substrait-java/blob/370f0713b01353f95298768563a8083840707fce/core/src/main/java/io/substrait/expression/Expression.java).

core/src/main/java/io/substrait/expression/proto/MaskExpressionProtoConverter.java
```diff
@@ -0,0 +1,203 @@
...
+import io.substrait.expression.MaskExpression.ListSelectItem;
+import io.substrait.expression.MaskExpression.MapSelect;
+import io.substrait.expression.MaskExpression.MaskExpr;
+import io.substrait.expression.MaskExpression.Select;
+import io.substrait.expression.MaskExpression.StructItem;
+import io.substrait.expression.MaskExpression.StructSelect;
+import io.substrait.proto.Expression;
+
+/** Converts between proto {@link Expression.MaskExpression} and POJO {@link MaskExpression}. */
+public final class MaskExpressionProtoConverter {
```
nielspardon, line 17: Mainly for consistency: the rest of the codebase currently separates the `toProto` and the `fromProto` conversion into two separate classes. 

In this case:
- `MaskExpressionProtoConverter` -> `toProto`
- `ProtoMaskExpressionConverter` -> `fromProto`

core/src/main/java/io/substrait/expression/MaskExpression.java
flex-seongmin: Thanks for the suggestion!

I refactored `Select` into a base interface with `accept()` method for visiting. And `StructSelect`, `ListSelect`, `MapSelect ` now extend it directly.

Also, Added `MaskExpressionVisitor` interface following the same project code pattern as you suggested.

core/src/main/java/io/substrait/expression/proto/MaskExpressionProtoConverter.java
flex-seongmin: I splited into `MaskExpressionProtoConverter - toProto` and `ProtoMaskExpressionConverter - fromProto`  to sync the project code convention. Thanks for notifying.

core/src/main/java/io/substrait/expression/MaskExpression.java
flex-seongmin: Thanks. I merged `MaskExpr` into `MaskExpression` interface. 

And sync this changes to what I've changed in this PR.

core/src/main/java/io/substrait/expression/proto/MaskExpressionProtoConverter.java
```diff
@@ -0,0 +1,107 @@
...
+
+  private MaskExpressionProtoConverter() {}
+
+  public static Expression.MaskExpression toProto(MaskExpression mask) {
+    return Expression.MaskExpression.newBuilder()
+        .setSelect(toProto(mask.getSelect()))
+        .setMaintainSingularStruct(mask.getMaintainSingularStruct())
+        .build();
+  }
```
nielspardon, line 62: Can you add a javadoc comment for this public function?

core/src/main/java/io/substrait/expression/proto/ProtoMaskExpressionConverter.java
```diff
@@ -0,0 +1,102 @@
...
+public final class ProtoMaskExpressionConverter {
+
+  private ProtoMaskExpressionConverter() {}
+
+  public static MaskExpression.Mask fromProto(Expression.MaskExpression proto) {
+    return MaskExpression.Mask.builder()
+        .select(fromProto(proto.getSelect()))
+        .maintainSingularStruct(proto.getMaintainSingularStruct())
+        .build();
+  }
```
nielspardon: Can you add a javadoc comment for this public function?

core/src/main/java/io/substrait/expression/proto/MaskExpressionProtoConverter.java
```diff
@@ -0,0 +1,107 @@
...
+  private static Expression.MaskExpression.Select toProtoSelect(Select select) {
+    Expression.MaskExpression.Select.Builder builder =
+        Expression.MaskExpression.Select.newBuilder();
+    if (select instanceof MaskExpression.StructSelect) {
+      builder.setStruct(toProto((MaskExpression.StructSelect) select));
+    } else if (select instanceof MaskExpression.ListSelect) {
+      builder.setList(toProtoListSelect((MaskExpression.ListSelect) select));
+    } else if (select instanceof MaskExpression.MapSelect) {
+      builder.setMap(toProtoMapSelect((MaskExpression.MapSelect) select));
+    }
```
nielspardon: this is where now you can use the `MaskExpressionVisitor` to avoid the `instanceof` checks

core/src/main/java/io/substrait/expression/MaskExpressionTypeProjector.java
```diff
@@ -0,0 +1,83 @@
...
+      return projectStruct((MaskExpression.StructSelect) select, (Type.Struct) fieldType);
+    }
+
+    if (select instanceof MaskExpression.ListSelect) {
+      return projectList((MaskExpression.ListSelect) select, (Type.ListType) fieldType);
+    }
+
+    if (select instanceof MaskExpression.MapSelect) {
+      return projectMap((MaskExpression.MapSelect) select, (Type.Map) fieldType);
+    }
```
nielspardon: this could also be a use case for using the `MaskExpressionVisitor`

core/src/main/java/io/substrait/expression/MaskExpressionTypeProjector.java
```diff
@@ -0,0 +1,126 @@
...
+  public static Type.Struct project(MaskExpression projection, Type.Struct baseStruct) {
+    return projectStruct(projection.getSelect(), baseStruct);
+  }
+
+  private static Type.Struct projectStruct(
+      MaskExpression.StructSelect structSelect, Type.Struct baseStruct) {
+    List<Type> fields = baseStruct.fields();
+    List<MaskExpression.StructItem> items = structSelect.getStructItems();
+
+    return TypeCreator.of(baseStruct.nullable())
```
nielspardon: minor: I would rename `baseStruct` here to `structType` which is a little more consistent with the other `projectList` and `projectMap` methods

```suggestion
      MaskExpression.StructSelect structSelect, Type.Struct structType) {
    List<Type> fields = structType.fields();
    List<MaskExpression.StructItem> items = structSelect.getStructItems();

    return TypeCreator.of(structType.nullable())
```

core/src/main/java/io/substrait/expression/proto/ProtoMaskExpressionConverter.java
nielspardon: for the javadoc you also want to have `@param` and `@return` tags:

```java
 /**
  * Converts a proto {@link Expression.MaskExpression} to its POJO representation.
  *
  * @param proto the proto {@link Expression.MaskExpression}
  * @return the POJO {@link MaskExpression.Mask}
  */
```

core/src/main/java/io/substrait/expression/proto/MaskExpressionProtoConverter.java
nielspardon: same here, can you add the `@param` and `@return` tags to the javadoc string

core/src/main/java/io/substrait/expression/proto/MaskExpressionProtoConverter.java
nielspardon: thanks, looks good

core/src/main/java/io/substrait/expression/MaskExpression.java
```diff
@@ -0,0 +1,407 @@
...
+  StructSelect getSelect();
+
+  /**
+   * When {@code true}, a struct that has only a single selected field will <em>not</em> be
+   * unwrapped into its child type.
+   */
+  @Value.Default
+  default boolean getMaintainSingularStruct() {
+    return false;
+  }
```
nielspardon, line 37: it's great that this method already has javadoc though it would be nice if you can add an `@return`

core/src/main/java/io/substrait/expression/MaskExpression.java
```diff
@@ -0,0 +1,407 @@
...
+
+  /** Selects a single field from a struct, with an optional nested child selection. */
+  @Value.Immutable
+  interface StructItem {
+    /** Zero-based field index within the struct. */
+    int getField();
+
+    /** Optional child selection for nested complex types. */
+    Optional<Select> getChild();
```
nielspardon: here we would also need the `@return` for both methods in order to have a valid javadoc

core/src/main/java/io/substrait/expression/MaskExpression.java
```diff
@@ -0,0 +1,407 @@
...
+  interface ListSelect extends Select {
+    /**
+     * Returns the list of selection items (individual elements or slices).
+     *
+     * @return the list of selection items
+     */
+    List<ListSelectItem> getSelection();
+
+    /** Optional child selection applied to each selected element. */
+    Optional<Select> getChild();
```
nielspardon: this one is also missing the `@return`

core/src/main/java/io/substrait/expression/MaskExpression.java
```diff
@@ -0,0 +1,407 @@
...
+    Optional<MapKey> getKey();
+
+    /**
+     * Returns the optional key expression for wildcard map selection.
+     *
+     * @return the optional map key expression
+     */
+    Optional<MapKeyExpression> getExpression();
+
+    /** Optional child selection applied to each selected map value. */
```
nielspardon: one more missing `@return` here

core/src/main/java/io/substrait/expression/MaskExpressionTypeProjector.java
```diff
@@ -0,0 +1,126 @@
...
+/**
+ * Applies a {@link MaskExpression} projection to a {@link Type.Struct}, returning a pruned struct.
+ */
+public final class MaskExpressionTypeProjector {
+
+  private MaskExpressionTypeProjector() {}
+
+  public static Type.Struct project(MaskExpression projection, Type.Struct baseStruct) {
+    return projectStruct(projection.getSelect(), baseStruct);
+  }
```
nielspardon: this public method could also deserve a javadoc comment

nielspardon:
if you can fix the javadoc and the minor renaming I think this is looking good

core/src/main/java/io/substrait/expression/MaskExpression.java
```diff
@@ -0,0 +1,425 @@
...
+    return ImmutableMaskExpression.Mask.builder();
+  }
+
+  // ---------------------------------------------------------------------------
+  // Top-level MaskExpression value
+  // ---------------------------------------------------------------------------
+
+  /** The concrete mask expression value holding the top-level struct selection and options. */
+  @Value.Immutable
+  interface Mask extends MaskExpression {}
```
nielspardon: sorry, found one more thing that can be simplified: you can simply add the `@Value.Immutable` annotation to the top-level interface so you don't need to create the alias interface `MaskExpression.Mask`.

```suggestion
@Value.Enclosing
@Value.Immutable
public interface MaskExpression {

  /** The top-level struct selection describing which fields to include. */
  StructSelect getSelect();

  /**
   * When {@code true}, a struct that has only a single selected field will <em>not</em> be
   * unwrapped into its child type.
   *
   * @return {@code true} if singular structs should be maintained, {@code false} otherwise
   */
  @Value.Default
  default boolean getMaintainSingularStruct() {
    return false;
  }

  /**
   * Creates a new builder for constructing a MaskExpression.
   *
   * @return a new builder instance
   */
  static ImmutableMaskExpression.Builder builder() {
    return ImmutableMaskExpression.builder();
  }
```

core/src/main/java/io/substrait/expression/proto/ProtoMaskExpressionConverter.java
```diff
@@ -0,0 +1,110 @@
...
+
+  private ProtoMaskExpressionConverter() {}
+
+  /**
+   * Converts a proto {@link Expression.MaskExpression} to its POJO representation.
+   *
+   * @param proto the proto {@link Expression.MaskExpression}
+   * @return the POJO {@link MaskExpression.Mask}
+   */
+  public static MaskExpression.Mask fromProto(Expression.MaskExpression proto) {
```
nielspardon: if you move the `@Value.Immutable` to `MaskExpression` and remove `MaskExpression.Mask` you would need to update this as well:
```suggestion
   * @return the POJO {@link MaskExpression}
   */
  public static MaskExpressionfromProto(Expression.MaskExpression proto) {
```

core/src/main/java/io/substrait/expression/MaskExpression.java
flex-seongmin: You're right. If we use `@Immutable` on MaskExpression on public interface, We do not need more inner `@Immutable` interface of MaskExpression.

core/src/main/java/io/substrait/expression/MaskExpression.java
```diff
@@ -0,0 +1,418 @@
...
+ * specification. It is used in {@code ReadRel} to describe column projection — the subset of a
+ * relation's schema that should actually be read.
+ *
+ * @see <a href="https://substrait.io/expressions/field_references/">Substrait Field References</a>
+ */
+@Value.Enclosing
+@Value.Immutable
+public interface MaskExpression {
+
+  /** The top-level struct selection describing which fields to include. */
```
nielspardon: I think this is also missing a `@return`

nielspardon:
```suggestion
  /**
    * The top-level struct selection describing which fields to include.
    *
    * @return the top-level struct selection
    */
```

flex-seongmin: I apologize that I didn't check it carefully. Adjusted It

nielspardon: no worries. I seem to have given you a code suggestion that isn't formatted properly by accident and you would need to run spotlessApply to fix the indendation

flex-seongmin: @nielspardon I executed `spotlessApply` task for lint

nielspardon: thanks

nielspardon (approved):
LGTM, thank you

PR #791, opened 2026-03-30
feat: add plan execution behavior

### LINKED ISSUE #801: add new Plan.ExecutionBehavior

nielspardon:
- Updates substrait Git submodule to v0.87.0
- Adds ExecutionBehavior field to Plan with VariableEvaluationMode enum
- Enhances SubstraitBuilder with overloaded plan() methods accepting custom execution behavior
- Updates SqlToSubstrait and Spark converters to set default execution behavior (VARIABLE_EVALUATION_MODE_PER_PLAN)

fixes https://github.com/substrait-io/substrait-java/issues/801

### DIFF
--- a/core/src/main/java/io/substrait/dsl/SubstraitBuilder.java
+++ b/core/src/main/java/io/substrait/dsl/SubstraitBuilder.java
@@ -16,7 +16,9 @@ import io.substrait.expression.WindowBound;
 import io.substrait.extension.DefaultExtensionCatalog;
 import io.substrait.extension.SimpleExtension;
 import io.substrait.function.ToTypeString;
+import io.substrait.plan.ImmutableExecutionBehavior;
 import io.substrait.plan.Plan;
+import io.substrait.plan.Plan.ExecutionBehavior.VariableEvaluationMode;
 import io.substrait.relation.AbstractWriteRel;
 import io.substrait.relation.Aggregate;
 import io.substrait.relation.Aggregate.Measure;
@@ -1540,13 +1542,54 @@ public class SubstraitBuilder {
   }
 
   /**
-   * Creates a plan from a plan root.
+   * Creates a plan from a plan root with default execution behavior.
+   *
+   * <p>The plan is created with {@link VariableEvaluationMode#VARIABLE_EVALUATION_MODE_PER_PLAN} as
+   * the default variable evaluation mode. To specify a custom execution behavior, use {@link
+   * #plan(Plan.ExecutionBehavior, Plan.Root)} instead.
    *
    * @param root the plan root
    * @return a new {@link Plan}
    */
   public Plan plan(Plan.Root root) {
-    return Plan.builder().addRoots(root).build();
+    return plan(
+        ImmutableExecutionBehavior.builder()
+            .variableEvaluationMode(VariableEvaluationMode.VARIABLE_EVALUATION_MODE_PER_PLAN)
+            .build(),
+        root);
+  }
+
+  /**
+   * Creates a plan from a plan root with custom execution behavior.
+   *
+   * @param executionBehavior the execution behavior for the plan
+   * @param root the plan root
+   * @return a new {@link Plan}
+   */
+  public Plan plan(Plan.ExecutionBehavior executionBehavior, Plan.Root root) {
+    return Plan.builder().executionBehavior(executionBehavior).addRoots(root).build();
+  }
+
+  /**
+   * Creates a plan from multiple plan roots with custom execution behavior.
+   *
+   * @param executionBehavior the execution behavior for the plan
+   * @param roots the plan roots
+   * @return a new {@link Plan}
+   */
+  public Plan plan(Plan.ExecutionBehavior executionBehavior, Plan.Root... roots) {
+    return Plan.builder().executionBehavior(executionBehavior).roots(Arrays.asList(roots)).build();
+  }
+
+  /**
+   * Creates a plan from multiple plan roots with custom execution behavior.
+   *
+   * @param executionBehavior the execution behavior for the plan
+   * @param roots the plan roots as an iterable
+   * @return a new {@link Plan}
+   */
+  public Plan plan(Plan.ExecutionBehavior executionBehavior, Iterable<Plan.Root> roots) {
+    return Plan.builder().executionBehavior(executionBehavior).roots(roots).build();
   }
 
   /**
--- a/core/src/main/java/io/substrait/dsl/SubstraitBuilder.java
+++ b/core/src/main/java/io/substrait/dsl/SubstraitBuilder.java
@@ -1544,8 +1544,8 @@ public class SubstraitBuilder {
   /**
    * Creates a plan from a plan root with default execution behavior.
    *
-   * <p>The plan is created with {@link VariableEvaluationMode#VARIABLE_EVALUATION_MODE_PER_PLAN} as
-   * the default variable evaluation mode. To specify a custom execution behavior, use {@link
+   * <p>The plan is created with {@link VariableEvaluationMode#PER_PLAN} as the default variable
+   * evaluation mode. To specify a custom execution behavior, use {@link
    * #plan(Plan.ExecutionBehavior, Plan.Root)} instead.
    *
    * @param root the plan root
@@ -1554,7 +1554,7 @@ public class SubstraitBuilder {
   public Plan plan(Plan.Root root) {
     return plan(
         ImmutableExecutionBehavior.builder()
-            .variableEvaluationMode(VariableEvaluationMode.VARIABLE_EVALUATION_MODE_PER_PLAN)
+            .variableEvaluationMode(VariableEvaluationMode.PER_PLAN)
             .build(),
         root);
   }
--- a/core/src/main/java/io/substrait/plan/Plan.java
+++ b/core/src/main/java/io/substrait/plan/Plan.java
@@ -21,6 +21,38 @@ public abstract class Plan {
 
   public abstract Optional<AdvancedExtension> getAdvancedExtension();
 
+  public abstract Optional<ExecutionBehavior> getExecutionBehavior();
+
+  /**
+   * Validates that the execution behavior is properly configured.
+   *
+   * <p>This validation method ensures that:
+   *
+   * <ul>
+   *   <li>The {@link ExecutionBehavior} field is present (not null or empty) - ExecutionBehavior is
+   *       a required field
+   *   <li>The {@link ExecutionBehavior.VariableEvaluationMode} is set to a valid value (not {@link
+   *       ExecutionBehavior.VariableEvaluationMode#VARIABLE_EVALUATION_MODE_UNSPECIFIED})
+   * </ul>
+   *
+   * @throws IllegalArgumentException if the execution behavior is not present, or if the variable
+   *     evaluation mode is set to {@link
+   *     ExecutionBehavior.VariableEvaluationMode#VARIABLE_EVALUATION_MODE_UNSPECIFIED}
+   */
+  @Value.Check
+  protected void check() {
+    if (!getExecutionBehavior().isPresent()) {
+      throw new IllegalArgumentException("ExecutionBehavior is required but was not set");
+    }
+    ExecutionBehavior behavior = getExecutionBehavior().get();
+    if (behavior.getVariableEvaluationMode()
+        == ExecutionBehavior.VariableEvaluationMode.VARIABLE_EVALUATION_MODE_UNSPECIFIED) {
+      throw new IllegalArgumentException(
+          "ExecutionBehavior requires a specified VariableEvaluationMode, but got: "
+              + behavior.getVariableEvaluationMode());
+    }
+  }
+
   public static ImmutablePlan.Builder builder() {
     return ImmutablePlan.builder();
   }
@@ -67,6 +99,21 @@ public abstract class Plan {
 
     public static ImmutableRoot.Builder builder() {
       return ImmutableRoot.builder();
+    }
+  }
+
+  @Value.Immutable
+  public abstract static class ExecutionBehavior {
+    public abstract VariableEvaluationMode getVariableEvaluationMode();
+
+    public static ImmutableExecutionBehavior.Builder builder() {
+      return ImmutableExecutionBehavior.builder();
+    }
+
+    public enum VariableEvaluationMode {
+      VARIABLE_EVALUATION_MODE_UNSPECIFIED,
+      VARIABLE_EVALUATION_MODE_PER_PLAN,
+      VARIABLE_EVALUATION_MODE_PER_RECORD
     }
   }
 }
--- a/core/src/main/java/io/substrait/plan/Plan.java
+++ b/core/src/main/java/io/substrait/plan/Plan.java
@@ -32,18 +32,17 @@ public abstract class Plan {
    *   <li>The {@link ExecutionBehavior} field is present (not null or empty) - ExecutionBehavior is
    *       a required field
    *   <li>The {@link ExecutionBehavior.VariableEvaluationMode} is set to a valid value (not {@link
-   *       ExecutionBehavior.VariableEvaluationMode#VARIABLE_EVALUATION_MODE_UNSPECIFIED})
+   *       ExecutionBehavior.VariableEvaluationMode#UNSPECIFIED})
    * </ul>
    *
    * @throws IllegalArgumentException if the execution behavior is not present, or if the variable
-   *     evaluation mode is set to {@link
-   *     ExecutionBehavior.VariableEvaluationMode#VARIABLE_EVALUATION_MODE_UNSPECIFIED}
+   *     evaluation mode is set to {@link ExecutionBehavior.VariableEvaluationMode#UNSPECIFIED}
    */
   @Value.Check
   protected void check() {
     ExecutionBehavior behavior = getExecutionBehavior();
     if (behavior.getVariableEvaluationMode()
-        == ExecutionBehavior.VariableEvaluationMode.VARIABLE_EVALUATION_MODE_UNSPECIFIED) {
+        == ExecutionBehavior.VariableEvaluationMode.UNSPECIFIED) {
       throw new IllegalArgumentException(
           "ExecutionBehavior requires a specified VariableEvaluationMode, but got: "
               + behavior.getVariableEvaluationMode());
@@ -108,9 +107,30 @@ public abstract class Plan {
     }
 
     public enum VariableEvaluationMode {
-      VARIABLE_EVALUATION_MODE_UNSPECIFIED,
-      VARIABLE_EVALUATION_MODE_PER_PLAN,
-      VARIABLE_EVALUATION_MODE_PER_RECORD
+      UNSPECIFIED(
+          io.substrait.proto.ExecutionBehavior.VariableEvaluationMode
+              .VARIABLE_EVALUATION_MODE_UNSPECIFIED),
+      PER_PLAN(
+          io.substrait.proto.ExecutionBehavior.VariableEvaluationMode
+              .VARIABLE_EVALUATION_MODE_PER_PLAN),
+      PER_RECORD(
+          io.substrait.proto.ExecutionBehavior.VariableEvaluationMode
+              .VARIABLE_EVALUATION_MODE_PER_RECORD);
+
+      private final io.substrait.proto.ExecutionBehavior.VariableEvaluationMode proto;
+
+      VariableEvaluationMode(io.substrait.proto.ExecutionBehavior.VariableEvaluationMode proto) {
+        this.proto = proto;
+      }
+
+      /**
+       * Converts this variable evaluation mode to its protobuf representation.
+       *
+       * @return the protobuf representation
+       */
+      public io.substrait.proto.ExecutionBehavior.VariableEvaluationMode toProto() {
+        return proto;
+      }
     }
   }
 }
--- a/core/src/main/java/io/substrait/plan/Plan.java
+++ b/core/src/main/java/io/substrait/plan/Plan.java
@@ -35,22 +35,14 @@ public abstract class Plan {
    *       ExecutionBehavior.VariableEvaluationMode#VARIABLE_EVALUATION_MODE_UNSPECIFIED})
    * </ul>
    *
-   * <p>The validation is currently triggered when the Plan gets serialized into protobuf in {@link
-   * PlanProtoConverter#toProto(Plan)}.
-   *
-   * <p>TODO: After the grace period of introducing this breaking change is over, the validation
-   * should also be triggered when the Plan gets deserialized from protobuf by adding the {@link
-   * Value.Check} annotation to this method and removing the call from {@link
-   * PlanProtoConverter#toProto(Plan)}.
-   *
    * @throws IllegalArgumentException if the execution behavior is not present, or if the variable
    *     evaluation mode is set to {@link
    *     ExecutionBehavior.VariableEvaluationMode#VARIABLE_EVALUATION_MODE_UNSPECIFIED}
    */
-  protected void validate() {
+  @Value.Check
+  protected void check() {
     if (!getExecutionBehavior().isPresent()) {
-      throw new IllegalArgumentException(
-          "ExecutionBehavior is required since Substrait v0.87.0 but was not set");
+      throw new IllegalArgumentException("ExecutionBehavior is required but was not set");
     }
     ExecutionBehavior behavior = getExecutionBehavior().get();
     if (behavior.getVariableEvaluationMode()
--- a/core/src/main/java/io/substrait/plan/Plan.java
+++ b/core/src/main/java/io/substrait/plan/Plan.java
@@ -21,7 +21,7 @@ public abstract class Plan {
 
   public abstract Optional<AdvancedExtension> getAdvancedExtension();
 
-  public abstract Optional<ExecutionBehavior> getExecutionBehavior();
+  public abstract ExecutionBehavior getExecutionBehavior();
 
   /**
    * Validates that the execution behavior is properly configured.
@@ -41,10 +41,7 @@ public abstract class Plan {
    */
   @Value.Check
   protected void check() {
-    if (!getExecutionBehavior().isPresent()) {
-      throw new IllegalArgumentException("ExecutionBehavior is required but was not set");
-    }
-    ExecutionBehavior behavior = getExecutionBehavior().get();
+    ExecutionBehavior behavior = getExecutionBehavior();
     if (behavior.getVariableEvaluationMode()
         == ExecutionBehavior.VariableEvaluationMode.VARIABLE_EVALUATION_MODE_UNSPECIFIED) {
       throw new IllegalArgumentException(
--- a/core/src/main/java/io/substrait/plan/Plan.java
+++ b/core/src/main/java/io/substrait/plan/Plan.java
@@ -35,14 +35,22 @@ public abstract class Plan {
    *       ExecutionBehavior.VariableEvaluationMode#VARIABLE_EVALUATION_MODE_UNSPECIFIED})
    * </ul>
    *
+   * <p>The validation is currently triggered when the Plan gets serialized into protobuf in {@link
+   * PlanProtoConverter#toProto(Plan)}.
+   *
+   * <p>TODO: After the grace period of introducing this breaking change is over, the validation
+   * should also be triggered when the Plan gets deserialized from protobuf by adding the {@link
+   * Value.Check} annotation to this method and removing the call from {@link
+   * PlanProtoConverter#toProto(Plan)}.
+   *
    * @throws IllegalArgumentException if the execution behavior is not present, or if the variable
    *     evaluation mode is set to {@link
    *     ExecutionBehavior.VariableEvaluationMode#VARIABLE_EVALUATION_MODE_UNSPECIFIED}
    */
-  @Value.Check
-  protected void check() {
+  protected void validate() {
     if (!getExecutionBehavior().isPresent()) {
-      throw new IllegalArgumentException("ExecutionBehavior is required but was not set");
+      throw new IllegalArgumentException(
+          "ExecutionBehavior is required since Substrait v0.87.0 but was not set");
     }
     ExecutionBehavior behavior = getExecutionBehavior().get();
     if (behavior.getVariableEvaluationMode()
--- a/core/src/main/java/io/substrait/plan/PlanProtoConverter.java
+++ b/core/src/main/java/io/substrait/plan/PlanProtoConverter.java
@@ -4,6 +4,8 @@ import io.substrait.extension.DefaultExtensionCatalog;
 import io.substrait.extension.ExtensionCollector;
 import io.substrait.extension.ExtensionProtoConverter;
 import io.substrait.extension.SimpleExtension.ExtensionCollection;
+import io.substrait.proto.ExecutionBehavior;
+import io.substrait.proto.ExecutionBehavior.VariableEvaluationMode;
 import io.substrait.proto.Plan;
 import io.substrait.proto.PlanRel;
 import io.substrait.proto.Rel;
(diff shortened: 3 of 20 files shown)


core/src/main/java/io/substrait/dsl/SubstraitBuilder.java
```diff
@@ -1523,13 +1525,54 @@ public Plan.Root root(Rel rel) {
...
+   * #plan(Plan.ExecutionBehavior, Plan.Root)} instead.
    *
    * @param root the plan root
    * @return a new {@link Plan}
    */
   public Plan plan(Plan.Root root) {
-    return Plan.builder().addRoots(root).build();
+    return plan(
+        ImmutableExecutionBehavior.builder()
+            .variableEvaluationMode(VariableEvaluationMode.VARIABLE_EVALUATION_MODE_PER_PLAN)
```
benbellick: Just an idea: I wonder if there is a way to hide `VariableEvaluationMode` until users actually work with `ExecutionContextVariable` expressions. 

For example, what if you configured it as part of the `ExecutionContextVariable` java POJO and then at plan build time, you just make sure all of those agree, and set the `VariableEvaluationMode` there?

nielspardon: @vbarua made a similar remark on the Substrait spec change: https://github.com/substrait-io/substrait/pull/945#discussion_r2998792520

There I expressed that I would not make this setting conditional on the presence or absence of execution context variables.

benbellick: I am not proposing removing it from the plan. I am proposing pushing the decision of evaluation mode to the location of `ExecutionContextVariable` construction. 

Still unsure if it is the right decision, just an idea to hide complexity from people that don't need to think about it.

benbellick: Here is one alternative thought, what if the default behavior was `VARIABLE_EVALUATION_MODE_UNSPECIFIED`, and then we consider it a plan build-time error if you didn't specify an evaluation mode with a plan that contains an `ExecutionContextVariable`? 

The pros of this is that: 
- it lets people ignore this setting when it isn't relevant to them without picking a real semantic-altering default, 
- it forces people who need these execution contexts to actually _pick_ their semantics instead of accidentally picking one by mistake. 

What do you think?

nielspardon: The semantic we agreed on in the Substrait specification is that it is a required field and `UNSPECIFIED` is not a valid value. What you are describing here is against what the spec says and would require a spec change.

benbellick: Just to be clear: are you saying a plan which has evaluation mode unspecified but has no `ExecutionContextVariable`s in it is considered an invalid plan, per the current spec?

nielspardon: Yes, that's my understanding. Especially after the conversation with Victor just before the PR got merged that I linked earlier: https://github.com/substrait-io/substrait/pull/945#discussion_r2998792520

nielspardon: Looking back at the proto and documentation I admit it's not 100% clear since in the proto it says `ExecutionBehavior` is required but it does not specify that `VariableEvaluationMode` is required and can not be set to `UNSPECIFIED`.

nielspardon: How do we make progress here @benbellick?

core/src/test/java/io/substrait/plan/PlanValidationTest.java
benbellick, line 1: nice tests 🚀

core/src/main/java/io/substrait/plan/ProtoPlanConverter.java
```diff
@@ -66,6 +66,29 @@ protected ProtoRelConverter getProtoRelConverter(final ExtensionLookup functionL
     return new ProtoRelConverter(functionLookup, this.extensionCollection, protoExtensionConverter);
   }
 
+  /**
```
benbellick, line 71: thanks for contributing these docstring comments

core/src/main/java/io/substrait/plan/Plan.java
```diff
@@ -21,6 +21,38 @@ public Version getVersion() {
...
+   *
+   * @throws IllegalArgumentException if the execution behavior is not present, or if the variable
+   *     evaluation mode is set to {@link
+   *     ExecutionBehavior.VariableEvaluationMode#VARIABLE_EVALUATION_MODE_UNSPECIFIED}
+   */
+  @Value.Check
+  protected void check() {
+    if (!getExecutionBehavior().isPresent()) {
+      throw new IllegalArgumentException("ExecutionBehavior is required but was not set");
+    }
```
benbellick: I think this is the behavior we will eventually want. 

However, if I understand correctly, this check will cause `substrait-java` to reject proto plans which don't contain the new field. Thus, I think this will inadvertently be a breaking change, and I think we should relax this check until the [core libraries](https://substrait.io/community/active_libraries/) all have supported this feature for some time. 

Perhaps we can for now just set the field if it is absent and then open a ticket for handling enforcement in the future?

benbellick (commented):
I think it generally looks good, but just the one comment about ensuring that old plans are still allowed through. Thanks!

core/src/main/java/io/substrait/dsl/SubstraitBuilder.java
benbellick: Sorry for the late review. If in the future you are waiting on a review from me, please feel free to ping me earlier 🙂 

I just left a review. In terms of moving forwards, I think this PR is generally good as is in that it implements the spec, and I think the questions I have remaining are more appropriate in the upstream so I will lay them to rest. Thanks for your patience!

core/src/main/java/io/substrait/plan/Plan.java
nielspardon: I get the idea. I vaguely remember we were discussing in one of the past community syncs about backwards compatibility of plans and I thought the consensus was that at least pre-1.0 we are not going to offer extensive backward compatibility. In that spirit I would introduce the exception with the substrait-java release which implements support for this feature. What we could do is to announce in the exception message which Substrait version introduces this requirement.

```java
      throw new IllegalArgumentException("ExecutionBehavior is required since Substrait v0.87.0 but was not set");
```

benbellick: We tried to capture the breaking change policy [here](https://substrait.io/spec/breaking_change_policy/#substrait-breaking-change-policy). 

The core bit being:
> Breaking changes should be made so that users can update systems independently and asynchronously.


For us, this change as is will cause our systems to break. `substrait-go` is producing plans upstream which are consumed by `substrait-java`, so the missing `ExecutionBehavior` in the `substrait-go`-produced plan will cause an exception to be thrown. I think the appropriate migration strategy here is to allow consumption of old plans, regardless of if they have `ExecutionBehavior` or not, but we can produce plans that always have it. Then in the future once all of the core libraries have been consistently producing the new feature for some time, we can then enforce the presence of the field on consumption. [This is how the URI -> URN migration was handled.](https://substrait.io/spec/breaking_change_policy/#uri-to-urn-migration)

nielspardon: in that sense you would not throw an exception when consuming a plan that has no execution behavior set but only throw an exception when a plan is being produced using substrait-java that does not have it set?

benbellick: yes, exactly. 

(Until later when all libraries have been producing the new `ExecutionBehavior` for a bit, at which point we can switch on exception-throwing for consumption of plans missing the new field.)

nielspardon: I removed the `@Valid.Check` annotation so it does not automatically get called by the immutables builder and added a call to `PlanProtoConverter.toProto()` so it gets called when the plan is serialized to protobuf. I guess that is the best way to catch it when producing a plan.

core/src/main/java/io/substrait/plan/Plan.java
```diff
@@ -21,6 +21,46 @@ public Version getVersion() {
...
+   *   <li>The {@link ExecutionBehavior} field is present (not null or empty) - ExecutionBehavior is
+   *       a required field
+   *   <li>The {@link ExecutionBehavior.VariableEvaluationMode} is set to a valid value (not {@link
+   *       ExecutionBehavior.VariableEvaluationMode#VARIABLE_EVALUATION_MODE_UNSPECIFIED})
+   * </ul>
+   *
+   * <p>The validation is currently triggered when the Plan gets serialized into protobuf in {@link
+   * PlanProtoConverter#toProto(Plan)}.
+   *
+   * <p>TODO: After the grace period of introducing this breaking change is over, the validation
```
benbellick: perhaps we might want to make an issue for this and link it here just to be overly cautious?

core/src/main/java/io/substrait/plan/ProtoPlanConverter.java
```diff
@@ -66,6 +66,29 @@ protected ProtoRelConverter getProtoRelConverter(final ExtensionLookup functionL
...
+   *   <li>Extension functions and types lookup
+   *   <li>Version information (major, minor, patch, git hash, producer)
+   *   <li>Advanced extensions if present
+   *   <li>Execution behavior configuration if present
+   * </ul>
+   *
+   * <p><b>Note:</b> While execution behavior is optional in the protobuf message, the Plan
+   * validation will still fail if ExecutionBehavior is not set or if it contains
+   * VARIABLE_EVALUATION_MODE_UNSPECIFIED. Ensure the protobuf Plan has a valid execution behavior
+   * before conversion.
```
benbellick: Is this still true? I think we now accept missing execution behavior.

core/src/main/java/io/substrait/plan/PlanProtoConverter.java
```diff
@@ -61,7 +63,30 @@ public PlanProtoConverter(
...
+   * <ul>
+   *   <li>All plan roots and their associated relations
+   *   <li>Extension functions and types used in the plan
+   *   <li>Version information
+   *   <li>Advanced extensions if present
+   *   <li>Execution behavior configuration if present
+   * </ul>
+   *
+   * <p>The execution behavior is optional. If present, it will be converted using {@link
+   * #toProtoExecutionBehavior(io.substrait.plan.Plan.ExecutionBehavior)}.
```
benbellick:
```suggestion
```

IMO I think this isn't necessary. I think it should be clear that the proto captures the whole plan and it becomes unclear what should and shouldn't be listed here.

core/src/main/java/io/substrait/plan/ProtoPlanConverter.java
```diff
@@ -66,6 +66,29 @@ protected ProtoRelConverter getProtoRelConverter(final ExtensionLookup functionL
...
+   *
+   * <p>This method performs a complete conversion from the protobuf representation, including:
+   *
+   * <ul>
+   *   <li>All plan relations and their roots
+   *   <li>Extension functions and types lookup
+   *   <li>Version information (major, minor, patch, git hash, producer)
+   *   <li>Advanced extensions if present
+   *   <li>Execution behavior configuration if present
+   * </ul>
```
benbellick: same here
```suggestion
```

core/src/main/java/io/substrait/plan/Plan.java
benbellick: I think a cleaner approach would be to keep the `@Value.Check` validation. 

My mental model is that our internal POJO representation should always be whatever the "new" thing is. So in this case, POJOS must have the required field set. Thus, if you construct a plan programmatically, you should have to set an `ExecutionBehavior` explicitly, or use a helper that supplies the default (which is what you have already done).

However for backwards compatibility, `ProtoPlanConverter.from(...)` can be the boundary. If an older proto plan does not have `execution_behavior`, we can set it to `VARIABLE_EVALUATION_MODE_PER_PLAN` during deserialization.

That keeps the compatibility handling invisible to callers and avoids having Java `Plan` objects that are only invalid when serialized back to protobuf.

I _think_ defaulting is safe here. If an old plan did not have this field, it could not have been relying on per-record variable evaluation semantics, so defaulting to the existing/per-plan behavior should preserve semantics. Do you agree with that approach?

benbellick (commented):
I left a comment about the behavior. I think we still want the `@Value.check` annotation. We can just add the field value explicitly to the POJO `Plan` when we do `FromProto`. 


Also just a minor aside, it makes it easier to review if you add new commits instead of force-pushing. We squash commit anyways so the history doesn't matter. However, if you force-push, the "show changes since your last review" button doesn't work anymore. Thanks!
image

nielspardon:
> Also just a minor aside, it makes it easier to review if you add new commits instead of force-pushing. We squash commit anyways so the history doesn't matter. However, if you force-push, the "show changes since your last review" button doesn't work anymore. Thanks!

I have to force push for a rebase, don't I? The commit are all still preservered just have different commit ids since I rebased.

core/src/main/java/io/substrait/plan/Plan.java
nielspardon: sure, we can do that though it gets a little tricky to differentiate between an old plan from when this setting didn't exist and a new plan that just ignored setting the value. We would have to start checking if the Substrait plan version is older than the version that introduced the feature. Even that is a little murky since we do have Substrait SDKs being updated to the newer spec version and released without all the changes fully reflected in the code, e.g. substrait-java is already producing plans claiming to be v0.87.0 since we already did the submodule bump which are technically invalid v0.87.0 plans. I don't think we can be a 100% clean here.

core/src/main/java/io/substrait/plan/Plan.java
nielspardon: I reverted this change.

core/src/main/java/io/substrait/plan/Plan.java
nielspardon: - I reverted the two commits that added the check only on serialization and tests in a separate commit. 
- Then I added a commit that checks whether the Substrait plan version is < 0.87.0 and set it to PER_PLAN if it is missing. Otherwise if the Substrait plan version is >= 0.87.0 the `@Value.Check` method will catch the missing setting and throw the exception.

core/src/main/java/io/substrait/plan/ProtoPlanConverter.java
nielspardon: I updated the comment to reflect the new behavior with the version check.

benbellick:
> > Also just a minor aside, it makes it easier to review if you add new commits instead of force-pushing. We squash commit anyways so the history doesn't matter. However, if you force-push, the "show changes since your last review" button doesn't work anymore. Thanks!
> 
> I have to force push for a rebase, don't I? The commit are all still preservered just have different commit ids since I rebased.

You can also merge `main` into your branch if you have merge conflicts. Yes the changes are still present, but the commit hash changes, which means GH no longer recognizes which commits came before a review and which come after.

core/src/main/java/io/substrait/plan/Plan.java
benbellick: Why does it matter to check the version of the plan here? Is there anything wrong with passing the value through if it is present, and defaulting to `VARIABLE_EVALUATION_MODE_PER_PLAN` if it is absent (regardless of version)?

I’m a little hesitant to trust the version as the enforcement boundary yet. Does the plan version mean the version of the proto definitions used to serialize the plan, the version of the spec the producer claims to support, or something else? Does it have a consistent meaning across the libraries?

We could also log a warning when defaulting, saying that future versions may reject plans without `ExecutionBehavior` and producers should start setting it explicitly.

nielspardon: The plan version is the Substrait spec version the producer claims the Plan has been built for and is a required field since a very long time: https://github.com/substrait-io/substrait/blob/38d87906472c04cd957ba5a2ba3880dcc2904ff8/proto/substrait/plan.proto#L30-L32

nielspardon: As I tried to express above the practice is different/messy:

> Even that is a little murky since we do have Substrait SDKs being updated to the newer spec version and released without all the changes fully reflected in the code, e.g. substrait-java is already producing plans claiming to be v0.87.0 since we already did the submodule bump which are technically invalid v0.87.0 plans. I don't think we can be a 100% clean here.

benbellick: Agreed it is messy here. This is why I am apprehensive to use the version as a gate. What are you worried about happening if we take this approach? 
> Is there anything wrong with passing the value through if it is present, and defaulting to VARIABLE_EVALUATION_MODE_PER_PLAN if it is absent (regardless of version)?

I personally am interpreting [this](https://github.com/substrait-io/substrait/blob/38d87906472c04cd957ba5a2ba3880dcc2904ff8/proto/substrait/plan.proto#L30-L32) to mean to proto version that plan was produced with, rather than the library's fully supported version. Since no library actually covers all of the spec, a library declaring that it supports the whole spec version is probably impossible at the moment 😅

nielspardon: > I personally am interpreting [this](https://github.com/substrait-io/substrait/blob/38d87906472c04cd957ba5a2ba3880dcc2904ff8/proto/substrait/plan.proto#L30-L32) to mean to proto version that plan was produced with, rather than the library's fully supported version. Since no library actually covers all of the spec, a library declaring that it supports the whole spec version is probably impossible at the moment 😅

Yes, that's the reality. I think there is a distinction between features that are opt-in via Substrait dialect like the support for execution context variables is opt-in but something like the Substrait plan version field is not opt-in via a dialect. This is a general requirement. 

With the current definition of execution behavior I would argue it is like plan version: if you claim you are producing plans using spec version 0.87.0 then you have to define execution behavior (until we change the spec to relax this requirement).

We can update the spec to say: default is per_plan and then always setting it to per_plan if absent would be strictly compliant with the spec.

My main concern is that this behavior would not be strictly compliant with the current spec and we would be setting a bad example if we started to deviate from what the released version of the spec says.

nielspardon: The other concern that feeds into this is that we introduced this in the spec 2+ months ago. If our turnaround for library adoption was shorter than we could react more quickly with spec changes, such findings could be updated in the spec sooner and we could adopt the modifications more quickly in the spec. In the last 2+ months we had multiple other modifications to the spec which are also on the backlog and if we are spending so much time on the adoption of all of the other spec changes.

vbarua:
I think we've drifted into two different discussions here:
1. Based on the deprecation policy, how should we implement support for ExecutionContextVariable in a backwards compatible way.
2. What does the version field mean?

For 1, the main constraint is that we want the library to read _protobuf_ plans that don't set the ExecutionContextVariable, as this allows users to update to this version without having to coordinate producer changes. As Ben points out, we can handle this at the protobuf conversion layer by setting a default value. This should be safe, because a system that doesn't set this variable isn't using it yet. Within the POJO code, we can assume that ExecutionContextVariable is set, and we can emit protobuf plans including it as other systems will safely ignore it until they are updated.

2 is a discussion that goes beyond this PR IMO. In observed practice, the plan version set is just the protobuf version and doesn't correspond to _specification_ support at the version. At this point in time we can't use it for anything beyond that.

nielspardon:
> 2 is a discussion that goes beyond this PR IMO. In observed practice, the plan version set is just the protobuf version and doesn't correspond to _specification_ support at the version. At this point in time we can't use it for anything beyond that.

I think this summary leaves out an important point. Sure, we can say: this is current practice and we just accept it. On the other hand since the project has such a high emphasis on interoperability and portability we should define what is the minimum required compliance with a given Substrait specification version in my opinion.

nielspardon:
I updated the PR to always set the default execution behavior when parsing a Substrait plan from proto.

vbarua:
> Sure, we can say: this is current practice and we just accept it. On the other hand since the project has such a high emphasis on interoperability and portability we should define what is the minimum required compliance with a given Substrait specification version in my opinion.

Sorry, I didn't mean to imply that we shouldn't work on 2 with my statement. I think it is worth pursuing, I just don't think it's something that we can resolve unilaterally in this PR discussion as there are a ton of moving parts. Filed an issue https://github.com/substrait-io/substrait/issues/1102 for it.

core/src/main/java/io/substrait/plan/ProtoPlanConverter.java
```diff
@@ -92,15 +116,87 @@ public Plan from(io.substrait.proto.Plan plan) {
...
+
+    // Set execution behavior if present
+    if (plan.hasExecutionBehavior()) {
+      planBuilder.executionBehavior(fromProtoExecutionBehavior(plan.getExecutionBehavior()));
+    } else {
+      // set default ExecutionBehavior for older plans
+      planBuilder.executionBehavior(
+          ExecutionBehavior.builder()
+              .variableEvaluationMode(VariableEvaluationMode.VARIABLE_EVALUATION_MODE_PER_PLAN)
+              .build());
```
vbarua, line 131: This is actually the only piece we need for backwards compatability IMO.

core/src/main/java/io/substrait/plan/ProtoPlanConverter.java
```diff
@@ -66,6 +68,28 @@ protected ProtoRelConverter getProtoRelConverter(final ExtensionLookup functionL
...
+   * <ul>
+   *   <li>For plans older than Substrait 0.87.0 (the version that introduced execution behavior), a
+   *       missing execution behavior is defaulted to {@link
+   *       io.substrait.plan.Plan.ExecutionBehavior.VariableEvaluationMode#VARIABLE_EVALUATION_MODE_PER_PLAN}.
+   *       Note that a plan without a version is treated as version 0.0.0, and therefore as older
+   *       than 0.87.0.
+   *   <li>For plans at or after 0.87.0, a missing execution behavior is left unset and Plan
+   *       validation will fail. The validation also fails if the execution behavior is present but
+   *       contains {@code VARIABLE_EVALUATION_MODE_UNSPECIFIED}.
+   * </ul>
```
vbarua: I'm not sure if its worth going into this level of detail here, especially given that we're not using the version number to perform this conditionally anymore. Something to the effect of:
> Note: If `Plan.ExecutionBehavior.VariableEvaluationMode` is not set, it will be defaulted to `VARIABLE_EVALUATION_MODE_PER_PLAN`. Once other producers populate this field correctly, this compatibility workaround will be removed.

core/src/main/java/io/substrait/plan/Plan.java
```diff
@@ -21,6 +21,38 @@ public Version getVersion() {
 
   public abstract Optional<AdvancedExtension> getAdvancedExtension();
 
+  public abstract Optional<ExecutionBehavior> getExecutionBehavior();
```
vbarua: If we include the code in the ProtoPlanConverter which automatically sets this when it's not set in the proto, I think that we can just make this:
```
public abstract ExecutionBehavior getExecutionBehavior();
```

Within the library, we can match the spec:
https://github.com/substrait-io/substrait/blob/7b5cfd9fc03cca6cca535bec060b565e54df027e/proto/substrait/plan.proto#L66-L67

and it's only at the external interface consuming protobuf plans that we need to worry about compatibility. That way, when we remove the shim we don't need to update the library API either.

core/src/main/java/io/substrait/plan/Plan.java
vbarua: > My main concern is that this behavior would not be strictly compliant with the current spec and we would be setting a bad example if we started to deviate from what the released version of the spec says.

I see where you're coming from with this, and I agree that it's not totally desirable. It's taking advantage of the fact that if this isn't being set, you as a producer likely don't care about it, but it's not strictly in the spec. The ideal way to have done this would probably have been to include something like: "If this is not set / UNSPECIFIED default to X", which is effectively what we're doing here behind the specs back.

If we don't do some form of compatibility shim though, we end up in a situation where a consumer updates substrait-java and then having every plan produced by a system that has not yet been updated fail the validation.

core/src/main/java/io/substrait/plan/ProtoPlanConverter.java
nielspardon: I changed the javadoc as you suggested

core/src/main/java/io/substrait/plan/Plan.java
nielspardon: makes sense. changed that.

bestbeforetoday:
> > I have to force push for a rebase, don't I? The commit are all still preservered just have different commit ids since I rebased.
> 
> You can also merge `main` into your branch if you have merge conflicts. Yes the changes are still present, but the commit hash changes, which means GH no longer recognizes which commits came before a review and which come after.

Just to add another alternative opinion... I have a pet hate for merging main into PR branches. It just ends up with such a messy commit history within the PR branch. I absolutely agree with adding additional commits to address PR comments so the specific changes can be viewed, but I much prefer rebasing to get the branch up-to-date with main and resolve any merge conflicts in the rebase so you only see the incremental changes from main in the PR commit history.

core/src/main/java/io/substrait/plan/PlanProtoConverter.java
```diff
@@ -97,6 +106,63 @@ public Plan toProto(final io.substrait.plan.Plan plan) {
 
     builder.setVersion(versionBuilder);
 
+    // Set execution behavior
+    builder.setExecutionBehavior(toProtoExecutionBehavior(plan.getExecutionBehavior()));
```
benbellick, line 109: Why are we setting the execution behavior here? I would expect that it is already part of the POJO that is being converted, no?

core/src/main/java/io/substrait/plan/Plan.java
```diff
@@ -69,4 +98,19 @@ public static ImmutableRoot.Builder builder() {
...
+    public static ImmutableExecutionBehavior.Builder builder() {
+      return ImmutableExecutionBehavior.builder();
+    }
+
+    public enum VariableEvaluationMode {
+      VARIABLE_EVALUATION_MODE_UNSPECIFIED,
+      VARIABLE_EVALUATION_MODE_PER_PLAN,
+      VARIABLE_EVALUATION_MODE_PER_RECORD
+    }
+  }
```
vbarua, line 135: There's an established pattern for representing these enums in the library that we should follow here for consistency:

https://github.com/substrait-io/substrait-java/blob/bb8d9229c6d7ef58af2c3702b44c92ca3d6773b8/core/src/main/java/io/substrait/expression/Expression.java#L1758-L1771

https://github.com/substrait-io/substrait-java/blob/bb8d9229c6d7ef58af2c3702b44c92ca3d6773b8/core/src/main/java/io/substrait/expression/Expression.java#L2161-L2174

core/src/main/java/io/substrait/plan/PlanProtoConverter.java
vbarua: Ah! `builder` here is the _protobuf_ builder. We need to set executionBehaviour on the _protobuf_ based on the value from the POJO `plan.getExecutionBehavior()`.

core/src/test/java/io/substrait/plan/PlanConverterTest.java
```diff
@@ -319,4 +342,320 @@ void nestedUserDefinedTypesShareExtensionCollector() {
...
+
+    assertNotNull(plan.getExecutionBehavior(), "Plan should have a default ExecutionBehavior");
+    assertEquals(
+        Plan.ExecutionBehavior.VariableEvaluationMode.VARIABLE_EVALUATION_MODE_PER_PLAN,
+        plan.getExecutionBehavior().getVariableEvaluationMode(),
+        "Default variable evaluation mode should be PER_PLAN");
+  }
+
+  /**
+   * Test case 6: Conversion from protobuf Plan with UNSPECIFIED mode fails validation.
```
vbarua: minor: none of the others tests you added in this file are numbered.

core/src/test/java/io/substrait/plan/PlanConverterTest.java
```diff
@@ -319,4 +342,320 @@ void nestedUserDefinedTypesShareExtensionCollector() {
...
+  }
+
+  /**
+   * Round-trip conversion with PER_PLAN mode.
+   *
+   * <p>Verifies that converting a Plan to protobuf and back preserves all data, including the
+   * execution behavior with PER_PLAN mode.
+   */
+  @Test
+  void testRoundTripWithExecutionBehaviorPerPlan() {
```
vbarua: minor: The to/from tests above feel somewhat redundant given that you also have roundtrip tests. I don't feel strongly enough to ask to keep one or the other, but thought it was worth mentioning at least.

core/src/test/java/io/substrait/plan/PlanConverterTest.java
```diff
@@ -319,4 +342,320 @@ void nestedUserDefinedTypesShareExtensionCollector() {
...
+  }
+
+  /**
+   * Round-trip conversion with PER_RECORD mode.
+   *
+   * <p>Verifies that converting a Plan with PER_RECORD mode to protobuf and back preserves the
+   * execution behavior correctly.
+   */
+  @Test
+  void testRoundTripWithExecutionBehaviorPerRecord() {
```
vbarua: minor: this test and one above differ only in the VariableEvaluationMode. This feels like a good place to use a parameterized test like:
```java
  @ParameterizedTest
  @EnumSource(
      value = Plan.ExecutionBehavior.VariableEvaluationMode.class,
      names = {"VARIABLE_EVALUATION_MODE_PER_PLAN", "VARIABLE_EVALUATION_MODE_PER_RECORD"})
  void testRoundTripWithExecutionBehaviorPerRecord(
      Plan.ExecutionBehavior.VariableEvaluationMode vem) {
```
to avoid duplication.

core/src/test/java/io/substrait/plan/PlanValidationTest.java
```diff
@@ -0,0 +1,120 @@
...
+
+  /**
+   * Test case 1: Valid execution behavior with proper variable evaluation mode.
+   *
+   * <p>This test verifies that a Plan with a properly configured ExecutionBehavior (with
+   * VariableEvaluationMode set to a valid value other than UNSPECIFIED) is successfully created
+   * without throwing any exceptions.
+   */
+  @Test
+  void testValidExecutionBehavior_PerPlan() {
```
vbarua: minor: This, with the test below, also feels like a good place for parameterized tests. Also, the `_` in the name doesn't feel very Java and isn't really consistent with this file or your other names.

core/src/test/java/io/substrait/plan/PlanConverterTest.java
```diff
@@ -319,4 +342,320 @@ void nestedUserDefinedTypesShareExtensionCollector() {
...
+  }
+
+  /**
+   * Empty plan with only execution behavior.
+   *
+   * <p>Verifies that a minimal Plan with only execution behavior (no roots, no type URLs) can be
+   * successfully converted.
+   */
+  @Test
+  void testRoundTripEmptyPlanWithExecutionBehavior() {
```
vbarua: minor: I don't if know if this provides any additional value given that you already have roundtrip tests above it.

vbarua (commented):
One final comment on the definition of `VariableEvaluationMode` which deviates from the established pattern of enum definitions in the codebase.

Also left some minor comments for your tests. I don't consider these blocking.

core/src/main/java/io/substrait/plan/Plan.java
nielspardon: I changed this accordingly. Shortened the enum names and added the proto mappings.

core/src/test/java/io/substrait/plan/PlanConverterTest.java
nielspardon: removed the numbering

core/src/test/java/io/substrait/plan/PlanConverterTest.java
nielspardon: should be addressed

core/src/test/java/io/substrait/plan/PlanConverterTest.java
nielspardon: uses parameterized tests now

core/src/test/java/io/substrait/plan/PlanValidationTest.java
nielspardon: uses parameterized tests now

core/src/test/java/io/substrait/plan/PlanConverterTest.java
nielspardon: should be addressed now

core/src/main/java/io/substrait/plan/PlanProtoConverter.java
benbellick: 🤦 ah yup... I see now. Thanks!

vbarua (approved):

vbarua:
Thanks for working on this with us.

PR #797, opened 2026-04-08
feat(core)!: bump to substrait v0.87.0

mbwhite:
Bump up to the latest version of the Substrait spec.
No code changes.

Contributes to #790.


benbellick:
Are you sure this bump doesn't bring in spec changes that need to be accounted for?

mbwhite:
Good question @benbellick so I went back and double checked. 

As far as I can see no changes are needed directly to accommodate the changes; corrections to the extension definitions will be picked up when they are used. (I've tested this locally - which to be transparent is my motivation!)

It is possible that changes could be made that would have not been possible before; though those probably best in their own PRs

benbellick:
I checked the diff and there are a few substrait spec changes that we should handle: 
- https://github.com/substrait-io/substrait/pull/1013
- https://github.com/substrait-io/substrait/pull/1014
- https://github.com/substrait-io/substrait/pull/945 (@nielspardon started working on this here https://github.com/substrait-io/substrait-java/pull/791)


No need to address all of these things here, but I think for good hygiene of this repo (and all of the substrait repos) it would be good to open tickets for these things if they don't exist already so that we can identify that they are spec gaps. Thanks!

nielspardon:
I had created issues across the 4 active libraries for this spec upgrade such as this one: https://github.com/substrait-io/substrait-java/issues/790

I tried to list out the modifications introduced in this spec version. Feel free to amend if you see something missing.

mbwhite:
thanks @benbellick @nielspardon   I've updated the title to indicate that this is **toleration** only #790 will be need for full support.  

Though please do close this if you'd rather it done 'all-in-one'

benbellick (approved):
> toleration

What does this word mean in this context? 

I'm fine with the issues @nielspardon created, thanks! This looks good to me, though I would rather not use the word `toleration` unless we have a clear definition. The title before looked good to me.

PR #812, opened 2026-04-13
fix(core): validate Plan.Root output name count

benbellick:
Adds a `Plan.Root` builder invariant for root output names.

Empty `Plan.Root.names` are still allowed for now, but emit a warning saying this will become an error in the next release. When names are present, their count is validated against the input relation's depth-first named-field count, matching the Substrait named-struct rules.

This also extracts named-field counting into a shared `NamedFieldCountingTypeVisitor`, adds dedicated tests for the documented named-struct examples, and updates the plan roundtrip fixtures to match the repo's current `Project` output semantics.


core/src/main/java/io/substrait/plan/Plan.java
```diff
@@ -65,6 +65,18 @@ public abstract static class Root {
 
     public abstract List<String> getNames();
 
+    @Value.Check
+    protected void check() {
+      final int expectedFieldCount = getInput().getRecordType().fields().size();
+      final int actualNameCount = getNames().size();
+      if (actualNameCount != expectedFieldCount) {
```
nielspardon, line 84: Something I learned about Substrait from Victor which I wasn't aware for a long time is that the names in `RelRoot` can be a nested set of names in depth-first order in case there are nested data structures in the record type:
```
message RelRoot {
  // A relation
  Rel input = 1;
  // Field names in depth-first order
  repeated string names = 2;
}
```
https://github.com/substrait-io/substrait/blob/f09071699edcb41d90e1080625e9ee25fce2b81c/proto/substrait/algebra.proto#L592-L600

which follows this:
https://substrait.io/types/named_structs/

We would probably need to make this validation more sophisticated in case someone uses more complex data structures.

benbellick: Yeah I just realized I forgot to account for that in this PR 🤦 

Thanks, working on a fix now!

core/src/main/java/io/substrait/plan/Plan.java
```diff
@@ -61,10 +64,31 @@ private static Version loadVersion() {
 
   @Value.Immutable
   public abstract static class Root {
+    private static final Logger LOGGER = LoggerFactory.getLogger(Root.class);
```
benbellick, line 67: I don't think this is a great solution. But considering that this update will result in stricter plan enforcement, I want to figure out a way where we can just warn on incorrect behavior for now, and then make it an actual error in the future. 

But I'm not sure introducing a logger here where one isn't used anywhere else makes sense.

core/src/main/java/io/substrait/plan/Plan.java
```diff
@@ -61,10 +64,31 @@ private static Version loadVersion() {
...
     public abstract List<String> getNames();
 
+    @Value.Check
+    protected void check() {
+      final int actualNameCount = getNames().size();
+      if (actualNameCount == 0) {
+        LOGGER.warn(
+            "Plan.Root built without output names; this will be an error in the next release");
+        return;
+      }
```
nielspardon, line 80: I have the feeling like we need to make the spec more clear on names being required for `Plan.Root`.

We probably also should flesh out the [Plan Root documentation](https://substrait.io/serialization/basics/#root-plans) a little bit more to describe the structure and reference to the `NamedStruct` documentation.

What do you think?

core/src/main/java/io/substrait/plan/Plan.java
nielspardon: I see. Left a comment on whether we want to make the spec more clear first. If the spec clearly says names are required then we can fire the exception straight away and we handle the change in behavior as a breaking change commit so it gets announced to consumers of substrait-java. That would then be sufficient in my opinion. Logging a warning to communicate the upcoming change is a nice gesture but who knows if consumers notice the warning message.

core/src/main/java/io/substrait/relation/VirtualTableScan.java
```diff
@@ -90,162 +90,4 @@ public <O, C extends VisitationContext, E extends Exception> O accept(
   public static ImmutableVirtualTableScan.Builder builder() {
     return ImmutableVirtualTableScan.builder();
   }
-
-  private static class NamedFieldCountingTypeVisitor
```
nielspardon, line 94: nice, I had no idea we already had such a visitor in the code base

core/src/main/java/io/substrait/plan/Plan.java
benbellick: I think you are right that the spec is a bit unclear. Maybe a more obvious validation per the current state of the spec is 
- if `names` is empty, preserve current behavior
- if `names` is non-empty, enforce that it matches depth-first output field count. 

What do you think about that?

nielspardon: We discussed this a little outside of this issue. 

Initially, I thought it was good idea to validate this but we may have wanted to sharpen the specification a bit on what the correct behavior is.

After some more consideration we identified that always requiring names to be set does not work since we have exceptions like `WriteRel`, `DdlRel` and `UpdateRel` which contain their own set of target field names and we would require them to duplicate this information into the `RelRoot.names` field which fields redundant.

Implementing conditional logic on the type of `Rel` does not work since besides the default `WriteRel`, `DdlRel` and `UpdateRel` exceptions we may deal with extension rels which represent DML- / DDL-like operations also carrying their own set of field names.

I think what we could do is the second option you suggested:

> * if `names` is non-empty, enforce that it matches depth-first output field count.

core/src/main/java/io/substrait/plan/Plan.java
benbellick: Opened a PR in upstream: https://github.com/substrait-io/substrait/pull/1101

core/src/main/java/io/substrait/plan/Plan.java
benbellick: https://github.com/substrait-io/substrait/pull/1101

I am happy to alter the PR. The current version has everything required. I _think_ this is actually okay because `WriteRel`, `UpdateRel`, and `DdlRel` do have their output schemas documented.

nielspardon: You're right. We have output schemas defined in the spec and all POJO Rels in substrait-java inherit the record type from `AbstractRel`: https://github.com/substrait-io/substrait-java/blob/594a1683bfea0f4175e3a10df8d40c8379ca4aa1/core/src/main/java/io/substrait/relation/AbstractRel.java#L27-L30

This even includes the extension Rels which I thought might have been another gap. Then we can just always enforce it.

nielspardon (approved):
LGTM, thanks for thinking this through with me.

benbellick:
> LGTM, thanks for thinking this through with me.

@Niels Pardon Thanks for the reviews and the approvals! Maybe we should wait until the [substrait](https://github.com/substrait-io/substrait/pull/1101) PR merges to get this in just to make sure we all agree?

bestbeforetoday:
@benbellick The related [substrait](https://github.com/substrait-io/substrait/pull/1101) PR has been merged so it seems this is good to go. Unfortunately there are some merge conflicts that need to be resolved first.

benbellick:
> @benbellick The related [substrait](https://github.com/substrait-io/substrait/pull/1101) PR has been merged so it seems this is good to go. Unfortunately there are some merge conflicts that need to be resolved first.

@bestbeforetoday getting around to this now! Thanks for pinging me.

nielspardon:
@benbellick you may want to merge main into your PR branch to get the OSV scanner patch also there seem to be a still some conflicts

PR #813, opened 2026-04-13
ci: use client-id instead of deprecated app-id

nielspardon:
Just noticed this deprecation warning by chance:
image
https://github.com/substrait-io/substrait-java/actions/runs/24298746175/job/70949184096

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -57,7 +57,7 @@ jobs:
       - uses: actions/create-github-app-token@v3
         id: app-token
         with:
-          app-id: ${{ secrets.RELEASER_ID }}
+          client-id: ${{ secrets.RELEASER_ID }}
           private-key: ${{ secrets.RELEASER_KEY }}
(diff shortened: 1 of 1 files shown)


benbellick:
I killed the CI job and reran it since it seemed to fail due to a network error.

.github/workflows/release.yml
```diff
@@ -57,7 +57,7 @@ jobs:
       - uses: actions/create-github-app-token@v3
         id: app-token
         with:
-          app-id: ${{ secrets.RELEASER_ID }}
```
benbellick, line 60: just for the history records, here is the relevant upstream pr: https://github.com/actions/create-github-app-token/pull/353

benbellick (approved):
LGTM once CI passes

.github/workflows/release.yml
nielspardon: oh, it's that recent. 3 days ago

PR #814, opened 2026-04-13
fix(core): correct mark join output schema

bvolpato:
## Summary

`Join.deriveRecordType()` produced the wrong output schema for `LEFT_MARK` and `RIGHT_MARK` joins:

1. **Mark column nullability is wrong.** The mark was typed `boolean NOT NULL`, but the [spec](https://substrait.io/relations/logical_relations/#join-types) defines it as `boolean NULL`. The mark is 3-valued -- `true` / `false` / `NULL` -- and a non-nullable type can't represent the "unknown match" state. This silently breaks `NOT IN` / `EXISTS` subquery lowering whenever the right side contains NULLs: the engine is forced to coerce NULL → false and returns wrong rows.

2. **`RIGHT_MARK` prepends the mark instead of appending.** The current code put the boolean in the left-side slot, so the emitted schema was `[mark, right_fields...]` instead of `[right_fields..., mark]`. Every downstream field index into the right side is off by one.

`LEFT_MARK` happened to place the column at the tail (only because the left side was emitted first), so the ordering bug was hidden there -- only the nullability was wrong.

### Fix

Refactor `deriveRecordType()` to compute `left`, `right`, and `mark` streams independently and concatenate them:

```
LEFT_MARK  -> left_fields  ++ [boolean NULL]
RIGHT_MARK -> right_fields ++ [boolean NULL]
```

### Proof

Added `JoinTest` which fails on `main` with the exact symptom:

```
RIGHT_MARK expected: [FP64, Str, I64, Bool{nullable=true}]
           but was:  [Bool{nullable=false}, FP64, Str, I64]

LEFT_MARK  expected: [I64, FP64, Str, Bool{nullable=true}]
           but was:  [I64, FP64, Str, Bool{nullable=false}]
```

and passes after the fix. Also covers proto roundtrip for both mark join types. Prior to this PR, `grep LEFT_MARK\|RIGHT_MARK` across the whole repo returned only the enum declaration and the two buggy switch cases -- zero test coverage.

Draft for visibility; happy to take review before un-drafting.

## Test plan

- [x] `./gradlew :core:test --tests 'io.substrait.relation.JoinTest'` -- 4 new tests, all pass after fix, all fail before
- [x] `./gradlew :core:test :isthmus:test` -- full core + isthmus suites green
- [x] `./gradlew build` -- full build including spark 3.4 / 3.5 / 4.0 variants green
- [x] `./gradlew spotlessApply` -- formatting clean

### DIFF
--- a/core/src/main/java/io/substrait/relation/Join.java
+++ b/core/src/main/java/io/substrait/relation/Join.java
@@ -65,7 +65,8 @@ public abstract class Join extends BiRel implements HasExtension {
   protected Type.Struct deriveRecordType() {
     Stream<Type> leftTypes = getLeftTypes();
     Stream<Type> rightTypes = getRightTypes();
-    return TypeCreator.REQUIRED.struct(Stream.concat(leftTypes, rightTypes));
+    Stream<Type> markType = getMarkType();
+    return TypeCreator.REQUIRED.struct(Stream.of(leftTypes, rightTypes, markType).flatMap(s -> s));
   }
 
   private Stream<Type> getLeftTypes() {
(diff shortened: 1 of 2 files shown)


nielspardon (approved):
LGTM, thanks for catching this

PR #816, opened 2026-04-15
chore(spark): remove unneccessary ‘compat’ functions

andrew-coleman:
When the Spark converter code was refactored to support all of spark v3.0, v3.5 and v4.0, a number of functions with incompatible APIs were abstracted into the ‘SparkCompat’ trait. In the event, `ScalarSubquery` and `Aggregate` didn’t need this treatment, so this commit restores the original (simpler) code.

### DIFF
--- a/spark/src/main/spark-4.0/io/substrait/spark/compat/SparkCompatImpl.scala
+++ b/spark/src/main/spark-4.0/io/substrait/spark/compat/SparkCompatImpl.scala
@@ -6,20 +6,6 @@ import org.apache.spark.sql.execution.datasources.{HadoopFsRelation, LogicalRela
 
 class SparkCompatImpl extends SparkCompat {
 
-  override def createScalarSubquery(plan: LogicalPlan): ScalarSubquery = {
-    // Spark 4.0 simplified constructor - no exprId needed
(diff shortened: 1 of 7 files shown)


spark/src/main/spark-4.0/io/substrait/spark/compat/SparkCompatImpl.scala
```diff
@@ -6,20 +6,6 @@ import org.apache.spark.sql.execution.datasources.{HadoopFsRelation, LogicalRela
 
 class SparkCompatImpl extends SparkCompat {
 
-  override def createScalarSubquery(plan: LogicalPlan): ScalarSubquery = {
-    // Spark 4.0 simplified constructor - no exprId needed
-    ScalarSubquery(plan)
```
bestbeforetoday, line 11: Note for others: I was confused that this single argument ScalarSubquery construction for Spark 4.0 is replaced by `ScalarSubquery(plan, exprId = NamedExpression.newExprId)`. Checking deeper, `NamedExpression.newExprId` is the default for the `exprId` parameter if none is supplied so the two are functionally equivalent.

bestbeforetoday (approved):

PR #818, opened 2026-04-15
fix(spark): correct handling of ‘remap’ property

andrew-coleman:
When running the substrait to spark converter on plans that were generated by another system, it was found that the emit/outputMappings (remap) property was not being processed correctly.  This commit fixes that.
Extra tests were added, containing substrait plans in JSON proto format that were generated by a third party system.


spark/src/main/scala/io/substrait/spark/logical/ToLogicalPlan.scala
```diff
@@ -310,17 +317,24 @@ class ToLogicalPlan(val spark: AnyRef = SparkCompat.instance.getOrCreateSparkSes
...
           .toSeq
+      }
       val projectList = if (names.size == projectExprs.size) {
         projectExprs.zip(names).map { case (expr, name) => Alias(expr, name)() }
       } else {
         projectExprs.map(toNamedExpression)
       }
       if (createProject) {
-        Project(projectList, child)
+        val ps = output.map(_.toAttribute) ++ projectList
```
nielspardon: My imagination falls a bit short in coming up with a meaning for `ps`. Is there are a more expressive variable name we can use here?

spark/src/test/scala/io/substrait/spark/SubstraitJsonSuite.scala
```diff
@@ -0,0 +1,123 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
```
nielspardon: since this is a new file and Substrait not being part of the Apache Software Foundation do we want to include a license header like this one?

spark/src/test/scala/io/substrait/spark/SubstraitJsonSuite.scala
```diff
@@ -0,0 +1,123 @@
...
+ * all the substrait plans are generated by the Spark to Substrait converter in this repo.
+ *
+ * The Substrait plan is loaded from a proto file (JSON format).
+ */
+class SubstraitJsonSuite extends SparkFunSuite {
+  var spark: SparkSession =
+    SparkSession.builder().config("spark.master", "local").enableHiveSupport().getOrCreate()
+
+  test("Aggregate") {
+    val substraitPlanPath = Path.of("../src/test/resources/substrait_plan_with_aggregate_op.json")
```
nielspardon, line 26: I'm a bit undecided on the testing approach for this specific feature. 

I feel like there would be a simpler testing approach where one could use `SubstraitBuilder` to construct Substrait plans containing remap and roundtripping them Substrait -> Spark -> Substrait without the need for going through JSON / add the test data as CSVs.

On the other hand @benbellick also added JSON plans as testcases to other modules in this repo and I have proposed adding actual execution of plans against Postgres in #700 also including test data.

Would be curious to know what others think.

nielspardon (approved):
LGTM

PR #826, opened 2026-04-30
fix(core): correct type short names for binary and fixedbinary

vbarua:
Type.Binary was emitting "binary" instead of the spec-mandated short name
"vbin", and Type.FixedBinary / ParameterizedType.FixedBinary were emitting
"fbinary" instead of "fbin". These mismatches caused function signature key
lookup failures for any function with binary or fixedbinary arguments.

### DIFF
--- a/core/src/test/java/io/substrait/function/ToTypeStringTest.java
+++ b/core/src/test/java/io/substrait/function/ToTypeStringTest.java
@@ -1,0 +1,73 @@
+package io.substrait.function;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+import io.substrait.type.Type;
+import io.substrait.type.TypeCreator;
+import java.util.stream.Stream;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.Arguments;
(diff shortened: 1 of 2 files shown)


core/src/test/java/io/substrait/function/ToTypeStringTest.java
```diff
@@ -0,0 +1,73 @@
...
+                    Arguments.of(c.decimal(10, 2), "dec"),
+                    Arguments.of(c.intervalDay(6), "iday"),
+                    Arguments.of(c.intervalCompound(3), "icompound"),
+                    Arguments.of(c.precisionTime(3), "pt"),
+                    Arguments.of(c.precisionTimestamp(3), "pts"),
+                    Arguments.of(c.precisionTimestampTZ(3), "ptstz"),
+                    Arguments.of(c.struct(c.I32), "struct"),
+                    Arguments.of(c.list(c.I32), "list"),
+                    Arguments.of(c.map(c.I32, c.I64), "map")));
+  }
```
vbarua, line 47: Not the most interesting test in world, but wanted to have something in the harness that would catch changes to these mappings.

bvolpato (approved):
LGTM, thanks Victor! 

nit: adding an integration test to the function-signature key generation path itself could be a good follow-up, instead of just focusing on the unit here

PR #846, opened 2026-06-04
feat(extensions): support deprecation info in extensions

### LINKED ISSUE #811: support deprecation info in extensions
See spec change: https://github.com/substrait-io/substrait/issues/1014

nielspardon:
BREAKING CHANGE: changes the visibility of the `ScalarFunctionVariant.resolve()` function from public to package-private

## Summary

Implements support for deprecation information in simple extensions, introduced in Substrait v0.86.0 ([substrait-io/substrait#1014](https://github.com/substrait-io/substrait/pull/1014)).

An optional `deprecated` object can now be parsed and exposed on:
- types
- scalar / aggregate / window functions (function-level)
- individual function implementations / overloads (impl-level)

The `deprecated` object carries a required `since` (core semantic version string), an optional `reason`, and optional `metadata`. As with the existing `metadata` support, consumers are not required to validate these fields — the information is surfaced so tooling can produce deprecation warnings.

Type variations are not modeled in substrait-java's extension schema (`ExtensionSignatures` only covers `types`, `scalar_functions`, `aggregate_functions`, `window_functions`), so type-variation deprecation is out of scope. I opened #847 to track adding type-variation support to the extension model (which would then also cover their deprecation info).

## Details

- New `SimpleExtension.DeprecationStatus` immutable value type (`since`, `reason`, `metadata`), following the existing `Option` / `VariadicBehavior` pattern.
- Added `Optional<DeprecationStatus> deprecated()` to `Type`, the `ScalarFunction`/`AggregateFunction`/`WindowFunction` parent classes, and the `Function` base class (per impl).
- Function-level deprecation is propagated into each resolved variant. When both a function-level and an impl-level deprecation are present, the impl-level value takes precedence. Aggregate functions registered as window functions retain their deprecation automatically.

## Testing

- New `DeprecationExtensionTest` (7 tests) covering type, function-level, and impl-level deprecation, deprecation `metadata`, an overload-specific deprecation where a sibling overload is not deprecated, and carry-over to aggregate-as-window functions.
- New `deprecation_extensions.yaml` test resource.

Closes #811

### DIFF
--- a/core/src/main/java/io/substrait/extension/SimpleExtension.java
+++ b/core/src/main/java/io/substrait/extension/SimpleExtension.java
@@ -124,7 +124,6 @@ public class SimpleExtension {
 
   /**
    * Deprecation information for an extension entry (type, function, or function implementation).
-   * See <a href="https://github.com/substrait-io/substrait/pull/1014">substrait#1014</a>.
    *
    * <p>Consumers of extension files are not required to understand or validate deprecation fields;
    * the information is provided so tooling can surface deprecation warnings.
--- a/core/src/main/java/io/substrait/extension/SimpleExtension.java
+++ b/core/src/main/java/io/substrait/extension/SimpleExtension.java
@@ -122,6 +122,32 @@ public class SimpleExtension {
     List<String> getValues();
   }
 
+  /**
+   * Deprecation information for an extension entry (type, function, or function implementation).
+   * See <a href="https://github.com/substrait-io/substrait/pull/1014">substrait#1014</a>.
+   *
+   * <p>Consumers of extension files are not required to understand or validate deprecation fields;
+   * the information is provided so tooling can surface deprecation warnings.
+   */
+  @JsonDeserialize(as = ImmutableSimpleExtension.DeprecationStatus.class)
+  @JsonSerialize(as = ImmutableSimpleExtension.DeprecationStatus.class)
+  @JsonIgnoreProperties(ignoreUnknown = true)
+  @Value.Immutable
+  public interface DeprecationStatus {
+    /**
+     * The version at which the entry was deprecated, as a core semantic version string (e.g. {@code
+     * "1.2.0"}).
+     */
+    @JsonProperty(required = true)
+    String since();
+
+    /** Optional human-readable description of why the entry was deprecated. */
+    Optional<String> reason();
+
+    /** Optional arbitrary data provided by the extension author. */
+    Optional<Map<String, Object>> metadata();
+  }
+
   @JsonSerialize(as = ImmutableSimpleExtension.ValueArgument.class)
   @JsonDeserialize(as = ImmutableSimpleExtension.ValueArgument.class)
   @Value.Immutable
@@ -280,6 +306,8 @@ public class SimpleExtension {
 
     public abstract Optional<Map<String, Object>> metadata();
 
+    public abstract Optional<DeprecationStatus> deprecated();
+
     public List<Argument> requiredArguments() {
       return requiredArgsSupplier.get();
     }
@@ -387,10 +415,13 @@ public class SimpleExtension {
 
     public abstract Optional<Map<String, Object>> metadata();
 
+    public abstract Optional<DeprecationStatus> deprecated();
+
     public abstract List<ScalarFunctionVariant> impls();
 
     public Stream<ScalarFunctionVariant> resolve(String urn) {
-      return impls().stream().map(f -> f.resolve(urn, name(), description(), metadata()));
+      return impls().stream()
+          .map(f -> f.resolve(urn, name(), description(), metadata(), deprecated()));
     }
   }
 
@@ -399,7 +430,11 @@ public class SimpleExtension {
   @Value.Immutable
   public abstract static class ScalarFunctionVariant extends Function {
     public ScalarFunctionVariant resolve(
-        String urn, String name, String description, Optional<Map<String, Object>> metadata) {
+        String urn,
+        String name,
+        String description,
+        Optional<Map<String, Object>> metadata,
+        Optional<DeprecationStatus> deprecated) {
       return ImmutableSimpleExtension.ScalarFunctionVariant.builder()
           .urn(urn)
           .name(name)
@@ -408,6 +443,7 @@ public class SimpleExtension {
           .args(args())
           .options(options())
           .metadata(metadata)
+          .deprecated(deprecated().isPresent() ? deprecated() : deprecated)
           .ordered(ordered())
           .variadic(variadic())
           .returnType(returnType())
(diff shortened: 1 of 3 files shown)


core/src/main/java/io/substrait/extension/SimpleExtension.java
```diff
@@ -122,6 +122,32 @@ public interface Option {
     List<String> getValues();
   }
 
+  /**
+   * Deprecation information for an extension entry (type, function, or function implementation).
+   * See <a href="https://github.com/substrait-io/substrait/pull/1014">substrait#1014</a>.
```
vbarua: I don't think it's worth linking to the PR where this was introduced. The Javadoc should stand on it's own, and it already does.

core/src/main/java/io/substrait/extension/SimpleExtension.java
```diff
@@ -399,7 +430,11 @@ public Stream<ScalarFunctionVariant> resolve(String urn) {
   @Value.Immutable
   public abstract static class ScalarFunctionVariant extends Function {
     public ScalarFunctionVariant resolve(
-        String urn, String name, String description, Optional<Map<String, Object>> metadata) {
+        String urn,
+        String name,
+        String description,
+        Optional<Map<String, Object>> metadata,
+        Optional<DeprecationStatus> deprecated) {
```
vbarua: As this is a public API, we should note the breaking change in the description.

core/src/main/java/io/substrait/extension/SimpleExtension.java
```diff
@@ -408,6 +443,7 @@ public ScalarFunctionVariant resolve(
           .args(args())
           .options(options())
           .metadata(metadata)
+          .deprecated(deprecated().isPresent() ? deprecated() : deprecated)
```
vbarua: If I understand what we're doing here, `resolve` is `ScalarFunction#resolve` which gets called with the function-level deprecation status. When we create the ScalarFunctionVariant, we first check if _it_ has an impl level deprecation status, and if not just use the function-level one.

This is definitely a little lossy in terms of representation, but works reasonable well. Is the reason to inject the deprecation status like this to make it easy to check if ScalarFunctionVariant is deprecated directly without having to access the ScalarFunction holding it? If so, could we do something like inject the function-level deprecation status and store _both_ on the ScalarFunctionImp? That way we still check the deprecation status easily with a method like `isDeprecated` that looks at the function-level and impl-level deprecation, but when we serialize ScalarFunctionImpl we can ignore the function-level deprecation to avoid setting it for every impl.

vbarua (commented):
Left some minor comments but overall seems reasonable.

core/src/main/java/io/substrait/extension/SimpleExtension.java
nielspardon: > This is definitely a little lossy in terms of representation, but works reasonable well. Is the reason to inject the deprecation status like this to make it easy to check if ScalarFunctionVariant is deprecated directly without having to access the ScalarFunction holding it?

I think the AI wanted to follow the existing pattern where we already copy the function name, description and metadata over from the function (`ScalarFunction`) to the impl (`ScalarFunctionVariant`). We could store a pointer to the `ScalarFunction` inside the `ScalarFunctionVariant` in general for accessing the function level metadata from the impl more easily.

nielspardon: The current behavior is already lossy since it overwrites the impl description with the function description.

nielspardon: I also spent a bit more time trying to understand the existing code. What seems to happen is that we are deserializing the extensions YAMLs into `SimpleExtension.ExtensionCollection` which then contains most of the structure in the YAML and then the `resolve()` method copies some of the function level details to the impl level `FunctionVariant` classes for easier consumption like `name`, `description`, `metadata`, `urn`, etc.

It feels to me like the resolve step is a bit redundant and we could add pointers to the parent fields instead.

core/src/main/java/io/substrait/extension/SimpleExtension.java
nielspardon: I removed that PR link from the comments. It suspect the AI wanted to follow an existing practice since this file contains another PR reference: https://github.com/substrait-io/substrait-java/blob/594a1683bfea0f4175e3a10df8d40c8379ca4aa1/core/src/main/java/io/substrait/extension/SimpleExtension.java#L180-L182

core/src/main/java/io/substrait/extension/SimpleExtension.java
nielspardon: I'm not even sure why this specific method is public while `AggregateFunctionVariant.resolve()` and `WindowFunctionVariant` are package private.

nielspardon: I'm tempted to change the visibility of that method to be in line with the others methods and mark the change of visibility as a breaking change. What do you think?

core/src/main/java/io/substrait/extension/SimpleExtension.java
vbarua: > It feels to me like the resolve step is a bit redundant and we could add pointers to the parent fields instead.

That seems like a better design IMO. I vaguely recalled that `resolve` was a mechanism to inject context from higher up in the YAML into child objects. Turns out I made some modifications around this like 3 years ago https://github.com/substrait-io/substrait-java/pull/149

If we just had pointers to the parents, we could avoid it entirely.

nielspardon: ok, I have a refactoring locally already. can push that up. maybe also include #848 while we refactor this area.

core/src/main/java/io/substrait/extension/SimpleExtension.java
vbarua: Now that you point it out, it is weird that this is `public`. Hiding this makes sense to me.

core/src/main/java/io/substrait/extension/SimpleExtension.java
vbarua: > The current behavior is already lossy since it overwrites the impl description with the function description.

Given that it's already lossy, I would be happy to see this merged and then have the refactor as a follow-up. No need to block this change, and it will make the refactor PR smaller.

vbarua (commented):
Circling back, if we hide the public `resolve` method I would be happy to see this shipped as is.

vbarua (approved):

core/src/main/java/io/substrait/extension/SimpleExtension.java
nielspardon: I changed the visibility and marked the visibility change as the breaking one

PR #848, opened 2026-06-04
feat(extensions): support impl-level description in extensions

### LINKED ISSUE #810: support optional description field to function implementations
See spec change: https://github.com/substrait-io/substrait/issues/1013

nielspardon:
## Summary

Substrait v0.86.0 added an optional `description` field to individual function implementations ([substrait#1013](https://github.com/substrait-io/substrait/pull/1013)). An impl-level description documents overload-specific behavior and can appear on each scalar/aggregate/window function implementation.

The variant already inherits `description()` from `Function`, so the impl-level value was being parsed but then discarded: `resolve()` unconditionally overwrote it with the parent function's description. This PR makes the impl-level description take precedence over the parent function's when present, mirroring the impl-level precedence used for deprecation info.

Closes #810

## Changes

- Add `Function.resolveDescription(...)` helper that prefers a non-empty impl-level description over the parent function's, and wire it into the scalar/aggregate/window `resolve()` methods.
- Add `DescriptionExtensionTest` and a `description_extensions.yaml` fixture covering function-level-only (inherited), impl-level overriding function-level, sibling overloads (one overriding, one inheriting), impl-level with no function-level, and aggregate-as-window carry-over.

## Testing

- `./gradlew :core:test` passes (including the new test).
- `./gradlew :core:spotlessCheck` passes.

### DIFF
--- a/core/src/main/java/io/substrait/extension/SimpleExtension.java
+++ b/core/src/main/java/io/substrait/extension/SimpleExtension.java
@@ -277,8 +277,7 @@ public class SimpleExtension {
     /**
      * Resolves the effective description for this implementation. An implementation-level
      * description (parsed into {@link #description()}) documents overload-specific behavior and
-     * takes precedence over the parent function's description when present. See <a
-     * href="https://github.com/substrait-io/substrait/pull/1013">substrait#1013</a>.
+     * takes precedence over the parent function's description when present.
      */
     String resolveDescription(String functionDescription) {
       String implDescription = description();
--- a/core/src/main/java/io/substrait/extension/SimpleExtension.java
+++ b/core/src/main/java/io/substrait/extension/SimpleExtension.java
@@ -274,6 +274,19 @@ public class SimpleExtension {
       return "";
     }
 
+    /**
+     * Resolves the effective description for this implementation. An implementation-level
+     * description (parsed into {@link #description()}) documents overload-specific behavior and
+     * takes precedence over the parent function's description when present. See <a
+     * href="https://github.com/substrait-io/substrait/pull/1013">substrait#1013</a>.
+     */
+    String resolveDescription(String functionDescription) {
+      String implDescription = description();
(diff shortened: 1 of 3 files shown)


nielspardon:
Claude thought it makes sense to overwrite the function level description with the impl specific description. I guess this is reasonable.

core/src/main/java/io/substrait/extension/SimpleExtension.java
```diff
@@ -274,6 +274,19 @@ public String description() {
       return "";
     }
 
+    /**
+     * Resolves the effective description for this implementation. An implementation-level
+     * description (parsed into {@link #description()}) documents overload-specific behavior and
+     * takes precedence over the parent function's description when present. See <a
+     * href="https://github.com/substrait-io/substrait/pull/1013">substrait#1013</a>.
+     */
+    String resolveDescription(String functionDescription) {
```
vbarua, line 282: I think the Javadoc is descriptive enough without referencing the PR.

core/src/test/java/io/substrait/extension/DescriptionExtensionTest.java
```diff
@@ -0,0 +1,93 @@
...
+import java.io.UncheckedIOException;
+import org.junit.jupiter.api.Test;
+
+/**
+ * Verifies that descriptions can be read from extension YAML files at both the function level and
+ * the function-implementation level. An implementation-level description documents
+ * overload-specific behavior and takes precedence over the parent function's description when both
+ * are present.
+ *
+ * <p>See <a href="https://github.com/substrait-io/substrait/pull/1013">substrait#1013</a>.
```
vbarua: I don't think the reference to the PR adds any value.

vbarua (approved):
Left some minor comments but other looks good to me.

This has the same lossy behaviour as https://github.com/substrait-io/substrait-java/pull/846 (overwriting the description in some cases), but is consistent with what we do today and we can improve it as a followup.

PR #850, opened 2026-06-05
ci: remove PR title check comment when check passes

nielspardon:
## Summary

The PR Title Check workflow posts an **ACTION NEEDED** comment when the PR title/description doesn't follow Conventional Commits, but it never removed that comment once the title was fixed and the check passed. This updates the workflow so the comment is deleted when the check passes.

## Changes

- Collapse the previous `failure()`-only logic into a single `github-script` step that branches on the commitlint outcome (`continue-on-error: true` + `core.setFailed(...)` keeps the check red on failure).
- When the check **passes**, delete any previously posted comment; when it **fails**, post the comment (if not already present).
- Identify the bot comment via a hidden `<!-- pr-title-check -->` marker rather than exact body matching, so detection/cleanup survives wording or whitespace changes.
- Declare `permissions: pull-requests: write` so the workflow can delete comments.

## Notes

Cleanup is marker-based, so it applies to comments posted by the new version going forward; pre-existing comments from the old workflow won't be auto-removed.

### DIFF
--- a/.github/workflows/pr_title.yaml
+++ b/.github/workflows/pr_title.yaml
@@ -3,6 +3,10 @@ name: PR Title Check
 on:
   pull_request_target:
     types: [opened, edited, synchronize, reopened]
+
+permissions:
+  pull-requests: write
+
 jobs:
   commitlint:
     name: PR title / description conforms to semantic-release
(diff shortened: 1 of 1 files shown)


nielspardon:
rebased on latest main and fixed the conflicts

bestbeforetoday (approved):
Normally I would want to see elevated permissions defined only for the job that need those permissions but this workflow has only one job so permissions at the top level seem OK.

A valid question was raised in the substrait-python repository about whether we need the comment at all, particularly as we then want to go to the trouble of removing it. That can be considered for the future though. This PR seems to add value by removing the visual clutter of a comment that is no longer applicable.

PR #854, opened 2026-06-08
ci: align commitlint config across pre-commit and CI

### LINKED ISSUE #640: CI: loosen commitlint restrictions

nielspardon:
## Summary

The PR title check workflow generated its own throwaway commitlint config inline, while `.commitlintrc.js` (used by the local pre-commit hook) maintained a separate, stricter rule set. The two diverged, so local commits were validated differently than PR titles in CI.

This makes `.commitlintrc.js` the single source of truth, mirroring the [substrait](https://github.com/substrait-io/substrait) repo:

- **`.commitlintrc.js`** now `extends` `@commitlint/config-conventional` (which supplies the same type list, case, and length rules that were previously hand-maintained), loosens the body/footer length rules, and ignores dependabot messages — including the `Updates the requirements on … to permit the latest version.` pattern that substrait has but we were missing.
- **`.github/workflows/pr_title.yaml`** no longer generates a config inline. It checks out the repo and runs `commitlint` against the checked-in config. Under `pull_request_target`, `actions/checkout` defaults to the trusted base branch, so a PR cannot supply its own commitlint rules.
- **`.pre-commit-config.yaml`** gains `additional_dependencies: ['@commitlint/config-conventional']` so the local commit-msg hook can resolve the now-extended config (the hook only bundles `@commitlint/cli`).

## Notes

Local commits now use the same loosened length rules as CI rather than the old strict 100-char limits — that is the intended convergence, since CI was the effective authority anyway. The commitizen `prompt` block was dropped to match substrait's config.

Closes #640

### DIFF
--- a/.commitlintrc.js
+++ b/.commitlintrc.js
@@ -1,128 +1,13 @@
 module.exports = {
+  "extends": ["@commitlint/config-conventional"],
   "rules": {
-    "body-leading-blank": [1, "always"],
-    "body-max-line-length": [2, "always", 100],
-    "footer-leading-blank": [1, "always"],
-    "footer-max-line-length": [2, "always", 100],
-    "header-max-length": [2, "always", 100],
-    "subject-case": [
-      2,
-      "never",
-      ["sentence-case", "start-case", "pascal-case", "upper-case"]
-    ],
-    "subject-empty": [2, "never"],
-    "subject-full-stop": [2, "never", "."],
-    "type-case": [2, "always", "lower-case"],
-    "type-empty": [2, "never"],
-    "type-enum": [
-      2,
-      "always",
-      [
-        "build",
-        "chore",
-        "ci",
-        "docs",
-        "feat",
-        "fix",
-        "perf",
-        "refactor",
-        "revert",
-        "style",
-        "test"
-      ]
-    ]
-  },
-  "prompt": {
-    "questions": {
-      "type": {
-        "description": "Select the type of change that you're committing",
-        "enum": {
-          "feat": {
-            "description": "A new feature",
-            "title": "Features",
-            "emoji": "✨"
-          },
-          "fix": {
-            "description": "A bug fix",
-            "title": "Bug Fixes",
-            "emoji": "🐛"
-          },
-          "docs": {
-            "description": "Documentation only changes",
-            "title": "Documentation",
-            "emoji": "📚"
-          },
-          "style": {
-            "description": "Changes that do not affect the meaning of the code (whitespace, formatting, missing semi-colons, etc)",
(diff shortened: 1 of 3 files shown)


nielspardon:
This should make the configuration consistent. If I read #640 correctly then we may actually remove the pre-commit configuration for commitlint since Ben suggests to not check commits but PR title / description only.

.commitlintrc.js
```diff
@@ -1,128 +1,13 @@
...
-    }
+    "body-max-line-length": [0, "always", Infinity],
+    "footer-max-line-length": [0, "always", Infinity],
+    "body-leading-blank": [0, "always"]
   },
   // Workaround for https://github.com/dependabot/dependabot-core/issues/5923
-  "ignores": [(message) => /^build\(deps\): bump .+ from .+ to .+$/s.test(message)]
+  "ignores": [
+    (message) => /^Bumps \[.+]\(.+\) from .+ to .+\.$/m.test(message),
+    (message) => /^Updates the requirements on .+ to permit the latest version\.$/m.test(message)
```
bestbeforetoday, line 11: This ignore doesn't look to have been used before. Is it necessary?

.github/workflows/pr_title.yaml
```diff
@@ -11,18 +11,9 @@ jobs:
       - uses: actions/setup-node@v6
         with:
           node-version: "24"
+      - uses: actions/checkout@v6
```
bestbeforetoday, line 14: This will certainly work but, if we only need the `.commitlintrc.js` file, we could avoid the work of checking out all the other repository files and submodules.

https://github.com/actions/checkout#fetch-only-a-single-file

bestbeforetoday (commented):
Looks like a good change to make. A couple of queries inline.

.github/workflows/pr_title.yaml
nielspardon: I changed this to a sparse checkout

.commitlintrc.js
nielspardon: Dependabot does not consistently use the same commit message style across all package types.

- Python packages from a dev dependency group use the later format, e.g. https://github.com/substrait-io/substrait/pull/1093
- Python packages from the default dependency use the former format, e.g. https://github.com/substrait-io/substrait/pull/1094

bestbeforetoday (approved):

PR #857, opened 2026-06-09
feat(core)!: remove AggregateRel.Grouping.grouping_expressions

### LINKED ISSUE #805: remove previously deprecated `Aggregate.Grouping.grouping_expressions` field

nielspardon:
BREAKING CHANGE: removes the deprecated Aggregate.Grouping.grouping_expressions field which was removed in Substrait v0.88.0

Closes #805 

I'm not sure how to best introduce these changes. We could update the Git submodule first which then breaks a whole lot of code which we would all need to migrate in one big PR or we gradually introduce the changes in PRs like this one and then update the git submodule.


vbarua (approved):
Removing this before the submodule bump makes sense to me.

This library has been dual writing the groupings outputs for a while now, so users should be able to migrate their consumers to handle the new grouping expression format if they haven't already.

PR #861, opened 2026-06-10
feat(core): add Substrait dialect support

nielspardon:
## What

Adds a typed model in the `core` Java SDK for **creating** and **consuming** Substrait *dialect* YAML files, faithful to `substrait/text/dialect_schema.yaml` (introduced in substrait v0.76.0). Until now the only producer of a dialect in this repo was the Scala `DialectGenerator` in the `spark` module, whose ad-hoc model isn't reusable.

## Usage

Build a dialect, serialize it to YAML, and parse it back:

```java
import io.substrait.dialect.Dialect;
import io.substrait.dialect.Dialect.*;

DialectDocument dialect =
    DialectDocument.builder()
        .name("Example Dialect")
        .putDependencies("arithmetic", "extension:io.substrait:functions_arithmetic")
        // Types: a bare entry and a configured one.
        .addSupportedTypes(SupportedType.of(TypeKind.BOOL))
        .addSupportedTypes(
            SupportedType.builder().type(TypeKind.PRECISION_TIMESTAMP).maxPrecision(9).build())
        // Relations: a bare entry and one carrying configuration.
        .addSupportedRelations(SupportedRelation.of(RelationKind.FILTER))
        .addSupportedRelations(
            SupportedRelation.builder()
                .relation(RelationKind.JOIN)
                .addJoinTypes(JoinType.INNER, JoinType.LEFT)
                .build())
        // Functions reference a dependency alias.
        .addSupportedScalarFunctions(
            DialectFunction.builder()
                .source("arithmetic")
                .name("add")
                .systemMetadata(
                    SystemFunctionMetadata.builder().name(\"+\").notation(Notation.INFIX).build())
                .addSupportedImpls("i32_i32", "i64_i64")
                .build())
        .build();

String yaml = Dialect.toYaml(dialect);          // create
DialectDocument parsed = Dialect.load(yaml);     // consume
```

The configuration-free entries serialize as bare enum strings and the configured ones as mappings:

```yaml
---
name: "Example Dialect"
dependencies:
  arithmetic: "extension:io.substrait:functions_arithmetic"
supported_types:
- "BOOL"
- type: "PRECISION_TIMESTAMP"
  max_precision: 9
supported_relations:
- "FILTER"
- relation: "JOIN"
  join_types:
  - "INNER"
  - "LEFT"
supported_scalar_functions:
- source: "arithmetic"
  name: "add"
  system_metadata:
    name: "+"
    notation: "INFIX"
  supported_impls:
  - "i32_i32"
  - "i64_i64"
```

A fuller example covering every union and configuration option lives in [`DialectRoundTripTest`](https://github.com/substrait-io/substrait-java/blob/par-dialect/core/src/test/java/io/substrait/dialect/DialectRoundTripTest.java).

## Design

- New `io.substrait.dialect` package with a `@Value.Enclosing` `Dialect` holder and nested Immutables types, mirroring the existing `SimpleExtension` pattern (Jackson + `@Value.Immutable`, static `load(...)`/`toYaml(...)` helpers).
- The three polymorphic unions (`supported_types` / `supported_relations` / `supported_expressions`) are `oneOf [bare-enum-string | config-object]` in the schema. Each is modeled as **one enum-tag class per category** (`SupportedType`/`SupportedRelation`/`SupportedExpression`) carrying a dialect-local kind enum plus typed config fields. Custom Jackson (de)serializers collapse config-free entries to a bare enum string and expand configured ones to objects.
- Config sub-enums (`JoinType`, `SetOperation`, ...) are **dialect-local** with exactly the schema's constants, keeping the dialect vocabulary decoupled from the relational-algebra model (whose `Join.JoinType`/`Set.SetOp` carry extra `UNKNOWN`/deprecated values).
- The existing `Type`/`Rel`/`Expression` hierarchies model full algebra *instances* — the wrong abstraction level for capability *tags* — so they are intentionally not reused for the kind enums.

## Validation & tests

- Schema validation is **test-scope only** (`networknt json-schema-validator`); the published `core` jar gains no new runtime dependency.
- `processTestResources` copies the dialect schema, the published `spark_dialect.yaml`, and the spec's per-section dialect fixtures onto the test classpath.
- Tests: a schema-validated build -> serialize -> validate -> parse -> assert-equal round-trip; bare-string collapse behavior; parsing/re-validating the real Spark dialect; and a parameterized round-trip over all five spec fixtures (`types`, `relations`, `expressions`, `functions`, `execution_behavior`).

The `spark` module is left unchanged; migrating its `DialectGenerator` onto this model is a natural follow-up.

🤖 Generated with AI


core/src/main/java/io/substrait/dialect/Dialect.java
```diff
@@ -0,0 +1,552 @@
...
+
+  private static ObjectMapper objectMapper() {
+    return new ObjectMapper(new YAMLFactory())
+        .registerModule(new Jdk8Module())
+        .enable(DeserializationFeature.ACCEPT_SINGLE_VALUE_AS_ARRAY)
+        // Omit absent Optionals and empty collections so that unset sections are not emitted.
+        // The custom (de)serializers for the polymorphic unions write their fields explicitly and
+        // are unaffected by this inclusion setting.
+        .setDefaultPropertyInclusion(JsonInclude.Include.NON_EMPTY);
+  }
```
bestbeforetoday: ObjectMapper is threadsafe and designed for reuse. Perhaps a single shared instance should be used instead of creating a new one each time.

core/src/main/java/io/substrait/dialect/Dialect.java
```diff
@@ -0,0 +1,552 @@
...
+ * <p>A dialect describes what a target execution system supports: which types, relations,
+ * expressions and functions, along with per-feature configuration (join types, set operations, cast
+ * failure behavior, ...). The model maps the schema published at {@code
+ * substrait/text/dialect_schema.yaml}.
+ *
+ * <p>Build a dialect with the nested builders and serialize it with {@link
+ * #toYaml(DialectDocument)}; parse one with {@link #load(String)} and friends.
+ */
+@Value.Enclosing
+public class Dialect {
```
bestbeforetoday: Dialect seems to be only used to provide an enclosing namespace for the nested classes. The `io.substrait.dialect` package is already providing a namespace for all the dialect handling. The enclosing Dialect class results in naming like `io.substrait.dialect.Dialect.DialectDocument`, which has a lot of repetition. Would it be practical to promote the nested classes to top-level classes within the dialect package? Perhaps DialectDocument could then be renamed to simply Dialect (or Document). The handful of static factory methods on Dialect that currently return Dialect.DialectDocument instances would then become factory methods on Dialect and return Dialect. This seems like a more typical static factory method pattern.

core/src/main/java/io/substrait/dialect/Dialect.java
```diff
@@ -0,0 +1,552 @@
...
+  }
+
+  /** Parse a dialect from a YAML stream. */
+  public static DialectDocument load(InputStream stream) {
+    try (Scanner scanner = new Scanner(stream, StandardCharsets.UTF_8.name())) {
+      scanner.useDelimiter(READ_WHOLE_FILE);
+      String content = scanner.hasNext() ? scanner.next() : "";
+      return load(content);
+    }
+  }
```
bestbeforetoday: ObjectMapper can read values directly from an InputStream. We do not need to read the entire InputStream content into a String first.

core/src/main/java/io/substrait/dialect/Dialect.java
```diff
@@ -0,0 +1,552 @@
...
+    try {
+      return objectMapper().readValue(content, DialectDocument.class);
+    } catch (IOException e) {
+      throw new IllegalStateException(e);
+    }
+  }
+
+  /** Parse a dialect from a YAML stream. */
+  public static DialectDocument load(InputStream stream) {
+    try (Scanner scanner = new Scanner(stream, StandardCharsets.UTF_8.name())) {
```
bestbeforetoday: I think the close of the Scanner at the end of the try-with-resources will also close the underlying InputStream. The caller might not expect this.

core/src/main/java/io/substrait/dialect/DialectJsonSupport.java
```diff
@@ -0,0 +1,81 @@
...
+
+  private DialectJsonSupport() {}
+
+  /** Read the {@code metadata} object on a node into a map, or {@code null} when absent. */
+  static Map<String, Object> readMetadata(JsonParser p, JsonNode node) {
+    JsonNode metadata = node.get("metadata");
+    if (metadata == null || metadata.isNull()) {
+      return null;
+    }
+    return ((ObjectMapper) p.getCodec()).convertValue(metadata, MAP_TYPE);
```
bestbeforetoday: The unchecked cast has the potential to cause unexpected runtime failures if other parts of the code are changed and the codec ends up not being an ObjectMapper. If all the dialect code used a single shared default (package) scoped ObjectMapper instance, perhaps that could be accessed directly by this code instead of calling `getCodec` on the parser.

bestbeforetoday: Similar pattern in SupportedTypeDeserializer.

core/src/main/java/io/substrait/dialect/DialectJsonSupport.java
```diff
@@ -0,0 +1,81 @@
...
+  }
+
+  /**
+   * Read a YAML/JSON array node into a list of enum constants. Missing nodes yield an empty list.
+   */
+  static <E extends Enum<E>> List<E> readEnums(JsonNode node, Class<E> type) {
+    List<E> result = new ArrayList<>();
+    if (node != null && node.isArray()) {
```
bestbeforetoday: `ACCEPT_SINGLE_VALUE_AS_ARRAY` is set on the ObjectMapper so Jackson's own deserializers will treat single values as arrays but this code requires an array and will silently ignore a single value. I am not sure whether this is an issue, just flagging a potential inconsistency in behavior.

core/src/main/java/io/substrait/dialect/SupportedExpressionSerializer.java
```diff
@@ -0,0 +1,45 @@
...
+      DialectJsonSupport.writeEnumArray(gen, "nested_types", value.nestedTypes());
+    }
+    if (!value.variableTypes().isEmpty()) {
+      DialectJsonSupport.writeEnumArray(gen, "variable_types", value.variableTypes());
+    }
+    // The schema forbids `metadata` on EXECUTION_CONTEXT_VARIABLE, so only emit it for others.
+    if (value.metadata().isPresent()
+        && value.expression() != ExpressionKind.EXECUTION_CONTEXT_VARIABLE) {
+      DialectJsonSupport.writeMetadata(gen, provider, value.metadata().get());
+    }
```
bestbeforetoday: SupportedExpressionDeserializer always reads metadata whereas the serialization here can exclude it. This means de/serialization is not always symmetric. Not sure if it is an issue or if checks when building a SupportedExpression should prevent it.

core/src/main/java/io/substrait/dialect/SupportedRelationSerializer.java
```diff
@@ -0,0 +1,54 @@
...
+      DialectJsonSupport.writeEnumArray(gen, "read_types", value.readTypes());
+    }
+    if (!value.operations().isEmpty()) {
+      DialectJsonSupport.writeEnumArray(gen, "operations", value.operations());
+    }
+    if (!value.writeTypes().isEmpty()) {
+      DialectJsonSupport.writeEnumArray(gen, "write_types", value.writeTypes());
+    }
+    if (!value.ddlWriteTypes().isEmpty()) {
+      DialectJsonSupport.writeEnumArray(gen, "write_types", value.ddlWriteTypes());
```
bestbeforetoday, line 38: If both values are present then they are both written to the same JSON field.

core/src/main/java/io/substrait/dialect/Dialect.java
```diff
@@ -0,0 +1,552 @@
...
+      scanner.useDelimiter(READ_WHOLE_FILE);
+      String content = scanner.hasNext() ? scanner.next() : "";
+      return load(content);
+    }
+  }
+
+  /** Parse a dialect from a YAML file on disk. */
+  public static DialectDocument loadFromFile(Path path) {
+    try {
+      return load(new String(Files.readAllBytes(path), StandardCharsets.UTF_8));
```
bestbeforetoday: ObjectMapper can read values directly from an File. We do not need to read the entire file content into a String first.

core/src/test/java/io/substrait/dialect/SpecDialectFixturesTest.java
```diff
@@ -0,0 +1,66 @@
...
+
+  private static String readResource(String resourcePath) {
+    try (InputStream stream = SpecDialectFixturesTest.class.getResourceAsStream(resourcePath)) {
+      if (stream == null) {
+        throw new IllegalStateException("Fixture not found on classpath: " + resourcePath);
+      }
+      try (Scanner scanner = new Scanner(stream, StandardCharsets.UTF_8.name())) {
+        scanner.useDelimiter("\\A");
+        return scanner.hasNext() ? scanner.next() : "";
+      }
```
bestbeforetoday: Test code compiles and runs with a newer Java version so you should be able to use InputStream.readAllBytes().

core/src/main/java/io/substrait/dialect/SupportedExpressionDeserializer.java
```diff
@@ -0,0 +1,46 @@
...
+import io.substrait.dialect.Dialect.SupportedExpression;
+import java.io.IOException;
+import java.util.Map;
+
+/**
+ * Deserializes a {@code supported_expressions} entry, which is either a bare enum string (e.g.
+ * {@code LITERAL}) or a configuration object (e.g. {@code {expression: CAST, failure_options:
+ * [RETURN_NULL]}}).
+ */
+public class SupportedExpressionDeserializer extends JsonDeserializer<SupportedExpression> {
```
bestbeforetoday: Perhaps could be default (package) scope instead of exposing in the public API.

core/src/main/java/io/substrait/dialect/SupportedExpressionSerializer.java
```diff
@@ -0,0 +1,45 @@
...
+import com.fasterxml.jackson.databind.SerializerProvider;
+import io.substrait.dialect.Dialect.ExpressionKind;
+import io.substrait.dialect.Dialect.SupportedExpression;
+import java.io.IOException;
+
+/**
+ * Serializes a {@code supported_expressions} entry as a bare enum string when it carries no
+ * configuration, or as a configuration object otherwise.
+ */
+public class SupportedExpressionSerializer extends JsonSerializer<SupportedExpression> {
```
bestbeforetoday: Perhaps could be default (package) scope instead of exposing in the public API.

core/src/main/java/io/substrait/dialect/SupportedRelationDeserializer.java
```diff
@@ -0,0 +1,56 @@
...
+import io.substrait.dialect.Dialect.RelationKind;
+import io.substrait.dialect.Dialect.SupportedRelation;
+import java.io.IOException;
+import java.util.Map;
+
+/**
+ * Deserializes a {@code supported_relations} entry, which is either a bare enum string (e.g. {@code
+ * FILTER}) or a configuration object (e.g. {@code {relation: JOIN, join_types: [INNER]}}).
+ */
+public class SupportedRelationDeserializer extends JsonDeserializer<SupportedRelation> {
```
bestbeforetoday: Perhaps could be default (package) scope instead of exposing in the public API.

core/src/main/java/io/substrait/dialect/SupportedRelationSerializer.java
```diff
@@ -0,0 +1,54 @@
...
+import com.fasterxml.jackson.databind.JsonSerializer;
+import com.fasterxml.jackson.databind.SerializerProvider;
+import io.substrait.dialect.Dialect.SupportedRelation;
+import java.io.IOException;
+
+/**
+ * Serializes a {@code supported_relations} entry as a bare enum string when it carries no
+ * configuration, or as a configuration object otherwise.
+ */
+public class SupportedRelationSerializer extends JsonSerializer<SupportedRelation> {
```
bestbeforetoday: Perhaps could be default (package) scope instead of exposing in the public API.

core/src/main/java/io/substrait/dialect/SupportedTypeDeserializer.java
```diff
@@ -0,0 +1,51 @@
...
+import io.substrait.dialect.Dialect.SystemTypeMetadata;
+import io.substrait.dialect.Dialect.TypeKind;
+import java.io.IOException;
+import java.util.Map;
+
+/**
+ * Deserializes a {@code supported_types} entry, which is either a bare enum string (e.g. {@code
+ * BOOL}) or a configuration object (e.g. {@code {type: PRECISION_TIMESTAMP, max_precision: 9}}).
+ */
+public class SupportedTypeDeserializer extends JsonDeserializer<SupportedType> {
```
bestbeforetoday: Perhaps could be default (package) scope instead of exposing in the public API.

core/src/main/java/io/substrait/dialect/SupportedTypeSerializer.java
```diff
@@ -0,0 +1,42 @@
...
+import com.fasterxml.jackson.databind.JsonSerializer;
+import com.fasterxml.jackson.databind.SerializerProvider;
+import io.substrait.dialect.Dialect.SupportedType;
+import java.io.IOException;
+
+/**
+ * Serializes a {@code supported_types} entry as a bare enum string when it carries no
+ * configuration, or as a configuration object otherwise.
+ */
+public class SupportedTypeSerializer extends JsonSerializer<SupportedType> {
```
bestbeforetoday: Perhaps could be default (package) scope instead of exposing in the public API.

bestbeforetoday (commented):
The programmatic building of a Dialect looks really nice. A whole load of inline comments; mostly very minor suggestions or queries.

core/src/main/java/io/substrait/dialect/Dialect.java
nielspardon: Done in fb74b41d — there's now a single package-scoped `DialectJsonSupport.MAPPER` reused for all (de)serialization instead of a per-call instance.

core/src/main/java/io/substrait/dialect/Dialect.java
nielspardon: Done in fb74b41d. Promoted the nested types to top-level classes in `io.substrait.dialect` and renamed `DialectDocument` → `Dialect`, which now carries the `builder()`/`load(...)`/`toYaml(...)` factory methods — so it reads `Dialect.builder()…build()`, `Dialect.toYaml(dialect)`, `Dialect.load(yaml)` now, and the `Dialect.DialectDocument` repetition is gone.

One trade-off worth flagging: this diverges from `SimpleExtension`, which still uses the `@Value.Enclosing` + nested-class shape. Flattening that similarly seems like a reasonable follow-up, but I left it out of scope here.

core/src/main/java/io/substrait/dialect/Dialect.java
nielspardon: Done in fb74b41d — reads straight from the stream now, no intermediate `String`.

core/src/main/java/io/substrait/dialect/Dialect.java
nielspardon: Good point. Done in fb74b41d: the `load(InputStream)` overload now reads with `AUTO_CLOSE_SOURCE` disabled so it no longer closes a caller-owned stream, and the Javadoc states the caller retains ownership.

core/src/main/java/io/substrait/dialect/DialectJsonSupport.java
nielspardon: Done in fb74b41d — the union (de)serializers reference the shared `DialectJsonSupport.MAPPER` directly, so the unchecked `getCodec()` casts are gone both here and in `SupportedTypeDeserializer`.

core/src/main/java/io/substrait/dialect/DialectJsonSupport.java
nielspardon: Good catch. Done in fb74b41d: `readEnums`/`readStrings` now accept a single scalar node as a one-element list, matching the mapper's configured behavior, so a single value is no longer silently dropped.

core/src/main/java/io/substrait/dialect/SupportedExpressionSerializer.java
nielspardon: Done in fb74b41d, by preventing the invalid state rather than tolerating it: `SupportedExpression` now has a `@Value.Check` rejecting `metadata` on `EXECUTION_CONTEXT_VARIABLE` (the schema's `execution_context_variable` is `additionalProperties: false` with no `metadata`). The serializer no longer special-cases it, so de/serialization is symmetric. Added `DialectValidationTest` to cover it.

core/src/main/java/io/substrait/dialect/SupportedRelationSerializer.java
nielspardon: Done in fb74b41d — `SupportedRelation` now has a `@Value.Check` forbidding both `writeTypes` and `ddlWriteTypes` from being set, since they share the single `write_types` field. Also covered in `DialectValidationTest`.

core/src/main/java/io/substrait/dialect/Dialect.java
nielspardon: Done in fb74b41d — `loadFromFile` hands the `File` straight to Jackson.

core/src/test/java/io/substrait/dialect/SpecDialectFixturesTest.java
nielspardon: Done in fb74b41d — replaced the `Scanner` slurp with `InputStream.readAllBytes()`.

core/src/main/java/io/substrait/dialect/SupportedExpressionDeserializer.java
nielspardon: Done in fb74b41d — made all six (de)serializer classes package-private. Resolving the related threads on the other five.

bestbeforetoday (approved):
Looks good to me. Nice updates, and including Javadoc is a great addition that I missed in my earlier review.

benbellick:
nice!

PR #874, opened 2026-06-16
build(core): exclude generated sources from javadoc

nielspardon:
The `core:javadoc` task linted the ANTLR- and version-generated sources (`SubstraitTypeParser`, `SubstraitTypeLexer`, `SubstraitLexer`, `SubstraitVersion`), producing ~640 "no comment" warnings for code that is regenerated on every build and cannot carry hand-written Javadoc.

This excludes these generated directories from the main Javadoc pass, alongside the already-excluded protobuf-generated sources, so the task only reports warnings for sources we can actually fix.

This is the first step in a series of changes to clean up the missing Javadoc reported by `./gradlew core:javadoc`, fixing the remaining warnings file by file in follow-up PRs.

🤖 Generated with AI

### DIFF
--- a/core/build.gradle.kts
+++ b/core/build.gradle.kts
@@ -369,9 +369,16 @@ tasks.named<Javadoc>("javadoc") {
   dependsOn("javadocProto")
   description = "Generate Javadoc for main sources (excludes protobuf-generated sources)."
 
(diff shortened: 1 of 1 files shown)


bestbeforetoday (approved):
Seems to work for me.

PR #953, opened 2026-06-17
fix(core): add compiled output to javadocImmutable classpath

nielspardon:
The `javadocImmutable` task only had `compileClasspath` (external dependency JARs) but not the project's own compiled output. Immutables-generated sources import their hand-written enclosing types (`Hint`, `Rel`, etc.) which live in the project's compiled classes, not in external JARs, causing 100+ `cannot find symbol` / `package does not exist` errors from javadoc.

Add `sourceSets["main"].output.classesDirs` to the classpath so javadoc can resolve cross-references between generated and hand-written sources.

---

🤖 Generated with AI

### DIFF
--- a/core/build.gradle.kts
+++ b/core/build.gradle.kts
@@ -337,8 +337,10 @@ tasks.register<Javadoc>("javadocImmutable") {
   // Only the generated proto sources
   setSource(fileTree(immuteableJavaDir) { include("**/*.java") })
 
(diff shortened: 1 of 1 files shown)


bestbeforetoday (approved):
Hard to tell if this successfully solves an issue without running locally since the current configuration suppresses all warnings but this does seem to solve issues so looks good to me.

PR #962, opened 2026-06-26
build(deps)!: update protobuf from v3 to v4

bestbeforetoday:
BREAKING CHANGE: Protobuf bindings are now generated using the v4 protobuf libraries, released in March 2024. Ensure that dependent projects also use the v4 protobuf-java libraries, and be aware of the breaking changes within the protobuf-java APIs documented at https://protobuf.dev/news/v26/#java-breaking-changes


nielspardon:
I feel that to be safe we should announce the major version increase as a BREAKING CHANGE since consumers of substrait-java might have re-declared the protobuf dependency with v3, pinned it to v3 or depend on v3 specific features in their code.

nielspardon (approved):
LGTM, thanks for adding the breaking change message

PR #963, opened 2026-06-26
fix: calcite optimization adds LITERAL_AGG(true)

mbwhite:
fix: calcite optimization adds literal_agg

I was testing running various Calcite optimisations; TPC/H 16 specifically was optimised in such a way that it wasn't possible for the Isthmus `SubstraitRelVisitor` to handle it. 

The issue was that calcite had added `LITERAL_AGG(true)`.

---
In Apache Calcite, `LITERAL_AGG(true)` is an internal aggregate function injected during the **decorrelation** process (specifically when handling `ANY`, `SOME`, or `IN` subqueries).

Here is why it is there and what it does:

### 1. The Problem: Handling Empty Subqueries

In your "before" plan, there is a `<> SOME(...)` correlated subquery. By SQL standards, if a subquery inside a quantified comparison (`SOME`/`ANY`) returns **zero rows**, the entire condition evaluates to `FALSE` (or `UNKNOWN`), not `NULL`.

### 2. The Solution: Flagging Matches

When Calcite transforms the correlated subquery into a join (`LogicalCorrelate` / `LogicalAggregate`), it needs a foolproof way to know if the subquery actually produced any rows for a given correlated key (`$cor0.L_PARTKEY`).

* `LITERAL_AGG(true)` evaluates the literal value `true` for every row entering the aggregate.
* Because it is an aggregate function, if the subquery returns **zero rows**, `LITERAL_AGG(true)` will return `NULL`.
* If the subquery returns **one or more rows**, it returns `true`.

### 3. How Calcite Uses It

Look closely at the massive `LogicalFilter(condition=[OR(...)]` right above the correlate in your "after" plan.

Calcite uses the output of `LITERAL_AGG(true)` (which maps to column `$19` or `$20` in that flattened row) to evaluate the exact short-circuiting logic of the `SOME` operator:

If it's NULL: The subquery was empty $\rightarrow$ handle as false/null.
If it's TRUE: The subquery returned data $\rightarrow$ proceed to check the actual value comparisons (like `COUNT` and `MAX`).

It essentially acts as a highly optimized, null-aware boolean flag for row existence during complex subquery unnesting.

--- 
:robot: built with assistance from AI


nielspardon:
for some reason your PR header had a trailing whitespace in it

mbwhite:
thanks @nielspardon

nielspardon:
Nice fix — the approach is correct: within a grouped aggregate every group is non-empty, so `LITERAL_AGG(true)` always yields `true`, and the null-when-empty semantics come from the decorrelation LEFT JOIN. Re-inserting the literal via a Project is the right call.

One bug, which I reproduced locally (PR branch, JDK 17):

**The wrapper's passthrough references are typed with the whole aggregate struct.** `newRootStructReference(int, Type)` assigns its argument verbatim as the reference type, so passing `aggRel.getRecordType()` types every passthrough column as the entire struct instead of the scalar field. The wrapper Project's record type comes out as:

```
[Struct{[I64,I64,I64,I64,Decimal]}, Struct{…}, Struct{…}, Struct{…}, Struct{…}, Bool]
```

i.e. all five passthroughs are the whole struct; only the re-inserted `Bool` literal is correct. Conversion doesn't throw (parents reference Calcite types), but the POJO carries wrong types and the wrapper subtree doesn't round-trip. Fix matches `fromGroupSet` just above:

```diff
-      projectExprs.add(FieldReference.newRootStructReference(i, aggRel.getRecordType()));
+      projectExprs.add(FieldReference.newInputRelReference(i, aggRel));
@@
-        projectExprs.add(
-            FieldReference.newRootStructReference(realAggIndex, aggRel.getRecordType()));
+        projectExprs.add(FieldReference.newInputRelReference(realAggIndex, aggRel));
```

With that change the columns become `[I64,I64,I64,I64,Decimal,Bool]` and everything round-trips.

The current test's structural assertions check the outer SELECT projection, not the wrapper, so they pass even with the bug. Suggested replacement that actually inspects the wrapper, plus a guard test for the `groupings.size() > 1` branch (both verified: red on this PR as-is, green with the fix):

```java
@Test
void conversionHandlesLiteralAggInsertedBySubQueryRemoveRule()
    throws SqlParseException, IOException {
  String query =
      "select e1.l_orderkey from lineitem e1 "
          + "where e1.l_quantity <> some ("
          + "  select l_quantity from lineitem e2 where e2.l_partkey = e1.l_partkey)";

  RelRoot relRoot = SubstraitSqlToCalcite.convertQuery(query, TPCH_CATALOG);
  HepPlanner hepPlanner =
      new HepPlanner(
          new HepProgramBuilder()
              .addRuleInstance(CoreRules.FILTER_SUB_QUERY_TO_CORRELATE)
              .build());
  hepPlanner.setRoot(relRoot.rel);
  RelNode decorrelated =
      RelDecorrelator.decorrelateQuery(
          hepPlanner.findBestExp(),
          RelFactories.LOGICAL_BUILDER.create(relRoot.rel.getCluster(), null));

  // Pin the trigger so a future Calcite bump can't silently stop exercising this path.
  assertTrue(containsLiteralAgg(decorrelated), "test setup no longer produces LITERAL_AGG");

  Plan.Root planRoot =
      assertDoesNotThrow(
          () ->
              SubstraitRelVisitor.convert(
                  RelRoot.of(decorrelated, relRoot.kind), EXTENSION_COLLECTION));

  // The fix inserts a Project directly over the Aggregate; inspect THAT, not the outer SELECT.
  Project wrapper =
      findProjectOverAggregate(planRoot.getInput())
          .orElseThrow(() -> new AssertionError("expected a Project wrapping the Aggregate"));

  assertTrue(
      wrapper.getExpressions().stream()
          .anyMatch(e -> e instanceof Expression.BoolLiteral && ((Expression.BoolLiteral) e).value()),
      "LITERAL_AGG(true) should be re-inserted as a boolean true literal");

  // Passthroughs must carry the scalar field type, not the whole aggregate struct.
  assertTrue(
      wrapper.getRecordType().fields().stream().noneMatch(f -> f instanceof Type.Struct),
      "wrapper columns must be scalar; fields=" + wrapper.getRecordType().fields());

  // The wrapper subtree must survive a proto round-trip with its schema intact.
  ExtensionCollector ec = new ExtensionCollector();
  io.substrait.proto.Rel proto = new RelProtoConverter(ec).toProto(wrapper);
  Rel rt = new ProtoRelConverter(ec, extensions).from(proto);
  assertEquals(wrapper.getRecordType(), rt.getRecordType(), "wrapper schema must round-trip");
}

@Test
void literalAggCombinedWithGroupingSetsIsRejected() {
  RelNode input = builder.values(new String[] {"a", "b"}, 1, 2, 3, 4).build();
  RexBuilder rexBuilder = creator.rex();
  AggregateCall literalAgg =
      AggregateCall.create(
          SqlInternalOperators.LITERAL_AGG, false, false, false,
          List.of(rexBuilder.makeLiteral(true)), List.of(), -1, null,
          RelCollations.EMPTY, typeFactory.createSqlType(SqlTypeName.BOOLEAN), "li");
  ImmutableBitSet g0 = ImmutableBitSet.of(0);
  ImmutableBitSet g1 = ImmutableBitSet.of(1);
  RelNode aggregate =
      LogicalAggregate.create(input, List.of(), g0.union(g1), List.of(g0, g1), List.of(literalAgg));

  UnsupportedOperationException ex =
      assertThrows(
          () -> SubstraitRelVisitor.convert(RelRoot.of(aggregate, SqlKind.SELECT), EXTENSION_COLLECTION));
  assertTrue(ex.getMessage().contains("GROUPING SETS"), ex.getMessage());
}

private static boolean containsLiteralAgg(RelNode node) {
  if (node instanceof org.apache.calcite.rel.core.Aggregate
      && ((org.apache.calcite.rel.core.Aggregate) node)
          .getAggCallList().stream()
              .anyMatch(c -> c.getAggregation().getKind() == SqlKind.LITERAL_AGG)) {
    return true;
  }
  return node.getInputs().stream().anyMatch(OptimizerIntegrationTest::containsLiteralAgg);
}

private static Optional<Project> findProjectOverAggregate(Rel rel) {
  if (rel instanceof Project && ((Project) rel).getInput() instanceof Aggregate) {
    return Optional.of((Project) rel);
  }
  return rel.getInputs().stream()
      .map(OptimizerIntegrationTest::findProjectOverAggregate)
      .filter(Optional::isPresent).map(Optional::get).findFirst();
}
```

nielspardon (commented):
A few suggestions from reviewing the LITERAL_AGG handling. The core rewrite (filtering the literal calls out of the measures and re-inserting them via a Project wrapper) looks correct — the field indexing and `Remap.offset` slice line up. These are refinements around the unsupported-combination guard and a couple of small cleanups.

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelVisitor.java
```diff
@@ -400,7 +411,57 @@ public Rel visit(org.apache.calcite.rel.core.Aggregate aggregate) {
...
     }
 
-    return builder.build();
+    Rel aggRel = builder.build();
+
+    if (literalAggCalls.isEmpty()) {
+      return aggRel;
+    }
+
+    if (groupings.size() > 1) {
```
nielspardon: This `groupings.size() > 1` rejection runs *after* the multi-grouping remap loop above, so for `GROUPING SETS + GROUP_ID + LITERAL_AGG` the loop is reached first. There, a `LITERAL_AGG` call hits the new `else if (literalAggCalls.contains(aggCall))` branch and is skipped (no `remap` entry added), but a subsequent call still does `remap.add(i + groupingFieldCount, ...)` using the original list index `i`. That leaves a gap and throws `IndexOutOfBoundsException` instead of this clear message — e.g. agg list `[LITERAL_AGG (i=0), GROUP_ID (i=1)]` with `groupingFieldCount=2` does `remap.add(3, ...)` on a size-2 list.

Suggest rejecting the combination up front, right after `literalAggCalls` is computed and before any remap work:

```java
if (!literalAggCalls.isEmpty() && groupings.size() > 1) {
      "LITERAL_AGG combined with GROUPING SETS / CUBE / ROLLUP is not supported");
}
```

With that, the `else if (literalAggCalls.contains(aggCall)) { /* skip */ }` branch in the remap loop and this post-build check both become dead code and can be dropped.

Also worth a test: `literalAggCombinedWithGroupingSetsIsRejected` only covers grouping sets *without* `GROUP_ID` (the `Remap.offset` branch), so the crashing path here isn't exercised. Adding a `GROUP_ID + LITERAL_AGG + grouping sets` case (literal ordered before another call) would pin it to the clean exception.

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelVisitor.java
```diff
@@ -400,7 +411,57 @@ public Rel visit(org.apache.calcite.rel.core.Aggregate aggregate) {
...
+
+    // Wrap the aggregate in a Project that replaces LITERAL_AGG output positions with their
+    // literal values and passes through all other fields via FieldReference.
+    //
+    // The aggregate output schema is: [grouping fields..., real agg measures...]
+    // The full output schema requested is: [grouping fields..., all agg calls (in original order)]
+    // For each position in the original agg call list:
+    //   - real measure  → FieldReference into the aggregate output
+    //   - LITERAL_AGG   → the literal value from aggCall.rexList
+    final int groupingFieldCount =
```
nielspardon: `groupingFieldCount` is now computed a third time in this method — with `.distinct()` here and in the no-GROUP_ID branch, but *without* `.distinct()` in the GROUP_ID branch. Suggest hoisting a single value near the top and reusing it everywhere:

```java
final int groupingFieldCount =
    Math.toIntExact(
        groupings.stream().flatMap(g -> g.getExpressions().stream()).distinct().count());
```

If the distinct/non-distinct divergence in the GROUP_ID branch is intentional, a one-line comment there would help; right now it reads like an oversight.

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelVisitor.java
```diff
@@ -348,10 +350,17 @@ public Rel visit(org.apache.calcite.rel.core.Aggregate aggregate) {
             .filter(c -> c.getAggregation().equals(SqlStdOperatorTable.GROUP_ID))
             .collect(Collectors.toList());
 
+    // get LITERAL_AGG() function calls — injected by SubQueryRemoveRule (CALCITE-6945) as a
+    // null-presence indicator; they carry a RexLiteral in rexList and have no Substrait binding.
+    List<AggregateCall> literalAggCalls =
```
nielspardon: `literalAggCalls` (and `groupIdCalls`) are only ever probed with `.contains()`, including inside the per-agg-call loop below (making it O(m·n)). Collecting them as a `Set` makes membership O(1) and signals intent:

```java
Set<AggregateCall> literalAggCalls =
    aggregate.getAggCallList().stream()
        .filter(c -> c.getAggregation().getKind() == SqlKind.LITERAL_AGG)
        .collect(Collectors.toSet());
```

Minor, but cheap.

nielspardon (approved):
LGTM

Issue #869, opened 2026-06-15
Update to Substrait v0.89.0


nielspardon:
# Release Notes
### ⚠ BREAKING CHANGES

* NOT A REAL BREAKING CHANGE. OuterReference.steps_out
field is now under oneof -- implicit optional to explicit optional. This
is not a real breaking change as Substrait always required non-zero
value.

Add optional RelCommon.id field and OuterReference.id_reference to
support unambiguous outer reference resolution in plans with common
subexpressions (ReferenceRel). The existing steps_out offset-based
mechanism remains for backward compatibility with tree-shaped plans.

The `RelCommon.rel_anchor` can be used for other purpose but for now
only required when it is used to resolve outer reference.

### Features

* add id-based outer reference resolution for DAG plans ([#1031](https://github.com/substrait-io/substrait/issues/1031)) ([ad664dc](https://github.com/substrait-io/substrait/commit/ad664dc9a4bc1a368ab8d2940084c77cdb7d9213)), closes [#1024](https://github.com/substrait-io/substrait/issues/1024)

See https://github.com/substrait-io/substrait/releases/tag/v0.89.0

PR #964, opened 2026-06-26
fix: outer reference confusing calcite optimizations

mbwhite:
fix: outer reference confusing post calcite optimizations

### DIFF
--- a/isthmus/src/main/java/io/substrait/isthmus/OuterReferenceResolver.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/OuterReferenceResolver.java
@@ -66,6 +66,27 @@ public class OuterReferenceResolver extends RelNodeVisitor<RelNode, RuntimeExcep
     for (CorrelationId id : filter.getVariablesSet()) {
       nestedDepth.putIfAbsent(id, 0);
     }
+    // Also register any CorrelationIds referenced directly in the condition expression.
+    // This covers partially-decorrelated plans where a Calcite optimizer (e.g. HepPlanner)
+    // rewrites correlated IN-subqueries into joins but leaves scalar-aggregate subquery
+    // correlations as RexFieldAccess($corN.col) inside a Filter condition without a
+    // surrounding Correlate node. In that case getVariablesSet() is empty but the condition
+    // still contains RexCorrelVariable references whose IDs must be in nestedDepth before
+    // rexVisitor.visitFieldAccess() is called.
+    filter
+        .getCondition()
+        .accept(
+            new RexShuttle() {
+              @Override
+              public RexNode visitFieldAccess(RexFieldAccess fieldAccess) {
+                if (fieldAccess.getReferenceExpr() instanceof RexCorrelVariable) {
+                  CorrelationId id =
+                      ((RexCorrelVariable) fieldAccess.getReferenceExpr()).id;
+                  nestedDepth.putIfAbsent(id, 0);
+                }
+                return fieldAccess;
+              }
+            });
     filter.getCondition().accept(rexVisitor);
     return super.visit(filter);
   }
--- a/isthmus/src/main/java/io/substrait/isthmus/OuterReferenceResolver.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/OuterReferenceResolver.java
@@ -80,8 +80,7 @@ public class OuterReferenceResolver extends RelNodeVisitor<RelNode, RuntimeExcep
               @Override
               public RexNode visitFieldAccess(RexFieldAccess fieldAccess) {
                 if (fieldAccess.getReferenceExpr() instanceof RexCorrelVariable) {
-                  CorrelationId id =
-                      ((RexCorrelVariable) fieldAccess.getReferenceExpr()).id;
+                  CorrelationId id = ((RexCorrelVariable) fieldAccess.getReferenceExpr()).id;
                   nestedDepth.putIfAbsent(id, 0);
                 }
                 return fieldAccess;
--- a/isthmus/src/main/java/io/substrait/isthmus/expression/RexExpressionConverter.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/expression/RexExpressionConverter.java
@@ -224,7 +224,23 @@ public class RexExpressionConverter implements RexVisitor<Expression> {
     switch (kind) {
       case CORREL_VARIABLE:
         {
-          int stepsOut = relVisitor.getFieldAccessDepth(fieldAccess);
+          Integer stepsOut = relVisitor.getFieldAccessDepth(fieldAccess);
+          // Substrait requires steps_out >= 1 for an outer reference
+          // (proto/substrait/algebra.proto, OuterReference.steps_out: "Must be >= 1.").
+          // A null or 0 depth means the correlation does not step out of any subquery,
+          // typically a plan that was only partially decorrelated (the owning Correlate
+          // was rewritten into a join but the $cor reference remains in this Filter).
+          // We can't represent that with steps_out, so fail clearly rather than emit
+          // an invalid steps_out=0 reference. Id-based resolution (substrait-io/substrait#1031)
+          // would handle this; tracked in #869.
+          if (stepsOut == null || stepsOut == 0) {
+            throw new UnsupportedOperationException(
+                String.format(
+                    "Cannot convert outer reference %s: it does not step out of any "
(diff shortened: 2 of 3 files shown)


nielspardon:
this seems to require a `./gradlew spotlessApply`

isthmus/src/main/java/io/substrait/isthmus/OuterReferenceResolver.java
```diff
@@ -66,6 +66,26 @@ public RelNode visit(Filter filter) throws RuntimeException {
...
+    // rexVisitor.visitFieldAccess() is called.
+    filter
+        .getCondition()
+        .accept(
+            new RexShuttle() {
+              @Override
+              public RexNode visitFieldAccess(RexFieldAccess fieldAccess) {
+                if (fieldAccess.getReferenceExpr() instanceof RexCorrelVariable) {
+                  CorrelationId id = ((RexCorrelVariable) fieldAccess.getReferenceExpr()).id;
+                  nestedDepth.putIfAbsent(id, 0);
```
nielspardon, line 84: Registering the id here is the right move — it removes the downstream NPE and lets genuinely-nested references (depth ≥ 1) resolve correctly.

The thing to be aware of: for a single top-level residual reference (no enclosing subquery), this anchors the id at depth 0 and the access is recorded at 0. Downstream that becomes `newRootStructOuterReference(..., stepsOut=0)`, which is an invalid Substrait plan — the proto requires `steps_out >= 1` (algebra.proto#L1649-L1653). See the guard proposal in the review summary so this fails loudly instead of emitting an invalid reference.

isthmus/src/test/java/io/substrait/isthmus/OuterReferenceResolverTest.java
```diff
@@ -142,4 +143,62 @@ void nestedApplyJoinQuery() throws SqlParseException {
...
+            .filter(correlCondition)
+            .join(JoinRelType.INNER, tpcDsRelBuilder.literal(true))
+            .project(tpcDsRelBuilder.field("SS_SOLD_DATE_SK"))
+            .build();
+
+    // Pre-fix: fieldAccessDepthMap would be empty because the CorrelationId was never registered.
+    // Post-fix: the Filter condition pre-scan registers $cor0, so the field access is recorded.
+    final Map<RexFieldAccess, Integer> fieldAccessDepthMap = buildOuterFieldRefMap(calciteRel);
+    Assertions.assertEquals(1, fieldAccessDepthMap.size());
+    validateOuterRef(fieldAccessDepthMap, "$cor0", "SS_ITEM_SK", 0);
```
nielspardon, line 201: This assertion enshrines the invalid `steps_out=0` output as the expected behaviour. Since the proto requires `steps_out >= 1`, depth 0 isn't representable.

**Proposal:** keep the assertion that the resolver now *registers* the correlation (it proves the NPE is gone), then assert that conversion is rejected rather than producing an invalid plan:

```java
// Resolver registers the dangling correlation (no longer silently dropped → no NPE)...
Map<RexFieldAccess, Integer> map = buildOuterFieldRefMap(calciteRel);
Assertions.assertEquals(1, map.size());
validateOuterRef(map, "$cor0", "SS_ITEM_SK", 0);
// ...but steps_out=0 is not a valid Substrait outer reference (must be >= 1),
// so conversion fails clearly rather than emitting an invalid plan.
UnsupportedOperationException ex =
    Assertions.assertThrows(
        () -> SubstraitRelVisitor.convert(calciteRel, converterProvider));
Assertions.assertTrue(ex.getMessage().contains("steps_out"));
```

nielspardon (changes requested):
Nice catch on the crash this fixes. When a Calcite optimizer only *partially* decorrelates a plan (the owning `Correlate` is rewritten into a join, but `$corN.col` is left in a `Filter` condition whose `getVariablesSet()` is now empty), the `CorrelationId` was never registered, the access was dropped from `fieldAccessDepthMap`, and `RexExpressionConverter.visitFieldAccess` then unboxed a `null` depth into an `int` → NPE. Registering the id in `visit(Filter)` removes that NPE, and for genuinely-nested references (depth ≥ 1) it produces correct outer references.

One concern with the result: for the common single-reference case the pre-scan registers the correlation at **depth 0**, so the converter emits `newRootStructOuterReference(..., stepsOut=0)`. That is an *invalid* Substrait plan, not just an odd one — the proto is explicit that `steps_out` **must be >= 1**:

> `OuterReference.steps_out`: "Number of subquery boundaries to traverse up for this field's reference. Must be >= 1."
> — https://github.com/substrait-io/substrait/blob/8675ba395ce966e593ca66575930f59dc359b419/proto/substrait/algebra.proto#L1649-L1653

Consistently, `FieldReference.isOuterReference()` is `outerReferenceStepsOut().orElse(0) > 0`, and on the reverse path a `0` reference satisfies neither `isSimpleRootReference()` nor `isOuterReference()` and falls through both branches of `ExpressionRexConverter.visit(FieldReference)`. (Emitting a plain non-outer reference instead is also wrong: the field index is in the correlated relation's coordinate system, not the Filter's input — e.g. `$cor0.SS_ITEM_SK` is index 2 in `STORE_SALES` but index 2 in `ITEM` is `I_REC_START_DATE`.)

So I'd suggest pairing the resolver fix with a guard so the unrepresentable depth-0 case fails loudly instead of producing an invalid plan.

**Proposal 1 — guard the converter** (`isthmus/.../expression/RexExpressionConverter.java`, `visitFieldAccess`, `CORREL_VARIABLE` case; that file isn't in this diff, hence here rather than inline):

```java
case CORREL_VARIABLE:
  {
    Integer stepsOut = relVisitor.getFieldAccessDepth(fieldAccess);
    // Substrait requires steps_out >= 1 for an outer reference
    // (proto/substrait/algebra.proto, OuterReference.steps_out: "Must be >= 1.").
    // A null/0 depth means the correlation does not step out of any subquery,
    // typically a plan that was only partially decorrelated (the owning Correlate
    // was rewritten into a join but the $cor reference remains in this Filter).
    // We can't represent that with steps_out, so fail clearly rather than emit an
    // invalid steps_out=0 reference. Id-based resolution (substrait#1031) would
    // handle this; tracked in #869.
    if (stepsOut == null || stepsOut == 0) {
          String.format(
              "Cannot convert outer reference %s: it does not step out of any "
                  + "subquery (steps_out=%s). This usually indicates a plan that was "
                  + "only partially decorrelated; Substrait requires steps_out >= 1.",
              fieldAccess, stepsOut));
    }
    return FieldReference.newRootStructOuterReference(
        fieldAccess.getField().getIndex(),
        typeConverter.toSubstrait(fieldAccess.getType()),
        stepsOut);
  }
```

This also subsumes the original NPE (the `null` case) under one clear message.

**Follow-up:** the proper long-term fix — id-based outer-reference resolution from substrait-io/substrait#1031, which would let these partially-decorrelated / DAG plans actually convert — is already tracked under #869 (Update to Substrait v0.89.0). No new issue needed; just point the guard comment at #869.

isthmus/src/main/java/io/substrait/isthmus/expression/RexExpressionConverter.java
```diff
@@ -224,7 +224,23 @@ public Expression visitFieldAccess(RexFieldAccess fieldAccess) {
...
-          int stepsOut = relVisitor.getFieldAccessDepth(fieldAccess);
+          Integer stepsOut = relVisitor.getFieldAccessDepth(fieldAccess);
+          // Substrait requires steps_out >= 1 for an outer reference
+          // (proto/substrait/algebra.proto, OuterReference.steps_out: "Must be >= 1.").
+          // A null or 0 depth means the correlation does not step out of any subquery,
+          // typically a plan that was only partially decorrelated (the owning Correlate
+          // was rewritten into a join but the $cor reference remains in this Filter).
+          // We can't represent that with steps_out, so fail clearly rather than emit
+          // an invalid steps_out=0 reference. Id-based resolution (substrait-io/substrait#1031)
+          // would handle this; tracked in #869.
```
nielspardon: Minor: this project keeps GitHub issue/PR numbers out of source (they belong in the commit/PR, which is what this fix is). The rest of the comment — the proto `steps_out >= 1` constraint and the partial-decorrelation explanation — is exactly the context a future reader needs, so let's keep that and just drop the `substrait-io/substrait#1031` and `#869` references.

```suggestion
          // an invalid steps_out=0 reference. Id-based outer-reference resolution
          // would handle this and is the proper long-term fix.
```

nielspardon (approved):
LGTM

PR #976, opened 2026-07-01
feat(isthmus): map execution context variables to/from Calcite

nielspardon:
## Summary

Adds Isthmus (Substrait ⇄ Apache Calcite) mappings for the three execution context variables introduced in `:core` by #867: `current_timestamp`, `current_date` and `current_timezone`. Previously any plan containing one of these failed to convert to/from Calcite.

## Changes

- **Substrait → Calcite** (`ExpressionRexConverter`): three `visit` overrides build niladic `RexCall`s whose return type is **forced** from the Substrait type so time-zone semantics, precision and required-nullability round-trip.
  - `CurrentTimestamp(p)` → `CURRENT_TIMESTAMP` typed `TIMESTAMP_WITH_LOCAL_TIME_ZONE(p)` — Substrait's `current_timestamp` is a `precision_timestamp_tz`, whereas Calcite's operator would otherwise infer a timezone-less `TIMESTAMP`.
  - `CurrentDate` → `CURRENT_DATE` typed `DATE`.
  - `CurrentTimezone` → the new `CurrentTimezoneFunction` typed non-nullable `VARCHAR`.
- **Calcite → Substrait**: new `CallConverters.EXECUTION_CONTEXT_VARIABLE` matches these operators by identity and produces the corresponding Substrait expressions. Registered in both `CallConverters.defaults()` and `ConverterProvider.getCallConverters()`.
- **Session timezone operator**: Calcite has no built-in operator for the session timezone, so this adds `CurrentTimezoneFunction` — a niladic `SqlBaseContextVariable` (modeled on Calcite's own `CURRENT_USER`/`CURRENT_ROLE`). It is registered in `SubstraitOperatorTable` via a separate scalar-operator table so that:
  - `SELECT current_timezone` parses from SQL text without parentheses (via Calcite's `SqlValidatorImpl.makeNullaryCall`), and
  - its generic `OTHER_FUNCTION` kind does **not** pollute `OVERRIDE_KINDS` (which would otherwise filter every standard `OTHER_FUNCTION` operator out of the operator list).

## Tests

- `ExecutionContextVariableConversionTest`: bidirectional expression ⇄ Rex round-trips for all three variables.
- `Substrait2SqlTest`: full SQL round-trips for `current_timestamp`, `current_date`, and `current_timezone`, including asserting that `current_timezone` parses and is emitted back as the bare keyword (no Calcite dialect changes required).

## Verification

`:core:spotlessCheck`, `:isthmus:spotlessCheck`, the full `:isthmus:test` suite, and compilation of `:spark:spark-3.5_2.12` and `:examples:substrait-spark` all pass.

🤖 Generated with AI

### DIFF
--- a/isthmus/src/test/java/io/substrait/isthmus/Substrait2SqlTest.java
+++ b/isthmus/src/test/java/io/substrait/isthmus/Substrait2SqlTest.java
@@ -3,6 +3,7 @@ package io.substrait.isthmus;
 import static org.junit.jupiter.api.Assertions.assertDoesNotThrow;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertInstanceOf;
+import static org.junit.jupiter.api.Assertions.assertTrue;
 
 import io.substrait.isthmus.utils.SetUtils;
 import io.substrait.plan.Plan;
@@ -36,6 +37,28 @@ class Substrait2SqlTest extends PlanTestBase {
     String query =
         "select l_partkey, l_discount from lineitem where l_orderkey > cast(100 as bigint)";
     assertFullRoundTrip(query);
+  }
+
+  @Test
+  void currentTimestamp() throws Exception {
+    assertFullRoundTrip("select current_timestamp from part");
+  }
+
+  @Test
+  void currentDate() throws Exception {
+    assertFullRoundTrip("select current_date from part");
+  }
+
+  @Test
+  void currentTimezone() throws Exception {
+    // CURRENT_TIMEZONE is a Substrait-specific niladic operator with no standard Calcite
+    // equivalent; it is registered in SubstraitOperatorTable so it parses without parentheses.
+    assertFullRoundTrip("select current_timezone from part");
+
+    // ...and it is emitted back as the bare niladic keyword (no dialect changes required).
+    Plan plan = toSubstraitPlan("select current_timezone from part", TPCH_CATALOG);
+    assertTrue(
+        toSql(plan).contains("CURRENT_TIMEZONE"), "expected CURRENT_TIMEZONE in emitted SQL");
   }
 
   @Test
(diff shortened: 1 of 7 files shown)


isthmus/src/test/java/io/substrait/isthmus/Substrait2SqlTest.java
```diff
@@ -38,6 +39,28 @@ void simpleTest2() throws Exception {
...
+  @Test
+  void currentTimezone() throws Exception {
+    // CURRENT_TIMEZONE is a Substrait-specific niladic operator with no standard Calcite
+    // equivalent; it is registered in SubstraitOperatorTable so it parses without parentheses.
+    assertFullRoundTrip("select current_timezone from part");
+
+    // ...and it is emitted back as the bare niladic keyword (no dialect changes required).
+    Plan plan = toSubstraitPlan("select current_timezone from part", TPCH_CATALOG);
+    assertTrue(
+        toSql(plan).contains("CURRENT_TIMEZONE"), "expected CURRENT_TIMEZONE in emitted SQL");
```
bestbeforetoday: Is this intended to just check that `CURRENT_TIMEZONE` appears in the output SQL or is it intended to check that it does not appear as a function with brackets, for example `CURRENT_TIMEZONE()`?

Either way, this might be clearer as a separate unit test. If the test is well named, the comment might not be necessary.

The failure message would be more helpful if it included the actual SQL. This would allow the problem to be diagnosed from the failure message without having to rerun the unit test in a debugger to see the failing SQL output.

bestbeforetoday (commented):
Looks good to me. A query/suggestion for one of the unit tests.

isthmus/src/test/java/io/substrait/isthmus/Substrait2SqlTest.java
nielspardon: Good catch — the "bare keyword" check was the intent, but you're right the old `.contains("CURRENT_TIMEZONE")` didn't enforce it. Split it into `currentTimezoneEmittedAsBareKeyword()`, which now also asserts `CURRENT_TIMEZONE(` is absent, with the emitted SQL in the failure messages. Thanks!

bestbeforetoday (approved):

PR #983, opened 2026-07-02
feat(isthmus)!: configurable SQL parser config via ConverterProvider builder

nielspardon:
## Summary

Makes the Calcite SQL parser configuration used by isthmus' SQL front-end customizable — most
importantly the casing applied to **unquoted** identifiers — by:

1. **Injecting a `ConverterProvider` into the SQL statement parsers**, so a single parser
   configuration is applied consistently across both CREATE-TABLE parsing (which builds the Calcite
   catalog) and query parsing; and
2. **Adding a `ConverterProvider.Builder`** that lets callers supply either a full Calcite
   `SqlParser.Config` or just an unquoted-casing override.

Default behaviour is unchanged (`Casing.TO_UPPER`, the DDL parser factory, and `LENIENT`
conformance).

## Motivation

isthmus hard-coded `Casing.TO_UPPER` in `getSqlParserConfig()`, and that config was not applied
consistently between the CREATE-statement parsing that builds the catalog and the query parsing.
`TO_UPPER` is the ANSI/Calcite default, but a large class of engines fold unquoted identifiers to
lower-case or preserve them (Presto/Trino, Spark, DuckDB, Postgres), so a plan generated from
`SELECT … FROM orders` comes out with a `NamedScan` named `ORDERS` that those engines can't resolve.
Users driving isthmus (the library and the CLI) with SQL that uses **unquoted** identifiers need to
control this. This PR makes the parser configuration a first-class, injectable input rather than
something reachable only by subclassing `ConverterProvider`.

## Approach (per review discussion)

Following the review thread (thanks @vbarua), the configuration is exposed through a **builder**
rather than a dedicated casing toggle on `ConverterProvider`:

- `ConverterProvider.builder()` returns a `Builder` with:
  - `sqlParserConfig(SqlParser.Config)` — supply a **full** Calcite parser config (any setting);
  - `unquotedCasing(Casing)` — a convenience for the common case, layered on the current config;
  - plus `extensions`, `typeFactory`, the function converters, `typeConverter`, and
    `executionBehavior`.
- `ConverterProvider.DEFAULT_SQL_PARSER_CONFIG` — the isthmus default
  (`SqlParser.Config.DEFAULT` + `TO_UPPER` + `SqlDdlParserImpl.FACTORY` + `LENIENT`), exposed as the
  recommended starting point for a customized config:
  `DEFAULT_SQL_PARSER_CONFIG.withUnquotedCasing(Casing.UNCHANGED)`.
- `getSqlParserConfig()` now returns the stored config. Subclassing `ConverterProvider` and
  overriding `getSqlParserConfig()` remains the escape hatch for fully dynamic behaviour.
- The **builder is now the sole public construction path**: every `ConverterProvider` constructor is
  `@Deprecated`, and a `protected ConverterProvider(Builder)` constructor is the stable seam for
  subclassing (see Breaking changes).

```java
// full Calcite parser config
ConverterProvider.builder()
    .sqlParserConfig(ConverterProvider.DEFAULT_SQL_PARSER_CONFIG.withUnquotedCasing(Casing.TO_LOWER))
    .build();

// common case
ConverterProvider.builder().unquotedCasing(Casing.UNCHANGED).build();
```

## Consistent propagation

The `ConverterProvider` (hence its parser config) is threaded through **both** CREATE-TABLE parsing
(`SubstraitCreateStatementParser`) and query parsing (`SubstraitSqlToCalcite` / `SqlToSubstrait`),
so the casing used to build the catalog matches the casing used to resolve names in queries. The
table name stored in a `NamedScan` therefore reflects the configured casing end-to-end.

## Breaking changes

- The `SqlParser.Config`-based parsing entry points (public since `v0.94.0`) are **removed** in
  favour of passing a `ConverterProvider` (built via `builder().sqlParserConfig(...)`):
  - `SubstraitSqlStatementParser.parseStatements(String, SqlParser.Config)`
  - `SubstraitSqlToCalcite.convertQueries(String, CatalogReader, SqlParser.Config)`
  - `SubstraitSqlToCalcite.convertQueries(String, CatalogReader, SqlValidator, RelOptCluster, SqlParser.Config)`
- **`ConverterProvider` construction now goes through `builder()`.** *All* public `ConverterProvider`
  constructors — the no-arg, `(extensions)`, `(extensions, typeFactory)`, and the 6-arg / 7-arg
  all-components ones — are **`@Deprecated`** in favour of `ConverterProvider.builder()` (or the
  shared `ConverterProvider.DEFAULT`). They still work and delegate to the builder path, so this is
  source-compatible (deprecation warnings only). `DEFAULT` is now `builder().build()`.
- **Subclassing seam.** A new `protected ConverterProvider(Builder)` constructor is the single,
  stable way for a subclass to seed base state, replacing the (previously private) positional master
  — so adding a component to `ConverterProvider` no longer changes any subclass's `super(...)` call.
  The two in-tree subclasses, `DynamicConverterProvider` and
  `AutomaticDynamicFunctionMappingConverterProvider`, now expose a public
  `(ConverterProvider.Builder)` constructor and call `super(builder)`; their positional constructors
  are `@Deprecated`.
- `SqlToSubstrait.convert(String, CatalogReader, SqlDialect)` remains `@Deprecated`: the `SqlDialect`
  argument is ignored and it delegates to `convert(String, CatalogReader)`.

## Notable changes

| Class | Change |
|-------|--------|
| `ConverterProvider` | Stores a `SqlParser.Config`; adds `DEFAULT_SQL_PARSER_CONFIG`, `builder()` and the nested `Builder`; `getSqlParserConfig()` returns the stored config. **All public constructors `@Deprecated`** in favour of `builder()`; adds a `protected ConverterProvider(Builder)` subclassing seam (the positional master is now internal), which owns the function-converter derivation; `DEFAULT = builder().build()` |
| `DynamicConverterProvider` | Public `(ConverterProvider.Builder)` constructor calling `super(builder)`; the no-arg / `(extensions)` / `(extensions, typeFactory)` constructors are `@Deprecated` |
| `AutomaticDynamicFunctionMappingConverterProvider` | Same treatment — public `(ConverterProvider.Builder)` constructor calling `super(builder)`; positional constructors `@Deprecated` |
| `SubstraitSqlStatementParser` | `parseStatements(String, ConverterProvider)` owns the `SqlParser` instantiation; `SqlParser.Config` overload removed |
| `SubstraitSqlToCalcite` | New `convertQueries(sql, catalog, ConverterProvider[, operatorTable])`; internal overloads route through `ConverterProvider`; `SqlParser.Config` overloads removed |
| `SubstraitCreateStatementParser` | New `processCreateStatements(ConverterProvider, …)` / `processCreateStatementsToCatalog(ConverterProvider, …)` overloads — same parser config as query parsing |
| `SqlToSubstrait` | `convert(sql, catalog)` uses the `ConverterProvider` path; `convert(sql, catalog, SqlDialect)` `@Deprecated` (arg ignored) |
| `SqlExpressionToSubstrait` / `SubstraitToSql` | Route through the provider / `ConverterProvider.DEFAULT` |
| `IsthmusEntryPoint` (CLI) | `--unquotedcasing` maps to `ConverterProvider.builder().unquotedCasing(…)`; anonymous subclass removed |
| `FromSql` (example) | Uses `ConverterProvider.builder().unquotedCasing(Casing.UNCHANGED)`, replacing the deprecated `convert(sql, catalog, SqlDialect)` call |

## Tests

- New `UnquotedCasingTest`: default casing; the `unquotedCasing(…)` convenience for each `Casing`;
  the full `sqlParserConfig(…)` path (verifying it retains isthmus' `LENIENT` conformance); and
  end-to-end — a plan built from `CREATE TABLE employees … / SELECT … FROM employees` produces a
  `NamedScan` named `EMPLOYEES` under `TO_UPPER` and `employees` under `UNCHANGED`.
- `CustomFunctionTest` and `UserDefinedLiteralRoundtripTest` migrated to the builder (they used the
  now-deprecated full-customization constructor).
- `DdlToSubstraitConversionTest` / `DdlToSubstraitConversionWithOptimizationTest` updated to the
  `ConverterProvider` API.
- Remaining tests and examples migrated off the now-deprecated constructors to `builder()` /
  `ConverterProvider.DEFAULT`: `PlanTestBase`, `RelExtensionRoundtripTest`,
  `PostgreSqlIntegrationTest`, `UdfSqlSubstraitTest`, `AnyValueFunctionTest`,
  `AutomaticDynamicFunctionMappingRoundtripTest`, and examples `ToOptimizedSql`, `DynamicFnToSql`,
  `CustomDialectDynamicFnToSql`.

🤖 Generated with AI


isthmus/src/main/java/io/substrait/isthmus/ConverterProvider.java
```diff
@@ -173,7 +260,7 @@ private static Plan.ExecutionBehavior createDefaultExecutionBehavior() {
    */
   public SqlParser.Config getSqlParserConfig() {
     return SqlParser.Config.DEFAULT
-        .withUnquotedCasing(Casing.TO_UPPER)
+        .withUnquotedCasing(unquotedCasing)
         .withParserFactory(SqlDdlParserImpl.FACTORY)
         .withConformance(SqlConformanceEnum.LENIENT);
   }
```
vbarua, line 264: I don't think we should have an `unquotedCasing` option like this on the ConverterProvider.

SqlParser.Config has a [ton of options](https://calcite.apache.org/javadocAggregate/org/apache/calcite/sql/parser/SqlParser.Config.html) on it. Historically, the issue that we ran into was that we added flags to toggle bits and pieces as we needed.

The reason this is structured like this to return a full `SqlParser.Config` is to allow users to extend the ConverterProvider to return their system specific parsing config.

vbarua:
> Previously the only way to change this was to subclass ConverterProvider and override getSqlParserConfig() — as IsthmusEntryPoint already did with an anonymous class. That workaround is now replaced by a first-class constructor parameter.

Left a code comment to this effect, but this actually not a feature I think we should support. SQL parsing is very much a Calcite concern, and for users that need to customize it I would push them to handle their parsing purely in Calcite and then hand the Calcite plans to the Isthmus converters.

Is there a specific usecase that you want this for?

nielspardon:
**TL;DR:** isthmus hard-codes `Casing.TO_UPPER`, so plans it generates carry upper-cased identifiers (`ORDERS`, `O_ORDERKEY`) that don't match engines which fold to lower-case (Presto/Trino, Spark, DuckDB, Postgres). The `isthmus` CLI already exposes `--unquotedcasing` for exactly this — but the only way it can implement that today is an anonymous `ConverterProvider` subclass, which every other consumer then has to duplicate. This PR makes casing first-class so isthmus-generated plans are portable to non-upper-casing engines. Casing is worth a dedicated knob (vs. the general `getSqlParserConfig()` override) because, unlike most `SqlParser.Config` options, it changes the *output* — the identifier names written into the plan — not just how SQL text is parsed. Happy to keep the `SqlParser.Config` overloads so it's non-breaking.

---

Hi Victor — thanks, both points are fair. Here's the concrete use case that drove this, and why casing specifically feels different from the general "one more `SqlParser.Config` toggle" concern.

**The specific use case: Substrait plans generated by isthmus don't round-trip against engines that don't fold to upper-case.**

isthmus hard-codes `Casing.TO_UPPER`. That's the ANSI fold direction and Calcite's default, but a large class of real engines fold unquoted identifiers to lower-case or preserve them — Presto/Trino, Spark, DuckDB, Postgres. So a plan generated from `SELECT … FROM orders` comes out with a `NamedScan` named `ORDERS` and columns like `O_ORDERKEY`. When that plan is handed to a lower-casing engine, none of the table/column names match what the engine actually has.

This isn't hypothetical — I ran into it head-on while prototyping a native Substrait consumer for Presto (Substrait → Presto `PlanNode` IR directly, no SQL bridge; still WIP on a branch in my fork). Presto normalizes unquoted identifiers to lower-case, so isthmus-generated plans came in upper-cased and didn't resolve. To make it work I had to:

- add case-insensitive name resolution + a lower-casing bridge in the plan converter (the `NamedScan` handling ended up with a comment: *"field names may differ in case from Presto's (lower-cased) column names"*), and
- deliberately write the TPC-H/TPC-DS test queries with **unquoted** names so isthmus's and Presto's identifier casing would line up at all.

The integration itself isn't upstream and isn't a dependency here — I'm raising it only as evidence that the mismatch is real: any consumer targeting a lower-casing engine has to reinvent that bridge, purely because the plan-generation side has no way to say "don't upper-case."

**The examples already demonstrate the same inconsistency.** `FromSql` generates a plan for DuckDB and needs the identifiers left as written — it was leaning on the deprecated `convert(sql, catalog, SqlDialect)` to get there. With casing configurable it just asks for `Casing.UNCHANGED` and the lower-case names survive end to end.

**On "push users to parse in Calcite themselves":** for most of `SqlParser.Config`, I agree completely. But these are precisely the consumers who *do* want isthmus's SQL front-end rather than hand-rolling Calcite — the `isthmus` CLI, the examples, and plan-generation/test harnesses. In fact the CLI **already ships a first-class `--unquotedcasing` flag**, so casing control is an established, supported isthmus feature. The only way `IsthmusEntryPoint` can implement it today is the anonymous `ConverterProvider` subclass overriding `getSqlParserConfig()` — which is exactly the workaround this PR removes. So the inconsistency is internal too: the CLI exposes casing, but the library API forces everyone else to duplicate that subclass boilerplate for the one setting we hit constantly.

**On the API shape (your inline comment):** your flag-creep worry is legitimate, and I don't want to relitigate the history of bolting individual parser toggles onto the API — `getSqlParserConfig()` returning a full config should stay the extension point for arbitrary Calcite parser settings. My argument is that **casing is not just a parse-time setting** the way most of `SqlParser.Config` is: it leaks into the *output*. It determines the identifier names written into the emitted Substrait plan, and therefore whether the plan is portable to the target engine at all. That's why it seems worth a first-class knob while everything else stays behind the override.

Happy to reshape to whatever you're comfortable with. A couple of options:

1. **Keep it minimal / non-breaking** — drop the constructor param and the removal of the `SqlParser.Config` overloads, and just document + use the `getSqlParserConfig()` override pattern in the CLI/examples. (Keeps the boilerplate, but no API surface change.)
2. **First-class `unquotedCasing` as in this PR** — because it's the one parser option that changes plan semantics, not just parsing — but I can keep the `SqlParser.Config` overloads so it isn't a breaking change.

I'd lean toward (2)-without-the-breaking-removal, but I'm glad to go with (1) if you'd rather not grow `ConverterProvider`'s surface. What's your preference?

vbarua:
> TL;DR: isthmus hard-codes Casing.TO_UPPER, so plans it generates carry upper-cased identifiers (ORDERS, O_ORDERKEY) that don't match engines which fold to lower-case (Presto/Trino, Spark, DuckDB, Postgres).

So what you're saying is that if we generate plans with upper-cased identifiers, that only works on some systems, and the same thing with lower-cased identifiers. This sounds like a deeper problem with the spec, because now we have plans that are basically identical except for the casing of names that don't work across engines. Making it easier to control the casing in the output papers over the issue here.

nielspardon:
> This sounds like a deeper problem with the spec, because now we have plans that are basically identical except for the casing of names that don't work across engines. Making it easier to control the casing in the output papers over the issue here.

From my perspective names in the spec are case-sensitive aka quoted identifiers in SQL-speak. We only have the problem when we parse SQL statements into Substrait plans. Currently, the `SqlParser.Config` is not aligned across the create statement parsing which creates the Calcite catalog and the other SQL to Substrait parsing. Further we are testing round-tripping by parsing SQL statements using **unquoted** identifiers in SQL and not **quoted** identifiers. Our users might have the same problem that they want to parse SQL with **unquoted** identifiers into Substrait plans so they very frequently want to configure the behavior for how to handle unquoted identifiers when parsing SQL statements.

vbarua:
> From my perspective names in the spec are case-sensitive aka quoted identifiers in SQL-speak.

That does feel reasonable. Anything else is a can-of-worms, and we might want to open it one day but it doesn't need to be today.

> We only have the problem when we parse SQL statements into Substrait plans. Currently, the SqlParser.Config is not aligned across the create statement parsing which creates the Calcite catalog and the other SQL to Substrait parsing.

Gotcha, from this I think it's perfectly reasonable to inject the ConvertProvider into the statement parsers.

> Our users might have the same problem that they want to parse SQL with unquoted identifiers into Substrait plans so they very frequently want to configure the behavior for how to handle unquoted identifiers when parsing SQL statements.

How are we expecting users to parse SQL? Are they doing so through the isthmus CLI, or are they integrating with the full Java library?

nielspardon:
> How are we expecting users to parse SQL? Are they doing so through the isthmus CLI, or are they integrating with the full Java library?

I would say both currently. For the Presto integration I used isthmus (the lib) to take their existing SQL test suite and create Substrait plans on the fly which are in Presto SQL dialect and may or may not use unquoted identifiers. We are also doing that for some of our internal work with a number of pre-existing SQL based benchmarks which may have been created with various SQL dialects.

The CLI is then just a thin wrapper around the lib which allows one to use a subset of the isthmus features without requiring Java.

vbarua (commented):
> For the Presto integration I used isthmus (the lib) to take their existing SQL test suite and create Substrait plans on the fly which are in Presto SQL dialect and may or may not use unquoted identifiers. 

Gotcha. How much of the pain points are resolved by allowing the injection of the ConverterProvider into the `SubstraitCreateStatementParser` and `SubstraitSqlStatementParser` like you've done.

I've been positioning them as being specific to Substrait, and that can be seen in the comments.
```
/**
 * Substrait flavoured SQL processor provided as a utility for testing and experimentation,
 * utilizing {@link SubstraitSqlStatementParser} and {@link SubstraitSqlValidator}
 */
```

but I do think that being able to inject the ConverterProvider is a reasonable feature to aid with some of your integrations.

I'm still a little wary of is getting too much into the SQL parsing game, mostly because Calcite already does a really good job of that and I'd rather push users to lean on Calcite directly, instead of awardly from us.

> Casing is worth a dedicated knob (vs. the general getSqlParserConfig() override) because, unlike most SqlParser.Config options, it changes the output

I'm not unswayed by this, and won't block it, but I would still prefer to have user provide a full SqlParser.Config with this option set directly rather than exposing a toggle. In another conversation you suggested having builder for ConverterProvider, which I think is a really good idea given how many options we've been tacking onto that, and something like that would reduce the barrier to customizing the conversion with a SqlParser.Config.

nielspardon:
Thanks — reshaped this around the builder as you suggested; just pushed.

- Dropped the casing toggle entirely (no `unquotedCasing` field/getter/constructors). `ConverterProvider` now just holds a full `SqlParser.Config`, and the builder covers both paths:
  ```java
  // full SqlParser.Config
  ConverterProvider.builder()
      .sqlParserConfig(ConverterProvider.DEFAULT_SQL_PARSER_CONFIG.withUnquotedCasing(Casing.TO_LOWER))
      .build();

  // casing-only convenience (just sugar over the config, not a first-class toggle)
  ConverterProvider.builder().unquotedCasing(Casing.UNCHANGED).build();
  ```
  `DEFAULT_SQL_PARSER_CONFIG` is the exposed base (keeps the DDL factory + `LENIENT`). Subclassing `getSqlParserConfig()` still works for anything dynamic.
- Kept the `ConverterProvider` injection into the parsers — that's what fixes the real issue (CREATE parsing and query parsing now share one config, so casing lines up end-to-end).

One call for you: I also `@Deprecated` the two verbose all-components constructors (6-/7-arg) in favour of the builder, since your point was partly about how many options we keep tacking on. Happy to keep or drop that. Default behaviour is unchanged.

nielspardon:
Follow-up on the constructor-deprecation question I floated earlier — I've taken it a step further, pushed in b3e1009a.

Rather than deprecating only the verbose 6-/7-arg constructors, the builder is now the **sole public construction path**: every `ConverterProvider` constructor is `@Deprecated` in favour of `builder()` (or the shared `DEFAULT`). To keep subclassing viable without re-exposing a fragile positional constructor, there's a new `protected ConverterProvider(Builder)` seam — it's the single place instances are created (`build()` and every delegating constructor route through it) and it owns the function-converter derivation. The advantage over a `protected` positional master: adding a component later touches only the builder + this one constructor, so no subclass's `super(...)` call breaks.

Both in-tree subclasses were migrated to match — `DynamicConverterProvider` and `AutomaticDynamicFunctionMappingConverterProvider` each expose a public `(ConverterProvider.Builder)` constructor calling `super(builder)`, with their positional constructors `@Deprecated`.

Everything here is source-compatible (deprecation warnings only; the old constructors still delegate to the builder path), and the repo's own callers/tests/examples are migrated to `builder()` / `DEFAULT`.

Happy to dial this back if you'd prefer a lighter touch — e.g. keeping the small `(extensions, typeFactory)` convenience constructor non-deprecated and only deprecating the fragile all-components ones.

🤖 Generated with AI

nielspardon:
Closing in favour of splitting this work into three smaller, stacked PRs, as @vbarua suggested — each is scoped to one concern and reviewable on its own:

1. #1035 — inject `ConverterProvider` into the SQL statement parsers
2. #1036 — introduce the `ConverterProvider` builder

They stack bottom-up (#1036 targets #1035's branch, #1037 targets #1036's), so please review/merge in order. The union of the three reproduces this PR's diff exactly.

🤖 Generated with AI

PR #986, opened 2026-07-03
feat(isthmus): map RIGHTSHIFT to Substrait shift_right

nielspardon:
## What

Adds a mapping so a right-shift can be converted to the Substrait `shift_right` function. Calcite defines the `<<` operator (`BIT_LEFT_SHIFT`) and the `LEFTSHIFT` function — both already mapped to Substrait `shift_left` — but has **no right-shift counterpart at all**. This PR defines an Isthmus `RIGHTSHIFT(value, shift)` `SqlFunction` (mirroring Calcite's `LEFTSHIFT`), registers it in `SubstraitOperatorTable`, and maps it to `shift_right`.

## Why

`shift_right` exists in `functions_arithmetic.yaml` (i32/i64, same shape as `shift_left`) but was unreachable from SQL because no Calcite operator produced it.

## Scope

- Covers the **function-call spelling** `RIGHTSHIFT(x, n)`. Round-trips to Substrait `shift_right` and back.
- The **`>>` infix operator is intentionally out of scope.** Calcite 1.42 has no `>>` token, no `SqlKind`, and no operator (only `<<`). Supporting `>>` requires isthmus to *generate* its own parser: Calcite exposes the needed extension hooks (`binaryOperatorsTokens` + `extraBinaryExpressions`, the same mechanism calcite-babel uses for the `::` infix cast), but isthmus currently consumes the prebuilt `SqlDdlParserImpl` and has no parser codegen. Tracking that as a separate follow-up; the cleanest long-term fix is likely to upstream `BIT_RIGHT_SHIFT` into Calcite core (a genuine gap, given it already has `<<`), which is being pursued in [apache/calcite#5073](https://github.com/apache/calcite/pull/5073) — once that lands, this reduces to a one-line mapping.

## Testing

- New round-trip test in `ArithmeticFunctionTest` (`i8`/`i16`/`i32`/`i64`), symmetric to the existing `leftshift` test.
- Full `:isthmus:test` suite passes locally.

🤖 Generated with AI

### DIFF
--- a/isthmus/src/main/java/io/substrait/isthmus/calcite/SubstraitOperatorTable.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/calcite/SubstraitOperatorTable.java
@@ -2,6 +2,7 @@ package io.substrait.isthmus.calcite;
 
 import io.substrait.isthmus.AggregateFunctions;
 import io.substrait.isthmus.expression.CurrentTimezoneFunction;
+import io.substrait.isthmus.expression.FunctionMappings;
 import java.util.EnumSet;
 import java.util.List;
 import java.util.Set;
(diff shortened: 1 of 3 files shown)


vbarua (approved):
Looks reasonable to me.

PR #989, opened 2026-07-06
fix(ci): pin semantic-release tooling and restore v0.95.0 and v0.95.1 notes

nielspardon:
## Summary

Fixes semantic-release changelog generation and backfills the missing `v0.95.0` and `v0.95.1` changelog entries. Both releases were published with empty notes — the `CHANGELOG.md` entry and the GitHub release contained only the header.

## Changes

- pin the semantic-release toolchain in [`ci/release/run.sh`](ci/release/run.sh) and [`ci/release/dry_run.sh`](ci/release/dry_run.sh), including [`@semantic-release/changelog@6.0.3`](ci/release/run.sh)
- migrate semantic-release config from [`.releaserc.json`](.releaserc.json) to [`.releaserc.mjs`](.releaserc.mjs)
- add a `writerOpts.transform` hook in [`.releaserc.mjs`](.releaserc.mjs) to strip git trailers like `Signed-off-by:` from breaking-change notes
- default the config to a dry run and only include the publishing plugins when `RELEASE_DRY_RUN=false`; [`ci/release/run.sh`](ci/release/run.sh) opts into a real release
- document in [`.releaserc.mjs`](.releaserc.mjs) why the preset import needs the `npx` fallback and why the publishing plugins are omitted via `RELEASE_DRY_RUN` rather than `--dry-run`
- restore the missing `v0.95.0` entry in [`CHANGELOG.md`](CHANGELOG.md) (and de-duplicate it — the restored block had been repeated)
- backfill the missing `v0.95.1` entry in [`CHANGELOG.md`](CHANGELOG.md), generated with the pinned toolchain

## Root cause

The empty release notes were caused by floating release-tool dependencies. In particular, `conventional-changelog-conventionalcommits@10.x` regressed changelog generation in this setup, while `9.3.1` still generated the expected sections.

## Validation

- reproduced the empty-notes behavior with the floating toolchain
- verified that pinning [`conventional-changelog-conventionalcommits@9.3.1`](ci/release/run.sh) restores proper release notes generation
- validated the trailer-stripping transform with a local semantic-release run against the `v0.95.0` release range
- regenerated the `v0.95.1` notes with the pinned toolchain over the `v0.95.0..v0.95.1` range
- re-ran a dry run after merging `main` to confirm the config still loads and produces valid notes


nielspardon:
This also ports the fix for stripping git trailers from https://github.com/substrait-io/substrait/pull/1099 to substrait-java

ci/release/run.sh
```diff
@@ -3,13 +3,13 @@
...
-  -p "@semantic-release/changelog" \
-  -p "@semantic-release/github" \
-  -p "@semantic-release/exec" \
-  -p "@semantic-release/git" \
-  -p "conventional-changelog-conventionalcommits" \
+RELEASE_DRY_RUN=false npx --yes \
+  -p "semantic-release@25.0.5" \
+  -p "@semantic-release/commit-analyzer@13.0.1" \
+  -p "@semantic-release/release-notes-generator@14.1.1" \
+  -p "@semantic-release/changelog@7.0.0-beta.1" \
```
bestbeforetoday: Using a prerelease beta version does not seem ideal unless there is some reason that release versions are not compatible or do not include some essential functionality. The latest recommended release is currently `6.0.3`.

.releaserc.mjs
```diff
@@ -0,0 +1,148 @@
...
+    return (await import("conventional-changelog-conventionalcommits")).default;
+  } catch {
+    const require = createRequire(pathToFileURL(process.argv[1]));
+    const resolved = pathToFileURL(
+      require.resolve("conventional-changelog-conventionalcommits")
+    ).href;
+    return (await import(resolved)).default;
+  }
+};
+const conventionalcommits = await loadPreset();
```
bestbeforetoday, line 44: I am not sure why the import handling is so complicated here. The [examples](https://conventional-changelog.js.org/presets/conventional-commits/#use-it-with-the-js-api) in the documentation are much simpler, using static rather than dynamic imports, and more what I would expect for ESM imports.

bestbeforetoday (commented):
Do we really need the `RELEASE_DRY_RUN` environment variable and programmatic manipulation of the semantic release configuration in addition to the standard `--dry-run` command-line argument? This already skips publishing, and it looks like the programmatic configuration manipulates (removes) publishing setting that (I guess) should not be used with `--dry-run` anyway. Without manually testing it is hard for me to tell for sure.

ci/release/run.sh
nielspardon: Good catch — there's no reason to be on the beta here. `@semantic-release/changelog@6.0.3` is the current `latest` dist-tag and declares `semantic-release >=18.0.0` as its peer dependency, so it's compatible with the pinned `semantic-release@25.0.5`. I've pinned `6.0.3` in both `run.sh` and `dry_run.sh` and validated it with a dry run: the preset resolves cleanly and the notes render correctly (`0.95.0` → `0.95.1`).

.releaserc.mjs
nielspardon: The dynamic import + fallback is unfortunately load-bearing given how we invoke the tool. The release scripts run semantic-release via `npx -p …`, which installs `conventional-changelog-conventionalcommits` into a *temporary* `node_modules` alongside the semantic-release binary. A bare static `import` from this config resolves relative to the config file's own directory (the repo root, which has no `node_modules`) and fails with `ERR_MODULE_NOT_FOUND` — so the `catch` re-resolves the preset relative to the running semantic-release binary (`process.argv[1]`). The static-import example in the docs assumes the preset is installed as a local devDependency, which isn't the case here.

This was ported from the spec repo but the explanatory comment got dropped in the port — I've restored it so the "why" lives next to the code.

nielspardon:
> Do we really need the `RELEASE_DRY_RUN` environment variable […] in addition to the standard `--dry-run` command-line argument?

Yes — this is a deliberate safety mechanism copied from the spec repo (substrait-io/substrait#1099), and `--dry-run` alone isn't sufficient. semantic-release still runs every plugin's `verifyConditions` step in dry-run mode; it skips only `prepare`, `publish`, `addChannel`, `success`, and `fail`. `@semantic-release/github`'s `verifyConditions` throws `ENOGHTOKEN` without a `GITHUB_TOKEN`, so a plain `--dry-run` with the full plugin set can't produce a credential-free preview. `RELEASE_DRY_RUN` lets us *omit* the side-effecting plugins entirely (and defaults to dry-run as a fail-safe, so a forgotten/typo'd var stays harmless) rather than merely guarding them behind `--dry-run`.

My dry run confirms this end to end — it completes token-free precisely because those plugins are dropped:

```
✔  Completed step "verifyConditions" of plugin "@semantic-release/exec"
✔  Completed step "verifyConditions" of plugin "@semantic-release/changelog"
ℹ  The next release version is 0.95.1
⚠  Skip step "prepare"  … in dry-run mode
⚠  Skip step "publish"  … in dry-run mode
⚠  Skip step "success"  … in dry-run mode
```

I've restored the comment explaining this (dropped in the port) and corrected it — the original claimed dry-run "only skips tag and push," which isn't accurate.

nielspardon:
Rebased on latest `main` via merge and backfilled the `0.95.1` notes, which were released empty (the CHANGELOG entry and the GitHub release both had just the header) since this fix hadn't merged yet:

- **Merged `upstream/main`** — brings the branch up to date; the only overlap was `CHANGELOG.md`.
- **Added the `0.95.1` entry.** Generated with the pinned tooling over the `v0.95.0..v0.95.1` range rather than hand-written; it contains just the one `fix` (`declare constants as final`, #996). The dependency bumps and the `ci` commit are omitted, as expected — the `conventionalcommits` preset hides those types unless they carry a breaking change (which is why `0.95.0` has a Build System line: its protobuf bump was breaking).
- **De-duplicated the `0.95.0` entry** — the restored block had been accidentally repeated twice under a single header; it now appears once.

.releaserc.mjs
```diff
@@ -0,0 +1,170 @@
...
+  "Co-authored-by",
+  "Co-developed-by",
+  "Reviewed-by",
+  "Acked-by",
+  "Tested-by",
+  "Reported-by",
+  "Suggested-by",
+  "Helped-by",
+  "Cc",
+];
```
andrew-coleman, line 59: Should we also add "Mentored-by" since it appears in this list:
https://git-scm.com/docs/SubmittingPatches#commit-trailers

nielspardon: Good idea — added in 376db89f. `Mentored-by` was indeed the only trailer from that git list missing here, so the strip list now covers all of them (plus `Cc`).

andrew-coleman (approved):

.releaserc.mjs
bestbeforetoday: Just for future reference, another less safe but more flexible option might have been to remove anything that fits the format of a Git trailer. Something like this:

```javascript
const trailerRegex = /\n(?:\n[A-Za-z-]+:\s[^\n]*)+(?:\s)*$/;
commitMessage.replace(trailerRegex, "");
```

PR #996, opened 2026-07-09
fix: declare constants as final

bestbeforetoday:
Some public static constants were not declared final, allowing them to be incorrectly modified at runtime. This change adds the final modifier to these constants.


benbellick (approved):
nice!

PR #997, opened 2026-07-09
ci: build only the Linux native image per PR

### LINKED ISSUE #768: ci: building isthmus native image takes long time
For a PR's CI checks to go through, we have to build Isthmus native image. [This takes ~15 min](https://github.com/substrait-io/substrait-java/actions/runs/23260941524/job/67640331513), which is quite a while for a relatively small repo. 

We should investigate either:
- making this build step faster, or
- dropping the requirement for something simpler / faster

nielspardon:
## What

Speeds up PR builds by taking the slow, 10x-billed **macOS** native-image build off the PR critical path, without losing macOS coverage.

The native image built on PRs is only used to run the smoke tests (`smoke.sh` / `tpch_smoke.sh`) and is uploaded as a throwaway PR artifact — the shipped release binaries are compiled separately in `release.yml`. That means the *runtime performance* of the PR binary is irrelevant, which unlocks the cheapest win below.

## Changes

1. **Quick build mode on PRs.** The Linux native image is now built with `--quick-build-native` (GraalVM `-Ob`), which skips the expensive optimizing compiles and typically cuts native-image build time ~30–40%. The tradeoff — a slower/larger binary — doesn't matter because the PR binary is only smoke-tested.
2. **Linux-only per PR.** The `macOS-latest` matrix entry is dropped from the PR native-image job. The Linux build already validates the GraalVM configuration and runs the full smoke suite; macOS-specific native regressions are rare and now covered out-of-band (see below). macOS runners are the long pole *and* bill at 10x.
3. **Runs in parallel with tests.** Removed `needs: [java, integration]` so the native-image job starts immediately instead of waiting for the Java + integration jobs to finish.

To keep macOS native coverage, a new `native-image-macos.yml` workflow builds and smoke-tests the macOS native image (full optimization, mirroring the release build) on **push to `main`** (per-merge signal) plus a **weekly** backstop and `workflow_dispatch`. Regressions are still caught promptly, and before the Sunday release build — just not on the per-PR critical path.

## Notes for reviewers

- **Required status checks:** the per-PR macOS native check (`Build Isthmus Native Image (macOS-latest)`) no longer runs. If branch protection requires it, that setting needs updating; the Linux check (`Build Isthmus Native Image (ubuntu-latest)`) is unchanged.
- `--quick-build-native` was verified to be a recognized option on the `nativeCompile` task (`gradle help --task nativeCompile` → "Enables quick build mode").
- `release.yml` is intentionally left untouched — release binaries stay fully optimized on both Linux and macOS.

Closes #768

🤖 Generated with AI

### DIFF
--- a/.github/workflows/native-image-macos.yml
+++ b/.github/workflows/native-image-macos.yml
@@ -1,0 +1,51 @@
+name: macOS Native Image
+
+# macOS native-image coverage is kept off the per-PR critical path: the macOS
+# build is the long pole and macOS runners bill at 10x. Instead we build and
+# smoke-test the macOS native image on every push to main (per-merge signal)
+# plus a weekly backstop, so macOS-specific native regressions are still caught
+# promptly before the Sunday release. PRs build only the Linux native image
+# (see pr.yml).
+
+on:
+  push:
+    branches:
+      - main
+  schedule:
+    # Saturday 04:23 UTC — backstop ahead of the Sunday release build.
+    - cron: "23 4 * * 6"
+  workflow_dispatch:
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  isthmus-native-image-macos:
+    name: Build Isthmus Native Image (macOS)
+    if: github.repository == 'substrait-io/substrait-java'
+    runs-on: macOS-latest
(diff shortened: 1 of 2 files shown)


benbellick (approved):
🚀 this has been a long time coming. Thanks for taking a look!

PR #998, opened 2026-07-10
feat(isthmus)!: remove invalid SetOp to Calcite relation mappings

### LINKED ISSUE #303: remove incorrect mappings from SetOps to Calcite
In https://github.com/substrait-io/substrait-java/pull/300 we added mappings for

nielspardon:
## Summary

Removes the two incorrect `SetOp` → Calcite relation mappings in the Substrait → Calcite converter (`SubstraitRelNodeConverter`):

- `MINUS_MULTISET` → Calcite `Minus(all=true)`
- `INTERSECTION_PRIMARY` → Calcite `Intersect(all=false)`

These mappings were incorrect — the Substrait spec behaviour of these set operations does not match the Calcite relations they were mapped to. They were retained temporarily (with a `TODO` referencing this issue) to let users migrate off of them after the correct mappings were introduced.

Neither `MINUS_MULTISET` nor `INTERSECTION_PRIMARY` has an equivalent Calcite relation, so converting a plan containing either to Calcite now throws `UnsupportedOperationException` (the `default` branch of the switch), consistent with how the test utilities already excluded them.

Note: `MINUS_MULTISET` and `INTERSECTION_PRIMARY` remain valid Substrait spec operations — this only touches the Isthmus (Calcite) conversion. The core POJO model, dialect enum, type-derivation, `SqlKindFromRel`, and the Calcite → Substrait direction (`SubstraitRelVisitor`, which already produces the correct ops) are unchanged.

## Changes

- Remove the `MINUS_MULTISET` and `INTERSECTION_PRIMARY` cases from `SubstraitRelNodeConverter.getRelBuilder`.
- Remove the stale `TODO` comment referencing the issue.
- Add tests asserting both ops now throw `UnsupportedOperationException` on conversion.
- Clarify the exclusion comment in `SetUtils`.

## Breaking change

Converting a Substrait plan that uses `MINUS_MULTISET` or `INTERSECTION_PRIMARY` to Calcite now throws `UnsupportedOperationException` instead of producing an incorrect Calcite relation. The supported set-difference operations are `MINUS_PRIMARY` (`Minus all=false`) and `MINUS_PRIMARY_ALL` (`Minus all=true`); the supported intersection operations are `INTERSECTION_MULTISET` (`Intersect all=false`) and `INTERSECTION_MULTISET_ALL` (`Intersect all=true`).

## Testing

- `./gradlew :isthmus:test --tests "io.substrait.isthmus.SubstraitRelNodeConverterTest" --tests "io.substrait.isthmus.Substrait2SqlTest" --tests "io.substrait.isthmus.ProtoPlanConverterTest"`
- `./gradlew spotlessApply :core:check :isthmus:check`

Closes #303

🤖 Generated with AI

### DIFF
--- a/isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
@@ -285,10 +285,6 @@ public class SubstraitRelNodeConverter
             input -> {
               relBuilder.push(input.accept(this, context));
             });
-    // TODO: MINUS_MULTISET and INTERSECTION_PRIMARY mappings are set to be removed as they do not
-    //   correspond to the Calcite relations they are associated with. They are retained for now
-    //   to enable users to migrate off of them.
-    //   See:  https://github.com/substrait-io/substrait-java/issues/303
     RelBuilder builder = getRelBuilder(set);
     RelNode node = builder.build();
     return applyRemap(node, set.getRemap());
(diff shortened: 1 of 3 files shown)


vbarua (approved):
Oh, hello my old issue. Makes sense to remove these if they are invalid.

PR #1000, opened 2026-07-10
docs: trim stale native-image CI note in AGENTS.md

nielspardon:
## Summary

Trims the AGENTS.md note about monitoring the native-image CI job down to the one durable, non-obvious fact.

The note originally existed to warn contributors/agents that a PR staying yellow was usually just waiting on the slow macOS `Build Isthmus Native Image` long-pole job. #997 removed that long pole — the per-PR native build is now **Linux only**, uses `--quick-build-native`, and runs **in parallel** with the `java`/`integration` jobs. So most of what the note (and my first pass at updating it) described was either the *absence* of a gotcha or detail already spelled out in `pr.yml`.

Per @bestbeforetoday's review feedback — this is guidance loaded into an agent's context every session, so it should earn its keep — I cut it to the single fact that isn't easily discoverable and that changes behavior:

> The macOS native image is not built on PRs (only Linux is), so macOS-specific native regressions surface on `main` or the weekly `native-image-macos.yml` run, not on the PR.

Docs-only; no code or workflow behavior is affected.

🤖 Generated with AI

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -205,11 +205,5 @@ compile — they have their own visitor implementors:
   proto message usually needs: POJO + visitor wiring + both proto converters + a
   round-trip test, and often `ExpressionCreator` factories and `dsl/SubstraitBuilder`
   helpers for ergonomics.
-- When monitoring PR checks: the per-PR native-image job (`Build Isthmus Native Image` in
-  `pr.yml`) builds **Linux only**, uses `--quick-build-native`, and runs **in parallel**
-  with the `java` + `integration` jobs (no `needs:`), so it's fast and not a long pole. The
-  **macOS** native image is *not* built on PRs — it runs out-of-band in
-  `.github/workflows/native-image-macos.yml` (push to `main`, a weekly backstop, and
-  `workflow_dispatch`), fully optimized to mirror `release.yml`. So a PR that stays yellow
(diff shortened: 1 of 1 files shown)


bestbeforetoday:
The information all looks correct to me. I wonder how useful it is to aid an AI agent in developing code. Is it worth taking up more of the context window with this information or would it be more effective to just remove it?

nielspardon:
Good push — agreed. As the thing this file is written for, I'll say the note was low value: it's guidance loaded into context every session, so it should earn its keep, and this one didn't. Two reasons beyond the general context-window cost:

- It was operational CI trivia, not something that changes how code gets written, and most of it (the `pr.yml` job layout, `--quick-build-native`, `needs:`) is discoverable by just reading the workflow.
- The note originally existed to warn "don't panic when a PR stays yellow — the macOS native job is a slow long pole." #997 removed that long pole, so the expanded version was largely documenting the *absence* of a gotcha.

So rather than remove it outright, I trimmed it to the one residual fact that isn't obvious and that actually changes behavior: macOS native isn't built per-PR, so those regressions land on `main`/the weekly run, not the PR. Everything else is gone. Updated the PR title/body to match.

bestbeforetoday (approved):

Issue #994, opened 2026-07-09
Provide semantics-aware plan comparison (order-insensitive equality)


nielspardon:
## Background

Split out from #203. That issue reported that a proto → POJO → proto roundtrip reorders extension declarations. The conclusion there was that this is **not a bug**: the Substrait spec doesn't constrain extension ordering, and `substrait-java` allocates extension anchors in discovery order during POJO → proto conversion (deterministic and harmless). We don't intend to guarantee that an externally-authored plan's extension ordering survives a roundtrip.

However, a distinct and still-open suggestion came out of that discussion (from @vbarua, @vibhatha, and @danepitkin): two plans that differ only in spec-irrelevant details (e.g. extension anchor numbering / declaration ordering) currently compare as **not equal**, because `io.substrait.plan.Plan` uses the stock order-sensitive `equals`/`hashCode` generated by Immutables.

> If we consider these plans to be equal, this would be a good argument for a smarter equals method.

## Proposal

Provide a **semantics-aware comparison** for plans — a way to determine that two plans are equivalent modulo details the spec does not constrain (extension anchor values, extension declaration ordering, etc.). Prior art: substrait-cpp's `ReferenceNormalizer` (normalize anchors in place, then compare).

Options to consider:
- A `normalize()` / plan-normalization utility that canonicalizes anchor numbering and declaration ordering, so normalized plans can be compared with ordinary `equals`.
- A dedicated comparator/equivalence helper rather than overriding `Plan.equals` (overriding `equals` on the Immutables POJO is likely too invasive and surprising).

## Additional case: associativity of variadic functions

Surfaced while reviewing #1001 (variadic `concat` conversion in Isthmus). Because Calcite's `||` (`SqlStdOperatorTable.CONCAT`) is strictly binary, a flat Substrait `concat(a, b, c)` is converted to a left-nested `CONCAT(CONCAT(a, b), c)`, and back to Substrait as `concat(concat(a, b), c)`. So a Substrait → Calcite → Substrait pass over an externally-authored flat variadic `concat` is a non-identity transform, even though the two plans are semantically equivalent (concat is associative).

This is a different — and larger — category than the anchor/ordering cases above:
- Anchor numbering / declaration ordering is **spec-irrelevant metadata**; normalization is a straightforward canonicalization.
[...]

PR #1001, opened 2026-07-11
fix(isthmus): support variadic concat conversion

### LINKED ISSUE #724: Substrait string concat with more than two arguments fails conversion to Calcite
The Substrait `concat` function is variadic so allows more than two arguments. When converting a Substrait concat expression with three (or more) arguments to Calcite, an error occurs since Calcite's corresponding `||` operator only allows two operands:

```
java.lang.AssertionError: wrong operand count 3 for ||
	at org.apache.calcite.util.Litmus.lambda$static$0(Litmus.java:31)
	at org.apache.calcite.sql.SqlBinaryOperator.validRexOperands(SqlBinaryOperator.java:229)
	at org.apache.calcite.rex.RexCall.<init>(RexCall.java:93)
	at org.apache.calcite.rex.RexBuilder.makeCall(RexBuilder.java:293)
	at org.apache.calcite.rex.RexBuilder.makeCall(RexBuilder.java:279)
	at io.substrait.isthmus.expression.ExpressionRexConverter.visit(ExpressionRexConverter.java:433)
```

This could be resolved by modifying the expression during conversion to be nested / chained expressions, each with only two arguments. For example, `concat(a, b, c)` becomes `concat(concat(a, b), c)`. This results in the expected Calcite (and SQL) output.

bvolpato:
## Summary

Convert variadic Substrait `concat` expressions into left-associated binary Calcite calls. This preserves argument order while satisfying Calcite `||` operand-count validation.

Closes #724.

## Changes

- Fold concat calls with more than two arguments into nested binary calls.
- Keep existing two-argument conversion unchanged.
- Add regression coverage for operator shape and operand order.

## Repro

Converting `concat("a", "b", "c")` previously passed three operands to Calcite `||` and failed with `wrong operand count 3 for ||`. It now produces `concat(concat("a", "b"), "c")`.

## Testing

- `./gradlew :isthmus:test --tests io.substrait.isthmus.FunctionConversionTest`
- `./gradlew spotlessApply`

Local `integrationTest` could not start because no Docker daemon is available; GitHub Actions runs that Testcontainers suite.

## Did this cause any problems?

No. Revert this commit to restore previous conversion behavior.

### DIFF
--- a/isthmus/src/main/java/io/substrait/isthmus/expression/ExpressionRexConverter.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/expression/ExpressionRexConverter.java
@@ -521,6 +521,13 @@ public class ExpressionRexConverter
             .collect(Collectors.toList());
 
     RelDataType returnType = typeConverter.toCalcite(typeFactory, expr.outputType());
+    if (operator == SqlStdOperatorTable.CONCAT && args.size() > 2) {
+      return args.stream()
+          .skip(1)
+          .reduce(
+              args.get(0),
+              (left, right) -> rexBuilder.makeCall(returnType, operator, List.of(left, right)));
+    }
     return rexBuilder.makeCall(returnType, operator, args);
   }
 
--- a/isthmus/src/test/java/io/substrait/isthmus/FunctionConversionTest.java
+++ b/isthmus/src/test/java/io/substrait/isthmus/FunctionConversionTest.java
@@ -20,6 +20,7 @@ import java.util.stream.Stream;
 import org.apache.calcite.rex.RexCall;
 import org.apache.calcite.rex.RexNode;
 import org.apache.calcite.sql.SqlKind;
+import org.apache.calcite.sql.fun.SqlStdOperatorTable;
 import org.junit.jupiter.api.Test;
 
 /**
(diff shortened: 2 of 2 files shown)


nielspardon (approved):
LGTM, the only issue this introduces is that we can not cleanly round-trip plans Substrait -> Calcite -> Subtrait if the initial Substrait plan contained a variadic concat function call. I don't think there is an elegant way to solve this during conversion and maybe should be considered in the semantic-aware plan comparison issue https://github.com/substrait-io/substrait-java/issues/994

vbarua:
I was looking at another PR and noticed that we'd added this special handling for the variadic Substrat concat:str function.

> Convert variadic Substrait concat expressions into left-associated binary Calcite calls

While we could convert the variadic concat into the binary concat operator, Calcite also has dedicated concat functions
* [CONCAT_FUNCTION](https://calcite.apache.org/javadocAggregate/org/apache/calcite/sql/fun/SqlLibraryOperators.html#CONCAT_FUNCTION)
* [CONCAT_FUNCTION_WITH_NULL](https://calcite.apache.org/javadocAggregate/org/apache/calcite/sql/fun/SqlLibraryOperators.html#CONCAT_FUNCTION)

These two functios are variadic, and there descriptions match the two possible behavours of [Substrait concat](https://github.com/substrait-io/substrait/blob/a254d75a8296f138871083bd4c125fd556d43e96/extensions/functions_string.yaml#L19-L21)
with null handling options of IGNORE_NULLS and ACCEPT_NULLS.

PR #1015, opened 2026-07-17
feat(isthmus): observe scalar function types

alexandrefimov:
## Summary

- add an opt-in type observer for Substrait-to-Calcite scalar function conversion
- report the source expression, supplied Substrait type, and either the independently inferred Calcite type or the inference failure
- expose observer configuration through `ConverterProvider` while keeping the default conversion path unchanged

Related to #379.

## Behavior

- observations are emitted for both matching and differing scalar function types
- the converted `RexNode` continues to use the type supplied by Substrait
- Calcite inference failures are reported without turning an otherwise successful conversion into a failure
- exceptions thrown by the observer are propagated to the caller
- variadic `CONCAT` produces one observation for the complete invocation

This PR intentionally covers scalar functions only. Window functions, field references, and aggregate functions remain separate follow-up work.

## Validation

- focused scalar observation tests covering matches, mismatches, inference failures, nested calls, provider injection, observer exceptions, and variadic `CONCAT`

Assisted-by: gpt-5.6-sol (OpenAI Codex)

### DIFF
--- a/isthmus/src/main/java/io/substrait/isthmus/expression/ExpressionRexConverter.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/expression/ExpressionRexConverter.java
@@ -26,8 +26,10 @@ import io.substrait.util.DecimalUtil;
 import java.math.BigDecimal;
 import java.util.Collections;
 import java.util.List;
+import java.util.Objects;
 import java.util.Set;
 import java.util.concurrent.TimeUnit;
+import java.util.function.Supplier;
 import java.util.stream.Collectors;
 import java.util.stream.IntStream;
 import java.util.stream.Stream;
@@ -95,6 +97,9 @@ public class ExpressionRexConverter
   /** Converter for Substrait window function invocations to Calcite {@link SqlOperator}s. */
   protected final WindowFunctionConverter windowFunctionConverter;
 
+  /** Observer for supplied and inferred expression types. */
+  protected final TypeObserver typeObserver;
+
   /** Converter for Substrait relational nodes to Calcite {@link RelNode}s, used for subqueries. */
   protected SubstraitRelNodeConverter relNodeConverter;
 
@@ -113,11 +118,35 @@ public class ExpressionRexConverter
       ScalarFunctionConverter scalarFunctionConverter,
       WindowFunctionConverter windowFunctionConverter,
       TypeConverter typeConverter) {
+    this(
+        typeFactory,
+        scalarFunctionConverter,
+        windowFunctionConverter,
+        typeConverter,
+        TypeObserver.NOOP);
+  }
+
+  /**
+   * Creates an {@code ExpressionRexConverter} with type observation enabled.
+   *
+   * @param typeFactory Calcite type factory for type creation
+   * @param scalarFunctionConverter converter for scalar function invocations
+   * @param windowFunctionConverter converter for window function invocations
+   * @param typeConverter converter for Substrait and Calcite type mappings
+   * @param typeObserver observer for supplied and independently inferred expression types
+   */
+  public ExpressionRexConverter(
+      RelDataTypeFactory typeFactory,
+      ScalarFunctionConverter scalarFunctionConverter,
+      WindowFunctionConverter windowFunctionConverter,
+      TypeConverter typeConverter,
+      TypeObserver typeObserver) {
     this.typeFactory = typeFactory;
     this.typeConverter = typeConverter;
     this.rexBuilder = new RexBuilder(typeFactory);
     this.scalarFunctionConverter = scalarFunctionConverter;
     this.windowFunctionConverter = windowFunctionConverter;
+    this.typeObserver = Objects.requireNonNull(typeObserver, "typeObserver");
   }
 
   /**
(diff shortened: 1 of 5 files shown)


nielspardon:
@vbarua can you take a look since you originally opened #379?

vbarua:
Did a quick pass and the direciton seems reasonable overall. Planning on taking a closer look tomorrow.

isthmus/src/main/java/io/substrait/isthmus/expression/ExpressionRexConverter.java
```diff
@@ -522,13 +551,49 @@ public RexNode visit(Expression.ScalarFunctionInvocation expr, Context context)
...
+      observeScalarType(
+          expr,
+          () ->
+              args.stream()
+                  .skip(1)
+                  .reduce(
+                      args.get(0),
+                      (left, right) -> rexBuilder.makeCall(operator, List.of(left, right))));
+      return suppliedCall;
+    }
```
vbarua: I got a little nerdsniped seeing that you had to work around this special case CONCAT handling, and am moving to just fully remove the handling which is both cleaner overall for dealing with CONCAT, and should simplify your work here.

https://github.com/substrait-io/substrait-java/pull/1025

isthmus/src/main/java/io/substrait/isthmus/expression/ExpressionRexConverter.java
```diff
@@ -522,13 +551,49 @@ public RexNode visit(Expression.ScalarFunctionInvocation expr, Context context)
...
+    RexNode suppliedCall = rexBuilder.makeCall(returnType, operator, args);
```
vbarua: minor: I would just suggest calling this something like `rexCallWithForcedSubstraitReturnType`, or just `rexCall`.

isthmus/src/main/java/io/substrait/isthmus/expression/ExpressionRexConverter.java
```diff
@@ -522,13 +551,49 @@ public RexNode visit(Expression.ScalarFunctionInvocation expr, Context context)
...
+    if (typeObserver == TypeObserver.NOOP) {
```
vbarua, line 554: A comment to the effect of
"If Type Observations are not needed, we return the call as is an avoid the work to recompute a RexCall with Calcite's inferred return type."
would be helpful here to explain why we're short-circuiting, and what we're doing below.

isthmus/src/main/java/io/substrait/isthmus/expression/ExpressionRexConverter.java
```diff
@@ -522,13 +551,49 @@ public RexNode visit(Expression.ScalarFunctionInvocation expr, Context context)
...
+
+  private void observeScalarType(
+      Expression.ScalarFunctionInvocation expression, Supplier<RexNode> inferredCallSupplier) {
+    TypeObservation observation;
+    RexNode inferredCall;
+    try {
+      inferredCall = inferredCallSupplier.get();
+    } catch (RuntimeException exception) {
+      observation =
+          TypeObservation.failure(TypeObservation.Source.SCALAR_FUNCTION, expression, exception);
```
vbarua: minor: I might suggest `failure` -> `inferenceFailure` or `calciteInferenceFailure` to make it clearer what failed. It took me a bit to realize we weren't checking for type differences here (which makes sense, that will come later).

I don't feel super strongly about it.

isthmus/src/main/java/io/substrait/isthmus/expression/TypeObservation.java
```diff
@@ -0,0 +1,109 @@
+package io.substrait.isthmus.expression;
+
+import io.substrait.expression.Expression;
+import io.substrait.type.Type;
+import java.util.Objects;
+import java.util.Optional;
+import org.apache.calcite.rel.type.RelDataType;
+
+/**
+ * The result of attempting to independently infer a Calcite type during expression conversion.
```
vbarua: I would specify Substrait to Calcite expression conversion... though could we ever do TypeObservations in the opposite direction (not something to implement in this PR).

isthmus/src/test/java/io/substrait/isthmus/SubstraitExpressionConverterTest.java
```diff
@@ -193,4 +361,47 @@ void assertTypeMatch(RelDataType actual, Type expected) {
     Type type = TypeConverter.DEFAULT.toSubstrait(actual);
     assertEquals(expected, type);
   }
+
+  private Expression.ScalarFunctionInvocation add(Type outputType) {
```
vbarua: I would give this a more specific name like `integerAddWithReturnType`

vbarua (commented):
Left some minor comments, but looks good overall.

alexandrefimov:
Thanks for the review! The five minor comments are clear. I’ll address them together with the CONCAT simplification after #1025 is merged.

vbarua:
CONCAT simplification is merged 🙂

isthmus/src/main/java/io/substrait/isthmus/expression/ExpressionRexConverter.java
alexandrefimov: Rebased on `main` after #1025 and removed the special-case CONCAT handling. Variadic CONCAT still has coverage asserting a single observation for the complete invocation.

isthmus/src/main/java/io/substrait/isthmus/expression/ExpressionRexConverter.java
alexandrefimov: Renamed this to `rexCall`.

isthmus/src/main/java/io/substrait/isthmus/expression/ExpressionRexConverter.java
alexandrefimov: Added the short-circuit explanation. I also added a regression test asserting that the default NOOP observer performs zero Calcite inference calls.

isthmus/src/main/java/io/substrait/isthmus/expression/ExpressionRexConverter.java
alexandrefimov: Renamed it to `inferenceFailure`.

isthmus/src/main/java/io/substrait/isthmus/expression/TypeObservation.java
alexandrefimov: Updated the Javadoc to specify Substrait-to-Calcite expression conversion.

isthmus/src/test/java/io/substrait/isthmus/SubstraitExpressionConverterTest.java
alexandrefimov: Renamed the helper to `integerAddWithReturnType`.

alexandrefimov:
@vbarua rebased on main and addressed all review comments. Ready for another look when you have a chance. Thanks!

alexandrefimov:
@vbarua, gentle follow-up when you have a chance :)

vbarua (approved):
Changes look good. Thanks for work on this 🙇

vbarua:
@alexandrefimov as a follow-up, if you wanted to make it easier to use the observer you've added, you could add it as a configurable setting in ConverterProvider.Builder
https://github.com/substrait-io/substrait-java/blob/243be42d193e305f6a6839735bbc337488c5aa0a/isthmus/src/main/java/io/substrait/isthmus/ConverterProvider.java#L546-L554

alexandrefimov:
Thanks!

PR #1025, opened 2026-07-23
feat(isthmus): map Substrait concat to Calcite CONCAT_FUNCTION

vbarua:
Updates the Substrait-Calcite mapping for the 
`extension:io.substrait:functions_string` `concat` from
`SqlStdOperatorTable.CONCAT` to `SqlLibraryOperators.CONCAT_FUNCTION`.
`CONCAT_FUNCTION` has the same semantics as `CONCAT`, but is variadic, like the
Substrait function.

Additionally, adds `ConcatFunctionMapper` to map
`SqlStdOperatorTable.CONCAT` (`||`) to `CONCAT_FUNCTION` when processing SQL,
which allows us to avoid special handling in the ExpressionRexConverter

Note that ONLY string concatenation (||) is mapped.

### DIFF
--- a/isthmus/src/main/java/io/substrait/isthmus/expression/ConcatFunctionMapper.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/expression/ConcatFunctionMapper.java
@@ -1,0 +1,47 @@
+package io.substrait.isthmus.expression;
+
+import io.substrait.expression.Expression;
+import io.substrait.expression.FunctionArg;
+import io.substrait.extension.DefaultExtensionCatalog;
+import io.substrait.extension.SimpleExtension.ScalarFunctionVariant;
+import java.util.List;
+import java.util.Optional;
+import java.util.stream.Collectors;
+import org.apache.calcite.rex.RexCall;
+import org.apache.calcite.sql.fun.SqlStdOperatorTable;
+
+/**
+ * Maps the SQL {@code ||} binary concatenation operator ({@link SqlStdOperatorTable#CONCAT}) to the
+ * Substrait {@code concat} function. This allows {@code ||} to continue working in the Calcite to
+ * Substrait direction while {@link org.apache.calcite.sql.fun.SqlLibraryOperators#CONCAT_FUNCTION}
+ * serves as the canonical Substrait-Calcite mapping via {@link FunctionMappings#SCALAR_SIGS}.
+ */
+final class ConcatFunctionMapper implements ScalarFunctionMapper {
+  private static final String CONCAT_FUNCTION_NAME = "concat";
+  private final List<ScalarFunctionVariant> concatFunctions;
+
+  ConcatFunctionMapper(List<ScalarFunctionVariant> functions) {
+    this.concatFunctions =
+        functions.stream()
+            .filter(
+                f ->
+                    CONCAT_FUNCTION_NAME.equals(f.name())
+                        && DefaultExtensionCatalog.FUNCTIONS_STRING.equals(f.urn()))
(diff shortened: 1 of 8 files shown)


isthmus/src/test/java/io/substrait/isthmus/DuplicateFunctionUrnTest.java
```diff
@@ -15,13 +16,22 @@
...
+import org.apache.calcite.sql.type.ReturnTypes;
 import org.junit.jupiter.api.Test;
 
 /** Tests to reproduce #562 */
 class DuplicateFunctionUrnTest extends PlanTestBase {
 
+  /** Test-only operator that maps to the {@code concatenate} function in the test extensions. */
+  static final SqlFunction TEST_CONCATENATE =
+      SqlBasicFunction.create(
+          "TEST_CONCATENATE", ReturnTypes.ARG0_NULLABLE, OperandTypes.STRING_SAME_SAME);
```
vbarua, line 33: I modified how we handled the concat function, which made these tests brake. I opted to introduce an all new test-only mapping to verify this to avoid future breakages like this.

cc: @benbellick

isthmus/src/main/java/io/substrait/isthmus/expression/ConcatFunctionMapper.java
```diff
@@ -0,0 +1,47 @@
...
+  private static final String CONCAT_FUNCTION_NAME = "concat";
+  private final List<ScalarFunctionVariant> concatFunctions;
+
+  ConcatFunctionMapper(List<ScalarFunctionVariant> functions) {
+    this.concatFunctions =
+        functions.stream()
+            .filter(
+                f ->
+                    CONCAT_FUNCTION_NAME.equals(f.name())
+                        && DefaultExtensionCatalog.FUNCTIONS_STRING.equals(f.urn()))
```
nielspardon, line 29: 🤖 _AI-generated_

Filtering on `FUNCTIONS_STRING` narrows `||` to only the standard concat. Previously `SqlStdOperatorTable.CONCAT` sat in `SCALAR_SIGS` and matched any `concat` by name regardless of URN, so a converter built with a custom-URN concat (and no default extensions) could still convert `||`; it now can't. Assuming this is intended — might be worth a release-note line since it's a subtle semantics change.

isthmus/src/test/java/io/substrait/isthmus/DuplicateFunctionUrnTest.java
```diff
@@ -37,7 +47,9 @@ class DuplicateFunctionUrnTest extends PlanTestBase {
       // Verify that the merged collection contains duplicate concat functions with different URNs
```
nielspardon: 🤖 _AI-generated_

Nit: stale after the rename — the comment still says "duplicate **concat** functions" and the local is still named `concatFunctions`, but this now filters `"concatenate"`.

nielspardon (approved):
LGTM and my AI friend found two nits

👨 human generated

isthmus/src/main/java/io/substrait/isthmus/expression/ConcatFunctionMapper.java
vbarua: This is intentional. We should avoid relying on mappings solely based on names. This is somewhat of an issue today because we don't include URNs in the function mappings.

The `concat` function in `extension:io.substrait:functions_string` is what we are mapping specifically to CONCAT_FUNCTION. A function named `concat` in another extension, if we want to map it, should have it's own explicit mapping.

PR #1033, opened 2026-07-23
feat(core)!: bump substrait to v0.98.0

### LINKED ISSUE #1018: Update to Substrait v0.98.0
# Release Notes

### ⚠ BREAKING CHANGES

* **protos:** The deprecated `left_keys` and `right_keys` fields have
been removed from `HashJoinRel` and `MergeJoinRel`. Producers must use
the `keys` field instead.
* **protos:** The deprecated `offset` and `count` fields have been
removed from `FetchRel`. Producers must use `offset_expr` and
`count_expr` instead.
* **protos:** remove deprecated VirtualTable.values field (#1131)
* **protos:** remove deprecated IntervalDayToSecond.microseconds field (#1116)

### Features

* **protos:** remove deprecated IntervalDayToSecond.microseconds field ([#1116](https://github.com/substrait-io/substrait/issues/1116)) ([a4ad9f4](https://github.com/substrait-io/substrait/commit/a4ad9f4208e827be845ecf3e307f58277e2cfc0f))
* **protos:** remove deprecated left_keys/right_keys from join rels ([#1129](https://github.com/substrait-io/substrait/issues/1129)) ([08b2736](https://github.com/substrait-io/substrait/commit/08b27364728822bf00bf13a82998f09f1c4b96d7))
* **protos:** remove deprecated offset/count from FetchRel ([#1130](https://github.com/substrait-io/substrait/issues/1130)) ([b7e8d02](https://github.com/substrait-io/substrait/commit/b7e8d027e01e21656ab498d23db4335098f50745))
* **protos:** remove deprecated VirtualTable.values field ([#1131](https://github.com/substrait-io/substrait/issues/1131)) ([7b0bd66](https://github.com/substrait-io/substrait/commit/7b0bd664b5e09bf6a230742b17511716040a9d21))

### Bug Fixes

* **tests:** correct decimal sum and nullif results ([#1048](https://github.com/substrait-io/substrait/issues/1048)) ([a750c25](https://github.com/substrait-io/substrait/commit/a750c25c3bd3c9e82fb10f1c1bbf3d0a273a0f46))

See https://github.com/substrait-io/substrait/releases/tag/v0.98.0

nielspardon:
Bump the `substrait` submodule from v0.97.0 to v0.98.0, which removes several long-deprecated proto fields.

## `FetchRel` → expression-based offset/count (breaking)
v0.98.0 removed `FetchRel`'s int64 `offset`/`count` in favor of the `offset_expr`/`count_expr` expression fields. `Fetch` now models these as `Optional<Expression> getOffset()`/`getCount()`, matching the existing `TopN`. Updated across both proto converters, `RelCopyOnWriteVisitor`, the DSL, and the **isthmus** and **spark** integrations (read + write). An unset offset is treated as `0`; an unset count signals `LIMIT ALL`.

**DSL** — `SubstraitBuilder` keeps the `long`-based `fetch`/`limit`/`offset` helpers (they wrap `i64` literals; offset `0` is left unset) and adds lambda `Function<Rel, Expression>` overloads that build the offset/count expression from the input relation (matching `filter`/`nestedLoopJoin`).

**Isthmus** — now passes offset/count **expressions** through to Calcite as `RexNode`s (via `sortLimit`) and back (via `toExpression`), instead of extracting integer literals. Non-literal offset/count (e.g. a dynamic-parameter `LIMIT ?`) now round-trips (covered by `FetchTest.limitWithDynamicParameterCount`).

## Other removed fields
- `HashJoinRel`/`MergeJoinRel` `left_keys`/`right_keys` — dropped the write, the read fallback, and the lossy-key helper (use `keys`).
- `ReadRel.VirtualTable.values` — dropped the value-conversion path (use `expressions`).
- `Expression.Literal.IntervalDayToSecond` `microseconds` and the `precision_mode` oneof — `precision` is now a plain, always-present field.

## IntervalDay type precision (spec v0.97.0)
The `IntervalDay` **type**'s `precision` is the only `optional` precision field, and v0.97.0 changed its guidance to "implementations should reject an unset precision." `ProtoTypeConverter` now rejects an `IntervalDay` type with an unset precision instead of defaulting to `6` (substrait-java always *emits* a precision, so this only affects reading externally-produced plans).

**BREAKING CHANGE:** `Fetch.getOffset()`/`getCount()` now return `Optional<Expression>` instead of `long`/`OptionalLong`. The generated protobuf API also drops `FetchRel.offset`/`count`, `HashJoinRel`/`MergeJoinRel.left_keys`/`right_keys`, `ReadRel.VirtualTable.values`, and `Expression.Literal.IntervalDayToSecond.microseconds`. Reading an `IntervalDay` type with an unset precision now throws instead of defaulting to 6.

## Verification
`./gradlew :core:check :isthmus:check :core:javadoc spotlessCheck`, both Spark Scala variants (`spark-3.5_2.12`, `spark-4.0_2.13`), the `:spark:spark-3.5_2.12:test` suite (TPC-H/DS limit queries round-trip), and `examples:substrait-spark` all pass. Adds `FetchRoundtripTest`, `FetchTest.limitWithDynamicParameterCount`, and `TestTypeRoundtrip.intervalDayWithoutPrecisionIsRejected`.

Closes #1018

🤖 Generated with AI

### DIFF
--- a/core/src/main/java/io/substrait/relation/ProtoRelConverter.java
+++ b/core/src/main/java/io/substrait/relation/ProtoRelConverter.java
@@ -906,11 +906,15 @@ public class ProtoRelConverter {
    */
   protected Fetch newFetch(FetchRel rel) {
     Rel input = from(rel.getInput());
-    ImmutableFetch.Builder builder = Fetch.builder().input(input).offset(rel.getOffset());
-    if (rel.getCount() != -1) {
-      // -1 is used as a sentinel value to signal LIMIT ALL
-      // count only needs to be set when it is not -1
-      builder.count(rel.getCount());
+    ProtoExpressionConverter converter =
+        new ProtoExpressionConverter(lookup, extensions, input.getRecordType(), this);
+    ImmutableFetch.Builder builder = Fetch.builder().input(input);
+    if (rel.hasOffsetExpr()) {
+      builder.offset(converter.from(rel.getOffsetExpr()));
+    }
+    if (rel.hasCountExpr()) {
+      // An unset count signals LIMIT ALL.
+      builder.count(converter.from(rel.getCountExpr()));
     }
 
     builder
--- a/core/src/main/java/io/substrait/relation/ProtoRelConverter.java
+++ b/core/src/main/java/io/substrait/relation/ProtoRelConverter.java
@@ -71,7 +71,6 @@ import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.stream.Collectors;
-import java.util.stream.IntStream;
 import org.jspecify.annotations.NonNull;
 
 /** Converts from {@link io.substrait.proto.Rel} to {@link io.substrait.relation.Rel} */
@@ -826,28 +825,6 @@ public class ProtoRelConverter {
   }
 
   /**
-   * Converts StructLiteral instances to NestedStruct for VirtualTableScan. This is a convenience
-   * method for migrating from the legacy StructLiteral-based VirtualTable API to the new
-   * NestedStruct-based API.
-   *
-   * @param nullable whether the resulting NestedStruct instances should be nullable
-   * @param structs the StructLiteral instances to convert
-   * @return a list of NestedStruct instances with the same field structure
-   */
-  private static List<Expression.NestedStruct> nestedStruct(
-      boolean nullable, Expression.StructLiteral... structs) {
-    List<Expression.NestedStruct> nestedStructs = new ArrayList<>();
-    for (Expression.StructLiteral struct : structs) {
-      nestedStructs.add(
-          Expression.NestedStruct.builder()
-              .nullable(nullable)
-              .addAllFields(struct.fields())
-              .build());
-    }
-    return nestedStructs;
-  }
-
-  /**
    * Converts the corresponding protobuf message to its POJO representation.
    *
    * @param rel the protobuf value to convert
@@ -855,24 +832,11 @@ public class ProtoRelConverter {
    */
   protected VirtualTableScan newVirtualTable(ReadRel rel) {
     ReadRel.VirtualTable virtualTable = rel.getVirtualTable();
-    // If both values and expressions are set, raise an error
-    if (virtualTable.getValuesCount() > 0 && virtualTable.getExpressionsCount() > 0) {
-      throw new IllegalArgumentException(
-          "VirtualTable cannot have both values and expressions set");
-    }
     NamedStruct virtualTableSchema = newNamedStruct(rel);
     ProtoExpressionConverter converter =
         new ProtoExpressionConverter(lookup, extensions, virtualTableSchema.struct(), this);
 
-    List<Expression.NestedStruct> expressions =
-        new ArrayList<>(virtualTable.getValuesCount() + virtualTable.getExpressionsCount());
-
-    // We cannot have a null row in VirtualTable, therefore we set the nullability to false
-    // nullability is also not supported at the Expression.Nested.Struct level
-    for (io.substrait.proto.Expression.Literal.Struct struct : virtualTable.getValuesList()) {
-      expressions.addAll(nestedStruct(false, converter.from(struct)));
-    }
-
+    List<Expression.NestedStruct> expressions = new ArrayList<>(virtualTable.getExpressionsCount());
     for (io.substrait.proto.Expression.Nested.Struct expr : virtualTable.getExpressionsList()) {
       expressions.add(converter.from(expr));
     }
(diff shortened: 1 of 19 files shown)


spark/src/main/scala/io/substrait/spark/logical/ToLogicalPlan.scala
```diff
@@ -258,8 +258,9 @@ class ToLogicalPlan(val spark: AnyRef = SparkCompat.instance.getOrCreateSparkSes
 
   override def visit(fetch: relation.Fetch, context: EmptyVisitationContext): LogicalPlan = {
     val child = fetch.getInput.accept(this, context)
-    val limit = fetch.getCount.orElse(-1).intValue() // -1 means unassigned here
-    val offset = fetch.getOffset.intValue()
+    // Offset/count are expressions; an unset count means LIMIT ALL (-1) and an unset offset means 0.
+    val limit = if (fetch.getCount.isPresent) asLong(fetch.getCount.get).toInt else -1
+    val offset = if (fetch.getOffset.isPresent) asLong(fetch.getOffset.get).toInt else 0
```
andrew-coleman: This would be slightly more idiomatic Scala:

```suggestion
   val limit = fetch.getCount.fold(-1)(asInt)
   val offset = fetch.getOffset.fold(0)(asInt)
```

spark/src/main/scala/io/substrait/spark/logical/ToLogicalPlan.scala
```diff
@@ -276,6 +277,13 @@ class ToLogicalPlan(val spark: AnyRef = SparkCompat.instance.getOrCreateSparkSes
     remap(plan, fetch.getRemap)
   }
 
+  private def asLong(e: SExpression): Long = e match {
+    case l: SExpression.I64Literal => l.value()
+    case l: SExpression.I32Literal => l.value().toLong
+    case other =>
+      throw new UnsupportedOperationException(s"Unsupported fetch offset/count expression: $other")
+  }
+
```
andrew-coleman: Saves casting it toInt above:

```suggestion
private def asInt(e: SExpression): Int = e match {
  case l: SExpression.I64Literal => l.value().toInt
  case l: SExpression.I32Literal => l.value().toInt
  case other =>
    throw new UnsupportedOperationException(s"Unsupported fetch offset/count expression: $other")
}

```

nielspardon: Good call — done in `0a5a797`. `asInt` now returns `Int` directly, so the `.toInt` casts drop off both call sites.

spark/src/main/scala/io/substrait/spark/logical/ToLogicalPlan.scala
nielspardon: For now I kept `if/else` rather than `.fold`, but wanted to check your preference. `getCount`/`getOffset` are Java `Optional`s (no `.fold`), so the `.fold` form needs `scala.jdk.OptionConverters` + `.toScala` — that compiles on 2.13 but not on 2.12, since only `scala.jdk.CollectionConverters` was backported to the 2.12 stdlib. Getting `.fold` to work across both variants would mean adding `scala-collection-compat` as an explicit dependency to the spark build.

Happy to do that if you'd rather have the more idiomatic form — do you want the extra dep, or is the `if/else` fine here?

andrew-coleman: I'm happy to leave it as it is.  No need to add an extra dependency just for this.

andrew-coleman (approved):

core/src/main/java/io/substrait/relation/ProtoRelConverter.java
```diff
@@ -906,11 +870,15 @@ protected VirtualTableScan newVirtualTable(ReadRel rel) {
...
+    ProtoExpressionConverter converter =
+        new ProtoExpressionConverter(lookup, extensions, input.getRecordType(), this);
+    ImmutableFetch.Builder builder = Fetch.builder().input(input);
+    if (rel.hasOffsetExpr()) {
+      builder.offset(converter.from(rel.getOffsetExpr()));
+    }
+    if (rel.hasCountExpr()) {
+      // An unset count signals LIMIT ALL.
+      builder.count(converter.from(rel.getCountExpr()));
     }
```
vbarua, line 882: Heads up here for the future. It looks like we never shipped a version of this that could process _both_ the old int64 based fields and the new expression based fields. We should have done that before bumping Substrait like this to make it possible for producers to migrate to the new expression based field.

Because we've fully removed support for the old fields, anyone who want to migrate will need to either:
1. Upgrade substrait-java in lockstep with their producers.
2. Upgrade substrait-java to v0.97 and implement a custom ProtoRelConsumer subclass the overrides `newFetch` to handle both fields.

PR #1035, opened 2026-07-24
feat(isthmus)!: inject ConverterProvider into the SQL statement parsers

nielspardon:
## Summary

Thread a single `ConverterProvider` through **both** the CREATE-statement parsing that builds the
Calcite catalog and the query parsing, so one parser configuration (the provider's
`getSqlParserConfig()`) is applied consistently end-to-end. Previously the CREATE-statement path and
the query path could apply different parser settings, so the identifier casing used to build a
`NamedScan` name could disagree with the casing used to resolve it in a query.

Also adds a shared `ConverterProvider.DEFAULT` instance, used as the default for the no-arg parsing
entry points.

## Breaking changes

The `SqlParser.Config`-based parsing entry points (public since `v0.94.0`) are **removed** in favour
of passing a `ConverterProvider`:

- `SubstraitSqlStatementParser.parseStatements(String, SqlParser.Config)`
- `SubstraitSqlToCalcite.convertQueries(String, CatalogReader, SqlParser.Config)` and the
  5-arg `SqlParser.Config` overload

`SqlToSubstrait.convert(String, CatalogReader, SqlDialect)` is **deprecated**: the `SqlDialect`
argument no longer influences parsing and it delegates to `convert(String, CatalogReader)`. Until the
builder arrives (next PR in the stack), parser configuration is customised by subclassing
`ConverterProvider` and overriding `getSqlParserConfig()`.

## Notable changes

| Class | Change |
|-------|--------|
| `SubstraitSqlStatementParser` | `parseStatements(String, ConverterProvider)` owns the `SqlParser` instantiation; `SqlParser.Config` overload removed |
| `SubstraitSqlToCalcite` | New `convertQueries(sql, catalog, ConverterProvider[, operatorTable])`; internal overloads route through `ConverterProvider`; `SqlParser.Config` overloads removed |
| `SubstraitCreateStatementParser` | New `processCreateStatements(ConverterProvider, …)` / `processCreateStatementsToCatalog(ConverterProvider, …)` overloads — same parser config as query parsing |
| `SqlToSubstrait` | `convert(sql, catalog)` uses the `ConverterProvider` path; `convert(sql, catalog, SqlDialect)` `@Deprecated` (arg ignored) |
| `SqlExpressionToSubstrait` | Routes CREATE parsing through the provider |
| `ConverterProvider` | Adds the shared `DEFAULT` instance |
| `IsthmusEntryPoint` (CLI) | Threads its provider through catalog and query parsing consistently |

## Split of #983

Per @vbarua's request, #983 is split into three stacked PRs (review/merge bottom-up):

1. **#1035 — inject `ConverterProvider` into the SQL statement parsers (this PR)**
2. #1036 — introduce the `ConverterProvider` builder

Replaces #983.

🤖 Generated with AI

### DIFF
--- a/isthmus/src/main/java/io/substrait/isthmus/SqlToSubstrait.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/SqlToSubstrait.java
@@ -5,11 +5,9 @@ import io.substrait.plan.ImmutablePlan.Builder;
 import io.substrait.plan.Plan;
 import io.substrait.plan.Plan.Version;
 import org.apache.calcite.prepare.Prepare;
-import org.apache.calcite.server.ServerDdlExecutor;
 import org.apache.calcite.sql.SqlDialect;
 import org.apache.calcite.sql.SqlOperatorTable;
 import org.apache.calcite.sql.parser.SqlParseException;
-import org.apache.calcite.sql.parser.SqlParser;
 
 /**
  * Take a SQL statement and a set of table definitions and return a substrait plan.
@@ -21,7 +19,7 @@ public class SqlToSubstrait extends SqlConverterBase {
 
   /** Creates a SQL-to-Substrait converter using the default configuration. */
   public SqlToSubstrait() {
-    this(new ConverterProvider());
+    this(ConverterProvider.DEFAULT);
   }
 
   /**
@@ -50,7 +48,9 @@ public class SqlToSubstrait extends SqlConverterBase {
     builder.executionBehavior(converterProvider.getExecutionBehavior());
 
     // TODO: consider case in which one sql passes conversion while others don't
-    SubstraitSqlToCalcite.convertQueries(sqlStatements, catalogReader, operatorTable).stream()
+    SubstraitSqlToCalcite.convertQueries(
+            sqlStatements, catalogReader, converterProvider, operatorTable)
+        .stream()
         .map(root -> SubstraitRelVisitor.convert(root, converterProvider))
         .forEach(root -> builder.addRoots(root));
 
@@ -60,31 +60,28 @@ public class SqlToSubstrait extends SqlConverterBase {
   /**
    * Converts one or more SQL statements into a Substrait {@link Plan}.
    *
+   * <p>The {@code sqlDialect} parameter was previously used to influence identifier casing during
+   * parsing. This is now controlled by the {@link ConverterProvider} supplied to this converter; to
+   * customise it, subclass {@link ConverterProvider} and override {@link
+   * ConverterProvider#getSqlParserConfig()}.
+   *
    * @param sqlStatements a string containing one more SQL statements
    * @param catalogReader the {@link Prepare.CatalogReader} for finding tables/views referenced in
    *     the SQL statements
    * @param sqlDialect The sql dialect to use for parsing.
    * @return the Substrait {@link Plan}
    * @throws SqlParseException if there is an error while parsing the SQL statements
+   * @deprecated Prefer constructing {@link SqlToSubstrait} with a {@link ConverterProvider}
+   *     configured for the desired casing and calling {@link #convert(String,
+   *     Prepare.CatalogReader)}. For fully custom parser behaviour, subclass {@link
+   *     ConverterProvider} and override {@link ConverterProvider#getSqlParserConfig()}.
    */
+  @Deprecated
   public Plan convert(
       final String sqlStatements,
       final Prepare.CatalogReader catalogReader,
       final SqlDialect sqlDialect)
       throws SqlParseException {
-    Builder builder = io.substrait.plan.Plan.builder();
-    builder.version(Version.builder().from(Version.DEFAULT_VERSION).producer("isthmus").build());
-    builder.executionBehavior(converterProvider.getExecutionBehavior());
-
-    final SqlParser.Config sqlParserConfig =
-        sqlDialect.configureParser(
-            SqlParser.config().withParserFactory(ServerDdlExecutor.PARSER_FACTORY));
-
-    // TODO: consider case in which one sql passes conversion while others don't
-    SubstraitSqlToCalcite.convertQueries(sqlStatements, catalogReader, sqlParserConfig).stream()
-        .map(root -> SubstraitRelVisitor.convert(root, converterProvider))
-        .forEach(root -> builder.addRoots(root));
-
-    return builder.build();
+    return convert(sqlStatements, catalogReader);
   }
 }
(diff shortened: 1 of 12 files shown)


isthmus/src/main/java/io/substrait/isthmus/sql/SubstraitSqlToCalcite.java
```diff
@@ -111,39 +122,48 @@ public static List<RelRoot> convertQueries(
...
   public static List<RelRoot> convertQueries(
-      String sqlStatements, Prepare.CatalogReader catalogReader) throws SqlParseException {
-    SqlValidator validator = new SubstraitSqlValidator(catalogReader);
-    return convertQueries(sqlStatements, catalogReader, validator, createDefaultRelOptCluster());
+      String sqlStatements,
+      Prepare.CatalogReader catalogReader,
+      ConverterProvider converterProvider,
+      SqlOperatorTable operatorTable)
+      throws SqlParseException {
+    SqlValidator validator = new SubstraitSqlValidator(catalogReader, operatorTable);
```
vbarua: Now that we're injecting the ConverterProvider, should we use the operatorTable available in ConverterProvider#getSqlOperator here?

isthmus/src/main/java/io/substrait/isthmus/SqlToSubstrait.java
```diff
@@ -60,31 +60,28 @@ public Plan convert(final String sqlStatements, final Prepare.CatalogReader cata
...
-        sqlDialect.configureParser(
-            SqlParser.config().withParserFactory(ServerDdlExecutor.PARSER_FACTORY));
-
-    // TODO: consider case in which one sql passes conversion while others don't
-    SubstraitSqlToCalcite.convertQueries(sqlStatements, catalogReader, sqlParserConfig).stream()
-        .map(root -> SubstraitRelVisitor.convert(root, converterProvider))
-        .forEach(root -> builder.addRoots(root));
-
-    return builder.build();
+    return convert(sqlStatements, catalogReader);
```
vbarua, line 80: Nice removal of duplication here 🧹

isthmus/src/main/java/io/substrait/isthmus/sql/SubstraitCreateStatementParser.java
```diff
@@ -133,19 +159,58 @@ private static SqlParseException fail(@Nullable final String message) {
...
-   * @param createStatements one or more SQL strings containing only CREATE statements; must not be
-   *     null
-   * @return a {@link CalciteSchema} generated from the CREATE statements
+   * @param converterProvider the converter provider whose parser config controls identifier casing
+   *     and other parser settings
+   * @param createStatements a SQL string containing only CREATE statements; must not be null
+   * @return list of {@link SubstraitTable}s generated from the CREATE statements
    * @throws SqlParseException if parsing fails or statements are invalid
    */
+  public static List<SubstraitTable> processCreateStatements(
```
vbarua: minor: would it make sense to move this  under the original `processCreateStatement` from which it was extracted? Would also have made the diff easier to review.

isthmus/src/main/java/io/substrait/isthmus/sql/SubstraitCreateStatementParser.java
```diff
@@ -101,7 +101,33 @@ public static CalciteCatalogReader processCreateStatementsToCatalog(
...
       @NonNull final String... createStatements) throws SqlParseException {
-    final CalciteSchema rootSchema = processCreateStatementsToSchema(createStatements);
+    final CalciteSchema rootSchema =
+        processCreateStatementsToSchema(ConverterProvider.DEFAULT, createStatements);
+    final List<String> defaultSchema = Collections.emptyList();
+    return new CalciteCatalogReader(
+        rootSchema,
+        defaultSchema,
+        SubstraitTypeSystem.TYPE_FACTORY,
+        SqlConverterBase.CONNECTION_CONFIG);
```
vbarua, line 152: (not for this PR) I wonder if it would make sense to get the RelTypeFactory and SqlConverterBase from the ConverterProvider as well. The former is already configured in the ConverterProvider.

isthmus/src/main/java/io/substrait/isthmus/sql/SubstraitSqlToCalcite.java
nielspardon: good catch. pushed a commit that refactors this part

isthmus/src/main/java/io/substrait/isthmus/sql/SubstraitCreateStatementParser.java
nielspardon: pushed a commit that moves the method as suggested

isthmus/src/main/java/io/substrait/isthmus/sql/SubstraitCreateStatementParser.java
nielspardon: can probably include that in the builder PR

vbarua (approved):
LGTM

I'll leave the merging to you. The PR message is a bit of mouthful for a commit message, especially because we'll try and generate the release notes from it as well.

isthmus/src/main/java/io/substrait/isthmus/sql/SubstraitCreateStatementParser.java
nielspardon: Revisiting this now that the builder PR (#1036) has merged — it landed about half of this. `processCreateStatementsToCatalog(converterProvider, ...)` now builds the `CalciteCatalogReader` with `converterProvider.getTypeFactory()` and `converterProvider.getCalciteConnectionConfig()`.

What it did not cover is the type derivation underneath, on that same injected path: `createSubstraitTable(...)` still builds the row type with `SubstraitTypeSystem.TYPE_FACTORY.createStructType(...)`, and the column types come from `col.dataType.deriveType(VALIDATOR)` where `VALIDATOR`/`EMPTY_CATALOG` are `public static final` and built on the global factory plus `SqlConverterBase.CONNECTION_CONFIG`. So a builder-configured `typeFactory` yields a catalog reader that uses it while the tables inside are typed by the global one — harmless today (the provider getter returns the same `CASE_SENSITIVE=false` config), but a custom type factory carrying a different type system would be silently bypassed.

Finishing it off in a focused follow-up PR rather than widening #1037, since making those two constants provider-derived is a public-API-shaped change worth its own review.

nielspardon: Opened #1050 for this.

PR #1036, opened 2026-07-24
feat(isthmus): introduce a ConverterProvider builder

nielspardon:
## Summary

Add `ConverterProvider.builder()` and a nested `Builder` — a readable, forward-compatible way to
configure a `ConverterProvider`, most importantly to supply a full Calcite `SqlParser.Config` now
that the parsers consume the provider's config (#1035). `ConverterProvider.DEFAULT_SQL_PARSER_CONFIG`
(`SqlParser.Config.DEFAULT` + `TO_UPPER` + `SqlDdlParserImpl.FACTORY` + `LENIENT`) is exposed as the
base for customized configs, e.g. `DEFAULT_SQL_PARSER_CONFIG.withUnquotedCasing(Casing.UNCHANGED)`.

## Sole construction path + subclassing seam

All public `ConverterProvider` constructors are `@Deprecated` in favour of `builder()`, and `DEFAULT`
is `builder().build()`. Subclassing routes through a new `protected ConverterProvider(Builder)` seam
— the single place instances are created — so adding a component later touches only the builder and
this constructor, never a subclass `super(...)` call. `DynamicConverterProvider` and
`AutomaticDynamicFunctionMappingConverterProvider` construct via `super(builder)`.

Now that the provider is fully configurable, call sites that still used global defaults are routed
through it: the builder's `typeConverter` reaches the scalar/aggregate/window converters, and the
CREATE-statement catalog reader uses the provider's type factory and connection config.

## Compatibility

Source-compatible — no public API removed, deprecated constructors still delegate to the builder,
and behavior is unchanged for the default provider.

## Stack (#983 split, merge bottom-up)

1. #1035 — inject `ConverterProvider` into the SQL statement parsers (merged)
2. **#1036 — this PR**

### DIFF
--- a/isthmus/src/main/java/io/substrait/isthmus/AutomaticDynamicFunctionMappingConverterProvider.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/AutomaticDynamicFunctionMappingConverterProvider.java
@@ -47,9 +47,14 @@ public class AutomaticDynamicFunctionMappingConverterProvider extends ConverterP
    *
    * <p>Uses {@link DefaultExtensionCatalog#DEFAULT_COLLECTION} for extensions and {@link
    * SubstraitTypeSystem#TYPE_FACTORY} for type operations.
+   *
+   * @deprecated Use {@link
+   *     #AutomaticDynamicFunctionMappingConverterProvider(ConverterProvider.Builder)} with {@link
+   *     ConverterProvider#builder()} instead.
    */
+  @Deprecated
   public AutomaticDynamicFunctionMappingConverterProvider() {
-    this(DefaultExtensionCatalog.DEFAULT_COLLECTION, SubstraitTypeSystem.TYPE_FACTORY);
+    this(ConverterProvider.builder());
   }
 
   /**
@@ -58,10 +63,15 @@ public class AutomaticDynamicFunctionMappingConverterProvider extends ConverterP
    * <p>Uses {@link SubstraitTypeSystem#TYPE_FACTORY} for type operations.
    *
    * @param extensions the extension collection containing function definitions
+   * @deprecated Use {@link
+   *     #AutomaticDynamicFunctionMappingConverterProvider(ConverterProvider.Builder)} instead, e.g.
+   *     {@code new AutomaticDynamicFunctionMappingConverterProvider(
+   *     ConverterProvider.builder().extensions(extensions))}.
    */
+  @Deprecated
   public AutomaticDynamicFunctionMappingConverterProvider(
       SimpleExtension.ExtensionCollection extensions) {
-    this(extensions, SubstraitTypeSystem.TYPE_FACTORY);
+    this(ConverterProvider.builder().extensions(extensions));
   }
 
   /**
@@ -72,10 +82,27 @@ public class AutomaticDynamicFunctionMappingConverterProvider extends ConverterP
    *
    * @param extensions the extension collection containing function definitions
    * @param typeFactory the type factory for creating and managing Calcite data types
+   * @deprecated Use {@link
+   *     #AutomaticDynamicFunctionMappingConverterProvider(ConverterProvider.Builder)} instead, e.g.
+   *     {@code new AutomaticDynamicFunctionMappingConverterProvider(
+   *     ConverterProvider.builder().extensions(extensions).typeFactory(typeFactory))}.
    */
+  @Deprecated
   public AutomaticDynamicFunctionMappingConverterProvider(
       SimpleExtension.ExtensionCollection extensions, RelDataTypeFactory typeFactory) {
-    super(extensions, typeFactory);
+    this(ConverterProvider.builder().extensions(extensions).typeFactory(typeFactory));
+  }
+
+  /**
+   * Creates a new provider from a {@link ConverterProvider.Builder}, seeding base state from the
+   * builder and then installing the automatically generated dynamic function mappings and operator
+   * table. This is the seam through which the other constructors — and subclasses, via {@code
+   * super(builder)} — route.
+   *
+   * @param builder the builder carrying the configured components
+   */
+  public AutomaticDynamicFunctionMappingConverterProvider(ConverterProvider.Builder builder) {
+    super(builder);
 
     List<SqlOperator> dynamicScalarOperators = getDynamicScalarOperators();
     this.scalarFunctionConverter = createScalarFunctionConverter(dynamicScalarOperators);
--- a/isthmus/src/main/java/io/substrait/isthmus/AutomaticDynamicFunctionMappingConverterProvider.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/AutomaticDynamicFunctionMappingConverterProvider.java
@@ -11,6 +11,7 @@ import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
+import java.util.Optional;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 import org.apache.calcite.rel.type.RelDataTypeFactory;
@@ -96,13 +97,14 @@ public class AutomaticDynamicFunctionMappingConverterProvider extends ConverterP
   /**
    * Creates a new provider from a {@link ConverterProvider.Builder}, seeding base state from the
    * builder and then installing the automatically generated dynamic function mappings and operator
-   * table. This is the seam through which the other constructors — and subclasses, via {@code
-   * super(builder)} — route.
+   * table.
    *
    * @param builder the builder carrying the configured components
+   * @throws IllegalArgumentException if the builder configures a scalar, aggregate or window
+   *     function converter, all of which this provider derives itself
    */
   public AutomaticDynamicFunctionMappingConverterProvider(ConverterProvider.Builder builder) {
-    super(builder);
+    super(requireDerivedFunctionConverters(builder));
 
     List<SqlOperator> dynamicScalarOperators = getDynamicScalarOperators();
     this.scalarFunctionConverter = createScalarFunctionConverter(dynamicScalarOperators);
@@ -118,6 +120,32 @@ public class AutomaticDynamicFunctionMappingConverterProvider extends ConverterP
             .flatMap(List::stream)
             .collect(Collectors.toList());
     this.operatorTable = buildOperatorTable(allOperators);
+  }
+
+  /**
+   * Rejects a builder that configures any function converter: this provider derives all three from
+   * the unmapped extension functions, so a configured one would be silently discarded.
+   *
+   * @param builder the builder to check
+   * @return the given builder
+   * @throws IllegalArgumentException if the builder configures a function converter
+   */
+  private static ConverterProvider.Builder requireDerivedFunctionConverters(
+      ConverterProvider.Builder builder) {
+    requireUnset(builder.getScalarFunctionConverter(), "scalarFunctionConverter");
+    requireUnset(builder.getAggregateFunctionConverter(), "aggregateFunctionConverter");
+    requireUnset(builder.getWindowFunctionConverter(), "windowFunctionConverter");
+    return builder;
+  }
+
+  private static void requireUnset(Optional<?> functionConverter, String setterName) {
+    if (functionConverter.isPresent()) {
+      throw new IllegalArgumentException(
+          "AutomaticDynamicFunctionMappingConverterProvider derives its own function converters "
+              + "from the unmapped extension functions; remove the ConverterProvider.Builder."
+              + setterName
+              + "(...) setting");
+    }
   }
 
   /**
--- a/isthmus/src/main/java/io/substrait/isthmus/ConverterProvider.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/ConverterProvider.java
@@ -40,20 +40,51 @@ import org.apache.calcite.tools.RelBuilder;
  *
  * <p>It is consumed by all conversion classes as their primary source of configuration.
  *
- * <p>The no argument constructor {@link #ConverterProvider()} provides reasonable system defaults.
+ * <p>{@link #DEFAULT} is a shared instance configured with reasonable system defaults, equivalent
+ * to {@code builder().build()}.
  *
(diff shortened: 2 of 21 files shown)


isthmus/src/main/java/io/substrait/isthmus/ConverterProvider.java
```diff
@@ -40,21 +40,52 @@
...
+ * <p>For customized conversion behaviour — including supplying a full Calcite {@link
+ * SqlParser.Config} for SQL parsing — use the {@link #builder()}.
  *
  * <p>More in-depth customization can be achieved by extending this class, as is done in {@link
  * DynamicConverterProvider}.
  */
 public class ConverterProvider {
 
+  /**
+   * The default Calcite {@link SqlParser.Config} used by isthmus: {@link SqlParser.Config#DEFAULT}
```
vbarua: super minor
```suggestion
   * The default Calcite {@link SqlParser.Config} used by Isthmus: {@link SqlParser.Config#DEFAULT}
```

isthmus/src/main/java/io/substrait/isthmus/ConverterProvider.java
```diff
@@ -152,13 +190,59 @@ public ConverterProvider(
...
+
+  /**
+   * Master constructor and the sole subclassing seam: derives any unset function converters from
+   * the configured extensions and type factory, then assigns all components. {@link
+   * Builder#build()} and every delegating public constructor route here, and subclasses (e.g.
+   * {@link DynamicConverterProvider}) invoke it via {@code super(builder)}.
+   *
+   * <p>Taking the {@link Builder} rather than a positional argument list keeps this seam stable as
+   * new components are added: a new field is a change to the builder and this constructor only, not
+   * to every subclass's {@code super(...)} call.
```
vbarua: minor/meta: This doc comment is explaining what a Builder is and why we're using it. The PR itself and the associated merge is where we're capturing the why, the code can just capture the current state. IMO this could be reduced to something like:


> Primary constructor for ConverterProvider. The ConverterProvider.Builder passed in can be used to further customize behaviours.

isthmus/src/main/java/io/substrait/isthmus/ConverterProvider.java
```diff
@@ -424,12 +509,150 @@ public TypeConverter getTypeConverter() {
...
+    public Builder executionBehavior(Plan.ExecutionBehavior executionBehavior) {
+      this.executionBehavior = executionBehavior;
+      return this;
+    }
+
+    /**
+     * Sets the full Calcite {@link SqlParser.Config} used for SQL parsing, replacing the default.
+     *
+     * <p>Use {@link ConverterProvider#DEFAULT_SQL_PARSER_CONFIG} as a starting point to retain
+     * isthmus' DDL parser factory and conformance while overriding individual settings.
```
vbarua:
```suggestion
     * Isthmus' DDL parser factory and conformance while overriding individual settings.
```

isthmus/src/main/java/io/substrait/isthmus/DynamicConverterProvider.java
```diff
@@ -67,10 +74,25 @@ public DynamicConverterProvider(SimpleExtension.ExtensionCollection extensions)
...
   public DynamicConverterProvider(
       SimpleExtension.ExtensionCollection extensions, RelDataTypeFactory typeFactory) {
-    super(extensions, typeFactory);
+    this(ConverterProvider.builder().extensions(extensions).typeFactory(typeFactory));
+  }
+
+  /**
+   * Creates a new DynamicConverterProvider from a {@link ConverterProvider.Builder}, seeding base
+   * state from the builder and then installing the converter that handles dynamic extension
+   * functions. This is the seam through which subclasses route via {@code super(builder)}.
```
vbarua: minor: I don't think the `seam` language adds any clarity here.
```suggestion
   * functions.
```

isthmus/src/main/java/io/substrait/isthmus/DynamicConverterProvider.java
```diff
@@ -67,10 +74,25 @@ public DynamicConverterProvider(SimpleExtension.ExtensionCollection extensions)
...
+   *
+   * @param builder the builder carrying the configured components
+   */
+  public DynamicConverterProvider(ConverterProvider.Builder builder) {
+    super(builder);
     this.scalarFunctionConverter = createScalarFunctionConverter();
```
vbarua, line 98: Given that we override the scalarFunctionConverter set in the builder, effectively ignoring it, would it be worth throwing an error if it is set?

isthmus/src/main/java/io/substrait/isthmus/ConverterProvider.java
```diff
@@ -424,12 +509,150 @@ public TypeConverter getTypeConverter() {
...
+   * #extensions(SimpleExtension.ExtensionCollection) extensions} and {@link
+   * #typeFactory(RelDataTypeFactory) type factory} when {@link #build()} is called.
+   */
+  public static class Builder {
+    private SimpleExtension.ExtensionCollection extensions =
+        DefaultExtensionCatalog.DEFAULT_COLLECTION;
+    private RelDataTypeFactory typeFactory = SubstraitTypeSystem.TYPE_FACTORY;
+    private ScalarFunctionConverter scalarFunctionConverter;
+    private AggregateFunctionConverter aggregateFunctionConverter;
+    private WindowFunctionConverter windowFunctionConverter;
```
vbarua: minor: given that scalarFunctionConverter, aggregateFunctionConverter and windowFunctionConverter are allowed to be unset, would it make sense to wrap them in Optional?

isthmus/src/main/java/io/substrait/isthmus/DynamicConverterProvider.java
```diff
@@ -67,10 +74,25 @@ public DynamicConverterProvider(SimpleExtension.ExtensionCollection extensions)
...
+
+  /**
+   * Creates a new DynamicConverterProvider from a {@link ConverterProvider.Builder}, seeding base
+   * state from the builder and then installing the converter that handles dynamic extension
+   * functions. This is the seam through which subclasses route via {@code super(builder)}.
+   *
+   * @param builder the builder carrying the configured components
+   */
+  public DynamicConverterProvider(ConverterProvider.Builder builder) {
+    super(builder);
```
vbarua: Potentially out of scope. With both a builder and subclassing available as extension mechanisms, we do start to open ourselves up a bit to questions about which to use. For example 
why use subclassing here when we could inject a modified ConverterProvider like:
```java
ConverterProvider
  .builder()
  .scalarFunctionConverter(createScalarFunctionConverter())
  .sqlOperatorTable(buildSqlOperatorTable())
```

isthmus/src/main/java/io/substrait/isthmus/ConverterProvider.java
```diff
@@ -152,13 +190,59 @@ public ConverterProvider(
...
+        builder.windowFunctionConverter != null
+            ? builder.windowFunctionConverter
+            : new WindowFunctionConverter(
+                builder.extensions.windowFunctions(),
+                List.of(),
+                builder.typeFactory,
+                builder.typeConverter);
+    this.typeConverter = builder.typeConverter;
+    this.executionBehavior = builder.executionBehavior;
+    this.sqlParserConfig = builder.sqlParserConfig;
```
vbarua, line 216: minor: I would suggest grouping all of the fields that can be set directly next to each other, and then the ones that need to have their creation deferred next.

isthmus/src/main/java/io/substrait/isthmus/sql/SubstraitSqlStatementParser.java
```diff
@@ -27,8 +27,9 @@ public static List<SqlNode> parseStatements(String sqlStatements) throws SqlPars
    * Parse one or more SQL statements to a list of {@link SqlNode}s, using the parser settings from
    * the given {@link ConverterProvider}.
    *
-   * <p>To use a custom parser configuration, subclass {@link ConverterProvider} and override {@link
-   * ConverterProvider#getSqlParserConfig()}.
+   * <p>To use a custom parser configuration, build the {@link ConverterProvider} via {@link
+   * ConverterProvider#builder()} and its {@code sqlParserConfig(...)}; for fully dynamic behaviour,
+   * subclass {@link ConverterProvider} and override {@link ConverterProvider#getSqlParserConfig()}.
```
vbarua: minor: I think we can just point users to using the builder for customization for this use case, and not full on extension.

isthmus/src/main/java/io/substrait/isthmus/AutomaticDynamicFunctionMappingConverterProvider.java
vbarua, line 122: Should it be an error to have any of the *FunctionConverters set here, given that we throw those settings away?

vbarua (approved):
Looks reasonable overall. I did leave a number of suggestions, but nothing I would consider blocking.

One bigger item I did want to point out is that now we have two ways of customizing the ConverterProvider using the builder or subclassing, and they can also be mixed, which does make this a bit confusing potentially.

isthmus/src/main/java/io/substrait/isthmus/ConverterProvider.java
nielspardon: Applied, thanks.

isthmus/src/main/java/io/substrait/isthmus/ConverterProvider.java
nielspardon: Good point — trimmed to that wording. The rationale lives in the commit message and the PR description instead.

isthmus/src/main/java/io/substrait/isthmus/ConverterProvider.java
nielspardon: Applied.

isthmus/src/main/java/io/substrait/isthmus/DynamicConverterProvider.java
nielspardon: Dropped the "seam" sentence here and the equivalent one on `AutomaticDynamicFunctionMappingConverterProvider`.

isthmus/src/main/java/io/substrait/isthmus/DynamicConverterProvider.java
nielspardon: Agreed — it now throws. `DynamicConverterProvider` rejects a configured `scalarFunctionConverter`, and `AutomaticDynamicFunctionMappingConverterProvider` rejects any of the three, both with a message naming the setter to remove.

The validation happens inside the `super(...)` argument (`super(requireDerivedScalarFunctionConverter(builder))`) because statements before `super()` are not available on JDK 17. That is also what motivated the `Optional` getters on `Builder` — the subclasses are separate top-level classes, so they could not otherwise see the builder's private fields. Covered by `ConverterProviderBuilderTest`.

isthmus/src/main/java/io/substrait/isthmus/ConverterProvider.java
nielspardon: Done — the three are now `Optional` fields defaulting to `Optional.empty()`, the setters wrap with `Optional.ofNullable`, and the primary constructor uses `orElseGet(...)` instead of a null check. They also get `Optional` getters, which is what lets the dynamic subclasses reject a configured converter.

isthmus/src/main/java/io/substrait/isthmus/ConverterProvider.java
nielspardon: Regrouped: `typeFactory`, `extensions`, `typeConverter`, `executionBehavior` and `sqlParserConfig` are assigned first, then the three derived converters. Did the same for the field declarations and setters on `Builder`, which is most of the diff churn in that class.

isthmus/src/main/java/io/substrait/isthmus/sql/SubstraitSqlStatementParser.java
nielspardon: Trimmed to point at the builder only.

isthmus/src/main/java/io/substrait/isthmus/AutomaticDynamicFunctionMappingConverterProvider.java
nielspardon: Yes — it now throws for all three, see the reply on `DynamicConverterProvider`.

isthmus/src/main/java/io/substrait/isthmus/DynamicConverterProvider.java
nielspardon: Fair concern, and I agree it is the real design question here. I have left it out of this PR and opened #1048 for it.

The reason the two dynamic providers subclass rather than configure is that the builder cannot currently express what they override: `getSqlOperatorTable()` (both chain generated operators onto the base table) and `getCallConverters()` (`DynamicConverterProvider` appends an extra `ScalarFunctionConverter`). Making them builder configurations means adding `sqlOperatorTable` and `callConverters` to `Builder` and exposing the derivation helpers, since those values depend on the resolved extensions and type factory. That is a bigger change than this PR, and #1037 is stacked on top.

In the meantime the overlap is at least no longer silent: passing a builder-configured function converter to either dynamic provider now throws instead of being discarded.

nielspardon:
Thanks for the review! Pushed ff3176da addressing the suggestions:

- Javadoc: `Isthmus` casing, the "why a builder" rationale removed from the primary constructor, the "seam" language dropped, and `SubstraitSqlStatementParser` now points at the builder only.
- `Builder`'s three deferred function converters are `Optional` with matching getters; the primary constructor derives via `orElseGet` instead of a null check.
- Builder fields, setters and constructor assignments regrouped: directly-set first, derived after.
- Both dynamic providers now throw `IllegalArgumentException` when handed a builder that configures a function converter they derive themselves, rather than silently discarding it. New `ConverterProviderBuilderTest` covers this.

On the bigger point about two customization mechanisms — I agree, and opened #1048 rather than growing this PR (the builder would need `sqlOperatorTable` and `callConverters` before the dynamic providers could be expressed as configuration, and #1037 is stacked on top).

PR #1037, opened 2026-07-24
feat(isthmus): configurable unquoted identifier casing via the ConverterProvider builder

nielspardon:
## Summary

Add `ConverterProvider.Builder.unquotedCasing(Casing)` — a convenience for the common case of
overriding only the unquoted-identifier casing, layered on top of the current parser configuration
(equivalent to `sqlParserConfig(current.withUnquotedCasing(casing))`).

## Motivation

`TO_UPPER` is the ANSI/Calcite default, but a large class of engines fold unquoted identifiers to
lower-case or preserve them (Presto/Trino, Spark, DuckDB, Postgres), so a plan generated from
`SELECT … FROM orders` comes out with a `NamedScan` named `ORDERS` that those engines can't resolve.
This gives library and CLI users a one-liner to control it, while the full `sqlParserConfig(...)`
path (previous PR) remains for anything else.

## Changes

- `ConverterProvider.Builder.unquotedCasing(Casing)` convenience.
- `IsthmusEntryPoint` (CLI): `--unquotedcasing` now maps to
  `builder().unquotedCasing(...)`.
- `FromSql` example: uses `builder().unquotedCasing(Casing.UNCHANGED)` (matching the lower-case
  identifiers in its `CREATE TABLE` statements), replacing the deprecated
  `convert(sql, catalog, SqlDialect)` call.

## Split of #983

Per @vbarua's request, #983 was split into three stacked PRs (review/merge bottom-up):

1. #1035 — inject `ConverterProvider` into the SQL statement parsers (merged)
2. #1036 — introduce the `ConverterProvider` builder (merged)
3. **#1037 — configurable unquoted identifier casing via the builder (this PR)**

Both dependencies have merged, so this PR is now rebased onto `main` and is the last commit in the
stack — no longer blocked on anything.

🤖 Generated with AI


nielspardon:
Rebased onto `main` now that #1035 and #1036 have merged. The branch previously carried pre-squash copies of both, which is where the conflicts came from — it is now a single commit on top of `main`.

The one real conflict was positional: #1036 gained review changes that regrouped the `Builder` members, so `unquotedCasing(...)` now sits directly after `sqlParserConfig(...)` (with the parser-config settings) rather than ahead of the derived function converters. Verified with `:isthmus:check`, `:isthmus:javadoc`, `:isthmus-cli:compileJava` and `:examples:isthmus-api:build` — all 6 `UnquotedCasingTest` tests pass. Ready for review.

isthmus/src/main/java/io/substrait/isthmus/ConverterProvider.java
```diff
@@ -606,6 +607,23 @@ public Builder sqlParserConfig(SqlParser.Config sqlParserConfig) {
...
+     * #sqlParserConfig(SqlParser.Config)} call replaces the whole configuration and discards the
+     * casing set here; set the full config first, then apply this convenience.
+     *
+     * @param unquotedCasing the casing to apply to unquoted SQL identifiers during parsing
+     * @return this builder
+     */
+    public Builder unquotedCasing(Casing unquotedCasing) {
+      this.sqlParserConfig = this.sqlParserConfig.withUnquotedCasing(unquotedCasing);
+      return this;
+    }
```
vbarua, line 631: Two Qs:
1. How much is this utility needed instead of just pushing folks to use `sqlParserConfig` directly as that's much easier now.
2. If we include this, do we need to protect against users calling both `sqlParserConfig` and `unquotedCasing` in the same builder? I could see that causing confusion if `unquotedCasing` override parts of an already set `sqlParserConfig`.

nielspardon: **1. Why keep it rather than pushing people at `sqlParserConfig`?**

The argument is not brevity, it is that the hand-rolled version has a footgun. `DEFAULT_SQL_PARSER_CONFIG` is `SqlParser.Config.DEFAULT` **plus** `SqlDdlParserImpl.FACTORY` and `LENIENT` conformance. So the natural thing to reach for:

```java
.sqlParserConfig(SqlParser.Config.DEFAULT.withUnquotedCasing(Casing.UNCHANGED))
```

silently drops the DDL parser factory — `CREATE TABLE` stops parsing — and drops `LENIENT`. You have to know to start from `DEFAULT_SQL_PARSER_CONFIG` instead. `unquotedCasing(...)` layers onto whatever is configured, so it cannot drop either.

That said, you are right that it is not *enabling* anything: `main` already handles `--unquotedcasing` fine via `sqlParserConfig(DEFAULT_SQL_PARSER_CONFIG.withUnquotedCasing(...))`, and this only makes it terser. And I take the flag-creep point from #983 — casing has no special claim over `quoting`, `caseSensitive` or `conformance`. If you would rather we hold the line at one setter and just document `DEFAULT_SQL_PARSER_CONFIG` as the required starting point, I am happy to drop it; this PR then reduces to the `FromSql` cleanup (which also removes the deprecated `convert(sql, catalog, SqlDialect)` call).

**2. Protecting against both being set**

Good catch, and it was worse than confusing — it was order-dependent. Rather than forbid the combination, I made it order-independent in cc55ddcc: the casing is held as an `Optional<Casing>` on the builder and applied over the parser config in the primary constructor, so

```java
builder().sqlParserConfig(cfg).unquotedCasing(x)   // same result
builder().unquotedCasing(x).sqlParserConfig(cfg)   // as this
```

both give `cfg` with `x` applied. The rule is just "the casing override wins over any casing the supplied config carries", and the Javadoc says that instead of documenting an ordering requirement.

I went this way rather than throwing because `sqlParserConfig(custom).unquotedCasing(x)` is the combination I would expect people to actually want — a custom config plus a casing tweak — and rejecting it would push them straight back into the derive-by-hand footgun above. Happy to switch to a hard error if you prefer the stricter contract.

isthmus/src/main/java/io/substrait/isthmus/ConverterProvider.java
```diff
@@ -213,7 +213,13 @@ protected ConverterProvider(Builder builder) {
...
     this.typeConverter = builder.typeConverter;
     this.executionBehavior = builder.executionBehavior;
-    this.sqlParserConfig = builder.sqlParserConfig;
+    // An unquoted casing set through the builder convenience is applied over the parser config
+    // here, so that the two setters are order-independent.
+    this.sqlParserConfig =
+        builder
+            .unquotedCasing
+            .map(builder.sqlParserConfig::withUnquotedCasing)
+            .orElse(builder.sqlParserConfig);
```
vbarua, line 222: I think this is a reasonable solution to the problem of setting both.

isthmus/src/main/java/io/substrait/isthmus/ConverterProvider.java
```diff
@@ -259,8 +265,8 @@ private static Plan.ExecutionBehavior createDefaultExecutionBehavior() {
    * identifier casing.
    *
    * <p>Defaults to {@link #DEFAULT_SQL_PARSER_CONFIG}. Provide a custom configuration via {@link
-   * Builder#sqlParserConfig(SqlParser.Config)}, or override this method in a subclass for fully
-   * dynamic behaviour.
+   * Builder#sqlParserConfig(SqlParser.Config)} (or the {@link Builder#unquotedCasing(Casing)}
+   * convenience), or override this method in a subclass for fully dynamic behaviour.
```
vbarua: Small doc suggestion (spotless compliant)

```suggestion
   * Builder#sqlParserConfig(SqlParser.Config)}, or override this method in a subclass for even more
   * control.
   *
   * <p>To override just the unquotedCasting, consider {@link Builder#unquotedCasing(Casing)}
```

isthmus/src/main/java/io/substrait/isthmus/ConverterProvider.java
```diff
@@ -519,7 +525,8 @@ public Plan.ExecutionBehavior getExecutionBehavior() {
    *
    * <p>The builder starts from reasonable system defaults (the same ones behind {@link #DEFAULT})
    * and lets callers override individual components — most notably the Calcite {@link
-   * SqlParser.Config} used for SQL parsing, via {@link Builder#sqlParserConfig(SqlParser.Config)}.
+   * SqlParser.Config} used for SQL parsing, via {@link Builder#sqlParserConfig(SqlParser.Config)}
+   * for full control or {@link Builder#unquotedCasing(Casing)} for the common casing-only case.
```
vbarua: minor/opinionated: I think we should mention even less in here
```java
   * <p>The builder starts from reasonable system defaults (the same ones behind {@link #DEFAULT})
   * and lets callers override individual components
```
and avoid calling out specific builders. If folks want details, they can look at the methods on this class.

isthmus/src/main/java/io/substrait/isthmus/ConverterProvider.java
```diff
@@ -606,6 +614,25 @@ public Builder sqlParserConfig(SqlParser.Config sqlParserConfig) {
       return this;
     }
 
+    /**
+     * Convenience for the common case of overriding only the unquoted-identifier casing, without
+     * having to restate the rest of the parser configuration.
+     *
+     * <p>Unlike deriving a config by hand, this cannot accidentally drop the DDL parser factory or
+     * conformance that {@link ConverterProvider#DEFAULT_SQL_PARSER_CONFIG} supplies.
```
vbarua: I think we should just remove this. The reasoning for doing this is capture in your PR comment, but I don't think it provides value to our users.
```suggestion
```

isthmus/src/test/java/io/substrait/isthmus/UnquotedCasingTest.java
```diff
@@ -0,0 +1,140 @@
...
+import org.junit.jupiter.api.Test;
+
+/**
+ * Verifies that {@link ConverterProvider#builder()} configures the Calcite {@link SqlParser.Config}
+ * used for SQL parsing — via the {@code unquotedCasing} convenience or a full {@code
+ * sqlParserConfig} — and that the configured casing is applied consistently across both CREATE
+ * TABLE parsing and query parsing, so that the table name stored in a Substrait {@link NamedScan}
+ * reflects the configured casing.
+ */
+class UnquotedCasingTest {
```
vbarua: I think this would be better named as `ConverterProviderTest` with an `@Nested` UnquotedCasing section. It's easier to see that these are related to the ConverterProvider, and not some general unquoted casing functionality.

edit: Further down, I do see that there are some test for SubstraitCreateStatementParser and SqlToSubstrait. I would advocate for having those be tests be in `SqlToSubstraitTest` instead.

isthmus/src/test/java/io/substrait/isthmus/UnquotedCasingTest.java
```diff
@@ -0,0 +1,140 @@
...
+        ConverterProvider.builder().unquotedCasing(Casing.UNCHANGED).build();
+    assertEquals(Casing.UNCHANGED, provider.getSqlParserConfig().unquotedCasing());
+  }
+
+  @Test
+  void builderCasingToLower() {
+    ConverterProvider provider =
+        ConverterProvider.builder().unquotedCasing(Casing.TO_LOWER).build();
+    assertEquals(Casing.TO_LOWER, provider.getSqlParserConfig().unquotedCasing());
+  }
```
vbarua: minor: `builderCasingUnchanged` and `builderCasingToLower` are redundant tests. One check to verify that explicit values set on the ConverterProvider are respected would be enough

isthmus/src/test/java/io/substrait/isthmus/UnquotedCasingTest.java
```diff
@@ -0,0 +1,140 @@
...
+    assertEquals(SqlConformanceEnum.PRAGMATIC_2003, casingFirst.getSqlParserConfig().conformance());
+  }
+
+  /**
+   * The convenience layers onto {@link ConverterProvider#DEFAULT_SQL_PARSER_CONFIG}, so it cannot
+   * drop the DDL parser factory or conformance the way hand-deriving from {@link
+   * SqlParser.Config#DEFAULT} would.
+   */
+  @Test
+  void unquotedCasingRetainsIsthmusParserDefaults() {
```
vbarua: I'm not sure what this is checking, and it feels kind of redundant with the checks you've already added above. `builderFullSqlParserConfig` already checks that default values on `ConverterProvider.DEFAULT_SQL_PARSER_CONFIG` (like LENIENT) are retained.

isthmus/src/test/java/io/substrait/isthmus/UnquotedCasingTest.java
```diff
@@ -0,0 +1,140 @@
...
+    assertEquals(SqlConformanceEnum.LENIENT, provider.getSqlParserConfig().conformance());
+    assertEquals(
+        ConverterProvider.DEFAULT_SQL_PARSER_CONFIG.parserFactory(),
+        provider.getSqlParserConfig().parserFactory());
+  }
+
+  /**
+   * With the default {@link Casing#TO_UPPER}, both CREATE TABLE and SELECT fold unquoted
+   * identifiers to upper-case. The resulting {@link NamedScan} table name is {@code EMPLOYEES}.
+   */
```
vbarua: I think this should be in SqlToSubstraitTest, and even then I think the property that we care to verify is that it respects whatever casing is set in the ConverterProvider.

vbarua (commented):
The implementation side looks good overall, left some comments on docs and tests. I don't think any of it is blocking though.

vbarua (approved):

isthmus/src/main/java/io/substrait/isthmus/ConverterProvider.java
nielspardon: Applied, with `unquoted casing` spelled out in prose.

isthmus/src/main/java/io/substrait/isthmus/ConverterProvider.java
nielspardon: Agreed — trimmed to just "and lets callers override individual components".

isthmus/src/main/java/io/substrait/isthmus/ConverterProvider.java
nielspardon: Removed. The reasoning stays in the PR description where it belongs.

isthmus/src/test/java/io/substrait/isthmus/UnquotedCasingTest.java
nielspardon: Done, with one adjustment: `ConverterProviderBuilderTest` already exists (from #1036), so rather than add a third overlapping class I put the builder-level assertions there as a `@Nested UnquotedCasing` section. Method names lose their now-redundant prefixes, so they read as `ConverterProviderBuilderTest > UnquotedCasing > defaultsToUpper`.

The end-to-end check moved out to a new `SqlToSubstraitTest` — there wasn't one yet, only `SqlToSubstraitJoinConversionTest`.

isthmus/src/test/java/io/substrait/isthmus/UnquotedCasingTest.java
nielspardon: Collapsed into a single `configuredCasingIsUsed`.

isthmus/src/test/java/io/substrait/isthmus/UnquotedCasingTest.java
nielspardon: Removed — you're right that `builderFullSqlParserConfig` (now `fullSqlParserConfigIsUsed`) already covers the retained defaults.

isthmus/src/test/java/io/substrait/isthmus/UnquotedCasingTest.java
nielspardon: Moved to the new `SqlToSubstraitTest` and reframed as exactly that property — one `@ParameterizedTest` over `TO_UPPER` / `TO_LOWER` / `UNCHANGED` asserting the `NamedScan` name follows whatever casing the provider carries, instead of two hand-written cases.

PR #1045, opened 2026-07-28
docs: refactor AGENTS.md into an entry point, move build mechanics to CONTRIBUTING

nielspardon:
Aligns this repo's `AGENTS.md` with the ecosystem convention introduced for the spec repo in substrait-io/substrait#1117: a lean entry point plus the notes agents specifically need, with generally-applicable build and contribution mechanics living in the human-facing docs so there is a single source of truth.

## What changes

- **`AGENTS.md` becomes an entry point.** The intro is now a "Start here" pointer to `README.md` and `CONTRIBUTING.md` rather than a standalone build guide.
- **Adds "Keep PR descriptions high-signal."** Tells agents to drop the noise they tend to add — lists of files touched, claims that CI-verified things pass, implicit process notes — and to keep commit bodies free of git trailers, since the PR title + body become the `semantic-release` commit message and feed `CHANGELOG.md`.
- **Moves the general build mechanics to `CONTRIBUTING.md`.** The build/test/format command reference, PMD ruleset and tripwires, and the `build-logic` included-build note land in a new "Building and testing" section; the JDK 17 daemon-consistency and GraalVM native-image details fold into the existing "Gradle & JDK 17" section. `AGENTS.md` keeps only the agent-specific verification workflow (build-before-push, cross-module compile check).

The codebase-specific architecture knowledge (module layout, Immutables/visitor/proto-converter patterns, Isthmus/Spark change-safety notes) stays in `AGENTS.md` — it is this repo's equivalent of the spec PR's "what agents need to get right." No content was lost; the build mechanics were relocated, not removed.

🤖 Generated with AI

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,8 +1,17 @@
 # AGENTS.md
 
-Guidance for AI agents working in the `substrait-java` repository. This complements
-`CONTRIBUTING.md` (commit conventions, style guide) with practical, codebase-specific
-knowledge.
+Entry point for AI agents working in the `substrait-java` repository. Read the shared,
+human-facing docs first, then keep the codebase-specific notes below in mind.
+
+## Start here
+
+- **[`README.md`](README.md)** — what the project is, the module overview, and how to build
+  and run it.
+- **[`CONTRIBUTING.md`](CONTRIBUTING.md)** — commit conventions, the style guide, and the
+  build / test / format / PMD command mechanics plus the JDK 17 daemon and GraalVM
+  native-image setup.
+
+For GitHub work (issues, PRs), use the `gh` CLI.
 
 ## What this project is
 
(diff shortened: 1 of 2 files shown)


AGENTS.md
```diff
@@ -1,8 +1,17 @@
...
+
+- **[`README.md`](README.md)** — what the project is, the module overview, and how to build
+  and run it.
+- **[`CONTRIBUTING.md`](CONTRIBUTING.md)** — commit conventions, the style guide, and the
+  build / test / format / PMD command mechanics plus the JDK 17 daemon and GraalVM
+  native-image setup.
+
+For GitHub work (issues, PRs), use the `gh` CLI.
 
 ## What this project is
```
benbellick: Also think we don't really need this now, it should just be in the README if it isn't already.

AGENTS.md
```diff
@@ -115,47 +124,20 @@ Proto conversion is split into two directions, and the class name tells you whic
...
+[`CONTRIBUTING.md`](CONTRIBUTING.md#building-and-testing). Two habits matter most for agents:
+
+- **Build the whole thing before pushing.** Narrower local tasks pass while CI fails: **PMD**
+  runs only via `check`/`build` and `javadoc` doclint only via `build`/`javadocJar` — never
+  via `compileJava` / `test` / `spotlessCheck` — while CI runs the full `./gradlew build
+  --rerun-tasks`. Run `./gradlew :core:check :core:javadoc` (and the module you touched)
+  before pushing.
+- **When you extend the public expression/visitor API, verify the dependent modules still
+  compile** — they have their own visitor implementors:
+  `./gradlew :core:spotlessCheck :isthmus:compileJava :spark:spark-3.5_2.12:compileScala :examples:substrait-spark:compileJava`
```
benbellick:
```suggestion
```

Do we need all of that since its already in the CONTRIBUTING.md? If we find the agent isn't doing this itself, then we can add a single line that says:
> run all tests and ensure they are passing before pushing

AGENTS.md
```diff
@@ -199,6 +181,19 @@ compile — they have their own visitor implementors:
   valid commit message). Scope tags seen in history: `feat(core)`, `feat(pojo)`,
```
benbellick: I also think this whole section can reasonably be removed. Its all stuff you'd want a human to know as well, and there is a section called `Commit Conventions` already in CONTRIBUTING.md.

AGENTS.md
```diff
@@ -199,6 +181,19 @@ compile — they have their own visitor implementors:
...
   `feat(isthmus)`, `feat(extensions)`, `build(deps)`, `chore(release)`. A `!` marks a
   breaking change.
+- **Keep PR descriptions high-signal.** The PR title and body together become the
+  squash-merge commit message that `semantic-release` uses to build `CHANGELOG.md`, so they
+  must form that valid conventional commit. Beyond that, leave out the noise agents tend to
+  add:
+  - **Lists of files touched** — they're in the diff.
+  - **Claims that CI-verified things pass** — e.g. "tests pass", "spotless clean". If they
+    didn't, the checks would be red.
+  - **Process notes that are already implicit** — e.g. "opened as draft pending review".
```
benbellick, line 171: ➕

Thanks for adding this bit

benbellick (approved):
Thanks! I think this is a great improvement. There may be some more places worth slimming down, but we can save it for a subsequent PR. 🚀

PR #1046, opened 2026-07-29
fix(isthmus): route native-image metadata queries through reflective provider

### LINKED ISSUE #251: [isthmus] CLI failures in native image due to undefined proxy classes
Queries like
```
./isthmus -c 'CREATE TABLE foo(a VARCHAR);' "SELECT a FROM foo"
```
work in Isthmus, however when a GROUP BY is added
```
./isthmus -c 'CREATE TABLE foo(a VARCHAR);' "SELECT a FROM foo GROUP BY a"
```

The CLI fails with
```
Proxy class defined by interfaces [interface org.apache.calcite.rel.metadata.BuiltInMetadata$ColumnUniqueness] not found.
```

This also appears to happen with `SELECT DISTINCT` queries.

There does appear to be some configuration around proxy classes, but it is possibly insufficient.
https://github.com/substrait-io/substrait-java/blob/ce0f1eea6b89b9bb733edab80c1bedea71345048/isthmus/proxies.json#L6

nielspardon:
## Problem

Queries like the following crash in the isthmus native image (#251):

```
./isthmus -c 'CREATE TABLE foo(a VARCHAR);' "SELECT a FROM foo GROUP BY a"
./isthmus -c 'CREATE TABLE foo(a VARCHAR);' "SELECT DISTINCT a FROM foo"
```

A grouping-only `GROUP BY` (and `SELECT DISTINCT`) goes through
`RelBuilder.aggregate_` → `alreadyUnique()` → `RelMetadataQuery.areColumnsUnique()`.
Calcite's default `RelMetadataQuery` obtains its handlers from
`JaninoRelMetadataProvider`, which **compiles a handler class at runtime with Janino**.
Runtime code generation is unavailable in a GraalVM native image (closed-world,
ahead-of-time compiled), so the query fails:

```
java.lang.AssertionError: Required type "Ljava/lang/AssertionError;" not found
    at org.codehaus.janino.IClassLoader.requireType
    at org.apache.calcite.rel.metadata.JaninoRelMetadataProvider.compile
```

An aggregate *with* a function (`count(...)`) skips the uniqueness check, so the
existing smoke test never exercised the broken path — which is why this stayed
uncaught.

## Fix

Calcite ships `ProxyingMetadataHandlerProvider`, a reflection-based (dynamic-proxy)
alternative to the Janino code generator. Dynamic proxies are supported in native
images once their interfaces are registered. `SqlConverterBase` already used this
provider — but only the SQL-expression path used that cluster. The SQL-query path
(`SubstraitSqlToCalcite`) and the Substrait→Calcite path
(`ConverterProvider.getRelBuilder`, `SubstraitRelNodeConverter.convert`) built
clusters with no metadata supplier and fell back to Janino.

- Extract the setup into `Utils.useReflectiveMetadataProvider(RelOptCluster)` and
  apply it to every cluster / `RelBuilder` used for Calcite conversion (removing the
  duplicated inline lambda in `SqlConverterBase`).
- The reflective provider proxies the **bare** `BuiltInMetadata$*` interfaces, not
  just the `$Handler` sub-interfaces already present, so register those in
  `reachability-metadata.json` (the safe superset — unused proxy registrations are
  harmless).
- Add grouping-only `GROUP BY` and `SELECT DISTINCT` cases to `smoke.sh` so the
  native image exercises the uniqueness path in CI.

Closes #251

🤖 Generated with AI

### DIFF
--- a/isthmus-cli/src/main/resources/META-INF/native-image/io.substrait/isthmus-cli/reachability-metadata.json
+++ b/isthmus-cli/src/main/resources/META-INF/native-image/io.substrait/isthmus-cli/reachability-metadata.json
@@ -16,7 +16,21 @@
     {
       "type": {
         "proxy": [
+          "org.apache.calcite.rel.metadata.BuiltInMetadata$AllPredicates"
+        ]
+      }
+    },
+    {
+      "type": {
+        "proxy": [
           "org.apache.calcite.rel.metadata.BuiltInMetadata$Collation$Handler"
+        ]
+      }
+    },
+    {
+      "type": {
+        "proxy": [
+          "org.apache.calcite.rel.metadata.BuiltInMetadata$Collation"
         ]
       }
     },
@@ -30,7 +44,21 @@
     {
       "type": {
         "proxy": [
+          "org.apache.calcite.rel.metadata.BuiltInMetadata$ColumnOrigin"
+        ]
+      }
+    },
+    {
+      "type": {
+        "proxy": [
           "org.apache.calcite.rel.metadata.BuiltInMetadata$ColumnUniqueness$Handler"
+        ]
+      }
+    },
+    {
+      "type": {
+        "proxy": [
+          "org.apache.calcite.rel.metadata.BuiltInMetadata$ColumnUniqueness"
         ]
       }
     },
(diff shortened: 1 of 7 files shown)


bvolpato (approved):
LGTM, thanks!

PR #1050, opened 2026-07-31
refactor(isthmus): derive CREATE-statement table types from the provider

nielspardon:
## Summary

Finishes the deferred half of @vbarua's [comment on #1035](https://github.com/substrait-io/substrait-java/pull/1035#discussion_r3646273565) —
*"I wonder if it would make sense to get the RelTypeFactory and SqlConverterBase from the
ConverterProvider as well"*. #1036 threaded the provider into the `CalciteCatalogReader`
construction; this covers the type derivation underneath it.

## The gap

On the provider-injected path (`processCreateStatementsToCatalog(provider, …)` →
`processCreateStatementsToSchema(provider, …)` → `createSubstraitTable(…)`):

- `createSubstraitTable` built the row type with `SubstraitTypeSystem.TYPE_FACTORY`, not the
  provider's type factory.
- Column types came from `col.dataType.deriveType(VALIDATOR)`, where `VALIDATOR` and
  `EMPTY_CATALOG` are `static final`, built on the global type factory plus
  `SqlConverterBase.CONNECTION_CONFIG`.

So a builder-configured `typeFactory` produced a catalog reader that used it while the tables
inside were typed by the global one. That is harmless today — `getCalciteConnectionConfig()`
returns the same case-insensitive config as `SqlConverterBase.CONNECTION_CONFIG` — but a custom
type factory carrying a different type system (different decimal precision limits, say) was
silently bypassed for table column types.

## Changes

- New private `createEmptyCatalog(ConverterProvider)` and `createValidator(ConverterProvider)`
  helpers. Because `SubstraitSqlValidator` takes its type factory from the catalog reader, fixing
  the empty catalog fixes column derivation as well.
- The validator is built once per call and passed, with the provider's type factory, into
  `createSubstraitTable`.
- `EMPTY_CATALOG` and `VALIDATOR` are kept and now defined via the same helpers against
  `ConverterProvider.DEFAULT`, so their values are unchanged and nothing public was removed.

## Note on scope

Two things worth calling out explicitly, both easy to drop if you would rather they were separate:

1. The validator is now built with `converterProvider.getSqlOperatorTable()` instead of a
   hardcoded `SubstraitOperatorTable.INSTANCE`, matching how #1035 made the query path source its
   operator table. Identical for the default provider; it only matters for a provider that adds
   operators.
2. `EMPTY_CATALOG` and `VALIDATOR` now have no callers inside the class either (each entry point
   derives its own from the provider). They are still `public`, so they are retained — but if you
   want them deprecated in favour of the provider-derived path, that is a one-line follow-up.

Independent of #1037 — no overlapping files, so the two can merge in either order.

🤖 Generated with AI

### DIFF
--- a/isthmus/src/test/java/io/substrait/isthmus/sql/CreateStatementParserProviderConfigTest.java
+++ b/isthmus/src/test/java/io/substrait/isthmus/sql/CreateStatementParserProviderConfigTest.java
@@ -8,11 +8,13 @@ import static org.junit.jupiter.api.Assertions.assertTrue;
 import io.substrait.isthmus.ConverterProvider;
 import io.substrait.isthmus.SubstraitTypeSystem;
 import io.substrait.isthmus.calcite.SubstraitTable;
+import java.util.ArrayList;
 import java.util.List;
 import org.apache.calcite.prepare.CalciteCatalogReader;
 import org.apache.calcite.rel.type.RelDataType;
 import org.apache.calcite.sql.parser.SqlParseException;
 import org.apache.calcite.sql.type.SqlTypeFactoryImpl;
+import org.apache.calcite.sql.type.SqlTypeName;
 import org.junit.jupiter.api.Test;
 
 /**
@@ -26,13 +28,18 @@ class CreateStatementParserProviderConfigTest {
       "CREATE TABLE employees (id BIGINT, name VARCHAR, salary DECIMAL(10, 2))";
 
   /**
-   * A type factory that records the struct types it is asked to build, so a test can tell which
-   * factory produced a table's row type and where its column types came from.
+   * A type factory that records the row types it is asked to build and the SQL types it is asked to
+   * create, so a test can tell which factory produced a table's row type and which one the
+   * validator derived its column types through.
+   *
+   * <p>Recording the requests is what makes the column-derivation check meaningful: {@code
+   * RelDataTypeFactoryImpl.canonize} interns through a {@code static} cache shared by every factory
+   * instance, so comparing interned type instances cannot distinguish one factory from another.
    */
   static final class RecordingTypeFactory extends SqlTypeFactoryImpl {
 
     private boolean createStructTypeCalled;
-    private List<RelDataType> recordedFieldTypes = List.of();
+    private final List<SqlTypeName> requestedSqlTypeNames = new ArrayList<>();
 
     RecordingTypeFactory() {
       super(SubstraitTypeSystem.TYPE_SYSTEM);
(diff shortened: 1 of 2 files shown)


isthmus/src/test/java/io/substrait/isthmus/sql/CreateStatementParserProviderConfigTest.java
```diff
@@ -0,0 +1,121 @@
...
+
+    SubstraitCreateStatementParser.processCreateStatementsToCatalog(provider, CREATE_STATEMENT);
+
+    List<RelDataType> fieldTypes = typeFactory.recordedFieldTypes();
+    assertEquals(3, fieldTypes.size());
+    for (RelDataType fieldType : fieldTypes) {
+      assertSame(
+          fieldType,
+          typeFactory.canonizeType(fieldType),
+          "column types should be interned in the provider's type factory");
```
bvolpato: Nice targeted coverage around both entry points. Non-blocking test thought: Calcite uses a static interner here, so canonizing a field type cannot distinguish the provider factory from the global factory. Recording createSqlType calls on RecordingTypeFactory, or asserting a custom type-system result, would make this check protect the validator path too.

bvolpato (approved):
LGTM! Clean provider-config propagation through CREATE parsing, with focused coverage for both entry points and defaults. Left one non-blocking test-strengthening note inline.

isthmus/src/test/java/io/substrait/isthmus/sql/CreateStatementParserProviderConfigTest.java
nielspardon: You are right, and thanks — I checked and `RelDataTypeFactoryImpl.canonize(RelDataType)` in Calcite 1.42.0 compiles to `getstatic DATATYPE_CACHE` on a `private static final Interner`, so interning is JVM-wide and the comparison held regardless of which factory produced the type. That assertion was decorative: the test only failed without the fix because its size assertion duplicated the row-type check the other two tests already cover, so the validator path was effectively untested.

Fixed in c58bea6b by recording the requested `SqlTypeName`s and asserting the declared column types are among them.

Worth noting for anyone reading later: matching the *specific* names is what makes it work. A plain "was `createSqlType` called" flag would have been a false positive — with the fix reverted, the provider factory still sees `[BOOLEAN, BOOLEAN, CHAR, NULL, ...]` from catalog setup, and never `BIGINT`/`VARCHAR`/`DECIMAL`. So the assertion now fails for the right reason.

Also rebased onto current `main` while I was in there, since the branch had fallen behind.

nielspardon: Filed #1057 for the other open question from this PR — `EMPTY_CATALOG` and `VALIDATOR` now have no callers anywhere, including inside their own class, so whether they stay public is worth deciding separately.

PR #1058, opened 2026-08-03
fix!: replace assert-based validation with real exceptions

### LINKED ISSUE #1047: Replace `assert`-based validation with real exceptions in `:core` and `:isthmus`
`VirtualTableScan.check()` validates its invariants with `assert` statements, which do not
run in a normal deployment — see the comment below for the concrete finding (Calcite's
asserts on WebSphere Liberty: nothing threw, the validation silently did not happen). Since
`assert` compiles behind `$assertionsDisabled`, these checks fire in Gradle's test JVMs
(which enable `-ea` by default) and nowhere else, so the invariant is enforced in CI and
not in production.

`VirtualTableScan` is not the only place, so this issue covers all `assert`-based
validation in the code we own — `:core` and `:isthmus`. Calcite's own asserts are out of
scope; we can only stop adding to the problem in our own converters.

The general rule proposed:

- **caller-facing invariant** (validating something the caller handed us: builder input,
  a proto message, a Calcite `RelNode`) → `IllegalArgumentException`
- **internal invariant / configuration** ("this cannot happen unless the converter is
  misconfigured") → `IllegalStateException`
- **no bare `assert` left in main sources.** An assert that can fail is worse than useless
  (it never runs where it matters); an assert that genuinely cannot fail costs nothing to
  write as an explicit throw.

### `:core`

| Site | Kind | Proposed |
| --- | --- | --- |
| `VirtualTableScan.check()` — name count vs depth-first named-field count; no null names/rows; row shape; rows not nullable; row field types match schema (#601) | caller-facing `@Value.Check` | `IllegalArgumentException` |
| `Expression.NestedList.check()` — non-empty; all values same type | caller-facing `@Value.Check` | `IllegalArgumentException` |
| `ExpressionProtoConverter.toLiteral()` — `assert e.getRexTypeCase() == LITERAL` | caller-facing (POJO → proto) | `IllegalArgumentException` |
| `VariadicParameterConsistencyValidator` — throws `new AssertionError(...)` explicitly | caller-facing validation | `IllegalArgumentException` |

### `:isthmus`

| Site | Kind | Proposed |
| --- | --- | --- |
| `SubstraitRelVisitor` (×2, INSERT/DELETE and UPDATE) — `modify.getTable() != null` | Calcite input | `IllegalArgumentException` |
| `SubstraitRelNodeConverter` (×2) — `relBuilder.getRelOptSchema() != null` | configuration | `IllegalStateException` |
| `SubstraitRelNodeConverter` — `catalogReader != null`, asserted on a value just assigned from a cast | dead check | drop, or fold into the `getRelOptSchema()` check |
| `SubstraitRelNodeConverter` — `targetTable != null` (`// checked by validation`) | catalog lookup miss | `IllegalStateException`, mirroring the existing `"Table not found in Calcite catalog: …"` message |
| `SqlMapValueConstructorCallConverter` — operand count even | Calcite input | `IllegalArgumentException` (also fix the `@throws AssertionError` Javadoc) |
| `CallConverters.CASE` — operand count odd | Calcite input | `IllegalArgumentException` |
| `FunctionConverter.matchKeys()` — `rexOperands.size() == opTypes.size()` | internal, private method | `IllegalStateException` |
| `CreateTable.copy()`, `CreateView.copy()` — `inputs.size() == 1` | `RelNode.copy` contract, our own rel | `IllegalArgumentException` |

### `:spark` is already fine

The Scala sources use `require(...)` (6 sites in `ToSubstraitType`, `ToLogicalPlan`,
`ToSparkExpression`), which throws `IllegalArgumentException` unconditionally — Scala's
`require` is not the Java `assert` keyword. There is no Scala `assert` in the Spark
sources, so nothing to change there.

### What the silent gap costs today

Three examples of what happens with assertions off, which is what a consumer sees:

- `ExpressionProtoConverter.toLiteral()` — a protobuf oneof getter returns the *default
  instance* when its case is not set, so a non-literal where a literal is required is
  silently converted to an empty literal. We emit a wrong plan instead of failing.
- `Expression.NestedList` — `getType()` reads `values().get(0)`, so an empty nested list
  throws `IndexOutOfBoundsException` from an unrelated place instead of the assert's
  "use `ExpressionCreator.emptyList()`" hint.
- `SubstraitRelNodeConverter` `targetTable` — a missing catalog entry becomes an NPE inside
  Calcite rather than the clear `"Table not found in Calcite catalog"` message the same
  file already produces 150 lines earlier.

### Tests to update

`assertThrows(AssertionError.class, …)` becomes the new exception type in
`core` `VirtualTableScanTest`, `NestedListExpressionTest`,
`VariadicParameterConsistencyTest`, and `isthmus` `VirtualTableScanTest`. (Other test
matches on `AssertionError` are `orElseThrow(AssertionError::new)` idioms and unaffected.)

### Things to decide

- **Wire-plan compatibility.** `ProtoRelConverter.newVirtualTable` builds through the
  `VirtualTableScan` builder, so a hard exception makes `substrait-java` reject externally
  produced plans whose virtual-table literals disagree with the base schema — plans that
  parse today whenever assertions are off. #812 handled comparable tightening with a
  `LOGGER.warn` plus a "this will be an error in the next release" grace period; the same
  may be warranted for this one and for `NestedList`.
- **Strictness of the row/schema type comparison.** It uses exact `Type.equals`, so
  nullability must match precisely. That is the strictest reading of the spec and the most
  likely thing to reject other producers' plans, so it is worth confirming before it
  becomes a hard failure.
- **Guarding against regressions.** The custom PMD ruleset
  (`build-logic/src/main/resources/substrait-pmd.xml`) could carry an XPath rule matching
  `AssertStatement` in main sources, so new asserts do not creep back in.
- **Scope of one PR vs several.** The `:core` `@Value.Check` conversions are the
  behaviour-visible part; the `:isthmus` ones are mostly mechanical. Splitting keeps the
  reviewable surface small.

nielspardon:
`assert` compiles behind `$assertionsDisabled`, so the invariant checks in `:core` and `:isthmus` fired in Gradle's test JVMs (which enable `-ea` by default) and nowhere else. The invariants were documented by the code and enforced in CI, but not in the place where a malformed plan actually causes damage. And when they *do* run, `AssertionError` is an `Error`, not an `Exception`, so a host that wraps plan conversion in `catch (Exception e)` to report a bad plan does not catch it — it propagates as a hard failure.

Two of these checks cost a consumer something today:

- `Expression.NestedList` — `getType()` reads `values().get(0)`, so an empty nested list throws `IndexOutOfBoundsException` from an unrelated place instead of the assert's "use `ExpressionCreator.emptyList()`" hint.
- `SubstraitRelNodeConverter` `targetTable` — a missing catalog entry passes `null` into `LogicalTableModify.create`, which NPEs inside Calcite, rather than producing the clear `"Table not found in Calcite catalog"` message the same file already produces 150 lines earlier.

The rest are invariants that today's call sites cannot violate. They stay as explicit guards — an assert that genuinely cannot fail costs nothing to write as a throw — but they are not fixes for anything.

## The rule applied

- **caller-facing invariant** (builder input, a proto message, a Calcite `RelNode`) → `IllegalArgumentException`
- **internal invariant / configuration** ("cannot happen unless the converter is misconfigured") → `IllegalStateException`
- **no bare `assert` left**, anywhere.

## `:core`

| Site | Now throws |
| --- | --- |
| `VirtualTableScan.check()` — name count vs depth-first named-field count, row shape, rows not nullable, row field types match schema | `IllegalArgumentException` |
| `Expression.NestedList.check()` — non-empty, all values the same type | `IllegalArgumentException` |
| `ExpressionProtoConverter.toLiteral()` — guard; all call sites are statically `Expression.Literal` | `IllegalArgumentException` |
| `VariadicParameterConsistencyValidator` (already threw `AssertionError` unconditionally) | `IllegalArgumentException` |

`VirtualTableScan.check()`'s compound assert is split into one check per invariant so the message names the mismatched counts, and the schema's field types are now derived once per relation instead of once per row.

`NestedList`'s homogeneity check compares element types **modulo nullability**. The `assert` it replaces compared types exactly, which under `-ea` rejected `SELECT ARRAY[not_null_col, nullable_col]`: SQL list constructors do not cast their values to a common type, so such a list legitimately holds values that differ only in nullability. Enforcing the old comparison in production would have broken queries that convert today.

## `:isthmus`

| Site | Now throws |
| --- | --- |
| `SubstraitRelNodeConverter` ×2 — `relBuilder.getRelOptSchema()` | `IllegalStateException` |
| `SubstraitRelNodeConverter` — `targetTable` | `IllegalStateException`, reusing the existing message |
| `SubstraitRelNodeConverter` — `catalogReader`, asserted on a value just assigned from a cast | dropped as dead |
| `SqlMapValueConstructorCallConverter` — operand count even | `IllegalArgumentException` |
| `CallConverters.CASE` — operand count odd | `IllegalArgumentException` |
| `CreateTable.copy()`, `CreateView.copy()` — `inputs.size() == 1` | `IllegalArgumentException` |
| `SubstraitRelVisitor` ×2 (INSERT/DELETE, UPDATE) — guard; Calcite's `TableModify` constructor already dereferences the table | `IllegalArgumentException` |
| `FunctionConverter.matchKeys()` — guard; both lists come from the same operand stream | `IllegalStateException` |

Two small refactors fall out of this: a `requireTable(TableModify)` helper in `SubstraitRelVisitor` (which also collapses the repeated `modify.getTable()` calls) and `requireRelOptSchema()` / `requireCatalogReader()` helpers in `SubstraitRelNodeConverter`, shared by the write and update paths and `protected` so subclasses overriding those `visit` methods can reuse them. The `targetTable` null check deliberately stays *after* the `switch` — the CTAS branch returns earlier and legitimately has no pre-existing table, so hoisting it to the lookup would break `handleCreateTableAs`.

`SqlMapValueConstructorCallConverter` also grows two fixes the operand-count check exposed: it ran *after* the cast that would already have thrown `ClassCastException`, and the map it builds is now insertion-ordered so a map literal keeps the key order written in the query.

The stale `@throws AssertionError` Javadoc tags are updated.

## Guarding against regressions

A custom PMD rule `AvoidAssertStatement` (`//AssertStatement`) in `substrait-pmd.xml` fails the build on any new `assert`, and its violation message states the `IllegalArgumentException` / `IllegalStateException` rule above at the offending line. PMD scans test source sets too, so the one assert in isthmus test code (`RepeatRel.copy()`) is converted as well.

## Two calls worth a second opinion

- **No grace period.** `Plan.Root.check()` is the precedent — hard `IllegalArgumentException` for the invariant, `LOGGER.warn` only for its one legacy allowance.
- **`VirtualTableScan` row/schema types still compare with exact `Type.equals`**, so nullability must match precisely there. That is the strictest reading of the spec and the most likely thing to reject another producer's plan, but relaxing it would be a semantic change rather than part of this one.

`:spark` needed no changes: its Scala sources use `require(...)`, not the Java `assert` keyword.

BREAKING CHANGE: validation that previously used `assert` now throws unconditionally. Callers catching `AssertionError` must catch `IllegalArgumentException` or `IllegalStateException` instead, and code running without `-ea` — i.e. most deployments — will now see these checks fire: `VirtualTableScan` and `Expression.NestedList` reject malformed input both from their builders and from `ProtoRelConverter` / `ProtoExpressionConverter`, and `VariadicParameterConsistencyValidator` throws `IllegalArgumentException` rather than `AssertionError`.

Closes #1047

🤖 Generated with AI

### DIFF
--- a/build-logic/src/main/resources/substrait-pmd.xml
+++ b/build-logic/src/main/resources/substrait-pmd.xml
@@ -21,4 +21,22 @@
   <rule ref="category/java/bestpractices.xml/UnusedPrivateMethod" />
   <rule ref="category/java/design.xml/AvoidThrowingRawExceptionTypes" />
   <rule ref="category/java/errorprone.xml/MissingSerialVersionUID" />
+
+  <rule name="AvoidAssertStatement" language="java"
+    message="Do not validate with assert: assertions only run when the JVM is started with -ea. Throw IllegalArgumentException for caller-facing invariants or IllegalStateException for internal ones."
+    class="net.sourceforge.pmd.lang.rule.xpath.XPathRule">
+    <description>
+      Java assertions are disabled unless the host JVM runs with -ea, which is the case for
+      Gradle's test JVMs and for almost nothing else. An assert-based check is therefore
+      enforced in CI and silently skipped in every real deployment, and AssertionError is an
+      Error rather than an Exception, so callers that catch Exception to report a bad plan
+      never see it.
+    </description>
+    <priority>3</priority>
+    <properties>
+      <property name="xpath">
+        <value>//AssertStatement</value>
+      </property>
+    </properties>
+  </rule>
 </ruleset>
--- a/core/src/main/java/io/substrait/expression/Expression.java
+++ b/core/src/main/java/io/substrait/expression/Expression.java
@@ -1790,7 +1790,11 @@ public interface Expression extends FunctionArg {
     public abstract List<Expression> values();
 
     /**
-     * Validates that the nested list is not empty and all values have the same type.
+     * Validates that the nested list is not empty and all values have the same type, disregarding
+     * nullability. Values of mixed nullability are allowed, because SQL list constructors do not
+     * cast their values to a common type: {@code ARRAY[not_null_column, nullable_column]} produces
+     * values that differ only in nullability. The list's element type is that of its first value,
+     * as {@link #getType()} shows.
      *
      * @throws IllegalArgumentException if the list is empty or its values have differing types
      */
@@ -1801,15 +1805,11 @@ public interface Expression extends FunctionArg {
             "To specify an empty list, use ExpressionCreator.emptyList()");
       }
 
-      List<Type> distinctTypes =
-          values().stream()
-              .map(Expression::getType)
-              .distinct()
-              .collect(java.util.stream.Collectors.toList());
-      if (distinctTypes.size() > 1) {
+      List<Type> types =
+          values().stream().map(Expression::getType).collect(java.util.stream.Collectors.toList());
+      if (types.stream().map(TypeCreator::asNullable).distinct().limit(2).count() > 1) {
         throw new IllegalArgumentException(
-            String.format(
-                "All values in NestedList must have the same type, found: %s", distinctTypes));
+            String.format("All values in NestedList must have the same type, found: %s", types));
       }
     }
 
--- a/core/src/main/java/io/substrait/expression/Expression.java
+++ b/core/src/main/java/io/substrait/expression/Expression.java
@@ -1789,13 +1789,28 @@ public interface Expression extends FunctionArg {
      */
     public abstract List<Expression> values();
 
-    /** Validates that the nested list is not empty and all values have the same type. */
+    /**
+     * Validates that the nested list is not empty and all values have the same type.
+     *
+     * @throws IllegalArgumentException if the list is empty or its values have differing types
+     */
     @Value.Check
     protected void check() {
-      assert !values().isEmpty() : "To specify an empty list, use ExpressionCreator.emptyList()";
+      if (values().isEmpty()) {
+        throw new IllegalArgumentException(
+            "To specify an empty list, use ExpressionCreator.emptyList()");
+      }
 
-      assert values().stream().map(Expression::getType).distinct().count() <= 1
-          : "All values in NestedList must have the same type";
+      List<Type> distinctTypes =
+          values().stream()
+              .map(Expression::getType)
+              .distinct()
+              .collect(java.util.stream.Collectors.toList());
+      if (distinctTypes.size() > 1) {
+        throw new IllegalArgumentException(
+            String.format(
+                "All values in NestedList must have the same type, found: %s", distinctTypes));
+      }
     }
 
     @Override
--- a/core/src/main/java/io/substrait/expression/proto/ExpressionProtoConverter.java
+++ b/core/src/main/java/io/substrait/expression/proto/ExpressionProtoConverter.java
@@ -477,7 +477,12 @@ public class ExpressionProtoConverter
 
   private Expression.Literal toLiteral(io.substrait.expression.Expression expression) {
     Expression e = toProto(expression);
-    assert e.getRexTypeCase() == Expression.RexTypeCase.LITERAL;
+    if (e.getRexTypeCase() != Expression.RexTypeCase.LITERAL) {
+      throw new IllegalArgumentException(
+          String.format(
+              "Expected a literal expression, but %s was converted to a %s",
+              expression.getClass().getSimpleName(), e.getRexTypeCase()));
+    }
     return e.getLiteral();
   }
 
--- a/core/src/main/java/io/substrait/relation/VirtualTableScan.java
+++ b/core/src/main/java/io/substrait/relation/VirtualTableScan.java
@@ -5,7 +5,6 @@ import io.substrait.type.NamedFieldCountingTypeVisitor;
 import io.substrait.type.Type;
 import io.substrait.util.VisitationContext;
 import java.util.List;
-import java.util.Objects;
 import org.immutables.value.Value;
 
 /** A read relation that produces an inline table from a fixed list of literal rows. */
@@ -23,8 +22,6 @@ public abstract class VirtualTableScan extends AbstractReadRel {
    * Checks the following invariants when construction a VirtualTableScan
    *
    * <ul>
-   *   <li>no null field names
-   *   <li>no null rows
    *   <li>row shape must match field-list
    *   <li>row field types must match schema types
    * </ul>
@@ -43,16 +40,10 @@ public abstract class VirtualTableScan extends AbstractReadRel {
               "VirtualTableScan schema names count (%d) does not match the depth-first named-field count of the schema struct (%d)",
               names.size(), schemaNameCount));
     }
-    if (names.stream().anyMatch(Objects::isNull)) {
-      throw new IllegalArgumentException("VirtualTableScan schema names must not contain null");
-    }
 
-    List<Expression.NestedStruct> rows = getRows();
-    if (rows.stream().anyMatch(Objects::isNull)) {
-      throw new IllegalArgumentException("VirtualTableScan rows must not contain null");
-    }
+    List<Type> schemaFieldTypes = getInitialSchema().struct().fields();
 
-    for (Expression.NestedStruct row : rows) {
+    for (Expression.NestedStruct row : getRows()) {
       // At the PROTOBUF layer, the Nested.Struct message does not carry nullability information.
       // Nullability is attached to the Nested message, which can contain a Nested.Struct.
       // The NestedStruct POJO flattens the Nested and Nested.Struct messages together, allowing
@@ -75,7 +66,7 @@ public abstract class VirtualTableScan extends AbstractReadRel {
                 rowNameCount, names.size()));
       }
 
-      validateRowConformsToSchema(row);
+      validateRowConformsToSchema(row, schemaFieldTypes);
     }
   }
 
@@ -83,11 +74,11 @@ public abstract class VirtualTableScan extends AbstractReadRel {
    * Validates that a row's field types conform to the table's schema.
    *
    * @param row the row to validate
+   * @param schemaFieldTypes the field types of the table's schema
    * @throws IllegalArgumentException if the row does not conform to the schema
    */
-  private void validateRowConformsToSchema(Expression.NestedStruct row) {
-    Type.Struct schemaStruct = getInitialSchema().struct();
-    List<Type> schemaFieldTypes = schemaStruct.fields();
+  private static void validateRowConformsToSchema(
+      Expression.NestedStruct row, List<Type> schemaFieldTypes) {
     List<Expression> rowFields = row.fields();
 
     if (rowFields.size() != schemaFieldTypes.size()) {
--- a/core/src/main/java/io/substrait/relation/VirtualTableScan.java
+++ b/core/src/main/java/io/substrait/relation/VirtualTableScan.java
@@ -28,34 +28,53 @@ public abstract class VirtualTableScan extends AbstractReadRel {
    *   <li>row shape must match field-list
    *   <li>row field types must match schema types
    * </ul>
+   *
+   * @throws IllegalArgumentException if any of these invariants is violated
    */
   @Value.Check
   protected void check() {
     List<String> names = getInitialSchema().names();
 
-    assert names.size()
-        == NamedFieldCountingTypeVisitor.countNames(this.getInitialSchema().struct());
-    List<Expression.NestedStruct> rows = getRows();
-
-    // At the PROTOBUF layer, the Nested.Struct message does not carry nullability information.
-    // Nullability is attached to the Nested message, which can contain a Nested.Struct.
-    // The NestedStruct POJO flattens the Nested and Nested.Struct messages together, allowing the
-    // nullability of a NestedStruct to be set directly.
-    //
-    // HOWEVER, the VirtualTable message contains a list of Nested.Struct messages, and as such
-    // the nullability cannot be set at the protobuf layer. To avoid users attaching meaningless
-    // nullability information in the POJOs, we restrict the nullability of NestedStructs to false
-    // when used in VirtualTableScans.
-    for (Expression.NestedStruct row : rows) {
-      assert !row.nullable();
+    int schemaNameCount =
+        NamedFieldCountingTypeVisitor.countNames(this.getInitialSchema().struct());
+    if (names.size() != schemaNameCount) {
+      throw new IllegalArgumentException(
+          String.format(
+              "VirtualTableScan schema names count (%d) does not match the depth-first named-field count of the schema struct (%d)",
+              names.size(), schemaNameCount));
+    }
+    if (names.stream().anyMatch(Objects::isNull)) {
+      throw new IllegalArgumentException("VirtualTableScan schema names must not contain null");
     }
 
-    assert names.stream().noneMatch(Objects::isNull)
-        && rows.stream().noneMatch(Objects::isNull)
-        && rows.stream()
-            .allMatch(r -> NamedFieldCountingTypeVisitor.countNames(r.getType()) == names.size());
+    List<Expression.NestedStruct> rows = getRows();
+    if (rows.stream().anyMatch(Objects::isNull)) {
+      throw new IllegalArgumentException("VirtualTableScan rows must not contain null");
+    }
 
     for (Expression.NestedStruct row : rows) {
+      // At the PROTOBUF layer, the Nested.Struct message does not carry nullability information.
+      // Nullability is attached to the Nested message, which can contain a Nested.Struct.
+      // The NestedStruct POJO flattens the Nested and Nested.Struct messages together, allowing
+      // the nullability of a NestedStruct to be set directly.
+      //
+      // HOWEVER, the VirtualTable message contains a list of Nested.Struct messages, and as such
+      // the nullability cannot be set at the protobuf layer. To avoid users attaching meaningless
+      // nullability information in the POJOs, we restrict the nullability of NestedStructs to
+      // false when used in VirtualTableScans.
+      if (row.nullable()) {
+        throw new IllegalArgumentException(
+            "VirtualTableScan rows must not be nullable; nullability cannot be represented for the Nested.Struct messages of a VirtualTable");
+      }
+
+      int rowNameCount = NamedFieldCountingTypeVisitor.countNames(row.getType());
+      if (rowNameCount != names.size()) {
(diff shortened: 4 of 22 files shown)


isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
```diff
@@ -734,17 +732,28 @@ public RelNode visit(NamedWrite write, Context context) {
...
+  }
+
+  /**
+   * Returns the {@link RelOptSchema} backing the {@link org.apache.calcite.tools.RelBuilder} used
+   * by this converter.
+   *
+   * @return the Calcite catalog to resolve table names against
+   * @throws IllegalStateException if the RelBuilder was created without a RelOptSchema
+   */
+  private RelOptSchema requireRelOptSchema() {
```
alexandrefimov: I don’t think this check is quite enough yet. It verifies that the `RelBuilder` has a `RelOptSchema`, but both callers ultimately need a `Prepare.CatalogReader`.

In the UPDATE path, the reader comes from `table.getRelOptSchema()`, which Calcite marks `@Nullable`. After dropping the old null assertion, we can still build a `LogicalTableModify` with a null catalog reader and get an NPE later from `getExpectedInputRowType()`. In the write path, a non-`CatalogReader` implementation will fail at the cast with `ClassCastException`, rather than the `IllegalStateException` this PR is aiming for.

Could we validate the actual value as a `Prepare.CatalogReader` before creating `LogicalTableModify`, and add a small test for the null and wrong-type cases?

nielspardon: You're right on both counts, and the null one is worse than I assumed — thanks for catching it.

I dropped the `catalogReader != null` assert because the issue's table listed it as a dead check on "a value just assigned from a cast". That was wrong: the value comes from `table.getRelOptSchema()`, which Calcite declares `@Nullable`, and `(Prepare.CatalogReader) null` casts happily. It's also a *different* schema from the `relBuilder.getRelOptSchema()` that `requireRelOptSchema()` validates, so that helper never covered it. `TableModify` then stores the reader with no null check (the field is non-`@Nullable`, unlike the ones right below it) and only dereferences it in `getExpectedInputRowType()`, exactly as you describe.

Fixed in 45780f29 with a `requireCatalogReader` helper used at both sites. It narrows via `instanceof`, so a single check covers null and wrong-type, and the message names the class it actually got. The schema *sources* are unchanged — UPDATE still reads it from the resolved table, the write path from the `RelBuilder` — so this is validation only.

Tests are in `CatalogReaderValidationTest`: a `RelOptSchema` that resolves tables through a real catalog but deliberately isn't a `CatalogReader`, handing out tables whose `getRelOptSchema()` is either that schema or null. Three cases: write path wrong-type, update path null, update path wrong-type. I checked they fail without the fix, and the null case is the interesting one:

```
namedUpdateRejectsTableSchemaThatIsNotACatalogReader() FAILED
    expected: <IllegalStateException> but was: <ClassCastException>
namedWriteRejectsSchemaThatIsNotACatalogReader() FAILED
    expected: <IllegalStateException> but was: <ClassCastException>
namedUpdateRejectsTableWithoutASchema() FAILED
    Expected java.lang.IllegalStateException to be thrown, but nothing was thrown.
```

That last line is the point: conversion completed and produced a `LogicalTableModify` carrying a null catalog reader, with the failure deferred to whoever touched it next.

alexandrefimov: Thanks for the detailed follow-up! The fix and tests look good to me.

andrew-coleman (commented):
One blocker: the `NestedList` check now rejects `SELECT ARRAY[nullable_col, notnull_col]`, which converts on main. Comments inline.

Three of the new checks are unreachable (`VirtualTableScan` null elements, `toLiteral`, `requireTable`) — the PR description cites them as motivating bugs.

Untested on the isthmus side: every new throw except `requireCatalogReader`.

core/src/main/java/io/substrait/expression/Expression.java
```diff
@@ -1789,13 +1789,28 @@ abstract class NestedList implements Nested {
...
+      List<Type> distinctTypes =
+          values().stream()
+              .map(Expression::getType)
+              .distinct()
+              .collect(java.util.stream.Collectors.toList());
+      if (distinctTypes.size() > 1) {
+        throw new IllegalArgumentException(
+            String.format(
+                "All values in NestedList must have the same type, found: %s", distinctTypes));
+      }
```
andrew-coleman: Rejects `SELECT ARRAY[nullable_col, notnull_col]`, which converts on main (the old `assert` was a no-op without `-ea`). Trigger is nullability, not type.

```suggestion
      List<Type> types =
          values().stream().map(Expression::getType).collect(java.util.stream.Collectors.toList());
      if (types.stream().map(TypeCreator::asNullable).distinct().limit(2).count() > 1) {
        throw new IllegalArgumentException(
            String.format("All values in NestedList must have the same type, found: %s", types));
      }
```

Needs a test with a nullability-only mismatch; `rejectNestedListWithElementsOfDifferentTypes` uses two different base types.

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
```diff
@@ -713,9 +713,8 @@ private RelNode handleCreateTableAs(NamedWrite namedWrite, Context context) {
   @Override
   public RelNode visit(NamedWrite write, Context context) {
     RelNode input = write.getInput().accept(this, context);
-    assert relBuilder.getRelOptSchema() != null;
-    final RelOptTable targetTable =
-        relBuilder.getRelOptSchema().getTableForMember(write.getNames());
+    final RelOptSchema relOptSchema = requireRelOptSchema();
+    final RelOptTable targetTable = relOptSchema.getTableForMember(write.getNames());
```
andrew-coleman, line 874: Pre-existing, not this PR — flagging for a separate issue. Line 715 converts the input, then `case CTAS` re-converts it in `handleCreateTableAs`. With a `rel_anchor` on the input that throws `UnsupportedOperationException: Duplicate rel_anchor=1`. Base behaves identically.

core/src/main/java/io/substrait/relation/VirtualTableScan.java
```diff
@@ -28,34 +28,53 @@ public abstract class VirtualTableScan extends AbstractReadRel {
...
+              names.size(), schemaNameCount));
+    }
+    if (names.stream().anyMatch(Objects::isNull)) {
+      throw new IllegalArgumentException("VirtualTableScan schema names must not contain null");
+    }
+
     List<Expression.NestedStruct> rows = getRows();
+    if (rows.stream().anyMatch(Objects::isNull)) {
+      throw new IllegalArgumentException("VirtualTableScan rows must not contain null");
+    }
```
andrew-coleman: Both null checks are unreachable: `ImmutableVirtualTableScan` (1202/1213/1236) and `ImmutableNamedStruct` (197/208/231) do `Objects.requireNonNull(element, ...)` on every add path.

```suggestion
    List<Expression.NestedStruct> rows = getRows();
```

core/src/main/java/io/substrait/relation/VirtualTableScan.java
```diff
@@ -64,26 +83,30 @@ protected void check() {
    * Validates that a row's field types conform to the table's schema.
    *
    * @param row the row to validate
-   * @throws AssertionError if the row does not conform to the schema
+   * @throws IllegalArgumentException if the row does not conform to the schema
    */
   private void validateRowConformsToSchema(Expression.NestedStruct row) {
     Type.Struct schemaStruct = getInitialSchema().struct();
     List<Type> schemaFieldTypes = schemaStruct.fields();
```
andrew-coleman: Called per row, so these two re-derive per row. Hoist above the loop in `check()` and pass them in.

```suggestion
  private void validateRowConformsToSchema(
      Expression.NestedStruct row, List<Type> schemaFieldTypes) {
```

core/src/main/java/io/substrait/expression/proto/ExpressionProtoConverter.java
```diff
@@ -477,7 +477,12 @@ public Expression visit(
 
   private Expression.Literal toLiteral(io.substrait.expression.Expression expression) {
     Expression e = toProto(expression);
-    assert e.getRexTypeCase() == Expression.RexTypeCase.LITERAL;
+    if (e.getRexTypeCase() != Expression.RexTypeCase.LITERAL) {
```
andrew-coleman, line 480: Unreachable: all four call sites are statically `Expression.Literal` (`Expression.java:927/1343`, `FieldReference.java:451`), and every `Literal` visit returns `lit(...)`. Fine as a guard, but the PR description presents it as a live bug.

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelVisitor.java
```diff
@@ -777,6 +778,7 @@ public Rel visit(org.apache.calcite.rel.core.Exchange exchange) {
    *
    * @param modify Calcite table modify node
    * @return Substrait write/update relation
+   * @throws IllegalArgumentException if the node has no target table.
    * @throws IllegalStateException if an update column is not found in the table schema.
```
andrew-coleman: `requireTable` can't fire — `TableModify`'s constructor dereferences `table` (`getRelOptSchema()` → `registerSchema`), so null NPEs at construction. Suggest dropping the tag and keeping the helper.

```suggestion
   * @throws IllegalStateException if an update column is not found in the table schema.
```

isthmus/src/main/java/io/substrait/isthmus/calcite/rel/CreateView.java
```diff
@@ -42,10 +42,14 @@ public RelWriter explainTerms(RelWriter pw) {
    * Returns the inputs to this node (single input).
    *
    * @return a list containing the input relation
+   * @throws IllegalArgumentException if given anything but exactly one input
```
andrew-coleman: This block documents `getInputs()`. `CreateTable.copy()` has the right text plus both `@param`s.

```suggestion
   * Copies this node with the given traits and input.
   *
   * @param traitSet the RelTraitSet
   * @param inputs List of RelNodes
   * @return a copy of this node with the given input
   * @throws IllegalArgumentException if given anything but exactly one input
```

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
```diff
@@ -734,19 +733,64 @@ public RelNode visit(NamedWrite write, Context context) {
...
   }
 
+  /**
+   * Returns the {@link RelOptSchema} backing the {@link org.apache.calcite.tools.RelBuilder} used
+   * by this converter.
+   *
+   * @return the Calcite catalog to resolve table names against
+   * @throws IllegalStateException if the RelBuilder was created without a RelOptSchema
+   */
+  private RelOptSchema requireRelOptSchema() {
```
andrew-coleman: Subclasses overriding `visit(NamedWrite)`/`visit(NamedUpdate)` can't reach these.

```suggestion
  protected RelOptSchema requireRelOptSchema() {
```

Same for `requireCatalogReader`.

isthmus/src/main/java/io/substrait/isthmus/expression/SqlMapValueConstructorCallConverter.java
```diff
@@ -53,8 +54,13 @@ private Optional<Expression> toMapLiteral(
         call.operands.stream()
             .map(t -> ((Expression.Literal) topLevelConverter.apply(t)))
             .collect(java.util.stream.Collectors.toList());
+    if (literals.size() % 2 != 0) {
+      throw new IllegalArgumentException(
+          String.format(
+              "SqlMapValueConstructor requires an even number of operands (key/value pairs), but got %d",
+              literals.size()));
+    }
```
andrew-coleman: The cast on line 55 throws `ClassCastException` before this check runs. Move it above line 53:

```java
    if (call.operands.size() % 2 != 0) {
          String.format(
              "SqlMapValueConstructor requires an even number of operands (key/value pairs), but got %d",
              call.operands.size()));
    }
```

Also line 63: `HashMap` reorders keys and silently drops duplicates — `LinkedHashMap` fixes the ordering.

build-logic/src/main/resources/substrait-pmd.xml
```diff
@@ -21,4 +21,22 @@
   <rule ref="category/java/bestpractices.xml/UnusedPrivateMethod" />
   <rule ref="category/java/design.xml/AvoidThrowingRawExceptionTypes" />
   <rule ref="category/java/errorprone.xml/MissingSerialVersionUID" />
+
+  <rule name="AvoidAssertStatement" language="java"
```
andrew-coleman, line 25: Please add `AvoidAssertStatement` to the rule list in `AGENTS.md` line 126-128.

isthmus/src/test/java/io/substrait/isthmus/CatalogReaderValidationTest.java
```diff
@@ -0,0 +1,259 @@
...
+    }
+
+    @Override
+    public void registerRules(RelOptPlanner planner) {
+      delegate.registerRules(planner);
+    }
+  }
+
+  /** A {@link RelOptTable} that delegates everything but the schema it claims to belong to. */
+  private static final class DelegatingRelOptTable implements RelOptTable {
```
andrew-coleman: `RelOptTableImpl.create(RelOptSchema, RelDataType, List<String>, Expression)` returns the schema you pass from `getRelOptSchema()` — replaces this class. Also `extend()`/`toRel()` currently forward to the undecorated `delegate`.

core/src/main/java/io/substrait/expression/proto/ExpressionProtoConverter.java
nielspardon: Corrected — the description now lists it with the other checks today's call sites cannot violate, rather than as a live bug.

andrew-coleman (approved):

PR #1059, opened 2026-08-04
feat(isthmus): observe window function return types

alexandrefimov:
## Summary

- report Calcite-inferred return types through `TypeObserver` during window-function conversion
- preserve the Substrait-supplied return type and skip extra inference for `TypeObserver.NOOP`
- allow callers to configure the observer through `ConverterProvider.Builder`

## Motivation

Scalar-function type observation was introduced in #1015. Issue #379 also identifies window functions as a conversion point where Substrait-supplied and Calcite-inferred types can diverge. This change extends the same observation side channel to window functions and makes it available without subclassing `ConverterProvider`.

## Validation

- `./gradlew :isthmus:spotlessCheck :isthmus:test --tests 'io.substrait.isthmus.ConverterProviderBuilderTest' --tests 'io.substrait.isthmus.SubstraitExpressionConverterTest' :isthmus:javadoc`

Related to #379.

### DIFF
--- a/isthmus/src/main/java/io/substrait/isthmus/expression/ExpressionRexConverter.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/expression/ExpressionRexConverter.java
@@ -39,6 +39,7 @@ import org.apache.calcite.rel.core.CorrelationId;
 import org.apache.calcite.rel.type.RelDataType;
 import org.apache.calcite.rel.type.RelDataTypeFactory;
 import org.apache.calcite.rex.RexBuilder;
+import org.apache.calcite.rex.RexCallBinding;
 import org.apache.calcite.rex.RexFieldCollation;
 import org.apache.calcite.rex.RexInputRef;
 import org.apache.calcite.rex.RexLambdaRef;
@@ -51,6 +52,7 @@ import org.apache.calcite.sql.SqlAggFunction;
 import org.apache.calcite.sql.SqlIntervalQualifier;
 import org.apache.calcite.sql.SqlKind;
 import org.apache.calcite.sql.SqlOperator;
+import org.apache.calcite.sql.SqlWindow;
 import org.apache.calcite.sql.fun.SqlStdOperatorTable;
 import org.apache.calcite.sql.parser.SqlParserPos;
 import org.apache.calcite.sql.type.SqlTypeName;
@@ -652,8 +654,37 @@ public class ExpressionRexConverter
     observeType(
         expr,
         TypeObservation.Source.WINDOW_FUNCTION,
-        () -> rexBuilder.deriveReturnType(operator, args));
+        () -> operator.inferReturnType(windowBinding(operator, args, lowerBound, upperBound)));
     return rexOver;
+  }
+
+  /**
+   * Builds the operand binding Calcite itself uses when inferring the return type of a windowed
+   * aggregate.
+   *
+   * <p>{@link RexBuilder#deriveReturnType(SqlOperator, java.util.List)} binds the operands with a
+   * plain {@link RexCallBinding}, whose {@link
+   * org.apache.calcite.sql.SqlOperatorBinding#hasEmptyGroup()} is always {@code false}. Calcite's
+   * validator instead derives that flag from the window bounds, and an operator whose return type
+   * strategy consults it widens its result to nullable over a frame that may be empty. Among the
+   * operators reachable here that is {@code FIRST_VALUE}, {@code LAST_VALUE} and {@code NTH_VALUE},
+   * which use {@code ARG0_NULLABLE_IF_EMPTY}; the aggregates usable in a window are unaffected
+   * because {@link io.substrait.isthmus.AggregateFunctions} already forces their results nullable.
+   * Binding without the flag would report a non-nullable inferred type for the former, so a genuine
+   * nullability deviation would be observed as a match.
+   */
+  private RexCallBinding windowBinding(
+      SqlOperator operator,
+      List<RexNode> operands,
+      RexWindowBound lowerBound,
+      RexWindowBound upperBound) {
+    boolean emptyGroup = !SqlWindow.isAlwaysNonEmpty(lowerBound, upperBound);
+    return new RexCallBinding(typeFactory, operator, operands, ImmutableList.of()) {
+      @Override
+      public boolean hasEmptyGroup() {
+        return emptyGroup;
+      }
+    };
   }
 
   private Set<SqlKind> asSqlKind(Expression.SortDirection direction) {
--- a/isthmus/src/main/java/io/substrait/isthmus/expression/ExpressionRexConverter.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/expression/ExpressionRexConverter.java
@@ -662,13 +662,11 @@ public class ExpressionRexConverter
    * <p>{@link RexBuilder#deriveReturnType(SqlOperator, java.util.List)} binds the operands with a
    * plain {@link RexCallBinding}, whose {@link
    * org.apache.calcite.sql.SqlOperatorBinding#hasEmptyGroup()} is always {@code false}. Calcite's
-   * validator instead derives that flag from the window bounds, and an operator whose return type
-   * strategy consults it widens its result to nullable over a frame that may be empty. Among the
-   * operators reachable here that is {@code FIRST_VALUE}, {@code LAST_VALUE} and {@code NTH_VALUE},
-   * which use {@code ARG0_NULLABLE_IF_EMPTY}; the aggregates usable in a window are unaffected
-   * because {@link io.substrait.isthmus.AggregateFunctions} already forces their results nullable.
-   * Binding without the flag would report a non-nullable inferred type for the former, so a genuine
-   * nullability deviation would be observed as a match.
+   * validator instead derives that flag from the window bounds, so any return type strategy that
+   * consults it (for example {@code ARG0_NULLABLE_IF_EMPTY} or {@code AGG_SUM}) widens its result
+   * to nullable over a frame that may be empty. Binding without the flag would report a
+   * non-nullable inferred type for such an operator, so a genuine nullability deviation would be
+   * observed as a match.
    */
   private RexCallBinding windowBinding(
       SqlOperator operator,
--- a/isthmus/src/main/java/io/substrait/isthmus/expression/ExpressionRexConverter.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/expression/ExpressionRexConverter.java
@@ -551,15 +551,10 @@ public class ExpressionRexConverter
 
     RelDataType returnType = typeConverter.toCalcite(typeFactory, expr.outputType());
     RexNode rexCall = rexBuilder.makeCall(returnType, operator, args);
-    // If type observations are not needed, avoid recomputing the RexCall with Calcite's
-    // independently inferred return type.
-    if (typeObserver == TypeObserver.NOOP) {
-      return rexCall;
-    }
     observeType(
         expr,
         TypeObservation.Source.SCALAR_FUNCTION,
-        () -> rexBuilder.makeCall(operator, args).getType());
+        () -> rexBuilder.deriveReturnType(operator, args));
     return rexCall;
   }
 
@@ -567,6 +562,11 @@ public class ExpressionRexConverter
       Expression expression,
       TypeObservation.Source source,
       Supplier<RelDataType> inferredTypeSupplier) {
+    // The conversion result never depends on the independently inferred type, so when no observer
+    // is installed skip Calcite's inference entirely.
+    if (typeObserver == TypeObserver.NOOP) {
+      return;
+    }
     TypeObservation observation;
     RelDataType inferredType;
     try {
@@ -648,9 +648,6 @@ public class ExpressionRexConverter
             nullWhenCountZero,
             distinct,
             ignoreNulls);
-    if (typeObserver == TypeObserver.NOOP) {
-      return rexOver;
-    }
     observeType(
         expr,
         TypeObservation.Source.WINDOW_FUNCTION,
--- a/isthmus/src/main/java/io/substrait/isthmus/expression/ExpressionRexConverter.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/expression/ExpressionRexConverter.java
@@ -554,26 +554,27 @@ public class ExpressionRexConverter
     if (typeObserver == TypeObserver.NOOP) {
       return rexCall;
     }
-    observeScalarType(expr, () -> rexBuilder.makeCall(operator, args));
+    observeType(
+        expr,
+        TypeObservation.Source.SCALAR_FUNCTION,
+        () -> rexBuilder.makeCall(operator, args).getType());
     return rexCall;
   }
 
-  private void observeScalarType(
-      Expression.ScalarFunctionInvocation expression, Supplier<RexNode> inferredCallSupplier) {
+  private void observeType(
+      Expression expression,
+      TypeObservation.Source source,
+      Supplier<RelDataType> inferredTypeSupplier) {
     TypeObservation observation;
-    RexNode inferredCall;
+    RelDataType inferredType;
     try {
-      inferredCall = inferredCallSupplier.get();
+      inferredType = inferredTypeSupplier.get();
     } catch (RuntimeException inferenceFailure) {
-      observation =
-          TypeObservation.failure(
-              TypeObservation.Source.SCALAR_FUNCTION, expression, inferenceFailure);
+      observation = TypeObservation.failure(source, expression, inferenceFailure);
       typeObserver.observe(observation);
       return;
     }
-    observation =
-        TypeObservation.success(
-            TypeObservation.Source.SCALAR_FUNCTION, expression, inferredCall.getType());
+    observation = TypeObservation.success(source, expression, inferredType);
     typeObserver.observe(observation);
   }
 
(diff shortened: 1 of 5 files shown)


alexandrefimov:
@vbarua this follows up on #1015 — it extends type observation to window functions (the second conversion point named in #379) and wires the observer into `ConverterProvider.Builder`, which is what you suggested after the merge. Would you have a chance to take a look?

isthmus/src/main/java/io/substrait/isthmus/expression/ExpressionRexConverter.java
```diff
@@ -631,19 +632,28 @@ public RexNode visit(Expression.WindowFunctionInvocation expr, Context context)
...
+            nullWhenCountZero,
+            distinct,
+            ignoreNulls);
+    if (typeObserver == TypeObserver.NOOP) {
+      return rexOver;
+    }
+    observeType(
+        expr,
+        TypeObservation.Source.WINDOW_FUNCTION,
+        () -> rexBuilder.deriveReturnType(operator, args));
```
vbarua: Claude did identify one potential bit of weirdness around nullability differences in certain cases. I still think this is useful overall for cases where the types are fully different, but it might not be reliable as-is to detect nullability deviations.

### Claude Summary
`rexBuilder.deriveReturnType(operator, args)` wraps args in a plain `RexCallBinding` where `hasEmptyGroup()` always returns false.

Calcite's actual window inference path (via `SqlOverOperator`) sets h`asEmptyGroup()` based on whether the window bounds guarantee non-emptiness. For SUM, MIN/MAX, AVG, and several other aggregate window functions, this flag controls result nullability.

Concrete example: `SUM(x BIGINT NOT NULL) OVER (...)` — the real Calcite inference returns BIGINT nullable; deriveReturnType returns BIGINT NOT NULL. The observation records the wrong inferred type, defeating the diagnostic purpose of the observer.

vbarua (commented):
Overall, the changes look reasonable.

There is one potential issue around nullability in the Calcite inferred output types. From a cursory look, I don't know how painful it would be to have the observation path follow the true Calcite inference path more closely. If it's not too bad, it would be worth doing in this PR. However, if it would require a fair bit of poking into Calcite internals to achieve, I would be happy with documenting the limitation for now and merging.

Let me know what you think @alexandrefimov

nielspardon (commented):
Nice, focused extension of the type-observation side channel from #1015 — the shape is right: observation stays advisory, the Substrait-supplied type still wins, and generalizing `observeScalarType` into `observeType(expr, source, Supplier<RelDataType>)` is the correct seam for aggregates to hang off later. Confirmed it earns its keep on the window path: a `sum:i32` window invocation reports `supplied=I64{nullable=true}` vs `inferred=INTEGER`, squarely the divergence #379 is about.

Two notes beyond the inline comments:

- **Aggregates remain unobserved**, and they're the one #379 conversion point this leaves open. `Aggregate.Measure` → `AggregateCall.create` happens in `SubstraitRelNodeConverter.fromMeasure`, entirely outside `ExpressionRexConverter`, so it needs a different hook rather than a third `observeType` call. Could you say in the description whether that's a deliberate follow-up? As written it reads as though #379's conversion points are now covered.
- **Please drop the "Validation" section from the description.** `AGENTS.md` calls out claims that CI-verified things pass as noise to leave out, and the description becomes the squash-merge commit message that `semantic-release` turns into the changelog. Summary and Motivation are exactly right.

Inline comments 1–3 are a single refactor split across three anchors — 1 and 3 delete the `NOOP` guard that 2 hoists into `observeType`, so they need to be applied together. I verified the whole set against `:isthmus:test`, `:isthmus:javadoc`, `:isthmus:pmdMain` and `:isthmus:spotlessCheck`.

isthmus/src/main/java/io/substrait/isthmus/expression/ExpressionRexConverter.java
```diff
@@ -554,26 +554,27 @@ public RexNode visit(Expression.ScalarFunctionInvocation expr, Context context)
     if (typeObserver == TypeObserver.NOOP) {
       return rexCall;
     }
-    observeScalarType(expr, () -> rexBuilder.makeCall(operator, args));
+    observeType(
+        expr,
+        TypeObservation.Source.SCALAR_FUNCTION,
+        () -> rexBuilder.makeCall(operator, args).getType());
```
nielspardon: **(1/3)** Two things, both consequences of the supplier now returning a `RelDataType`.

`rexBuilder.makeCall(operator, args).getType()` allocates a throwaway `RexCall` for nothing: in Calcite 1.42, `makeCall(op, exprs)` *is* `new RexCall(pos, deriveReturnType(op, exprs), op, exprs)` (`RexBuilder.java:303-324`), so calling `deriveReturnType` directly is equivalent and makes both sources visibly symmetric.

The `NOOP` guard belongs in `observeType` — see the next comment.

The comment on lines 552-553 is stale as well and should go along with this ("recomputing the RexCall" describes the pre-refactor supplier), but it sits outside the diff so I can't fold it into the suggestion.

```suggestion
    observeType(
        expr,
        TypeObservation.Source.SCALAR_FUNCTION,
        () -> rexBuilder.deriveReturnType(operator, args));
```

isthmus/src/main/java/io/substrait/isthmus/expression/ExpressionRexConverter.java
```diff
@@ -554,26 +554,27 @@ public RexNode visit(Expression.ScalarFunctionInvocation expr, Context context)
...
     return rexCall;
   }
 
-  private void observeScalarType(
-      Expression.ScalarFunctionInvocation expression, Supplier<RexNode> inferredCallSupplier) {
+  private void observeType(
+      Expression expression,
+      TypeObservation.Source source,
+      Supplier<RelDataType> inferredTypeSupplier) {
     TypeObservation observation;
```
nielspardon, line 570: **(2/3)** The `NOOP` short-circuit is the load-bearing part of this design — it's what keeps the default path free — but it's now duplicated at two call sites with the explanatory comment at only one. Hoisting it here states the invariant once, and means the guard can't be omitted when aggregates get wired up.

```suggestion
      Supplier<RelDataType> inferredTypeSupplier) {
    // The conversion result never depends on the independently inferred type, so when no observer
    // is installed skip Calcite's inference entirely.
    if (typeObserver == TypeObserver.NOOP) {
      return;
    }
    TypeObservation observation;
```

isthmus/src/main/java/io/substrait/isthmus/expression/ExpressionRexConverter.java
nielspardon: **(3/3)** Guard dropped here since the previous comment hoists it into `observeType`.

```suggestion
    observeType(
        expr,
        TypeObservation.Source.WINDOW_FUNCTION,
        () -> rexBuilder.deriveReturnType(operator, args));
```

isthmus/src/test/java/io/substrait/isthmus/SubstraitExpressionConverterTest.java
```diff
@@ -371,8 +412,64 @@ void useSubstraitReturnTypeDuringWindowFunctionConversion() {
             WindowBound.UNBOUNDED,
             sb.i32(42));
 
-    RexNode calciteExpr = expr.accept(expressionRexConverter, Context.newContext());
+    RexNode calciteExpr = expr.accept(observingConverter, Context.newContext());
+
+    assertEquals(SqlTypeName.INTEGER, observed.get().inferredType().orElseThrow().getSqlTypeName());
```
nielspardon: Asserting only `getSqlTypeName()` drops the more interesting half of what this test observes. `SqlLeadLagAggFunction` applies `FORCE_NULLABLE` when there's no third (default) operand, so with a single argument the inferred type is nullable `INTEGER` — pinning the whole `RelDataType` documents that, and matches how the sibling window tests assert.

```suggestion
    // LAG with no default operand forces a nullable return type.
    assertEquals(
        TypeConverter.DEFAULT.toCalcite(typeFactory, N.I32),
        observed.get().inferredType().orElseThrow());
```

isthmus/src/test/java/io/substrait/isthmus/SubstraitExpressionConverterTest.java
```diff
@@ -279,12 +280,7 @@ void observeVariadicConcatOnce() {
   void injectTypeObserverThroughConverterProvider() {
     AtomicReference<TypeObservation> observed = new AtomicReference<>();
     ConverterProvider observingProvider =
-        new ConverterProvider() {
-          @Override
-          public TypeObserver getTypeObserver() {
-            return observed::set;
-          }
-        };
+        ConverterProvider.builder().typeObserver(observed::set).build();
```
nielspardon, line 284: Converting this from the anonymous-subclass form to the builder form is right for the builder's own coverage, but it leaves the override path with no test at all — and `getTypeObserver()` still advertises overriding in its Javadoc.

That matters because correctness now rests on the single `getTypeObserver()` read in `ConverterProvider#getExpressionRexConverter`. Nothing else reads the `typeObserver` field directly, so it's correct as written, but a future refactor reading the field there would silently break every overriding subclass with nothing red to show it. Cheap insurance to keep both:

```java
  @Test
  void injectTypeObserverByOverridingConverterProvider() {
    AtomicReference<TypeObservation> observed = new AtomicReference<>();
    ConverterProvider observingProvider =
        new ConverterProvider() {
          @Override
          public TypeObserver getTypeObserver() {
            return observed::set;
          }
        };
    Expression.ScalarFunctionInvocation expr =
        sb.scalarFn(
            DefaultExtensionCatalog.FUNCTIONS_ARITHMETIC,
            "add:i32_i32",
            R.I32,
            sb.i32(7),
            sb.i32(42));
    Project query = sb.project(input -> List.of(expr), sb.emptyVirtualTableScan());

    new SubstraitToCalcite(observingProvider).convert(query);

    assertEquals(R.I32, observed.get().suppliedType());
  }
```

alexandrefimov:
@vbarua fixed rather than documented — it turned out cheap. Both pieces needed to swap that binding are public: `RexCallBinding`'s constructor, and `SqlWindow.isAlwaysNonEmpty(RexWindowBound, RexWindowBound)`, which takes exactly the bounds already built for `makeOver`. `hasEmptyGroup()` is now `!SqlWindow.isAlwaysNonEmpty(lowerBound, upperBound)` — the predicate `SqlOverOperator.deriveType` itself uses.

One correction: **`SUM` isn't affected here.** Isthmus substitutes `AggregateFunctions.SUM`, which forces `ARG0_FORCE_NULLABLE`, and likewise for `MIN`/`MAX`/`AVG` and the stddev/variance family, so the empty-group flag cannot change their result. (Correction from the thread below: that does not extend to every aggregate in `WINDOW_SIGS` — `COUNT` and `APPROX_COUNT_DISTINCT` are stock `SqlStdOperatorTable` operators and `SUM0` forces non-null `BIGINT`. They are unaffected too, but because their return types never consult the flag.) The real cases are `first_value`/`last_value`/`nth_value`, which map to stock Calcite operators on `ARG0_NULLABLE_IF_EMPTY`: `first_value(x)` over `ROWS BETWEEN 1 PRECEDING AND 1 PRECEDING` inferred `INTEGER NOT NULL` before, `INTEGER` now. Both directions tested; the possibly-empty one fails without the change.

The predicate inherits Calcite's conservatism — offset-bounded `RexWindowBounds` leave `getOrderKey()` at -1, so any row-offset frame counts as possibly-empty. That's Calcite's own answer, which is what we want to observe.

---

@nielspardon thanks — (1/3), (2/3) and both test comments applied. The guard now lives in `observeType` alone, the scalar supplier is `deriveReturnType` with the stale comment above it gone, and the override path has its own test again.

(3/3) I've taken the guard removal but kept the window binding: `deriveReturnType` is precisely what can't be used here, since its `RexCallBinding` hardcodes `hasEmptyGroup()` to false. The two sources are asymmetric on purpose now — a scalar call has no frame, a window call does — so the reason lives in a javadoc on `windowBinding` rather than at the call site.

On the LAG test: confirmed `SqlLeadLagAggFunction.transformType` applies `FORCE_NULLABLE` whenever `getOperandCount() < 3`, independent of the empty-group flag, so `N.I32` holds either way.

isthmus/src/main/java/io/substrait/isthmus/expression/ExpressionRexConverter.java
alexandrefimov: Fixed rather than documented — `hasEmptyGroup()` now comes from `SqlWindow.isAlwaysNonEmpty(lowerBound, upperBound)`, the same predicate `SqlOverOperator.deriveType` uses. One correction to the example: `SUM` is not affected here, since Isthmus substitutes `AggregateFunctions.SUM` with a forced-nullable return type. The reachable cases are `first_value`/`last_value`/`nth_value`.

isthmus/src/main/java/io/substrait/isthmus/expression/ExpressionRexConverter.java
alexandrefimov: Applied, and the stale comment above it is gone with it.

isthmus/src/main/java/io/substrait/isthmus/expression/ExpressionRexConverter.java
alexandrefimov: Applied — the guard lives in `observeType` only now.

isthmus/src/main/java/io/substrait/isthmus/expression/ExpressionRexConverter.java
alexandrefimov: Took the guard removal, kept the window binding: `deriveReturnType` hardcodes `hasEmptyGroup()` to false, which is exactly the deviation this PR exists to observe. The two suppliers are asymmetric on purpose now, with the reason in a javadoc on `windowBinding`. Leaving this thread open for your call.

isthmus/src/test/java/io/substrait/isthmus/SubstraitExpressionConverterTest.java
alexandrefimov: Applied. Confirmed `SqlLeadLagAggFunction.transformType` forces nullable whenever `getOperandCount() < 3`, independent of the empty-group flag, so `N.I32` holds either way.

isthmus/src/test/java/io/substrait/isthmus/SubstraitExpressionConverterTest.java
alexandrefimov: Added back as `injectTypeObserverByOverridingConverterProvider`, alongside the builder one.

isthmus/src/main/java/io/substrait/isthmus/expression/ExpressionRexConverter.java
```diff
@@ -631,19 +634,54 @@ public RexNode visit(Expression.WindowFunctionInvocation expr, Context context)
...
+   * <p>{@link RexBuilder#deriveReturnType(SqlOperator, java.util.List)} binds the operands with a
+   * plain {@link RexCallBinding}, whose {@link
+   * org.apache.calcite.sql.SqlOperatorBinding#hasEmptyGroup()} is always {@code false}. Calcite's
+   * validator instead derives that flag from the window bounds, and an operator whose return type
+   * strategy consults it widens its result to nullable over a frame that may be empty. Among the
+   * operators reachable here that is {@code FIRST_VALUE}, {@code LAST_VALUE} and {@code NTH_VALUE},
+   * which use {@code ARG0_NULLABLE_IF_EMPTY}; the aggregates usable in a window are unaffected
+   * because {@link io.substrait.isthmus.AggregateFunctions} already forces their results nullable.
+   * Binding without the flag would report a non-nullable inferred type for the former, so a genuine
+   * nullability deviation would be observed as a match.
```
nielspardon: The conclusion is right, but both membership claims are pinned to `WINDOW_SIGS` / `AGGREGATE_SIGS`, so they will rot as those grow — and two are already off: `COUNT` and `APPROX_COUNT_DISTINCT` are plain `SqlStdOperatorTable` operators rather than `AggregateFunctions` ones, and `AggregateFunctions.SUM0` forces non-null `BIGINT` rather than nullable. All three are unaffected, just not for the stated reason.

Since the binding is operator-agnostic and the two new tests already pin the behavior, the census is not load-bearing. Stating the mechanism against Calcite's contract instead cannot go stale as the mappings grow — it only moves on a Calcite bump:

```suggestion
   * validator instead derives that flag from the window bounds, so any return type strategy that
   * consults it (for example {@code ARG0_NULLABLE_IF_EMPTY} or {@code AGG_SUM}) widens its result
   * to nullable over a frame that may be empty. Binding without the flag would report a
   * non-nullable inferred type for such an operator, so a genuine nullability deviation would be
   * observed as a match.
```

`AGG_SUM` is worth naming because it also consults the flag, which shows the fix is broader than `ARG0_NULLABLE_IF_EMPTY` even though the Isthmus substitutions make that moot today.

nielspardon (commented):
All five threads addressed. Verified the new `windowBinding` against Calcite 1.42.0 source — it mirrors `SqlOverOperator.deriveType`'s `hasEmptyGroup()` override exactly, and that is the only flag that can diverge here. One nit on the new Javadoc.

isthmus/src/main/java/io/substrait/isthmus/expression/ExpressionRexConverter.java
nielspardon: Keep it. Scalar functions have no frame, so there is nothing to bind — symmetry would mean inventing an empty-group notion that does not exist, and Calcite exposes no public "deriveReturnType with an empty-group override".

Confirmed the binding matches `deriveReturnType` exactly apart from the override (same constructor, same empty collations), and that `hasEmptyGroup()` is the complete lever: `getGroupCount()` has no consumers in `sql/type/` or `sql/fun/`, and the `hasFilter()` half of `ARG0_NULLABLE_IF_EMPTY` is legitimately false since `WindowFunctionInvocation` has no `filter()`.

isthmus/src/main/java/io/substrait/isthmus/expression/ExpressionRexConverter.java
alexandrefimov: Applied. Both corrections hold — `AGGREGATE_SIGS` maps `count` and `approx_count_distinct` straight to `SqlStdOperatorTable`, and `SUM0` infers `ReturnTypes.BIGINT`, so "AggregateFunctions forces them nullable" was wrong for three of the entries I was counting. The same overreach is in my summary comment above; the conclusion stands but the reason does not.

Naming the strategies is the better anchor anyway, and `AGG_SUM` is worth having there for the reason you give.

isthmus/src/main/java/io/substrait/isthmus/expression/ExpressionRexConverter.java
alexandrefimov: Thanks for checking `getGroupCount()` and the `hasFilter()` half — that settles that `hasEmptyGroup()` is the whole lever. Resolving.

nielspardon (approved):
LGTM, though I guess we want to confirmation from @vbarua on his feedback

vbarua (approved):
Thanks for the updates!

PR #1060, opened 2026-08-05
feat(isthmus-cli): explain common SQL conversion errors instead of dumping stack traces

### LINKED ISSUE #113: [ISTHMUS] Add friendly tips for the CLI about using -c option to pass DDL SQL when parsing DML SQL
For beginners, if a user only parses a DML SQL via isthmus will meet the calcite exception, like this:

```
org.apache.calcite.runtime.CalciteContextException: From line 1, column 17 to line 1, column 31: Object 'XXX' not found
        at java.lang.reflect.Constructor.newInstanceWithCaller(Constructor.java:499)
        at java.lang.reflect.Constructor.newInstance(Constructor.java:480)
        at org.apache.calcite.runtime.Resources$ExInstWithCause.ex(Resources.java:505)
        at org.apache.calcite.sql.SqlUtil.newContextException(SqlUtil.java:932)
        at org.apache.calcite.sql.SqlUtil.newContextException(SqlUtil.java:917)
        at org.apache.calcite.sql.validate.SqlValidatorImpl.newValidationError(SqlValidatorImpl.java:5266)
        at org.apache.calcite.sql.validate.IdentifierNamespace.resolveImpl(IdentifierNamespace.java:183)
        at org.apache.calcite.sql.validate.IdentifierNamespace.validateImpl(IdentifierNamespace.java:188)
        at org.apache.calcite.sql.validate.AbstractNamespace.validate(AbstractNamespace.java:89)
        at org.apache.calcite.sql.validate.SqlValidatorImpl.validateNamespace(SqlValidatorImpl.java:1100)
        at org.apache.calcite.sql.validate.SqlValidatorImpl.validateQuery(SqlValidatorImpl.java:1071)
        at org.apache.calcite.sql.validate.SqlValidatorImpl.validateFrom(SqlValidatorImpl.java:3375)
        at org.apache.calcite.sql.validate.SqlValidatorImpl.validateSelect(SqlValidatorImpl.java:3639)
        at org.apache.calcite.sql.validate.SelectNamespace.validateImpl(SelectNamespace.java:64)
        at org.apache.calcite.sql.validate.AbstractNamespace.validate(AbstractNamespace.java:89)
        at org.apache.calcite.sql.validate.SqlValidatorImpl.validateNamespace(SqlValidatorImpl.java:1100)
        at org.apache.calcite.sql.validate.SqlValidatorImpl.validateQuery(SqlValidatorImpl.java:1071)
        at org.apache.calcite.sql.SqlSelect.validate(SqlSelect.java:247)
        at org.apache.calcite.sql.validate.SqlValidatorImpl.validateScopedExpression(SqlValidatorImpl.java:1046)
        at org.apache.calcite.sql.validate.SqlValidatorImpl.validate(SqlValidatorImpl.java:752)
        at org.apache.calcite.sql2rel.SqlToRelConverter.convertQuery(SqlToRelConverter.java:586)
        at io.substrait.isthmus.SqlToSubstrait.getBestExpRelRoot(SqlToSubstrait.java:124)
        at io.substrait.isthmus.SqlToSubstrait.lambda$sqlToRelNode$1(SqlToSubstrait.java:103)
        at java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:197)
        at java.util.AbstractList$RandomAccessSpliterator.forEachRemaining(AbstractList.java:720)
        at java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:509)
        at java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:499)
        at java.util.stream.ReduceOps$ReduceOp.evaluateSequential(ReduceOps.java:921)
        at java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234)
        at java.util.stream.ReferencePipeline.collect(ReferencePipeline.java:682)
        at io.substrait.isthmus.SqlToSubstrait.sqlToRelNode(SqlToSubstrait.java:104)
        at io.substrait.isthmus.SqlToSubstrait.executeInner(SqlToSubstrait.java:74)
        at io.substrait.isthmus.SqlToSubstrait.execute(SqlToSubstrait.java:43)
        at io.substrait.isthmus.PlanEntryPoint.call(PlanEntryPoint.java:59)
        at io.substrait.isthmus.PlanEntryPoint.call(PlanEntryPoint.java:16)
        at picocli.CommandLine.executeUserObject(CommandLine.java:1953)
        at picocli.CommandLine.access$1300(CommandLine.java:145)
        at picocli.CommandLine$RunLast.executeUserObjectOfLastSubcommandWithSameParent(CommandLine.java:2352)
        at picocli.CommandLine$RunLast.handle(CommandLine.java:2346)
        at picocli.CommandLine$RunLast.handle(CommandLine.java:2311)
        at picocli.CommandLine$AbstractParseResultHandler.execute(CommandLine.java:2179)
        at picocli.CommandLine.execute(CommandLine.java:2078)
        at io.substrait.isthmus.PlanEntryPoint.main(PlanEntryPoint.java:51)
Caused by: org.apache.calcite.sql.validate.SqlValidatorException: Object 'XXX' not found
        at java.lang.reflect.Constructor.newInstanceWithCaller(Constructor.java:499)
        at java.lang.reflect.Constructor.newInstance(Constructor.java:480)
        at org.apache.calcite.runtime.Resources$ExInstWithCause.ex(Resources.java:505)
        at org.apache.calcite.runtime.Resources$ExInst.ex(Resources.java:599)
        ... 40 more
```

So, IMO, it would be better to give a friendly tip when parsing DML SQL to let the users know it could give the definition via the `-c` option.

nielspardon:
`isthmus "SELECT * FROM foo"` answered a beginner's first attempt with a 40-line `CalciteContextException` stack trace that never mentioned `-c` / `--create`. Every other way to get the input wrong behaved the same way, up to an outright `NullPointerException` when no query was passed at all.

```
$ isthmus "SELECT * FROM foo"
Error: From line 1, column 15 to line 1, column 17: Object 'FOO' not found

Hint: table definitions are not part of the query. Pass a CREATE TABLE
statement for each table it references using -c / --create:

  isthmus -c "CREATE TABLE FOO (col1 INT, col2 VARCHAR)" "SELECT * FROM FOO"

Unquoted identifiers are upper-cased unless --unquotedcasing says otherwise.
```

## What changes

- **Failures caused by the input are reported as a message**, with a hint naming the option to reach for wherever the mistake is identifiable: an undefined table points at `-c` / `--create`, naming the table and the schema Calcite looked in; an unresolved column or an unresolved identifier in `-e` / `--expression` explains where columns come from; a `CREATE TABLE` given as the query, or a query given to `-c`, points at the other one; and a query swallowed by the greedy `-e` says so. Every hint that quotes an identifier back also notes that unquoted identifiers are upper-cased — unless `--unquotedcasing` was given, where the note would be both wrong and beside the point.
- **Anything that is not recognizably an input problem keeps its stack trace**, so defects still reach bug reports intact. A `SqlParseException` carrying no position did not come from the grammar, and counts as a defect rather than as a mistake in the SQL. The new `--stacktrace` prints the message *and* the full trace for the recognized failures.
- **Missing and contradictory input are usage errors.** With neither a query nor `-e` the CLI threw a `NullPointerException`; with both, it silently discarded the query. Both now print a message plus the usage and exit 2. `main` no longer parses arguments ahead of `execute()` either, so an unquoted or mistyped argument gets picocli's usage error instead of an `UnmatchedArgumentException` trace.
- **A `CREATE TABLE` with no column list no longer throws a `NullPointerException`.** Calcite's DDL grammar makes the column list and the `AS query` independently optional, so both `-c "CREATE TABLE foo"` and `-c "CREATE TABLE foo AS SELECT 1"` reached a `@NonNull` parameter with a null column list, on both of the parser's two paths — only one of which had a guard, and only against the CTAS half.
- **Syntax errors report their position in the native image.** Calcite assembles the expected-token list for a `SqlParseException` by reflectively calling three grammar productions on the parser, which the native image was never told to keep, so every syntax error came out as `RuntimeException: While building token lists` instead of as a position in the SQL. That was invisible while every error printed a trace anyway, and it is what made this feature a no-op for malformed SQL in the shipped binary.

The isthmus-cli readme now documents the three exit codes, and its `--help` block reflows beyond the added line because the new option widens picocli's option column.

Closes #113

BREAKING CHANGE: `isthmus` with no arguments reports the missing SQL on stderr and exits 2, where it used to print the usage on stdout and exit 0; `--help` still does the latter. Passing a query together with `-e` / `--expression` is now rejected instead of silently ignoring the query.

### DIFF
--- a/isthmus-cli/src/main/java/io/substrait/isthmus/cli/IsthmusEntryPoint.java
+++ b/isthmus-cli/src/main/java/io/substrait/isthmus/cli/IsthmusEntryPoint.java
@@ -17,8 +17,11 @@ import org.apache.calcite.avatica.util.Casing;
 import org.apache.calcite.prepare.Prepare;
 import picocli.CommandLine;
 import picocli.CommandLine.Command;
+import picocli.CommandLine.Model.CommandSpec;
 import picocli.CommandLine.Option;
+import picocli.CommandLine.ParameterException;
 import picocli.CommandLine.Parameters;
+import picocli.CommandLine.Spec;
 
 /** Isthmus CLI entry point. */
 @Command(
@@ -59,33 +62,57 @@ public class IsthmusEntryPoint implements Callable<Integer> {
       description = "Calcite's casing policy for unquoted identifiers: ${COMPLETION-CANDIDATES}")
   private Casing unquotedCasing = Casing.TO_UPPER;
 
+  @Option(
+      names = {"--stacktrace"},
+      description = "Print the full stack trace of any error, not just its message")
+  private boolean stackTrace;
+
+  @Spec private CommandSpec spec;
+
   /**
    * Standard Java Main method invoked by the isthmus CLI command.
    *
    * @param args Isthmus CLI arguments.
    */
   public static void main(String... args) {
+    CommandLine commandLine = createCommandLine();
+    if (args.length == 0) { // If no arguments print usage help
+      commandLine.usage(commandLine.getOut());
+      System.exit(CommandLine.ExitCode.OK);
+    }
+    System.exit(commandLine.execute(args));
+  }
+
+  /**
+   * Creates the {@link CommandLine} driving the isthmus CLI. Errors caused by the given SQL are
+   * reported by {@link IsthmusExecutionExceptionHandler} rather than as a stack trace.
+   *
+   * @return the configured {@link CommandLine}
+   */
+  static CommandLine createCommandLine() {
     CommandLine commandLine = new CommandLine(new IsthmusEntryPoint());
     commandLine.setCaseInsensitiveEnumValuesAllowed(true);
-    CommandLine.ParseResult parseResult = commandLine.parseArgs(args);
-    if (parseResult.originalArgs().isEmpty()) { // If no arguments print usage help
-      commandLine.usage(System.out);
-      System.exit(0);
-    }
-    if (commandLine.isUsageHelpRequested()) {
-      commandLine.usage(System.out);
-      System.exit(0);
-    }
-    if (commandLine.isVersionHelpRequested()) {
-      commandLine.printVersionHelp(System.out);
-      System.exit(0);
-    }
-    int exitCode = commandLine.execute(args);
-    System.exit(exitCode);
+    commandLine.setExecutionExceptionHandler(new IsthmusExecutionExceptionHandler());
+    return commandLine;
+  }
+
+  /**
+   * Reports whether the full stack trace of an error was asked for on the command line.
+   *
+   * @return true if {@code --stacktrace} was given
+   */
+  boolean isStackTraceRequested() {
+    return stackTrace;
   }
 
   @Override
   public Integer call() throws Exception {
+    if (sqlExpressions == null && sql == null) {
+      throw new ParameterException(
+          spec.commandLine(),
+          "Missing SQL to convert: pass a SQL query as the first argument, "
+              + "or SQL expressions with -e / --expression");
+    }
     ConverterProvider provider = ConverterProvider.builder().unquotedCasing(unquotedCasing).build();
     // Isthmus image is parsing SQL Expression if that argument is defined
     if (sqlExpressions != null) {
--- a/isthmus-cli/src/main/java/io/substrait/isthmus/cli/IsthmusEntryPoint.java
+++ b/isthmus-cli/src/main/java/io/substrait/isthmus/cli/IsthmusEntryPoint.java
@@ -64,7 +64,7 @@ public class IsthmusEntryPoint implements Callable<Integer> {
 
   @Option(
       names = {"--stacktrace"},
-      description = "Print the full stack trace of any error, not just its message")
+      description = "Print the full stack trace of a conversion failure, not just its message")
   private boolean stackTrace;
 
   @Spec private CommandSpec spec;
@@ -75,12 +75,7 @@ public class IsthmusEntryPoint implements Callable<Integer> {
    * @param args Isthmus CLI arguments.
    */
   public static void main(String... args) {
-    CommandLine commandLine = createCommandLine();
-    if (args.length == 0) { // If no arguments print usage help
-      commandLine.usage(commandLine.getOut());
-      System.exit(CommandLine.ExitCode.OK);
-    }
-    System.exit(commandLine.execute(args));
+    System.exit(createCommandLine().execute(args));
   }
 
   /**
@@ -96,15 +91,6 @@ public class IsthmusEntryPoint implements Callable<Integer> {
     return commandLine;
   }
 
-  /**
-   * Reports whether the full stack trace of an error was asked for on the command line.
-   *
-   * @return true if {@code --stacktrace} was given
-   */
-  boolean isStackTraceRequested() {
-    return stackTrace;
-  }
-
   @Override
   public Integer call() throws Exception {
     if (sqlExpressions == null && sql == null) {
@@ -112,6 +98,13 @@ public class IsthmusEntryPoint implements Callable<Integer> {
           spec.commandLine(),
           "Missing SQL to convert: pass a SQL query as the first argument, "
               + "or SQL expressions with -e / --expression");
+    }
+    if (sqlExpressions != null && sql != null) {
+      throw new ParameterException(
+          spec.commandLine(),
+          "Give either a SQL query or -e / --expression, not both: the query '"
+              + sql
+              + "' would be ignored");
     }
     ConverterProvider provider = ConverterProvider.builder().unquotedCasing(unquotedCasing).build();
     // Isthmus image is parsing SQL Expression if that argument is defined
--- a/isthmus-cli/src/main/java/io/substrait/isthmus/cli/IsthmusExecutionExceptionHandler.java
+++ b/isthmus-cli/src/main/java/io/substrait/isthmus/cli/IsthmusExecutionExceptionHandler.java
@@ -1,0 +1,165 @@
+package io.substrait.isthmus.cli;
+
+import java.io.PrintWriter;
+import java.util.Optional;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import org.apache.calcite.runtime.CalciteContextException;
+import org.apache.calcite.sql.parser.SqlParseException;
+import picocli.CommandLine;
+
+/**
+ * Reports mistakes in the SQL handed to the CLI as a short message, with a hint about the option to
+ * reach for whenever the mistake can be identified.
+ *
+ * <p>Only failures that are recognizably caused by the input are reported this way. Anything else
+ * is rethrown so that its stack trace is still printed, as is the full stack trace of a recognized
+ * failure when {@code --stacktrace} is given.
+ */
+class IsthmusExecutionExceptionHandler implements CommandLine.IExecutionExceptionHandler {
+
+  /** Matches Calcite's complaint about a table (or other object) that the catalog does not hold. */
+  private static final Pattern OBJECT_NOT_FOUND =
+      Pattern.compile("(?:Object|Table) '([^']+)' not found");
+
+  /** Matches Calcite's complaint about a column that none of the known tables holds. */
+  private static final Pattern COLUMN_NOT_FOUND = Pattern.compile("Column '([^']+)' not found");
+
+  /** Matches Calcite's complaint about an identifier that an expression cannot be resolved to. */
+  private static final Pattern UNKNOWN_IDENTIFIER = Pattern.compile("Unknown identifier '([^']+)'");
+
+  /** The message the CREATE statement parser reports for anything that is not a CREATE TABLE. */
+  private static final String NOT_A_CREATE_TABLE = "Not a valid CREATE TABLE statement.";
+
+  /** The message the CREATE statement parser reports for a CREATE TABLE AS SELECT. */
+  private static final String CTAS_NOT_SUPPORTED = "CTAS not supported.";
+
+  /** The message the DDL converter reports for a CREATE TABLE without a query. */
+  private static final String CTAS_ONLY = "Only create table as select statements are supported";
+
+  // The hints are hard-wrapped for a terminal rather than joined into single long lines.
+
+  private static final String CREATE_HINT =
+      """
+      Hint: table definitions are not part of the query. Pass a CREATE TABLE
+      statement for each table it references using -c / --create:
+
+        isthmus -c "CREATE TABLE %1$s (col1 INT, col2 VARCHAR)" "SELECT * FROM %1$s"
+
+      Unquoted identifiers are upper-cased unless --unquotedcasing says otherwise.""";
+
+  private static final String COLUMN_HINT =
+      """
+      Hint: '%s' is not a column of any table defined with -c / --create. Check
+      the column names in the CREATE TABLE statement; unquoted identifiers are
+      upper-cased unless --unquotedcasing says otherwise.""";
+
+  private static final String EXPRESSION_HINT =
+      """
+      Hint: identifiers in a -e / --expression must be columns of a table defined
+      with -c / --create:
+
+        isthmus -c "CREATE TABLE T (%1$s INT)" -e "%1$s + 1\"""";
+
+  private static final String QUERY_ARGUMENT_HINT =
+      """
+      Hint: -c / --create takes plain CREATE TABLE statements; the query itself is
+      the first argument:
+
+        isthmus -c "CREATE TABLE FOO (col1 INT)" "SELECT * FROM FOO\"""";
+
+  @Override
+  public int handleExecutionException(
+      Exception ex, CommandLine cmd, CommandLine.ParseResult parseResult) throws Exception {
+    if (!isInputError(ex)) {
+      // Not something the user can act on: keep the stack trace, it belongs in a bug report.
+      throw ex;
+    }
+
+    PrintWriter err = cmd.getErr();
+    err.println(cmd.getColorScheme().errorText("Error: " + ex.getMessage()));
+    Optional<String> hint = hint(ex);
+    if (hint.isPresent()) {
+      err.println();
+      err.println(hint.get());
+    }
+    err.flush();
+
+    if (stackTraceRequested(cmd)) {
+      throw ex;
+    }
+    return cmd.getCommandSpec().exitCodeOnExecutionException();
+  }
+
+  /**
+   * Reports whether the given exception was caused by the SQL given to the CLI rather than by a
+   * defect.
+   *
+   * @param ex the exception thrown while converting
+   * @return true if the exception describes a problem with the input
+   */
+  private static boolean isInputError(final Exception ex) {
+    // CalciteContextException is, by construction, a complaint about the SQL at a line and column.
+    return ex instanceof SqlParseException
+        || ex instanceof CalciteContextException
+        || isPlainCreateTableQuery(ex);
+  }
+
+  /**
+   * Reports whether the given exception is the DDL converter's rejection of a CREATE TABLE
+   * statement passed as the query.
+   *
+   * @param ex the exception thrown while converting
+   * @return true if a plain CREATE TABLE statement was given as the query
+   */
+  private static boolean isPlainCreateTableQuery(final Exception ex) {
+    return ex instanceof IllegalArgumentException && CTAS_ONLY.equals(ex.getMessage());
+  }
+
+  /**
+   * Returns the hint to print for the given exception, if its message identifies a mistake we can
+   * advise on.
+   *
+   * @param ex the exception thrown while converting
+   * @return the hint to print below the error message, or empty if the mistake is not recognized
+   */
+  private static Optional<String> hint(final Exception ex) {
+    if (isPlainCreateTableQuery(ex)) {
+      return Optional.of(QUERY_ARGUMENT_HINT);
+    }
+
+    String message = ex.getMessage();
+    if (message == null) {
+      return Optional.empty();
+    }
+    if (message.contains(NOT_A_CREATE_TABLE) || message.contains(CTAS_NOT_SUPPORTED)) {
+      return Optional.of(QUERY_ARGUMENT_HINT);
+    }
+
+    Matcher objectNotFound = OBJECT_NOT_FOUND.matcher(message);
+    if (objectNotFound.find()) {
+      return Optional.of(CREATE_HINT.formatted(objectNotFound.group(1)));
+    }
+    Matcher columnNotFound = COLUMN_NOT_FOUND.matcher(message);
+    if (columnNotFound.find()) {
+      return Optional.of(COLUMN_HINT.formatted(columnNotFound.group(1)));
+    }
+    Matcher unknownIdentifier = UNKNOWN_IDENTIFIER.matcher(message);
+    if (unknownIdentifier.find()) {
+      return Optional.of(EXPRESSION_HINT.formatted(unknownIdentifier.group(1)));
+    }
+    return Optional.empty();
+  }
+
+  /**
+   * Reports whether the full stack trace was asked for on the command line.
+   *
+   * @param cmd the command line that failed
+   * @return true if {@code --stacktrace} was given
+   */
+  private static boolean stackTraceRequested(final CommandLine cmd) {
+    Object command = cmd.getCommand();
+    return command instanceof IsthmusEntryPoint
+        && ((IsthmusEntryPoint) command).isStackTraceRequested();
+  }
+}
--- a/isthmus-cli/src/main/java/io/substrait/isthmus/cli/IsthmusExecutionExceptionHandler.java
+++ b/isthmus-cli/src/main/java/io/substrait/isthmus/cli/IsthmusExecutionExceptionHandler.java
@@ -18,9 +18,12 @@ import picocli.CommandLine;
  */
 class IsthmusExecutionExceptionHandler implements CommandLine.IExecutionExceptionHandler {
 
-  /** Matches Calcite's complaint about a table (or other object) that the catalog does not hold. */
+  /**
+   * Matches Calcite's complaint about a table (or other object) that the catalog does not hold. The
+   * second group holds the schema it looked in, for the qualified form of the message.
+   */
   private static final Pattern OBJECT_NOT_FOUND =
-      Pattern.compile("(?:Object|Table) '([^']+)' not found");
+      Pattern.compile("(?:Object|Table) '([^']+)' not found(?: within '([^']+)')?");
 
   /** Matches Calcite's complaint about a column that none of the known tables holds. */
   private static final Pattern COLUMN_NOT_FOUND = Pattern.compile("Column '([^']+)' not found");
@@ -28,11 +31,18 @@ class IsthmusExecutionExceptionHandler implements CommandLine.IExecutionExceptio
   /** Matches Calcite's complaint about an identifier that an expression cannot be resolved to. */
   private static final Pattern UNKNOWN_IDENTIFIER = Pattern.compile("Unknown identifier '([^']+)'");
 
+  /** Matches Calcite's complaint about a query given where an expression was expected. */
+  private static final Pattern STATEMENT_AS_EXPRESSION =
+      Pattern.compile("Incorrect syntax near the keyword '(?:SELECT|WITH|VALUES|TABLE)'");
+
   /** The message the CREATE statement parser reports for anything that is not a CREATE TABLE. */
   private static final String NOT_A_CREATE_TABLE = "Not a valid CREATE TABLE statement.";
 
   /** The message the CREATE statement parser reports for a CREATE TABLE AS SELECT. */
   private static final String CTAS_NOT_SUPPORTED = "CTAS not supported.";
+
+  /** The message the CREATE statement parser reports for a CREATE TABLE without a column list. */
+  private static final String COLUMNS_REQUIRED = "Column definitions are required.";
 
   /** The message the DDL converter reports for a CREATE TABLE without a query. */
   private static final String CTAS_ONLY = "Only create table as select statements are supported";
@@ -44,15 +54,19 @@ class IsthmusExecutionExceptionHandler implements CommandLine.IExecutionExceptio
       Hint: table definitions are not part of the query. Pass a CREATE TABLE
       statement for each table it references using -c / --create:
 
-        isthmus -c "CREATE TABLE %1$s (col1 INT, col2 VARCHAR)" "SELECT * FROM %1$s"
-
-      Unquoted identifiers are upper-cased unless --unquotedcasing says otherwise.""";
+        isthmus -c "CREATE TABLE %1$s (col1 INT, col2 VARCHAR)" "SELECT * FROM %1$s\"""";
 
   private static final String COLUMN_HINT =
       """
-      Hint: '%s' is not a column of any table defined with -c / --create. Check
-      the column names in the CREATE TABLE statement; unquoted identifiers are
-      upper-cased unless --unquotedcasing says otherwise.""";
+      Hint: '%1$s' is not a column of any table defined with -c / --create. Check
+      the column names in the CREATE TABLE statement.""";
+
+  private static final String COLUMN_WITHOUT_CREATE_HINT =
+      """
+      Hint: '%1$s' is not a column of any table, and no table was defined. Pass a
+      CREATE TABLE statement for each table the query reads using -c / --create:
+
+        isthmus -c "CREATE TABLE T (%1$s INT)" "SELECT %1$s FROM T\"""";
 
   private static final String EXPRESSION_HINT =
       """
@@ -61,12 +75,29 @@ class IsthmusExecutionExceptionHandler implements CommandLine.IExecutionExceptio
 
         isthmus -c "CREATE TABLE T (%1$s INT)" -e "%1$s + 1\"""";
 
+  private static final String EXPRESSION_ARGUMENT_HINT =
+      """
+      Hint: -e / --expression consumes every following argument, so a query
+      written after it is parsed as an expression. Pass the query on its own:
+
+        isthmus "SELECT * FROM FOO\"""";
+
   private static final String QUERY_ARGUMENT_HINT =
       """
       Hint: -c / --create takes plain CREATE TABLE statements; the query itself is
       the first argument:
 
         isthmus -c "CREATE TABLE FOO (col1 INT)" "SELECT * FROM FOO\"""";
+
+  private static final String COLUMN_LIST_HINT =
+      """
+      Hint: -c / --create needs the columns of the table, with their types:
+
+        isthmus -c "CREATE TABLE FOO (col1 INT, col2 VARCHAR)" "SELECT * FROM FOO\"""";
+
+  /** Appended to the hints whose advice depends on how unquoted identifiers are cased. */
+  private static final String CASING_NOTE =
+      "\n\nUnquoted identifiers are upper-cased unless --unquotedcasing says otherwise.";
 
   @Override
   public int handleExecutionException(
@@ -78,14 +109,14 @@ class IsthmusExecutionExceptionHandler implements CommandLine.IExecutionExceptio
 
     PrintWriter err = cmd.getErr();
     err.println(cmd.getColorScheme().errorText("Error: " + ex.getMessage()));
-    Optional<String> hint = hint(ex);
+    Optional<String> hint = hint(ex, parseResult);
     if (hint.isPresent()) {
       err.println();
       err.println(hint.get());
     }
     err.flush();
 
-    if (stackTraceRequested(cmd)) {
+    if (parseResult.hasMatchedOption("--stacktrace")) {
       throw ex;
     }
     return cmd.getCommandSpec().exitCodeOnExecutionException();
@@ -99,10 +130,15 @@ class IsthmusExecutionExceptionHandler implements CommandLine.IExecutionExceptio
    * @return true if the exception describes a problem with the input
    */
   private static boolean isInputError(final Exception ex) {
+    if (ex instanceof SqlParseException) {
+      // Calcite funnels whatever the parser threw into a SqlParseException, leaving the position
+      // null unless the cause was a grammar or lexer error
+      // (SqlAbstractParserImpl.convertException). A missing position therefore means a defect
+      // rather than a problem with the SQL.
+      return ((SqlParseException) ex).getPos() != null;
+    }
     // CalciteContextException is, by construction, a complaint about the SQL at a line and column.
-    return ex instanceof SqlParseException
-        || ex instanceof CalciteContextException
-        || isPlainCreateTableQuery(ex);
+    return ex instanceof CalciteContextException || isPlainCreateTableQuery(ex);
   }
 
   /**
@@ -117,13 +153,15 @@ class IsthmusExecutionExceptionHandler implements CommandLine.IExecutionExceptio
   }
 
   /**
-   * Returns the hint to print for the given exception, if its message identifies a mistake we can
-   * advise on.
+   * Returns the hint to print for the given exception, if the message and the options it was given
+   * identify a mistake we can advise on.
    *
    * @param ex the exception thrown while converting
+   * @param parseResult the parsed command line the failure came from
    * @return the hint to print below the error message, or empty if the mistake is not recognized
    */
-  private static Optional<String> hint(final Exception ex) {
+  private static Optional<String> hint(
+      final Exception ex, final CommandLine.ParseResult parseResult) {
     if (isPlainCreateTableQuery(ex)) {
       return Optional.of(QUERY_ARGUMENT_HINT);
     }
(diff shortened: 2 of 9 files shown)


isthmus/src/main/java/io/substrait/isthmus/sql/SubstraitCreateStatementParser.java
```diff
@@ -215,6 +215,11 @@ private static CalciteSchema processCreateStatementsToSchema(
         }
 
         final SqlCreateTable create = (SqlCreateTable) parsed;
+
+        if (create.query != null) {
+          throw fail("CTAS not supported.", create.name.getParserPosition());
+        }
```
andrew-coleman: The NPE this guard was added to remove is still reachable. Calcite's DDL grammar makes *both* `TableElementList()` and `AS query` optional, so a bare `CREATE TABLE FOO` parses to a `SqlCreateTable` with `columnList == null` **and** `query == null` — straight past this check into `createSubstraitTable(..., create.columnList)` on line 235, whose loop at 266 iterates it:

```
$ isthmus -c "CREATE TABLE FOO" "SELECT * FROM FOO"
java.lang.NullPointerException: Cannot invoke "org.apache.calcite.sql.SqlNodeList.iterator()" because "columnList" is null
	at io.substrait.isthmus.sql.SubstraitCreateStatementParser.createSubstraitTable(SubstraitCreateStatementParser.java:266)
	at io.substrait.isthmus.sql.SubstraitCreateStatementParser.processCreateStatementsToSchema(SubstraitCreateStatementParser.java:234)
```

Exit 1 with the full trace, unrecognised by the new handler — exactly what this PR set out to remove. `CREATE TABLE IF NOT EXISTS FOO` and `CREATE OR REPLACE TABLE FOO` do the same.

Note that `columnList == null` is *not* a replacement for the `query != null` check: `CREATE TABLE FOO (a INT) AS SELECT 1` parses with both fields present, and dropping the query check would make that silently create `FOO(a INT)` and discard the query. Both are needed:

```suggestion
        if (create.query != null) {
          throw fail("CTAS not supported.", create.name.getParserPosition());
        }

        if (create.columnList == null) {
          throw fail("Column definitions are required.", create.name.getParserPosition());
        }
```

The sibling loop in `processCreateStatements` has the identical hole — `isthmus -c "CREATE TABLE FOO" -e "1"` NPEs at line 92 via `SqlExpressionToSubstrait.registerCreateTablesForExtendedExpression`. It's outside the diff so I can't attach a suggestion there, but lines 87-89 need the same pair:

```java
      if (create.query != null) {
        throw fail("CTAS not supported.", create.name.getParserPosition());
      }

      if (create.columnList == null) {
        throw fail("Column definitions are required.", create.name.getParserPosition());
      }
```

Three checks are now duplicated between the two loops (and only the first also checks `create.name.names.size() > 1`), which is precisely why this gap is absent from both. A shared helper is probably the better landing spot:

```java
  private static void validateCreateTable(@NonNull final SqlCreateTable create)
      throws SqlParseException {
    if (create.query != null) {
      throw fail("CTAS not supported.", create.name.getParserPosition());
    }

    if (create.columnList == null) {
      throw fail("Column definitions are required.", create.name.getParserPosition());
    }
  }
```

Either shape gives `Error: Column definitions are required.`, exit 1, no trace — verified on both paths.

andrew-coleman (commented):
The direction here is right and the hint text reads well. Three things I think need to change before this lands, plus a set of smaller ones inline.

**The NPE is still reachable.** Calcite's grammar makes the column list optional independently of `AS query`, so `isthmus -c "CREATE TABLE FOO" "SELECT * FROM FOO"` still dies with the `columnList` NPE and a full stack trace — on both the query and the `-e` path.

**Internal defects are now silently swallowed.** `SqlParser` wraps whatever the parser threw into a `SqlParseException`, so `isthmus ""` prints `Error: Index 0 out of bounds for length 0` with no trace and no hint, where it used to print a trace that located the bug. That contradicts the promise the README adds in this PR.

**`-e` swallows the positional query.** `isthmus -e "1 + 1" "SELECT 1"` parses the query as an expression, and `isthmus "SELECT 1" -e "1 + 1"` exits 0 having silently discarded the query.

Several hints also fire for the wrong command line — they are selected purely from message text while `parseResult` is passed in and never consulted, so `-e` advice appears when `-e` was not used, the `--unquotedcasing` note appears when the casing was explicitly chosen, and the column hint tells you to check a CREATE TABLE statement that may not exist.

<sub>How this was checked: built `installDist` at `da2e943e` and reproduced each case from the CLI, then applied the suggested fixes and re-ran. With all of them applied, `:isthmus:test` is 961/961 and `:isthmus-cli:test` is 21/21 (up from 11), with `pmdMain`, `spotlessCheck` and `javadoc` clean. Every suggestion block below comes from that tree, so they should apply as-is. The suggestions on `IsthmusExecutionExceptionHandler` touch overlapping methods (`hint` gains a `parseResult` parameter) and are easiest to take together. The one thing I did not run end to end is the `--stacktrace` addition to `smoke.sh`, since that needs a native build.</sub>

isthmus-cli/src/main/java/io/substrait/isthmus/cli/IsthmusExecutionExceptionHandler.java
```diff
@@ -0,0 +1,165 @@
...
+   *
+   * @param ex the exception thrown while converting
+   * @return true if the exception describes a problem with the input
+   */
+  private static boolean isInputError(final Exception ex) {
+    // CalciteContextException is, by construction, a complaint about the SQL at a line and column.
+    return ex instanceof SqlParseException
+        || ex instanceof CalciteContextException
+        || isPlainCreateTableQuery(ex);
+  }
```
andrew-coleman, line 142: Treating every `SqlParseException` as an input error swallows genuine defects. `SqlParser.parseStmtList` funnels whatever the parser threw into a `SqlParseException` carrying the raw message, so:

```
$ isthmus ""
Error: Index 0 out of bounds for length 0
```

That is an `ArrayIndexOutOfBoundsException` from inside the parser, reported as if the user had written bad SQL — no hint, no trace, nothing to act on. `isthmus -c "" "SELECT 1"` does the same. Before this PR the same input printed a trace that located it, so this is a regression in the opposite direction, and it contradicts the README line the PR adds: *"Anything that is not a recognizable problem with the input is always reported with its stack trace."*

Calcite gives a clean discriminator. `SqlAbstractParserImpl.convertException` sets the position only when the cause was a `ParseException` or `TokenMgrError`, and leaves it null otherwise:

```
[]               pos=null              cause=java.lang.ArrayIndexOutOfBoundsException
[SELECT 1 FROM]  pos=line 1, column 13 cause=org.apache.calcite.sql.parser.ddl.ParseException
[SELECT (1]      pos=line 1, column 9  cause=org.apache.calcite.sql.parser.ddl.ParseException
```

`SubstraitCreateStatementParser.fail(...)` uses `SqlParserPos.ZERO`, so every "CTAS not supported." / "Not a valid CREATE TABLE statement." case stays recognised:

```suggestion
  private static boolean isInputError(final Exception ex) {
    if (ex instanceof SqlParseException) {
      // Calcite funnels whatever the parser threw into a SqlParseException, leaving the position
      // null unless the cause was a grammar or lexer error
      // (SqlAbstractParserImpl.convertException).
      // A missing position therefore means a defect rather than a problem with the SQL.
      return ((SqlParseException) ex).getPos() != null;
    }
    // CalciteContextException is, by construction, a complaint about the SQL at a line and column.
    return ex instanceof CalciteContextException || isPlainCreateTableQuery(ex);
  }
```

Separately, this only inspects the top-level exception. It happens to line up today because `SqlToSubstrait.convert` and `SqlExpressionToSubstrait.convert` propagate unwrapped, but the first layer that wraps will silently revert the whole category with no test failing.

isthmus-cli/src/main/java/io/substrait/isthmus/cli/IsthmusEntryPoint.java
```diff
@@ -59,33 +62,57 @@ enum OutputFormat {
...
   }
 
   @Override
   public Integer call() throws Exception {
+    if (sqlExpressions == null && sql == null) {
+      throw new ParameterException(
+          spec.commandLine(),
+          "Missing SQL to convert: pass a SQL query as the first argument, "
+              + "or SQL expressions with -e / --expression");
+    }
```
andrew-coleman, line 101: This guard covers "neither given", but the two adjacent mistakes are still silent. `sqlExpressions` is `arity = "1..*"`, so it greedily eats the positional:

```
$ isthmus -e "1 + 1" "SELECT 1"
Error: Incorrect syntax near the keyword 'SELECT' at line 1, column 1.
```

The query became a second *expression*, and nothing points at why. In the other order the query is discarded without a word:

```
$ isthmus "SELECT 1" -e "1 + 1"      # exit 0, converts only the expression
```

`--` does not help either — `isthmus -e "1 + 1" -- "SELECT 1"` also exits 0 having converted only the expression. This is the right place for the mutual-exclusion check:

```suggestion
    if (sqlExpressions == null && sql == null) {
      throw new ParameterException(
          spec.commandLine(),
          "Missing SQL to convert: pass a SQL query as the first argument, "
              + "or SQL expressions with -e / --expression");
    }
    if (sqlExpressions != null && sql != null) {
      throw new ParameterException(
          spec.commandLine(),
          "Give either a SQL query or -e / --expression, not both: the query '"
              + sql
              + "' would be ignored");
    }
```

That fixes the discard case (exit 2 with `...the query 'SELECT 1' would be ignored`) but not the greedy one, where `sql` is already null by the time `call()` runs. Since the ambiguity is inherent to a variable-arity option sitting next to a positional, the cheapest cure is a hint in the handler:

```java
  private static final Pattern STATEMENT_AS_EXPRESSION =
      Pattern.compile("Incorrect syntax near the keyword '(?:SELECT|WITH|VALUES|TABLE)'");

  private static final String EXPRESSION_ARGUMENT_HINT =
      """
      Hint: -e / --expression consumes every following argument, so a query
      written after it is parsed as an expression. Pass the query on its own:

        isthmus "SELECT * FROM FOO\"""";
```

gated on `parseResult.hasMatchedOption("-e")` and checked before the identifier patterns. Verified working on both cases.

isthmus-cli/src/main/java/io/substrait/isthmus/cli/IsthmusEntryPoint.java
```diff
@@ -59,33 +62,57 @@ enum OutputFormat {
       description = "Calcite's casing policy for unquoted identifiers: ${COMPLETION-CANDIDATES}")
   private Casing unquotedCasing = Casing.TO_UPPER;
 
+  @Option(
+      names = {"--stacktrace"},
+      description = "Print the full stack trace of any error, not just its message")
+  private boolean stackTrace;
```
andrew-coleman, line 68: `--stacktrace` is described as applying to "any error", but it is silently ignored for every usage error — including the one this PR adds. picocli routes `ParameterException`s to `getParameterExceptionHandler()`, which the PR does not override, and the catch range in `CommandLine.execute` covers the execution strategy, so even the one thrown from `call()` lands there:

```
$ isthmus --stacktrace -c "CREATE TABLE FOO(A INT)"
Missing SQL to convert: ...          # exit 2, message plus usage, no trace
```

A user asked to "re-run with `--stacktrace`" for a bug report gets nothing. Printing a trace for a `ParameterException` would only be picocli internals, so narrowing the claim is the honest fix:

```suggestion
  @Option(
      names = {"--stacktrace"},
      description = "Print the full stack trace of a conversion failure, not just its message")
```

`isthmus-cli/README.md` line 59 needs the same narrowing — as written, "Anything that is not a recognizable problem with the input is always reported with its stack trace" reads as covering usage errors too.

isthmus-cli/src/main/java/io/substrait/isthmus/cli/IsthmusExecutionExceptionHandler.java
```diff
@@ -0,0 +1,165 @@
...
+      return Optional.of(CREATE_HINT.formatted(objectNotFound.group(1)));
+    }
+    Matcher columnNotFound = COLUMN_NOT_FOUND.matcher(message);
+    if (columnNotFound.find()) {
+      return Optional.of(COLUMN_HINT.formatted(columnNotFound.group(1)));
+    }
+    Matcher unknownIdentifier = UNKNOWN_IDENTIFIER.matcher(message);
+    if (unknownIdentifier.find()) {
+      return Optional.of(EXPRESSION_HINT.formatted(unknownIdentifier.group(1)));
+    }
```
andrew-coleman, line 201: `Unknown identifier` also comes out of `-c` validation, so this fires when `-e` was never used. A typo in a column *type* is enough:

```
$ isthmus -c "CREATE TABLE FOO(A NOSUCHTYPE)" "SELECT 1"
Error: From line 1, column 20 to line 1, column 29: Unknown identifier 'NOSUCHTYPE'

Hint: identifiers in a -e / --expression must be columns of a table defined
with -c / --create:

  isthmus -c "CREATE TABLE T (NOSUCHTYPE INT)" -e "NOSUCHTYPE + 1"
```

It is telling the user to use the bogus type name as a column name, via an option they did not pass. `handleExecutionException` already receives `parseResult` (line 73) and never consults it:

```suggestion
    Matcher unknownIdentifier = UNKNOWN_IDENTIFIER.matcher(message);
    if (parseResult.hasMatchedOption("-e") && unknownIdentifier.find()) {
      return Optional.of(EXPRESSION_HINT.formatted(unknownIdentifier.group(1)));
    }
```

This needs `hint` to take the parse result — `private static Optional<String> hint(final Exception ex, final CommandLine.ParseResult parseResult)` at line 126, and `hint(ex, parseResult)` at line 81. Verified: the misleading hint disappears and `unknownIdentifierInExpressionSuggestsCreateOption` still passes.

The deeper fix is a typed exception per failure site, so the handler does not have to infer origin from message text at all — this gating is still wrong if someone passes both `-e` and a bad `-c`.

isthmus-cli/src/main/java/io/substrait/isthmus/cli/IsthmusExecutionExceptionHandler.java
```diff
@@ -0,0 +1,165 @@
...
+ *
+ * <p>Only failures that are recognizably caused by the input are reported this way. Anything else
+ * is rethrown so that its stack trace is still printed, as is the full stack trace of a recognized
+ * failure when {@code --stacktrace} is given.
+ */
+class IsthmusExecutionExceptionHandler implements CommandLine.IExecutionExceptionHandler {
+
+  /** Matches Calcite's complaint about a table (or other object) that the catalog does not hold. */
+  private static final Pattern OBJECT_NOT_FOUND =
+      Pattern.compile("(?:Object|Table) '([^']+)' not found");
```
andrew-coleman: For a schema-qualified table the hint drops the qualifier, so the suggested command cannot fix the query:

```
$ isthmus -c "CREATE TABLE s.t(a INT)" "SELECT * FROM s.u"
Error: From line 1, column 15 to line 1, column 17: Object 'U' not found within 'S'

  isthmus -c "CREATE TABLE U (col1 INT, col2 VARCHAR)" "SELECT * FROM U"
```

`processCreateStatementsToCatalog`'s own javadoc says it "expects the use of fully qualified table names", so the suggestion needs to be `S.U`. Capturing the schema fixes this variant:

```suggestion
  /**
   * Matches Calcite's complaint about a table (or other object) that the catalog does not hold. The
   * second group holds the schema it looked in, for the qualified form of the message.
   */
  private static final Pattern OBJECT_NOT_FOUND =
      Pattern.compile("(?:Object|Table) '([^']+)' not found(?: within '([^']+)')?");
```

with a small helper, and `CREATE_HINT.formatted(objectName(objectNotFound))` at line 141:

```java
  private static String objectName(final Matcher objectNotFound) {
    String schema = objectNotFound.group(2);
    return schema == null ? objectNotFound.group(1) : schema + "." + objectNotFound.group(1);
  }
```

Verified → `isthmus -c "CREATE TABLE S.U (col1 INT, col2 VARCHAR)" "SELECT * FROM S.U"`.

Heads-up that this does not cover the *other* qualified form. When the schema does not exist either, Calcite reports only the unresolved prefix — `SELECT * FROM s.t` gives `Object 'S' not found` — and the message alone cannot tell you `S` was a schema, so the hint still suggests `CREATE TABLE S`. Fixing that needs the identifier text rather than the message: `CalciteContextException` carries `getPosColumn()`/`getEndPosColumn()`, and for `SELECT * FROM s.t` columns 15-17 are exactly `s.t`.

isthmus-cli/src/main/java/io/substrait/isthmus/cli/IsthmusExecutionExceptionHandler.java
```diff
@@ -0,0 +1,165 @@
...
+
+        isthmus -c "CREATE TABLE %1$s (col1 INT, col2 VARCHAR)" "SELECT * FROM %1$s"
+
+      Unquoted identifiers are upper-cased unless --unquotedcasing says otherwise.""";
+
+  private static final String COLUMN_HINT =
+      """
+      Hint: '%s' is not a column of any table defined with -c / --create. Check
+      the column names in the CREATE TABLE statement; unquoted identifiers are
+      upper-cased unless --unquotedcasing says otherwise.""";
```
andrew-coleman: This hint points at a CREATE TABLE statement that may not exist. `Column '...' not found in any table` also fires with no `-c` at all:

```
$ isthmus "SELECT bar"
Error: From line 1, column 8 to line 1, column 10: Column 'BAR' not found in any table

Hint: 'BAR' is not a column of any table defined with -c / --create. Check
the column names in the CREATE TABLE statement; ...
```

There is no CREATE TABLE statement to check — the user needs to be told to write one. A second variant selected on `parseResult.hasMatchedOption("-c")` covers it:

```suggestion
  private static final String COLUMN_HINT =
      """
      Hint: '%1$s' is not a column of any table defined with -c / --create. Check
      the column names in the CREATE TABLE statement.""";

  private static final String COLUMN_WITHOUT_CREATE_HINT =
      """
      Hint: '%1$s' is not a column of any table, and no table was defined. Pass a
      CREATE TABLE statement for each table the query reads using -c / --create:

        isthmus -c "CREATE TABLE T (%1$s INT)" "SELECT %1$s FROM T\"""";
```

The casing sentence moves out to a `withCasingNote(...)` helper — see the comment on `CREATE_HINT`. Selection at line 144:

```java
    if (columnNotFound.find()) {
      String hint = creates ? COLUMN_HINT : COLUMN_WITHOUT_CREATE_HINT;
      return Optional.of(withCasingNote(hint.formatted(columnNotFound.group(1)), parseResult));
    }
```

Verified: `SELECT bar` now suggests `isthmus -c "CREATE TABLE T (BAR INT)" "SELECT BAR FROM T"`, and `undefinedColumnIsExplained` still passes.

isthmus-cli/src/main/java/io/substrait/isthmus/cli/IsthmusExecutionExceptionHandler.java
```diff
@@ -0,0 +1,165 @@
...
+  // The hints are hard-wrapped for a terminal rather than joined into single long lines.
+
+  private static final String CREATE_HINT =
+      """
+      Hint: table definitions are not part of the query. Pass a CREATE TABLE
+      statement for each table it references using -c / --create:
+
+        isthmus -c "CREATE TABLE %1$s (col1 INT, col2 VARCHAR)" "SELECT * FROM %1$s"
+
+      Unquoted identifiers are upper-cased unless --unquotedcasing says otherwise.""";
```
andrew-coleman: The closing sentence is false in the one run most likely to print it — when the user has already chosen the casing:

```
$ isthmus --unquotedcasing UNCHANGED "SELECT * FROM foo"
Error: From line 1, column 15 to line 1, column 17: Object 'foo' not found
...
Unquoted identifiers are upper-cased unless --unquotedcasing says otherwise.
```

Identifiers were *not* upper-cased, and it steers them toward the one knob they already turned. Lifting the sentence out of the templates lets it be conditional:

```suggestion
  private static final String CREATE_HINT =
      """
      Hint: table definitions are not part of the query. Pass a CREATE TABLE
      statement for each table it references using -c / --create:

        isthmus -c "CREATE TABLE %1$s (col1 INT, col2 VARCHAR)" "SELECT * FROM %1$s\"""";
```

with:

```java
  /** Appended to the hints whose advice depends on how unquoted identifiers are cased. */
  private static final String CASING_NOTE =
      "\n\nUnquoted identifiers are upper-cased unless --unquotedcasing says otherwise.";

  private static String withCasingNote(
      final String hint, final CommandLine.ParseResult parseResult) {
    return parseResult.hasMatchedOption("--unquotedcasing") ? hint : hint + CASING_NOTE;
  }
```

Verified: the sentence disappears when `--unquotedcasing` is given and is unchanged otherwise, so `undefinedTableSuggestsCreateOption` and `undefinedColumnIsExplained` both still pass.

isthmus-cli/src/main/java/io/substrait/isthmus/cli/IsthmusEntryPoint.java
```diff
@@ -59,33 +62,57 @@ enum OutputFormat {
...
    * @param args Isthmus CLI arguments.
    */
   public static void main(String... args) {
+    CommandLine commandLine = createCommandLine();
+    if (args.length == 0) { // If no arguments print usage help
+      commandLine.usage(commandLine.getOut());
+      System.exit(CommandLine.ExitCode.OK);
+    }
+    System.exit(commandLine.execute(args));
+  }
```
andrew-coleman, line 79: Two spellings of the same mistake now disagree on both stream and exit code:

```
$ isthmus                                  # usage on stdout, exit 0
$ isthmus -c "CREATE TABLE FOO(A INT)"     # message plus usage on stderr, exit 2
```

A wrapper checking `$?` or piping stdout gets opposite answers for the same user error. The branch is also untested: `IsthmusEntryPointTest.run()` calls `createCommandLine()`/`execute()` and never `main()`, so lines 78-82 have no coverage at all.

Dropping the special case routes the zero-arg case through the new `ParameterException`, making it consistent and testable:

```suggestion
  public static void main(String... args) {
    System.exit(createCommandLine().execute(args));
  }
```

This is a deliberate behaviour change — bare `isthmus` goes from usage/exit 0 to `Missing SQL to convert` plus usage/exit 2. If you would rather keep exit 0 for the zero-arg case, then the `-c`-only case should match it, otherwise the inconsistency just moves. Verified with:

```java
  @Test
  void noArgumentsReportsUsageError() {
    Run run = run();

    assertEquals(CommandLine.ExitCode.USAGE, run.statusCode);
    run.assertErrContains("Missing SQL to convert");
    run.assertNoStackTrace();
  }
```

isthmus-cli/src/main/java/io/substrait/isthmus/cli/IsthmusExecutionExceptionHandler.java
```diff
@@ -0,0 +1,165 @@
...
+    Optional<String> hint = hint(ex);
+    if (hint.isPresent()) {
+      err.println();
+      err.println(hint.get());
+    }
+    err.flush();
+
+    if (stackTraceRequested(cmd)) {
+      throw ex;
+    }
```
andrew-coleman: `parseResult` already carries this, so the reach-through through `cmd.getCommand()` is not needed:

```suggestion
    if (parseResult.hasMatchedOption("--stacktrace")) {
      throw ex;
    }
```

That lets `stackTraceRequested` (lines 154-164) and `IsthmusEntryPoint.isStackTraceRequested()` (lines 99-106, which exists only to feed it) both go. It also removes a failure mode: the current `instanceof` fails *open* to "no stack trace", so if a subcommand is ever added, `ex.getCommandLine().getCommand()` is the subcommand's user object, `--stacktrace` quietly stops working, and nothing reports why.

Verified: `stackTraceOptionKeepsTheStackTrace` still passes.

isthmus-cli/src/main/java/io/substrait/isthmus/cli/IsthmusExecutionExceptionHandler.java
```diff
@@ -0,0 +1,165 @@
...
+  private static final Pattern UNKNOWN_IDENTIFIER = Pattern.compile("Unknown identifier '([^']+)'");
+
+  /** The message the CREATE statement parser reports for anything that is not a CREATE TABLE. */
+  private static final String NOT_A_CREATE_TABLE = "Not a valid CREATE TABLE statement.";
+
+  /** The message the CREATE statement parser reports for a CREATE TABLE AS SELECT. */
+  private static final String CTAS_NOT_SUPPORTED = "CTAS not supported.";
+
+  /** The message the DDL converter reports for a CREATE TABLE without a query. */
+  private static final String CTAS_ONLY = "Only create table as select statements are supported";
```
andrew-coleman, line 48: This literal is a copy of the one in `isthmus/src/main/java/io/substrait/isthmus/calcite/rel/DdlSqlToRelConverter.java:91` — a different Gradle module — matched with `.equals()`. Reword it there (a plausible, unrelated change) and `isthmus "CREATE TABLE foo(a INT)"` silently goes back to dumping a stack trace, with everything still compiling and every test in `isthmus` still green. `NOT_A_CREATE_TABLE` and `CTAS_NOT_SUPPORTED` have the same problem against the inline literals in `SubstraitCreateStatementParser`.

A dedicated exception type is the real fix; failing that, `public static final` constants exposed by the throwing classes would at least turn a rename into a compile error.

Minor, same area: `isPlainCreateTableQuery(ex)` is evaluated twice per failure — once in `isInputError` (line 105) and again at the top of `hint()` (line 127).

isthmus/src/test/java/io/substrait/isthmus/sql/SubstraitCreateStatementParserTest.java
```diff
@@ -119,4 +119,13 @@ void testToCatalogWithMultipleCreateTableForSameTableThrowsException() throws Sq
...
   }
+
+  @Test
+  void testToCatalogWithCreateTableAsSelectThrowsException() {
+    assertThrows(
+        SqlParseException.class,
+        () ->
+            SubstraitCreateStatementParser.processCreateStatementsToCatalog(
+                "create table src1 as select 1"));
+  }
```
andrew-coleman, line 167: This passes on any `SqlParseException`, including "Not a valid CREATE TABLE statement." or a raw parser error, so it does not pin the guard it was added for. That matters more than usual here: `IsthmusExecutionExceptionHandler.hint()` line 135 dispatches on `message.contains(CTAS_NOT_SUPPORTED)`, so a message change would keep this test green while silently dropping the CLI hint.

```suggestion
  @Test
  void testToCatalogWithCreateTableAsSelectThrowsException() {
    final SqlParseException e =
        assertThrows(
            SqlParseException.class,
            () ->
                SubstraitCreateStatementParser.processCreateStatementsToCatalog(
                    "create table src1 as select 1"));

    assertEquals("CTAS not supported.", e.getMessage());
  }

  @Test
  void testToCatalogWithCreateTableAsSelectWithColumnListThrowsException() {
    final SqlParseException e =
        assertThrows(
            SqlParseException.class,
            () ->
                SubstraitCreateStatementParser.processCreateStatementsToCatalog(
                    "create table src1 (intcol int) as select 1"));

    assertEquals("CTAS not supported.", e.getMessage());
  }

  @Test
  void testToCatalogWithCreateTableWithoutColumnsThrowsException() {
    final SqlParseException e =
        assertThrows(
            SqlParseException.class,
            () ->
                SubstraitCreateStatementParser.processCreateStatementsToCatalog(
                    "create table src1"));

    assertEquals("Column definitions are required.", e.getMessage());
  }

  @Test
  void testCreateStatementsWithCreateTableWithoutColumnsThrowsException() {
    final SqlParseException e =
        assertThrows(
            SqlParseException.class,
            () -> SubstraitCreateStatementParser.processCreateStatements("create table src1"));

    assertEquals("Column definitions are required.", e.getMessage());
  }
}
```

The explicit-column CTAS case is the one that shows why the guard has to stay `query != null` rather than becoming `columnList == null`. The last two require the `columnList` fix in the main class.

isthmus-cli/src/test/script/smoke.sh
```diff
@@ -39,3 +39,19 @@ echo "${LINEITEM}"
...
+fi
+echo "${error}"
+if ! grep -q -- '--create' <<<"${error}"; then
+  echo "expected a hint about --create"
+  exit 1
+fi
+if grep -q 'CalciteContextException' <<<"${error}"; then
+  echo "expected a message instead of a stack trace"
+  exit 1
+fi
```
andrew-coleman: Neither of these can fail the way it is meant to.

`grep -q -- '--create'` matches picocli's own usage block (`-c, --create=<createStatements>`), so if the hint machinery regresses to printing usage instead of the message, both checks still pass. And `grep -q 'CalciteContextException'` rejects one class name, so a trace from the NPE in `SubstraitCreateStatementParser`, a `SqlParseException`, or picocli's `ExecutionException` all sail through. The property being asserted is "no stack trace", which `\tat ` captures for every exception type — `IsthmusEntryPointTest.Run.assertNoStackTrace()` already uses exactly that form, so the two layers would then be checking the same thing.

```suggestion
if ! grep -q 'table definitions are not part of the query' <<<"${error}"; then
  echo "expected a hint about --create"
  exit 1
fi
if grep -q $'\tat ' <<<"${error}"; then
  echo "expected a message instead of a stack trace"
  exit 1
fi

# Error path - --stacktrace keeps the trace as well as the message
if trace=$("${CMD}" --stacktrace 'select * from lineitem' 2>&1); then
  echo "expected a query without a table definition to fail"
  exit 1
fi
if ! grep -q $'\tat ' <<<"${trace}"; then
  echo "expected --stacktrace to print the stack trace"
  exit 1
fi
```

The `--stacktrace` half is the other side of the new behaviour and is currently unexercised. Flagging that I checked the shell logic but not against a native build, so that addition is the one thing here I have not run end to end.

isthmus-cli/src/main/java/io/substrait/isthmus/cli/IsthmusExecutionExceptionHandler.java
```diff
@@ -0,0 +1,165 @@
...
+    if (hint.isPresent()) {
+      err.println();
+      err.println(hint.get());
+    }
+    err.flush();
+
+    if (stackTraceRequested(cmd)) {
+      throw ex;
+    }
+    return cmd.getCommandSpec().exitCodeOnExecutionException();
```
andrew-coleman, line 122: Recognised input errors return `ExitCode.SOFTWARE` — picocli's *internal software error* code — which is also what an unrecognised crash returns:

```
$ isthmus "SELECT * FROM foo"                          # bad input, message printed -> 1
$ isthmus -c "CREATE TABLE FOO" "SELECT * FROM FOO"    # NPE, trace printed        -> 1
```

Meanwhile missing input deliberately exits 2 (`ExitCode.USAGE`). So two members of the newly created "the input is wrong" category get different codes, and a script wrapping isthmus cannot tell "fix your SQL" from "file a bug". The new tests bake `ExitCode.SOFTWARE` in, so the split is now pinned.

Worth deciding explicitly: either input errors join missing input on 2, or the distinction gets documented. Not a blocker, but easier to settle before release than after.

isthmus-cli/src/main/java/io/substrait/isthmus/cli/IsthmusExecutionExceptionHandler.java
nielspardon: Kept the literals, because CI does catch the rename: the CLI tests assert on all three messages end to end (`ctasPassedToCreateOptionSuggestsQueryArgument`, `queryPassedToCreateOptionSuggestsQueryArgument`, and `createTableAsQuerySuggestsCreateOption` for `CTAS_ONLY`, which only prints its hint when the message matches exactly), so rewording any of them turns `:isthmus-cli:test` red in the same build. Exposing them as `public static final` in `:isthmus` would make error-message text part of that module's API, which I would rather not do for a coupling the tests already pin — happy to add the constants, or a dedicated exception type, if you disagree.

Left the double `isPlainCreateTableQuery` evaluation: it is an `instanceof` plus a `String.equals` on a path that is about to print and exit, and hoisting it means either threading a boolean through `hint()` or reordering the two checks into one method that does both jobs.

isthmus-cli/src/main/java/io/substrait/isthmus/cli/IsthmusExecutionExceptionHandler.java
nielspardon: Settled it as documentation rather than a move to 2: 1 for "the SQL could not be converted", 2 for "the command line itself was wrong". That keeps `--stacktrace`-worthy defects and bad SQL on the same code, which is the part your example is really about, but it is the split a compiler makes too — bad source is a normal failure, not a usage error — and it means adding a hint for a failure never silently changes a script's exit code. The three codes are now a table in the isthmus-cli readme.

nielspardon:
Thanks — everything reproduced, and all of it is fixed in 696ae64 except the two threads I replied to inline (the cross-module message literals, which the CLI tests already pin, and the exit-code split, which is now documented rather than unified).

Two things worth calling out beyond the review:

Bare `isthmus` now reports `Missing SQL to convert` on stderr and exits 2 instead of printing usage to stdout and exiting 0. That is your option 1 from the "two spellings disagree" thread, and it is a behavior change for anyone running the binary with no arguments to get its usage — `--help` still does that.

Chasing your note about not having run the `--stacktrace` addition against a native build turned up a native-only defect that the feature was quietly sitting on top of. Calcite builds a syntax error's expected-token list by reflectively calling three grammar productions on the parser (`SqlAbstractParserImpl.MetadataImpl.initList`), and those were not registered, so in the native image *every* syntax error came out as `RuntimeException: While building token lists` caused by `NoSuchMethodException: SqlDdlParserImpl.ReservedFunctionName()` — never as a `SqlParseException`, and so never as a friendly message. It was invisible before this PR because every error printed a trace anyway. `RegisterAtRuntime` now registers the three productions, and smoke.sh asserts a syntax error names its position, since the JVM tests cannot see this.

isthmus-cli/src/main/java/io/substrait/isthmus/cli/IsthmusExecutionExceptionHandler.java
```diff
@@ -0,0 +1,228 @@
...
+    Matcher columnNotFound = COLUMN_NOT_FOUND.matcher(message);
+    if (columnNotFound.find()) {
+      String hint = parseResult.hasMatchedOption("-c") ? COLUMN_HINT : COLUMN_WITHOUT_CREATE_HINT;
+      return Optional.of(withCasingNote(hint.formatted(columnNotFound.group(1)), parseResult));
+    }
+    Matcher unknownIdentifier = UNKNOWN_IDENTIFIER.matcher(message);
+    if (expressions && unknownIdentifier.find()) {
+      // Without -e this is a complaint about a -c statement, where naming the identifier as a
+      // column of a new table would be the wrong advice.
+      return Optional.of(EXPRESSION_HINT.formatted(unknownIdentifier.group(1)));
```
andrew-coleman: Non-blocking, but this is the one identifier hint that did not pick up `withCasingNote`, and it is arguably where casing bites hardest. `isthmus -e "col + 1"` reports `Unknown identifier 'COL'` and then suggests `-e "COL + 1"` — the input has been silently upper-cased in the very command being recommended, with nothing to explain why:

```
Error: From line 1, column 1 to line 1, column 3: Unknown identifier 'COL'

Hint: identifiers in a -e / --expression must be columns of a table defined
with -c / --create:

  isthmus -c "CREATE TABLE T (COL INT)" -e "COL + 1"
```

The other three hints all carry the note now, and the PR body still promises it for this one specifically ("an unresolved column or an unresolved identifier in `-e` / `--expression` ... mentions `--unquotedcasing`").

```suggestion
      return Optional.of(
          withCasingNote(EXPRESSION_HINT.formatted(unknownIdentifier.group(1)), parseResult));
```

Verified: `spotlessCheck` and `pmdMain` clean, `:isthmus-cli:test` still 21/21, and the note appears.

Separately, on the comment just above — agreed that gating on `-e` is the right call, and noting for the record that the residual case is `-e` *plus* a bad `-c`: `isthmus -e "1" -c "CREATE TABLE foo(a NOSUCHTYPE)"` still suggests `CREATE TABLE T (NOSUCHTYPE INT)`. Lower severity than before since the user did pass `-e`, and the real cure is the typed exception, so I would leave it.

andrew-coleman (approved):
Approving. All fourteen threads are addressed, and the two you pushed back on I now agree with.

I re-ran everything rather than reading the diff. Each of the original reproductions against the new binary: the bare `CREATE TABLE` NPE is gone on both parser paths and also for `IF NOT EXISTS` and `OR REPLACE`, while `CREATE TABLE FOO (a INT) AS SELECT 1` still reports `CTAS not supported.`; `isthmus ""` keeps its trace; the query-after-`-e` case gets the new hint and query-plus-`-e` exits 2 naming the discarded query; the `-e` hint no longer fires without `-e`; the qualified name comes out as `S.U`; the casing note disappears once `--unquotedcasing` is given; bare `isthmus` exits 2 while `--help` still exits 0 on stdout. `:isthmus:test` 961/961, `:isthmus-cli:test` 21/21, `pmdMain`, `spotlessCheck`, `javadoc` clean. I also emulated both new `smoke.sh` assertion blocks against the JVM build and they pass.

**On the message literals** — your argument holds and I checked both halves of it. `pr.yml` runs `./gradlew build --rerun-tasks`, one build across all modules, so a rename in `:isthmus` does turn `:isthmus-cli:test` red in the same run; and the pinning is real, including for `CTAS_ONLY`, where asserting the hint pins the message because the hint only prints on an exact match. Dropping it. One note for whenever a typed exception does happen: `COLUMNS_REQUIRED` makes it four call sites rather than three.

**On the exit codes** — the README table matches the binary, and "bad source is a normal failure, not a usage error" is the line a compiler draws too. Settled.

**On the native-image find** — good catch, and your account is exact. I could not build a native image locally (Temurin here, not GraalVM), so I corroborated the chain from bytecode instead: `SqlDdlParserImpl.getMetadata()` constructs `MetadataImpl`, whose constructor calls `initList` with exactly those three production names, which reaches `virtualCall` → `Class.getMethod` → `Method.invoke`, and whose `catch (Throwable)` handler is literally `new RuntimeException("While building token lists", cause)` — the message you quoted, verbatim. All three productions are `public final` no-arg on `SqlDdlParserImpl`, so the `getMethod` call in `registerMethods` is a real build-time guard; it throws `NoSuchMethodException`, which the enclosing `catch (Exception)` rethrows as `IllegalStateException` out of `beforeAnalysis`, so the javadoc's "fails to build" promise holds. `compileOnly` is the right scope, since `:isthmus` already declares `calcite-server` as `implementation`. And it explains itself: `getMetadata()` is only reached for a `ParseException`, never a `CalciteContextException`, which is why the existing `select * from lineitem` smoke case never exposed it.

One inline comment on the casing note, and one thing that has no line to attach to:

**The PR body has fallen behind the branch, and it is what gets committed.** `AGENTS.md` lines 163-165: the title and body together become the squash-merge commit message that `semantic-release` builds `CHANGELOG.md` from, so the fix commit's (very good) message is discarded on merge. As it stands the body:

- claims the `-e` identifier hint mentions `--unquotedcasing`, which is not true today — see the inline comment;
- does not mention that bare `isthmus` changed from usage/exit 0 to error/exit 2. You called it out in a comment here, but a comment is not the commit message, and this is the change most likely to surprise someone scripting the binary;
- does not mention that a query passed together with `-e` is now rejected;
- does not mention the native-image reflection fix at all — a distinct defect from the one in the title, and the kind of thing that wants its own changelog line;
- still ends with `🤖 Generated with AI`, which `AGENTS.md` lines 174-176 lists among the tool-attribution lines to keep out of commit bodies.

None of that blocks the code, so I am approving rather than holding it — but the body is worth a pass before you squash.

isthmus-cli/src/main/java/io/substrait/isthmus/cli/IsthmusExecutionExceptionHandler.java
nielspardon: Taken in c35c7ed — you are right that this is where the casing bites hardest, since the hint hands back a spelling the user never typed. Extended `unknownIdentifierInExpressionSuggestsCreateOption` to assert the note so it cannot drift off again.

Agreed on leaving the `-e` plus bad `-c` residual: the identifier hint is wrong there, but the user did ask for expressions, and separating the two needs the exception type rather than another message match.

nielspardon:
Thanks for re-running it all rather than reading the diff — and for corroborating the native-image chain from bytecode, which is a better check than my "it works now" was.

The casing note is on the `-e` hint in c35c7ed. The PR body has had the pass you asked for: the AI attribution line is gone, the `--unquotedcasing` claim is true again rather than dropped, and it now covers the bare-`isthmus` change, the query-plus-`-e` rejection, and the native-image reflection fix as its own item. The two script-visible changes are also a `BREAKING CHANGE` footer, so they get a line in the release notes instead of only living in the body.

Noted on `COLUMNS_REQUIRED` being a fourth call site whenever the typed exception happens.

PR #1061, opened 2026-08-05
fix(core)!: re-derive field reference types when copy-on-write replaces a relation

### LINKED ISSUE #185: stale type in FieldReference after RelCopyOnWrite modifications

nielspardon:
A `FieldReference` caches the type of the field it references. `RelCopyOnWriteVisitor` copied those references over verbatim, so replacing a subtree with one that emits a different record type left every reference above it carrying the type of the relation that is no longer there — and the record types derived from those references, such as a `Project`'s, were wrong in turn.

The visitor now tracks the record type that each relation's own expressions resolve against and re-derives the cached type of every field reference it rewrites from it. Inputs are rewritten before their relation's expressions so the scope is the type a replaced input emits, and enclosing scopes are tracked per subquery boundary so an outer reference re-derives against the relation it steps out to.

Each scope is the one the reference actually resolves against, which is not always the relation's inputs concatenated:

- a join condition, post-join filter and residual expression resolve against the two inputs combined;
- a hash or merge join **key** resolves against the single side it selects from — its offsets are side-relative, which is why proto conversion types each side with its own converter;
- the filter of a read relation resolves against the schema being read, not against any enclosing scope;
- a lateral join's right input resolves the references it makes to the current left row against the left record type, under the join's rel anchor. The left input is rewritten first, so that type is the one a replaced left input emits — mirroring `ProtoRelConverter.newLateralJoin`, which registers the same scope before converting the right input.

Re-derivation does not depend on anything having been replaced: a reference whose cached type disagrees with the root it resolves against is corrected either way, whether that root is a relation or another expression the rewrite left alone.

References that resolve against something outside the tracked scopes keep the type they have: a lambda parameter, and an outer reference naming a rel anchor that no enclosing relation exposes, which needs plan-wide context this visitor does not have.

## Resolving a reference is total

Re-deriving a type must not become a new way for a rewrite to fail. A rewrite that drops a column, reshapes a nested type, or changes a container kind can leave a reference selecting something its input no longer has; the resulting tree is invalid either way, so such a reference keeps its cached type instead.

Delivering that needed the resolution itself to be total, not a guard in front of a throwing derivation — a guard that checks only the outermost segment still lets a nested reference reach the derivation and throw. `FieldReference.resolveType` reports the type a chain of segments selects, or nothing, at any depth and for every kind of segment. It sits beside the finders whose rules it mirrors so the two cannot drift apart silently, and it mirrors them exactly, including the two asymmetries that matter: a list element offset is not bounds-checked, because the length of a list is not part of its type, and a map key type is compared exactly, nullability included.

## Also fixed here

One neighbouring bug surfaced while wiring this up, load-bearing for the retyping above: `visitFieldReference` built its replacement without copying the original, dropping the segments and the type. Since the type is mandatory, rewriting any reference rooted at an expression threw instead of returning the rewritten reference.

## Not fixed here

The types cached on function invocations are not re-derived — that needs the function declarations, which the visitor does not have — so a relation whose record type comes from a measure or window function can still be stale. `Expression.ScalarSubquery` caches its type the same way. Whether these types should be cached on the POJOs at all is the broader question the issue raises; this is the short-term fix it asks for.

The off-by-one bound check in `StructFieldFinder` stays as it is (#1068 — the exception it produces is part of what `ProtoExpressionConverter` reports for a malformed plan, so changing it is not a free fix).

## Behaviour worth calling out

- Where a narrowing rewrite previously threw from inside the visitor, it now yields a plan in which the unresolvable reference keeps its stale type. Anyone relying on that exception as a validation signal loses it — validation belongs in a validator, not in a copy-on-write rewrite.
- Tracking the scope makes a visitor instance stateful for the duration of a traversal, so an instance can no longer be used to visit several relation trees concurrently. Sequential reuse is unaffected.
- `visitComparisonJoinKey` takes the two side record types now. It could not previously return a usable reference at all, so nothing can have depended on the old signature.

Closes #185

BREAKING CHANGE: `RelCopyOnWriteVisitor.visitComparisonJoinKey` now takes the record types of the join's two sides alongside the key, so an override must adopt the new signature. A rewrite that leaves a field reference unable to resolve against its input no longer throws from inside the visitor; it returns a plan in which that reference keeps its stale type, so callers relying on that exception as a validation signal must validate separately. A visitor instance now carries the scope of the traversal it is running, so a single instance can no longer visit several relation trees concurrently; sequential reuse is unaffected.

### DIFF
--- a/core/src/main/java/io/substrait/relation/ExpressionCopyOnWriteVisitor.java
+++ b/core/src/main/java/io/substrait/relation/ExpressionCopyOnWriteVisitor.java
@@ -12,7 +12,6 @@ import io.substrait.expression.ImmutableExpression;
 import io.substrait.expression.ImmutableFieldReference;
 import io.substrait.type.Type;
 import io.substrait.util.EmptyVisitationContext;
-import java.util.ArrayList;
 import java.util.List;
 import java.util.Optional;
 
@@ -435,10 +434,20 @@ public class ExpressionCopyOnWriteVisitor<E extends Exception>
    * cached type from the root it resolves against. Re-deriving the type is what keeps references
    * correct when a relation's input is replaced by one emitting a different record type.
    *
+   * <p>A reference that no longer resolves against its root, at any segment depth, is left exactly
+   * as it is, including the type it has cached. Such a reference makes the relation tree invalid
+   * whatever this visitor does with it, so re-deriving its type is not turned into a failure of the
+   * rewrite.
+   *
    * <p>The type of a reference to a lambda parameter, or of an outer reference identified by the
    * {@link io.substrait.relation.Rel#getRelAnchor() rel anchor} of the relation it is rooted on, is
    * left as it is: neither resolves against a relation in the enclosing scopes tracked during the
    * traversal.
+   *
+   * <p>Override this rather than {@link #visit(FieldReference, EmptyVisitationContext)} to change
+   * how references are rewritten: this is what the positions that hold a reference rather than an
+   * arbitrary expression — a {@link io.substrait.relation.physical.ScatterExchange}'s fields and a
+   * {@link io.substrait.relation.physical.ComparisonJoinKey}'s sides — are rewritten through.
    *
    * @param fieldReference the field reference to visit
    * @param context the visitation context
@@ -453,12 +462,12 @@ public class ExpressionCopyOnWriteVisitor<E extends Exception>
       if (!inputExpression.isPresent()) {
         return Optional.empty();
       }
+      // The reference is returned rewritten even when its type could not be re-derived: the
+      // expression it is rooted at changed, and that change is not lost because the type is stale.
       ImmutableFieldReference.Builder rewritten =
           ImmutableFieldReference.builder().from(fieldReference).inputExpression(inputExpression);
-      if (!fieldReference.segments().isEmpty()) {
-        rewritten.type(
-            FieldReference.ofExpression(inputExpression.get(), segmentsOf(fieldReference)).type());
-      }
+      FieldReference.resolveType(inputExpression.get().getType(), fieldReference.segments())
+          .ifPresent(rewritten::type);
       return Optional.of(rewritten.build());
     }
     return retypeRootReference(fieldReference);
@@ -476,40 +485,21 @@ public class ExpressionCopyOnWriteVisitor<E extends Exception>
     Type.Struct rootType =
         getRelCopyOnWriteVisitor()
             .inputTypeStepsOut(fieldReference.outerReferenceStepsOut().orElse(0));
-    // A rewrite that drops fields can leave a reference selecting a field its input no longer has.
-    // The resulting relation tree is invalid either way, so leave the reference as it is rather
-    // than turn re-deriving its type into a failure of the rewrite.
-    if (rootType == null || !selectsFieldOf(fieldReference, rootType)) {
+    if (rootType == null) {
       return Optional.empty();
     }
-    Type type = FieldReference.ofRoot(rootType, segmentsOf(fieldReference)).type();
-    if (type.equals(fieldReference.type())) {
+    // A rewrite that drops a field, or that reshapes a nested one, can leave a reference selecting
+    // something its input no longer has. The resulting relation tree is invalid either way, so
+    // leave
+    // the reference as it is rather than turn re-deriving its type into a failure of the rewrite.
+    // Resolution is total, at every segment depth and for every kind of segment, so this is a
+    // decision the rewrite makes rather than an exception it has to recover from.
+    Optional<Type> type = FieldReference.resolveType(rootType, fieldReference.segments());
+    if (!type.isPresent() || type.get().equals(fieldReference.type())) {
       return Optional.empty();
     }
-    return Optional.of(ImmutableFieldReference.builder().from(fieldReference).type(type).build());
-  }
-
-  /**
-   * Returns whether the outermost segment of the given reference selects a field that the given
-   * record type has. The segments are held innermost first, so the outermost one is the last.
-   */
-  private static boolean selectsFieldOf(FieldReference fieldReference, Type.Struct rootType) {
-    List<FieldReference.ReferenceSegment> segments = fieldReference.segments();
-    if (segments.isEmpty()) {
-      return false;
-    }
-    FieldReference.ReferenceSegment outermost = segments.get(segments.size() - 1);
-    return outermost instanceof FieldReference.StructField
-        && ((FieldReference.StructField) outermost).offset() < rootType.fields().size();
-  }
-
-  /**
-   * Returns the segments of the given reference as a mutable list, which is what {@link
-   * FieldReference#ofRoot} and {@link FieldReference#ofExpression} require: they reverse the list
-   * they are given.
-   */
-  private static List<FieldReference.ReferenceSegment> segmentsOf(FieldReference fieldReference) {
-    return new ArrayList<>(fieldReference.segments());
+    return Optional.of(
+        ImmutableFieldReference.builder().from(fieldReference).type(type.get()).build());
   }
 
   @Override
--- a/core/src/main/java/io/substrait/relation/ExpressionCopyOnWriteVisitor.java
+++ b/core/src/main/java/io/substrait/relation/ExpressionCopyOnWriteVisitor.java
@@ -9,7 +9,9 @@ import io.substrait.expression.ExpressionVisitor;
 import io.substrait.expression.FieldReference;
 import io.substrait.expression.FunctionArg;
 import io.substrait.expression.ImmutableExpression;
+import io.substrait.expression.ImmutableFieldReference;
 import io.substrait.expression.WindowBound;
+import io.substrait.type.Type;
 import io.substrait.util.EmptyVisitationContext;
 import java.util.List;
 import java.util.Optional;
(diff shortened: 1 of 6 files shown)


core/src/main/java/io/substrait/relation/ExpressionCopyOnWriteVisitor.java
```diff
@@ -424,39 +426,108 @@ protected Optional<Expression.MultiOrListRecord> visitMultiOrListRecord(
...
+    return retypeRootReference(fieldReference);
+  }
 
-    if (allEmpty(inputExpression)) {
+  /**
+   * Re-derives the type of a reference into the record type of the relation it is rooted on, which
+   * may have been replaced by one emitting a different record type.
+   */
+  private Optional<FieldReference> retypeRootReference(FieldReference fieldReference) {
+    if (fieldReference.isLambdaParameterReference()
```
alexandrefimov: Would registering the lateral-join scope here be enough for the case that matters? `visit(LateralJoin)` rewrites left before right, so an anchor-to-record-type map filled at that point would cover the same references `ProtoRelConverter.newLateralJoin` covers, without needing anything plan-wide.

If the answer is "yes but not in this PR", a sentence in "Not fixed here" naming the lateral join would keep the next reader from concluding anchors are simply out of reach.

core/src/main/java/io/substrait/relation/ExpressionCopyOnWriteVisitor.java
```diff
@@ -424,39 +426,108 @@ protected Optional<Expression.MultiOrListRecord> visitMultiOrListRecord(
...
+   * @param context the visitation context
+   * @return Optional containing the modified field reference, or empty if no changes
+   * @throws E if an error occurs during visitation
+   */
+  public Optional<FieldReference> visitFieldReference(
+      FieldReference fieldReference, EmptyVisitationContext context) throws E {
+    if (fieldReference.inputExpression().isPresent()) {
+      Optional<Expression> inputExpression =
+          fieldReference.inputExpression().get().accept(this, context);
+      if (!inputExpression.isPresent()) {
```
alexandrefimov: This returns before `resolveType` when the root expression is unchanged, so an expression-rooted reference never gets the unconditional correction that `retypeRootReference` applies.

Defensible — an unchanged root has an unchanged type — but worth a word here or in the description, since the two branches now differ in a way the summary does not distinguish.

alexandrefimov (commented):
Read this against `origin/main` and checked the parts that could go wrong on their own; three things, only the first of which I think needs a decision.

**The anchor case is the lateral-join case.** `retypeRootReference` leaves an outer reference identified by `rel_anchor` alone, on the grounds that resolving an anchor needs plan-wide context. That is true of anchors in general — one can point into a `ReferenceRel`-shared subtree — but the common producer of them is not general: `ProtoRelConverter.newLateralJoin` registers `anchorScopes.put(anchor, left.getRecordType())` *before* converting the right input, precisely because that is where the references live, and `outerReferenceScope` special-cases `LateralJoin` to the left record type for the same reason. `visit(LateralJoin)` here already rewrites left before right, so the same registration would fit in the same place.

That matters more than the general case, because a lateral join's right input is exactly where a rewrite of the left changes the type its references resolve against — the bug this PR is about, left unfixed for the relation where correlation is most common. I am not sure it is worth doing in this PR rather than the next one; I am fairly sure "Not fixed here" should name it, since as written the reader is left thinking anchors are unresolvable rather than that one resolvable case was deferred.

Checked while looking at this and it holds: `inputTypeStepsOut` indexes the enclosing stack exactly as the shipped `ProtoRelConverter.outerScopeForStepsOut` does, including agreeing on `stepsOut == 0`, and nothing pushes a `steps_out` scope for a lateral join in either place — so the two mechanisms have the same shape rather than two conventions.

**Re-derivation is not quite unconditional.** The description says a reference whose cached type disagreed with its input is corrected even when nothing was replaced, and `retypeRootReference` does that. The expression-rooted branch does not: if the root expression comes back unchanged, `visitFieldReference` returns empty before `resolveType` runs. There is a good argument that this is right — the root is right there, and if it did not change neither did its type — but then the claim holds for root references only, and it is the kind of asymmetry that reads as an oversight later.

**Is this a breaking release?** The title is `fix(core)` without `!`, and three things in the description look like they belong on the other side of that line: `visitComparisonJoinKey` changes a public signature, a rewrite that previously threw now returns a plan with a stale type, and a visitor instance acquires a no-concurrent-reuse contract it did not have. #1058 took the same shape — validation behaviour that consumers could be relying on — and shipped as breaking, and I made the same argument on #1074 for `Fetch`. Being wrong about this is cheap in one direction and not in the other.

alexandrefimov (approved):
Rechecked `fd18dcc4`. The lateral join now exposes its rewritten left record type under the anchor while visiting the right input. References rooted at an unchanged expression also have their cached type corrected.

Both regression tests fail when the corresponding fixes are removed and pass on this head. The full core test task passes (797 passed, 32 skipped). Both of my earlier comments are addressed.

core/src/main/java/io/substrait/relation/RelCopyOnWriteVisitor.java
```diff
@@ -672,6 +787,132 @@ protected Optional<ConsistentPartitionWindow.WindowRelFunctionInvocation> visitW
...
+   *
+   * <p>Only the fields of the result are ever read, so its own nullability is not meaningful.
+   *
+   * @param inputs the relations the expressions are evaluated over, in field order
+   * @return the combined record type
+   */
+  protected static Type.Struct recordTypeOf(Rel... inputs) {
+    return TypeCreator.REQUIRED.struct(
+        Arrays.stream(inputs).flatMap(input -> input.getRecordType().fields().stream()));
+  }
```
andrew-coleman, line 804: Runs in every expression-bearing `visit`, even with nothing to rewrite. `Set.deriveRecordType()` throws for inputs with differing field counts — a state `Set` deliberately does not validate at construction — so a pure traversal over such a tree now throws where it returned on `main`. Checked both sides against `0b206ab4`.

```suggestion
  protected static Type.Struct recordTypeOf(Rel... inputs) {
    try {
      return TypeCreator.REQUIRED.struct(
          Arrays.stream(inputs).flatMap(input -> input.getRecordType().fields().stream()));
    } catch (RuntimeException e) {
      // A relation whose record type cannot be derived -- a Set over inputs with differing field
      // counts is not rejected at construction -- leaves the references in its parent's own
      // expressions alone, rather than making a traversal fail where it used to return.
      return null;
    }
  }
```

core/src/main/java/io/substrait/relation/RelCopyOnWriteVisitor.java
```diff
@@ -211,18 +263,29 @@ public Optional<Rel> visit(Join join, EmptyVisitationContext context) throws E {
...
-    Optional<Expression> condition = visitOptionalExpression(lateralJoin.getCondition(), context);
+    Rel newLeft = left.orElse(lateralJoin.getLeft());
+    // The right input is evaluated once per left row, and the references it makes to that row are
+    // identified by this join's anchor, so the left record type is in scope under that anchor for
+    // exactly the rewrite of the right input.
+    Optional<Rel> right =
+        inAnchorScope(
+            lateralJoin.getRelAnchor(),
+            recordTypeOf(newLeft),
+            () -> lateralJoin.getRight().accept(this, context));
```
andrew-coleman, line 274: The anchor scope covers the id form, but the offset form (`steps_out=1`) of the same reference falls through to `inputTypeStepsOut` and reads whatever enclosing scope is on the stack. Verified: nested in a correlated subquery it resolves against the unrelated outer relation instead of the left row.

Needs the subquery boundary too (uses the overload in the `inSubqueryScope` comment below):

```suggestion
    Type.Struct leftType = recordTypeOf(newLeft);
    // The offset form of the same reference (steps_out=1) resolves against the left row too, so the
    // right input is a subquery boundary as well as an anchor scope. Without the boundary such a
    // reference reads whatever enclosing scope happens to be on the stack.
    Optional<Rel> right =
        inSubqueryScope(
            leftType,
            () ->
                inAnchorScope(
                    lateralJoin.getRelAnchor(),
                    leftType,
                    () -> lateralJoin.getRight().accept(this, context)));
```

core/src/main/java/io/substrait/relation/RelCopyOnWriteVisitor.java
```diff
@@ -672,6 +787,132 @@ protected Optional<ConsistentPartitionWindow.WindowRelFunctionInvocation> visitW
...
+  protected <T> T outsideInputScope(CopyOnWriteUtils.ThrowingSupplier<T, E> rewrite) throws E {
+    return inInputScope(null, rewrite);
+  }
+
+  /**
+   * Records the scope currently being rewritten as an enclosing one for the duration of the given
+   * rewrite, so that an outer reference within it resolves against the right relation.
+   */
+  <T> T inSubqueryScope(CopyOnWriteUtils.ThrowingSupplier<T, E> rewrite) throws E {
+    outerInputTypes.add(currentInputType());
```
andrew-coleman, line 847: Package-private, so an out-of-package subclass overriding `visit(ScalarSubquery/SetPredicate/InPredicate)` cannot push the boundary — a nest that misses one push silently retypes a `steps_out=1` reference against the wrong relation.

`protected` is not enough here: the callers subclass `ExpressionCopyOnWriteVisitor`, not this class, so protected access does not reach across the package. Confirmed by compiling one — it needs `public`. That also makes the javadoc tags mandatory under `-Werror`.

```suggestion
   * <p>Public rather than {@code protected} because the callers that need it are subclasses of
   * {@link ExpressionCopyOnWriteVisitor} overriding a subquery {@code visit}, which are not
   * subclasses of this class and so cannot reach a protected member of it.
   *
   * @param <T> the type of the rewrite's result
   * @param rewrite the rewrite to run within the subquery boundary
   * @return the result of the rewrite
   * @throws E if the rewrite fails
   */
  public <T> T inSubqueryScope(CopyOnWriteUtils.ThrowingSupplier<T, E> rewrite) throws E {
    return inSubqueryScope(currentInputType(), rewrite);
  }

  /**
   * As {@link #inSubqueryScope(CopyOnWriteUtils.ThrowingSupplier)}, but with the enclosing scope
   * given explicitly, for a boundary whose outer references resolve against something other than
   * the scope currently being rewritten.
   *
   * @param <T> the type of the rewrite's result
   * @param outerType the record type an outer reference within the boundary resolves against
   * @param rewrite the rewrite to run within the subquery boundary
   * @return the result of the rewrite
   * @throws E if the rewrite fails
   */
  public <T> T inSubqueryScope(
      Type.Struct outerType, CopyOnWriteUtils.ThrowingSupplier<T, E> rewrite) throws E {
    outerInputTypes.add(outerType);
```

core/src/main/java/io/substrait/relation/RelCopyOnWriteVisitor.java
```diff
@@ -697,27 +939,33 @@ protected Optional<List<Expression>> visitExprList(
...
       FieldReference fieldReference, EmptyVisitationContext context) throws E {
-    Optional<Expression> inputExpression =
-        visitOptionalExpression(fieldReference.inputExpression(), context);
-    if (allEmpty(inputExpression)) {
-      return Optional.empty();
-    }
-
-    return Optional.of(FieldReference.builder().inputExpression(inputExpression).build());
+    return getExpressionCopyOnWriteVisitor().visitFieldReference(fieldReference, context);
   }
```
andrew-coleman, line 943: Bypasses `visit(FieldReference)`, so a subclass overriding the standard `ExpressionVisitor` hook — as `OuterReferenceConverter.ExprRewriter` does at :270 — is skipped for `ScatterExchange` fields and `ComparisonJoinKey` sides. Verified the hook is never called in those positions. Your own test comment notes the same asymmetry and works around it by overriding both.

```suggestion
  public Optional<FieldReference> visitFieldReference(
      FieldReference fieldReference, EmptyVisitationContext context) throws E {
    // Dispatched through visit(FieldReference) rather than straight to visitFieldReference, so that
    // a subclass overriding the standard ExpressionVisitor hook is honoured in the positions that
    // hold a reference rather than an arbitrary expression.
    return getExpressionCopyOnWriteVisitor()
        .visit(fieldReference, context)
        .map(
            rewritten -> {
              if (!(rewritten instanceof FieldReference)) {
                throw new IllegalStateException(
                    "A field reference position can only be rewritten to another field reference; "
                        + "got "
                        + rewritten.getClass().getName());
              }
              return (FieldReference) rewritten;
            });
  }
```

core/src/main/java/io/substrait/relation/RelCopyOnWriteVisitor.java
```diff
@@ -237,7 +300,8 @@ public Optional<Rel> visit(Set set, EmptyVisitationContext context) throws E {
 
   @Override
   public Optional<Rel> visit(NamedScan namedScan, EmptyVisitationContext context) throws E {
-    Optional<Expression> filter = visitOptionalExpression(namedScan.getFilter(), context);
+    Optional<Expression> filter =
+        outsideInputScope(() -> visitOptionalExpression(namedScan.getFilter(), context));
```
andrew-coleman, line 304: The root type is on the relation (`getInitialSchema().struct()`, matching `ProtoRelConverter:484`), so a stale reference type here is never corrected — the case `multiBucketExchangeExpressionIsRetypedWithAnUnchangedInput` pins elsewhere. Pre-existing on `main`, so not a regression, just not closed.

Same for `LocalFiles` (:315), `VirtualTableScan` (:576), `ExtensionTable` (:613), and `NamedUpdate` (:400, via `getTableSchema().struct()`). `bestEffortFilter` is still not traversed at all.

```suggestion
    Optional<Expression> filter =
        inInputScope(
            namedScan.getInitialSchema().struct(),
            () -> visitOptionalExpression(namedScan.getFilter(), context));
```

andrew-coleman (commented):
Scope tracking checks out, and the `visitFieldReference` bug it fixes is real.

Five comments below. One is a regression — `recordTypeOf` runs unconditionally, so a pure traversal over a `Set` with mismatched field counts now throws where it returned on `main`. The rest are silent mis-typing or a bypassed extension point.

I applied all five and confirmed each reproduces on head and flips with the fix; `:core:test` 829 pass, spotless and javadoc clean. Two corrections from testing: `inSubqueryScope` needs `public` not `protected`, and the read-relation gap is pre-existing rather than new. Patch available if that beats five suggestions.

Issue #847, opened 2026-06-04
Support type variations in simple extension YAML files


nielspardon:
## Context

While implementing deprecation info support for extensions (#811, PR #846), I noticed that substrait-java does not model **type variations** declared in simple extension YAML files at all.

The Substrait simple-extension schema supports a `type_variations` section (see [`text/simple_extensions_schema.yaml`](https://github.com/substrait-io/substrait/blob/main/text/simple_extensions_schema.yaml)), but substrait-java's `ExtensionSignatures` only parses `types`, `scalar_functions`, `aggregate_functions`, `window_functions`, and `metadata`:

https://github.com/substrait-io/substrait-java/blob/main/core/src/main/java/io/substrait/extension/SimpleExtension.java

Because there is no model object for a type variation, any `type_variations` entries in a loaded extension file are silently ignored.

## Impact

This came up directly in PR #846: Substrait v0.86.0 ([substrait-io/substrait#1014](https://github.com/substrait-io/substrait/pull/1014)) added `deprecated` support to type variations among other extension entries. That deprecation could be implemented for types and functions, but **not** for type variations, since type variations are not represented in the model — there is nothing to attach the deprecation info to.

## Proposed change

- Add a `TypeVariation` model class (with `name`, `description`, `metadata`, `deprecated`, and the variation's base type / nullability behavior as defined by the schema) to `SimpleExtension`.
- Parse the `type_variations` section in `ExtensionSignatures`.
- Expose the parsed type variations from `ExtensionCollection` (lookup by anchor, similar to `getType`).

## Note

This is distinct from #567, which concerns `type_variation_reference` on user-defined types at the proto/plan level. This issue is about parsing/representing the `type_variations` **declarations** from extension YAML files.

PR #1062, opened 2026-08-05
feat(core)!: support nested map expressions and import nested structs

### LINKED ISSUE #375: support for Nested expressions

nielspardon:
Proto `Expression.Nested` has three arms — struct, list and map — but only `list` round-tripped. `Expression.NestedMap` did not exist at all, so the `map` arm was unreachable, and `ProtoExpressionConverter.from(Nested)` handled only `case LIST`: a plan carrying a nested struct failed on import with `UnsupportedOperationException: Unimplemented nested type: STRUCT`, even though the POJO to proto direction emitted one. (`VirtualTableScan` rows were unaffected — they use the separate `from(Nested.Struct)` overload.)

This adds the `NestedMap` POJO with the usual visitor and converter wiring, and imports all three nested kinds on the way back from proto.

Two details worth calling out:

- **The pairs are a list, not a map.** Proto `Nested.Map` is a `repeated KeyValue key_values` with no uniqueness constraint, so a map expression may repeat a key, and the order of the pairs is part of the value. `NestedMap` holds `List<NestedMap.KeyValue>` accordingly. A `Map`-backed accessor turned a valid plan into a different one on import, since the second of two equal keys overwrote the first and a pair disappeared. The pre-existing `MapLiteral` has the same limitation, tracked separately in #1081.
- **`NestedList` and `NestedMap` now agree on what a homogeneous collection is**, through one shared `commonType` helper used by both `check()` and `getType()`. It compares element types with nullability disregarded, because SQL collection constructors do not cast their operands to a common type — `ARRAY[not_null_column, nullable_column]` is a legal plan, and an importer must not reject it — and it returns the nullable union, so the reported element type does not depend on the order of the values. Validation throws `IllegalArgumentException` rather than asserting, matching the direction #1058 set.

Isthmus still cannot convert nested structs or maps to and from Calcite; that follows in a stacked PR, which closes #375.

`Expression.Nested.type_variation_reference` is still dropped on import, as it is for literals and for every type but `Type.UserDefined`; that gap is model-wide rather than specific to the new arms, and is tracked in #1091.

Part of #375

BREAKING CHANGE: `ExpressionVisitor` gains `visit(Expression.NestedMap)`. Direct implementors must add it; implementors extending `AbstractExpressionVisitor` inherit the `visitFallback` default and need no change. `NestedList.getType()` now reports a nullable element type when any value is nullable, rather than the type of the first value. Importing an `Expression.Nested` with no arm set now throws `IllegalArgumentException` rather than `UnsupportedOperationException`.


core/src/main/java/io/substrait/expression/Expression.java
```diff
@@ -1819,6 +1819,64 @@ public static ImmutableExpression.NestedList.Builder builder() {
...
+   * EmptyMapLiteral}.
+   */
+  @Value.Immutable
+  abstract class NestedMap implements Nested {
+    /**
+     * Returns the key-value pairs in this nested map, in the order they were added.
+     *
+     * @return the key-value pairs
+     */
+    public abstract Map<Expression, Expression> values();
```
alexandrefimov: Could we avoid representing `key_values` as `Map<Expression, Expression>` here? Substrait maps allow repeated keys. During proto import, `LinkedHashMap.put()` overwrites an earlier entry when two key expressions are equal, so a valid plan is changed by a round trip. I reproduced this with two `i32(1)` keys and different values: the output had one `key_values` entry instead of two.

An ordered list of key/value pair objects would preserve repeated keys and serialization order. The existing literal-map POJO may have the same limitation, but I do not think we should add it to this new public API.

nielspardon: Good catch, and reproduced — two `i32(1)` keys with different values came back as one entry, and it was the *first* pair that was dropped. `NestedMap` now holds `List<NestedMap.KeyValue>` rather than a `Map`, mirroring the proto's `repeated KeyValue key_values`, so repeated keys and pair order both survive.

#1063 needed the same treatment in the other direction: `MAP['a', 1, 'a', 2]` was collapsing to a `MapLiteral` and losing a pair, so that collapse now requires distinct keys too.

You're right about the literal map, and it's a bit worse than lossy — proto import there uses `Collectors.toMap`, which throws `IllegalStateException` on a duplicate key. Agreed it doesn't belong in this PR; filed as #1081.

alexandrefimov: Nice, thanks — I checked the latest revision and the duplicate-key case now round-trips correctly. This resolves my concern.

alexandrefimov (approved):

core/src/main/java/io/substrait/expression/Expression.java
```diff
@@ -1819,6 +1819,104 @@ public static ImmutableExpression.NestedList.Builder builder() {
...
+      if (keyValues().isEmpty()) {
+        throw new IllegalArgumentException(
+            "To specify an empty map, use ExpressionCreator.emptyMap");
+      }
+      if (keyValues().stream().map(keyValue -> keyValue.key().getType()).distinct().count() > 1) {
+        throw new IllegalArgumentException("All keys in a NestedMap must have the same type");
+      }
+      if (keyValues().stream().map(keyValue -> keyValue.value().getType()).distinct().count() > 1) {
+        throw new IllegalArgumentException("All values in a NestedMap must have the same type");
+      }
```
andrew-coleman: Reached from `from(Nested)` line 504, so these reject legal plans on import:

```
map, values i32? and i32            | IllegalArgumentException: All values in a NestedMap must have the same type
map, keys string? and string        | IllegalArgumentException: All keys in a NestedMap must have the same type
map, values DEC<10,2> and DEC<12,2> | IllegalArgumentException: All values in a NestedMap must have the same type
```

The `LIST` arm accepts the same variance today, so the two arms disagree.

```suggestion
      if (keyValues().stream()
              .map(keyValue -> TypeCreator.asNullable(keyValue.key().getType()))
              .distinct()
              .limit(2)
              .count()
          > 1) {
        throw new IllegalArgumentException("All keys in a NestedMap must have the same type");
      }
      if (keyValues().stream()
              .map(keyValue -> TypeCreator.asNullable(keyValue.value().getType()))
              .distinct()
              .limit(2)
              .count()
          > 1) {
        throw new IllegalArgumentException("All values in a NestedMap must have the same type");
      }
```

`NestedList` needs the same normalization (#1058 converts its `assert` to a throw) — worth one shared helper that `check()` and `getType()` both use. No test covers a nullability-only mismatch.

core/src/main/java/io/substrait/expression/Expression.java
```diff
@@ -1819,6 +1819,104 @@ public static ImmutableExpression.NestedList.Builder builder() {
...
+    }
+
+    @Override
+    public Type getType() {
+      KeyValue first = keyValues().get(0);
+      return Type.withNullability(nullable()).map(first.key().getType(), first.value().getType());
```
andrew-coleman: Type comes from pair 0 with no union, so once `check()` accepts nullability variance this is order-dependent: `MAP['a', 1, 'b', NULL::i32]` reports `map<string, i32>`, the reordered equivalent `map<string, i32?>`. Should be the nullable union of all keys and all values.

core/src/main/java/io/substrait/expression/Expression.java
```diff
@@ -1819,6 +1819,104 @@ public static ImmutableExpression.NestedList.Builder builder() {
...
+
+    /**
+     * Validates that the nested map is not empty and that all keys, and all values, have a single
+     * common type.
+     */
+    @Value.Check
+    protected void check() {
+      if (keyValues().isEmpty()) {
+        throw new IllegalArgumentException(
+            "To specify an empty map, use ExpressionCreator.emptyMap");
```
andrew-coleman: This runs as the importer too, where "use `ExpressionCreator.emptyMap`" is not actionable — the caller is reading someone else's plan. Same for the two type messages below.

core/src/main/java/io/substrait/expression/proto/ProtoExpressionConverter.java
```diff
@@ -481,13 +481,29 @@ private WindowBound toWindowBound(io.substrait.proto.Expression.WindowFunction.B
    */
   public Expression.Nested from(io.substrait.proto.Expression.Nested nested) {
```
andrew-coleman, line 482: `Expression.Nested` carries `type_variation_reference`, which no arm reads and the POJO cannot represent. Round trip drops it:

```
tvr struct in=7 out=0 DROPPED
tvr list   in=7 out=0 DROPPED
tvr map    in=7 out=0 DROPPED
```

Pre-existing for LIST; this extends it to STRUCT and MAP. Either model it or reject a non-zero value on import.

core/src/main/java/io/substrait/expression/proto/ProtoExpressionConverter.java
```diff
@@ -481,13 +481,29 @@ private WindowBound toWindowBound(io.substrait.proto.Expression.WindowFunction.B
...
         return ExpressionCreator.nestedList(nested.getNullable(), list);
+      case MAP:
+        // The pairs are kept in a list, in the order the producer emitted them, so that a map
+        // repeating a key keeps both of its pairs.
+        List<Expression.NestedMap.KeyValue> keyValues =
+            nested.getMap().getKeyValuesList().stream()
+                .map(
+                    keyValue ->
+                        Expression.NestedMap.KeyValue.of(
+                            from(keyValue.getKey()), from(keyValue.getValue())))
```
andrew-coleman: `hasKey()`/`hasValue()` are unchecked, so a pair missing either side lands on line 364 instead:

```
map pair with no value | IllegalArgumentException: Unknown type: REXTYPE_NOT_SET
map pair with no key   | IllegalArgumentException: Unknown type: REXTYPE_NOT_SET
```

Names neither the map nor the side.

core/src/main/java/io/substrait/expression/proto/ProtoExpressionConverter.java
```diff
@@ -481,13 +481,29 @@ private WindowBound toWindowBound(io.substrait.proto.Expression.WindowFunction.B
...
+                .collect(Collectors.toList());
+        return ExpressionCreator.nestedMap(nested.getNullable(), keyValues);
       default:
-        throw new UnsupportedOperationException(
-            "Unimplemented nested type: " + nested.getNestedTypeCase());
+        throw new IllegalStateException("Unexpected nested type: " + nested.getNestedTypeCase());
```
andrew-coleman: Base threw `UnsupportedOperationException("Unimplemented nested type: …")` here; this is a public-API change not in the BREAKING CHANGE note. The only reachable case is now `NESTEDTYPE_NOT_SET` — bad input — where this file's convention is `IllegalArgumentException` (line 472).

```suggestion
      default:
        throw new IllegalArgumentException(
            "Unsupported nested type: " + nested.getNestedTypeCase());
```

andrew-coleman (commented):
One blocker: `NestedMap.check()` rejects keys or values differing only in nullability, and it runs on the proto import path, so a legal plan fails to import. Comments inline.

Also lossy: `type_variation_reference` on `Expression.Nested` is dropped on import for all three arms.

core/src/main/java/io/substrait/relation/ExpressionCopyOnWriteVisitor.java
```diff
@@ -402,6 +403,26 @@ public Optional<Expression> visit(Expression.NestedList expr, EmptyVisitationCon
             Expression.NestedList.builder().from(expr).values(expressionList).build());
   }
 
+  @Override
+  public Optional<Expression> visit(Expression.NestedMap expr, EmptyVisitationContext context)
+      throws E {
```
andrew-coleman, line 407: Every other composite here is `transformList(items, context, this::visitXxx)` with an overridable `protected visitXxx` (`visitSwitchClause:246`, `visitIfClause:284`, `visitSortField:340`, `visitMultiOrListRecord:383`). This inlines the changed-flag/ArrayList logic and exposes no hook, so a subclass rewriting pairs must re-override the whole method. No test exercises it.

core/src/main/java/io/substrait/relation/ExpressionCopyOnWriteVisitor.java
```diff
@@ -402,6 +403,26 @@ public Optional<Expression> visit(Expression.NestedList expr, EmptyVisitationCon
...
+    boolean changed = false;
+    List<Expression.NestedMap.KeyValue> keyValues = new ArrayList<>();
+    for (Expression.NestedMap.KeyValue keyValue : expr.keyValues()) {
+      Optional<Expression> key = keyValue.key().accept(this, context);
+      Optional<Expression> value = keyValue.value().accept(this, context);
+      changed |= !allEmpty(key, value);
+      keyValues.add(
+          Expression.NestedMap.KeyValue.of(
+              key.orElse(keyValue.key()), value.orElse(keyValue.value())));
```
andrew-coleman: `builder().from(original)` is the convention here (`visitMultiOrListRecord:433`, `visit(NestedList):403`) — `of(...)` drops any attribute `KeyValue` later gains.

```suggestion
      keyValues.add(
          Expression.NestedMap.KeyValue.builder()
              .from(keyValue)
              .key(key.orElse(keyValue.key()))
              .value(value.orElse(keyValue.value()))
              .build());
```

Same at `ProtoExpressionConverter.java:501`.

core/src/main/java/io/substrait/expression/proto/ExpressionProtoConverter.java
```diff
@@ -616,6 +616,23 @@ public Expression visit(
...
+        bldr -> {
+          Expression.Nested.Map.Builder mapBldr = Expression.Nested.Map.newBuilder();
+          for (io.substrait.expression.Expression.NestedMap.KeyValue keyValue : expr.keyValues()) {
+            mapBldr.addKeyValues(
+                Expression.Nested.Map.KeyValue.newBuilder()
+                    .setKey(toProto(keyValue.key()))
+                    .setValue(toProto(keyValue.value())));
+          }
+          bldr.setMap(mapBldr).setNullable(expr.nullable());
+        });
```
andrew-coleman, line 641: `visit(NestedStruct):409` and `visit(NestedList):605` both stream + `addAll`; this one mutates a local builder in a loop.

```suggestion
    return nested(
        bldr ->
            bldr.setMap(
                    Expression.Nested.Map.newBuilder()
                        .addAllKeyValues(
                            expr.keyValues().stream()
                                .map(
                                    keyValue ->
                                        Expression.Nested.Map.KeyValue.newBuilder()
                                            .setKey(toProto(keyValue.key()))
                                            .setValue(toProto(keyValue.value()))
                                            .build())
                                .collect(Collectors.toList())))
                .setNullable(expr.nullable()));
```

core/src/main/java/io/substrait/expression/ExpressionCreator.java
```diff
@@ -571,6 +571,22 @@ public static Expression.NestedStruct nestedStruct(boolean nullable, Expression.
...
+   * <p>Note: This method cannot be used to construct an empty map. To create an empty map, use
+   * {@link ExpressionCreator#emptyMap(boolean, Type, Type)} which returns an {@link
+   * Expression.EmptyMapLiteral}.
+   *
+   * @param nullable whether the map can be null
+   * @param keyValues the key-value pairs in the nested map, in the order they should be serialized
+   * @return a NestedMap expression
+   */
+  public static Expression.NestedMap nestedMap(
+      boolean nullable, List<Expression.NestedMap.KeyValue> keyValues) {
```
andrew-coleman: `nestedStruct` has `Iterable` and varargs overloads (571, and the one above it); `nestedMap(false, KeyValue.of(k, v))` does not compile. Worth matching.

core/src/test/java/io/substrait/type/proto/NestedMapExpressionTest.java
```diff
@@ -0,0 +1,137 @@
...
+    Expression.NestedMap nestedMap =
+        Expression.NestedMap.builder().addAllKeyValues(keyValues).build();
+
+    assertEquals(keyValues, nestedMap.keyValues());
+    verifyRoundTrip(projectOf(nestedMap));
+  }
+
+  private Project projectOf(Expression expression) {
+    return Project.builder().addExpressions(expression).input(sb.emptyVirtualTableScan()).build();
+  }
```
andrew-coleman: `NestedStructExpressionTest:65` has the same helper, and `NestedListExpressionTest` inlines it four times. Moving `projectOf` to `TestBase` removes all three.

nielspardon:
Thanks — all of it applied in 3db51b4e, on top of a merge of `main` (so #1058 is in). Pushed as a separate commit rather than an amend so the delta is reviewable on its own.

The validation comments turned out to be one change, not three. `NestedList` and `NestedMap` now share a `commonType` helper that both `check()` and `getType()` call: it normalizes nullability before comparing, so the two arms agree and a legal plan mixing nullable and non-nullable operands imports; and it returns the nullable union, so the reported element type is order-independent. `NestedList.getType()` changing shape is a behavior change to what #1058 documented, so it is in the BREAKING CHANGE footer along with the `UnsupportedOperationException` → `IllegalArgumentException` one you flagged. The helper is a package-private class rather than a private interface method because `:core` main compiles with `--release 8`.

`DEC<10,2>` vs `DEC<12,2>` stays rejected — nullability is the only variance SQL constructors introduce without a cast, and differing precision is a genuinely different type. There is a test for it now, and for a nullability-only mismatch in both directions.

Two places where I did not take the suggestion:

- `type_variation_reference` — filed #1091 rather than handling it here. substrait-java drops it for `Expression.Literal` and for all 26 messages in `type.proto` too; `Type.UserDefined` is the only place it is modelled. Rejecting a non-zero value on just the `Nested` arm would make us refuse plans we accept today while still silently dropping the same field one level down in the element types, so it reads as a model-wide change rather than part of this one. #1091 proposes modelling it in all three places. (Distinct from #847, which is about parsing the `type_variations` declarations out of the extension YAML.)
- `KeyValue.of(...)` at `ProtoExpressionConverter` — kept, since there is no original pair to copy from on the import path; `builder().from(original)` is what the copy-on-write visitor needed and it now uses it.

core/src/main/java/io/substrait/expression/proto/ProtoExpressionConverter.java
nielspardon: Filed #1091 rather than handling it here — the field is dropped for `Expression.Literal` and for every message in `type.proto` too, so rejecting it on just this arm would refuse plans we accept today while still dropping the same field in the element types. Details in the PR comment.

core/src/main/java/io/substrait/relation/ExpressionCopyOnWriteVisitor.java
nielspardon: Applied here. Left `of(...)` at `ProtoExpressionConverter:501` — on the import path there is no original pair to copy from.

andrew-coleman (approved):

PR #1063, opened 2026-08-05
feat(isthmus): convert nested struct and map expressions to and from Calcite

### LINKED ISSUE #375: support for Nested expressions

nielspardon:
Builds on #1062, now merged, which added the `NestedMap` POJO and the proto import this needs.

Of the three nested expression kinds, only nested lists survived a trip through Calcite. In the Substrait to Calcite direction `ExpressionRexConverter` had no case for `NestedStruct` or `NestedMap`, so both reached `visitFallback` and threw `UnsupportedOperationException`. Coming back the other way, `CallConverters.ROW` rejected any ROW whose fields were not all literals with "ROW operands must be literals.", and the map value constructor cast every operand to `Expression.Literal` — so `SELECT ROW(a + 1, b)` and `SELECT MAP['key', a + 1]` failed with an `IllegalArgumentException` and a `ClassCastException` respectively.

Nested structs now become Calcite ROW calls and nested maps become `MAP_VALUE_CONSTRUCTOR` calls, each carrying its own nullability in the call type. In reverse, a ROW or map constructor whose operands are not all literals falls back to a `NestedStruct` or `NestedMap` — the same treatment array value constructors already gave nested lists — while the all-literal cases still collapse to a `StructLiteral` or `MapLiteral`.

A map constructor has to clear one more bar than a ROW to collapse: its keys must also be distinct. A `MapLiteral` holds its pairs in a `Map`, so `MAP['key', 1, 'key', 2]` came back with one pair instead of two; it now stays a `NestedMap`, which keeps its pairs in a list. Same limitation as the one #1062 fixed, reached from the SQL side instead of from proto.

## Nullability has to survive that collapse

`nullable` on a Substrait literal marks the literal's *type* as nullable, not the value as null — a value that is actually null is a `NullLiteral`. The ROW and map converters were hardcoding `nullable=false`, so a nullable struct or map of literals silently came back non-nullable. They now take it from the call type, which is what the array value constructor already did for lists.

The same gap existed in the other direction: the list and map literal visitors let Calcite infer the container type from the elements, which discards the container's own nullability, so they now pass the converted type like their struct and nested counterparts.

Struct-level nullability is deliberately kept *inside* `StructLiteral` rather than routed out to `NestedStruct`. Routing looks tidier and gives an exact round trip, but it breaks the user-defined type struct encoding, which recognises its payload by it being a `StructLiteral`: a UDT literal with a nullable struct field produces a nullable inner ROW, that ROW becomes a `NestedStruct`, the enclosing UDT ROW then fails its all-literals test too, and `REINTERPRET` can no longer extract the literal — `IllegalArgumentException: Unable to convert call Reinterpret(struct<struct<i32>?, i32>)`. `structEncodedUdtWithNullableStructFieldRoundTrip` locks that shape down.

One consequence to be aware of: Calcite makes every field of a nullable record type nullable, so a nullable struct widens its field types on the way through. That is Calcite's representation of a nullable row, not something this change can avoid.

## Mixed literal and non-literal entries, and what remains of #1066

Calcite types every non-null literal NOT NULL while preserving nullability on everything else, so a nested value that mixes a nullable literal with a nullable non-literal of the same declared type came back with entry types that no longer matched. `NestedList` and `NestedMap` used to reject that outright. As of #1062 they compare their element types disregarding nullability and report the nullable union, so such a value converts, and the container's own type survives the round trip — `list<i32?>` in, `list<i32?>` out.

What is still lost is the nullability of the individual literal, which comes back non-nullable. So the round trip is not identity, and that part of #1066 stays open; fixing it needs literals to carry their nullability across the Calcite boundary, which is a change to every literal visitor rather than to the nested arms. `nestedListMixingNullableLiteralAndFieldReferenceTest` and its map counterpart pin down where the behaviour stands.

## Incidental fix in the map converter

The literal map is built in operand order rather than `HashMap` order, so the emitted `key_values` no longer depend on hash iteration order.

Closes #375

### DIFF
--- a/isthmus/src/main/java/io/substrait/isthmus/expression/CallConverters.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/expression/CallConverters.java
@@ -127,14 +127,18 @@ public class CallConverters {
           };
 
   /**
-   * Converts Calcite ROW constructors into Substrait {@link Expression.StructLiteral}s.
+   * Converts Calcite ROW constructors into Substrait {@link Expression.StructLiteral}s, or into
+   * {@link Expression.NestedStruct}s when the fields are not all literals.
    *
-   * <p>ROW values are always concrete (never null themselves) - if a value is actually null, use
-   * NullLiteral instead of StructLiteral. Therefore, the resulting StructLiteral always has
-   * nullable=false. The ROW's type may be nullable (for regular structs) or non-nullable (for UDT
-   * struct encoding), but the value itself is always concrete.
+   * <p>Either way the struct takes its nullability from the ROW's type, which is where Calcite
+   * keeps it. On a Substrait literal, {@code nullable} marks the literal's type as nullable rather
+   * than the value as null - a value that is actually null is a NullLiteral - so a nullable ROW of
+   * literals does not have to give up being a StructLiteral. The UDT struct encoding depends on
+   * that: it builds a deliberately non-nullable ROW, keeping the user-defined type's own
+   * nullability in the REINTERPRET target type, and so still arrives here as a StructLiteral.
    *
-   * <p>Each literal's nullability is set to match its field type's nullability.
+   * <p>Each literal's nullability is set to match its field type's nullability. Note that Calcite
+   * makes every field of a nullable record type nullable, so a nullable ROW widens its fields.
    */
   public static final SimpleCallConverter ROW =
       (call, visitor) -> {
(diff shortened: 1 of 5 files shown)


bvolpato (approved):
LGTM!

PR #1065, opened 2026-08-05
fix(isthmus)!: use schema types for virtual table literals

### LINKED ISSUE #1064: LogicalValues conversion fails when tuple literal type is narrower than row schema
## Problem

`SubstraitRelVisitor.visit(Values)` currently uses the `LogicalValues` row schema
only to override literal nullability. The remaining Substrait literal type is
derived from the tuple's `RexLiteral` type.

Calcite can represent a `LogicalValues` field with a wider row type than its
literal. For example, an `INTEGER` row field may contain a `TINYINT` literal.
Converting that relation creates an `I32` virtual-table schema and an `I8` row
field. `VirtualTableScan` then rejects the mismatch:

```text
Row field type (I8{nullable=false}) does not match schema field type (I32{nullable=false})
```

This is the same class of schema/literal mismatch addressed for nullability by
#683 and #684, but it also affects type width and other schema type attributes.

## Minimal reproducer

```java
RelDataType rowType =
    typeFactory.builder().add("col1", SqlTypeName.INTEGER).build();
RexLiteral literal =
    builder
        .getRexBuilder()
        .makeExactLiteral(
            BigDecimal.ONE, typeFactory.createSqlType(SqlTypeName.TINYINT));
LogicalValues values =
    LogicalValues.create(
        builder.getCluster(), rowType, ImmutableList.of(ImmutableList.of(literal)));

SubstraitRelVisitor.convert(values, converterProvider);
```

## Expected behavior

The virtual-table row should use the complete type of the corresponding
`LogicalValues` schema field. The example should produce an `I32` literal with
value `1`, matching the `I32` schema.

Schema validation should remain strict; this change should normalize compatible
literals to the schema type rather than relaxing `VirtualTableScan` validation.

## Proposed direction

Pass the schema field's `RelDataType` to `LiteralConverter` when converting
`LogicalValues` tuples. Use that type for width, precision, scale and
nullability while retaining the literal's value.

alexandrefimov:
## Summary

- convert `LogicalValues` tuple literals using the complete row-schema field type;
- preserve the existing literal-based behavior for other expression conversions;
- add a regression test for a `TINYINT` tuple literal in an `INTEGER` field.

`VirtualTableScan` requires row field types to exactly match its schema. Calcite
may infer a narrower type for a tuple literal than for the corresponding
`LogicalValues` row field — `Values.assertRowType()` only asserts that the field
type `canAssignFrom` the literal type, so a `TINYINT` literal is a legal member of
an `INTEGER` row. Previously Isthmus copied only schema nullability, so the
conversion could construct an `I8` row under an `I32` schema and fail its own
validation.

`LiteralConverter.convert` now takes the schema `RelDataType`, and derives
nullability from it rather than from a separate flag. Existing callers continue to
derive the result type from the literal itself.

Fixes #1064

BREAKING CHANGE: `io.substrait.isthmus.expression.LiteralConverter#convert(RexLiteral, boolean)` is removed. Use `convert(RexLiteral, RelDataType)`, which takes nullability from the supplied Calcite type; callers needing a nullability other than the literal's own should widen the type with `RelDataTypeFactory#createTypeWithNullability` first.

### DIFF
--- a/isthmus/src/main/java/io/substrait/isthmus/SubstraitRelVisitor.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/SubstraitRelVisitor.java
@@ -163,19 +163,19 @@ public class SubstraitRelVisitor extends RelNodeVisitor<Rel, RuntimeException> {
     NamedStruct type = typeConverter.toNamedStruct(values.getRowType());
 
     LiteralConverter literalConverter = new LiteralConverter(typeConverter);
-    List<Type> schemaFieldTypes = type.struct().fields();
     List<Expression.NestedStruct> structs =
         values.getTuples().stream()
             .map(
                 list -> {
-                  // Use schema nullability since Calcite infers non-nullable for all non-null
-                  // values
+                  // Calcite may infer a narrower type for a tuple literal than for its row field.
+                  // Virtual table rows must use the complete schema type.
                   List<Expression> fields =
                       IntStream.range(0, list.size())
                           .mapToObj(
                               i ->
                                   literalConverter.convert(
-                                      list.get(i), schemaFieldTypes.get(i).nullable()))
+                                      list.get(i),
+                                      values.getRowType().getFieldList().get(i).getType()))
                           .collect(Collectors.toUnmodifiableList());
                   return ExpressionCreator.nestedStruct(false, fields);
                 })
--- a/isthmus/src/main/java/io/substrait/isthmus/SubstraitRelVisitor.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/SubstraitRelVisitor.java
@@ -56,6 +56,7 @@ import org.apache.calcite.rel.core.TableModify;
 import org.apache.calcite.rel.logical.LogicalProject;
 import org.apache.calcite.rel.type.RelDataType;
 import org.apache.calcite.rel.type.RelDataTypeFactory;
+import org.apache.calcite.rel.type.RelDataTypeField;
 import org.apache.calcite.rex.RexBuilder;
 import org.apache.calcite.rex.RexInputRef;
 import org.apache.calcite.rex.RexNode;
@@ -161,6 +162,7 @@ public class SubstraitRelVisitor extends RelNodeVisitor<Rel, RuntimeException> {
   @Override
   public Rel visit(org.apache.calcite.rel.core.Values values) {
     NamedStruct type = typeConverter.toNamedStruct(values.getRowType());
+    List<RelDataTypeField> rowFields = values.getRowType().getFieldList();
 
     LiteralConverter literalConverter = new LiteralConverter(typeConverter);
     List<Expression.NestedStruct> structs =
@@ -168,14 +170,12 @@ public class SubstraitRelVisitor extends RelNodeVisitor<Rel, RuntimeException> {
             .map(
                 list -> {
                   // Calcite may infer a narrower type for a tuple literal than for its row field.
-                  // Virtual table rows must use the complete schema type.
+                  // Virtual table rows must use the row field type.
                   List<Expression> fields =
                       IntStream.range(0, list.size())
                           .mapToObj(
                               i ->
-                                  literalConverter.convert(
-                                      list.get(i),
-                                      values.getRowType().getFieldList().get(i).getType()))
+                                  literalConverter.convert(list.get(i), rowFields.get(i).getType()))
                           .collect(Collectors.toUnmodifiableList());
                   return ExpressionCreator.nestedStruct(false, fields);
                 })
--- a/isthmus/src/main/java/io/substrait/isthmus/expression/LiteralConverter.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/expression/LiteralConverter.java
@@ -96,39 +96,23 @@ public class LiteralConverter {
   }
 
   /**
-   * Converts a RexLiteral to a Substrait Literal with the specified nullability.
-   *
-   * <p>This overload is useful when the target nullability should come from the schema rather than
-   * the literal's own type. For example, Calcite's LogicalValues may have literals with
-   * non-nullable types even when the schema field is nullable.
-   *
-   * @param literal the RexLiteral to convert
-   * @param nullable the nullability to use for the resulting Substrait literal
-   * @return the converted Substrait Literal
-   */
-  public Expression.Literal convert(RexLiteral literal, boolean nullable) {
-    return convert(literal, literal.getType(), nullable);
-  }
-
-  /**
-   * Converts a RexLiteral to a Substrait Literal using the specified result type.
+   * Converts a RexLiteral to a Substrait Literal carrying the given result type.
    *
    * <p>This overload is useful when the target type comes from a containing schema rather than the
    * literal itself. Calcite may infer a narrower type for a value in a LogicalValues tuple than for
-   * the corresponding row field.
+   * the corresponding row field. Nullability is taken from {@code resultType}, so callers that need
+   * a nullability other than the literal's own should widen the Calcite type with {@link
+   * org.apache.calcite.rel.type.RelDataTypeFactory#createTypeWithNullability} before calling.
    *
    * @param literal the RexLiteral to convert
    * @param resultType the Calcite type required by the containing schema
    * @return the converted Substrait Literal
    */
   public Expression.Literal convert(RexLiteral literal, RelDataType resultType) {
-    Type type = typeConverter.toSubstrait(resultType);
-    return convert(literal, resultType, type.nullable());
-  }
-
-  private Expression.Literal convert(RexLiteral literal, RelDataType resultType, boolean nullable) {
+    // convert type first to guarantee we can handle the value.
+    final Type type = typeConverter.toSubstrait(resultType);
+    final boolean nullable = type.nullable();
     if (literal.isNull()) {
-      final Type type = typeConverter.toSubstrait(resultType);
       final Type typeWithNullability =
           nullable ? TypeCreator.asNullable(type) : TypeCreator.asNotNullable(type);
       return ExpressionCreator.typedNull(typeWithNullability);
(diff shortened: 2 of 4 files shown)


alexandrefimov:
@nielspardon would you mind taking a look at this one? Small isthmus fix for #1064: `VirtualTableScan` requires row field types to match its schema exactly, but Calcite can infer a narrower type for a tuple literal than for the corresponding `LogicalValues` row field — so the conversion could build an `I8` row under an `I32` schema and then fail its own validation.

isthmus/src/main/java/io/substrait/isthmus/expression/LiteralConverter.java
```diff
@@ -108,14 +107,34 @@ public Expression.Literal convert(RexLiteral literal) {
    * @return the converted Substrait Literal
    */
   public Expression.Literal convert(RexLiteral literal, boolean nullable) {
```
vbarua: I would advocate for getting rid of this method entirely actually.

It's not used internally after your changes, and any user that wants to set the nullability explicitly on the Calcite type can use the `RelDataTypeFactory#createTypeWithNullability` to set it, and then pass it to the new method below.

isthmus/src/main/java/io/substrait/isthmus/expression/LiteralConverter.java
```diff
@@ -108,14 +107,34 @@ public Expression.Literal convert(RexLiteral literal) {
...
+   * @param literal the RexLiteral to convert
+   * @param resultType the Calcite type required by the containing schema
+   * @return the converted Substrait Literal
+   */
+  public Expression.Literal convert(RexLiteral literal, RelDataType resultType) {
+    Type type = typeConverter.toSubstrait(resultType);
+    return convert(literal, resultType, type.nullable());
+  }
+
+  private Expression.Literal convert(RexLiteral literal, RelDataType resultType, boolean nullable) {
```
vbarua: Getting rid of the `boolean nullable` would let us get rid of the `nullable` flag here and just rely on the `resultType` as being the actual desired result type with the correct nullability.

vbarua:
To confirm two things.

In your test, you construct a LogicalValue containing literals with different but compatible(?) types than the row type of the LogicalValue relation, and that ends up being valid because Calcite checks if the type can be assigned to value?
https://github.com/apache/calcite/blob/052a35015c56db2927727afd7e08ed7a18a26ace/core/src/main/java/org/apache/calcite/rel/core/Values.java#L169-L190. The issue that you've identified is that Calcite isn't as strict as Substrait, so when we convert from Calcite we can't rely on the literal types and need to override them with the row type?

In Substrait, we assume that all literals in VirtualTables have the same types as given in the schema. When we convert from Substrait to Calcite, do the tuple values in the Calcite LogicalValues relation this, or is that conversion also potentially lossy?

alexandrefimov:
@vbarua both confirmed.

On strictness — the asymmetry is `canAssignFrom` against `VirtualTableScan`'s `@Value.Check`, which requires `rowFieldType.equals(schemaFieldType)`, nullability included.

The reverse direction isn't lossy, and structurally so: that same check makes a mismatched `VirtualTableScan` unconstructible, so nothing reaches the converter to lose. `visit(VirtualTableScan)` then derives the row type and every literal from the same `TypeConverter` over the same Substrait types. Checked i8, i32, i64, decimal(10,2), varchar(5), char(3) and string — every literal type is equal to its row field type, not merely assignable.

Both inline suggestions applied; the javadoc now points at `createTypeWithNullability` for callers wanting a nullability other than the literal's own. That does make it an API-visible removal — `convert(RexLiteral, boolean)` has been public since #684, so by the standard just applied in #1014 it's breaking. Retitled `fix(isthmus)!` with a footer; `releaseRules` maps that to a minor bump. Say the word if you'd rather split the removal out and keep this one non-breaking.

isthmus/src/main/java/io/substrait/isthmus/expression/LiteralConverter.java
alexandrefimov: Removed. It has been public since #684, so the PR is now `fix(isthmus)!` with a footer pointing at `convert(RexLiteral, RelDataType)`.

isthmus/src/main/java/io/substrait/isthmus/expression/LiteralConverter.java
alexandrefimov: Done — nullability comes from `resultType` alone, and the javadoc points at `createTypeWithNullability` for callers wanting something else.

alexandrefimov:
@vbarua ping when you get a chance. The overload is gone, so the PR is now `fix(isthmus)!` with a footer pointing at `convert(RexLiteral, RelDataType)` — that thread is left open because the call on removing a public API is yours.

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelVisitor.java
```diff
@@ -163,19 +163,19 @@ public Rel visit(org.apache.calcite.rel.core.Values values) {
...
+                  // Calcite may infer a narrower type for a tuple literal than for its row field.
+                  // Virtual table rows must use the complete schema type.
                   List<Expression> fields =
                       IntStream.range(0, list.size())
                           .mapToObj(
                               i ->
                                   literalConverter.convert(
-                                      list.get(i), schemaFieldTypes.get(i).nullable()))
+                                      list.get(i),
+                                      values.getRowType().getFieldList().get(i).getType()))
```
vbarua: minor: could we extract `values.getRowType().getFieldList()` to its own variable alongside `type` above.

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelVisitor.java
```diff
@@ -163,19 +163,19 @@ public Rel visit(org.apache.calcite.rel.core.Values values) {
...
     LiteralConverter literalConverter = new LiteralConverter(typeConverter);
-    List<Type> schemaFieldTypes = type.struct().fields();
     List<Expression.NestedStruct> structs =
         values.getTuples().stream()
             .map(
                 list -> {
-                  // Use schema nullability since Calcite infers non-nullable for all non-null
-                  // values
+                  // Calcite may infer a narrower type for a tuple literal than for its row field.
+                  // Virtual table rows must use the complete schema type.
```
vbarua: minor: clarify that we want the literal to match the row field type specifically.

```suggestion
                  // Virtual table rows must use the row field type.
```

isthmus/src/test/java/io/substrait/isthmus/VirtualTableScanTest.java
```diff
@@ -111,6 +118,26 @@ void mixedNullabilityRoundTrip() {
...
+        LogicalValues.create(
+            builder.getCluster(), rowType, ImmutableList.of(ImmutableList.of(literal)));
+
+    VirtualTableScan converted =
+        assertInstanceOf(
+            VirtualTableScan.class, SubstraitRelVisitor.convert(values, converterProvider));
+    assertEquals(R.I32, converted.getInitialSchema().struct().fields().get(0));
+    Expression.I32Literal convertedLiteral =
+        assertInstanceOf(Expression.I32Literal.class, converted.getRows().get(0).fields().get(0));
+    assertEquals(1, convertedLiteral.value());
```
vbarua: Small suggestion for this test. Add a second nullable column, and make values something like
```
(1, null)
(1, 5)
```
to check that null literals, and literals in nullable columns, also get converted with the correct nullability.

isthmus/src/main/java/io/substrait/isthmus/expression/LiteralConverter.java
```diff
@@ -246,21 +249,29 @@ public Expression.Literal convert(RexLiteral literal, boolean nullable) {
         {
           List<RexLiteral> literals = (List<RexLiteral>) literal.getValue();
           return ExpressionCreator.struct(
-              nullable, literals.stream().map(this::convert).collect(Collectors.toList()));
+              nullable,
+              IntStream.range(0, literals.size())
+                  .mapToObj(
+                      i -> convert(literals.get(i), resultType.getFieldList().get(i).getType()))
+                  .collect(Collectors.toList()));
```
vbarua, line 256: Do we have a tests that covers ROW column conversions? If not, I think it would be worth adding one to verify this, especially now that we need the ROW type contents to line up with the row literals.

vbarua (commented):
Left some minor comments, but this looks great overall. I think we can definitely get this merged this week.

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelVisitor.java
alexandrefimov: Thanks, Viktor! Done — I extracted the row fields alongside the converted schema.

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelVisitor.java
alexandrefimov: Done — I updated the comment to say that virtual table rows use the row field type.

isthmus/src/test/java/io/substrait/isthmus/VirtualTableScanTest.java
alexandrefimov: Done — the test now has a nullable second column with `(1, null)` and `(1, 5)`, and checks both the typed null and the non-null nullable literal.

isthmus/src/main/java/io/substrait/isthmus/expression/LiteralConverter.java
alexandrefimov: Added `tStructUsesResultFieldTypes` to cover a ROW literal whose TINYINT field is converted using a nullable INTEGER result field type.

vbarua (approved):
Changes look good. Thanks for working on this.

PR #1069, opened 2026-08-05
feat(core): handle RelCommon data for every relation type

### LINKED ISSUE #297: handle hints for all relation types
A Hint POJO was added as part of https://github.com/substrait-io/substrait-java/pull/295, but was not set up for every relation. We should wire Hint handling for every relation, and even better would be a way to simplify setting the data from RelCommon across all relations.

nielspardon:
`Hint` was wired into most but not all relations, and each of the ~31 `newXxx` methods in `ProtoRelConverter` repeated the same four `RelCommon` setters by hand, which is how they drift: `Cross` dropped its hint on the way back from proto, and `newTopN` dropped its rel anchor (fixed separately in #1088).

The proto → POJO direction is now centralised in `ProtoRelConverter.applyRelCommon`, mirroring `RelProtoConverter.common` on the way out, so the emit mapping, common extension, hint and rel anchor are applied in one place. This needs type-agnostic `withRemap` / `withCommonExtension` / `withHint` copy methods on `Rel` alongside the existing `withRelAnchor`; they are useful for plan rewriting in their own right. `applyRelCommon` copies a field only when it differs, so it stays a no-op for a `common { direct {} }` message — including for custom, non-Immutables relations, which inherit `Rel`'s throwing defaults.

One data-loss bug surfaced along the way: `RelProtoConverter.common` never wrote `RelCommon.Hint.advanced_extension`, so `Hint.getExtension()` was silently dropped on every relation even though the inbound direction read it back.

`UpdateRel` gained its `common` field in spec v0.101.0, so every modeled relation message now has one and `NamedUpdate` goes through the same two helpers as everything else — no carve-out, and no relation left that can hold `RelCommon` data the wire cannot carry. (`ReferenceRel` still has no `common` field, but no POJO models it.)

`RelCommonRoundtripTest` covers one sample of every `RelVisitor` relation type and fails when a new relation has no sample. It is keyed on POJO types rather than proto `oneof` cases, so a new sub-case mapping onto an existing POJO type still needs care — noted in AGENTS.md.

Closes #297


core/src/main/java/io/substrait/relation/RelProtoConverter.java
```diff
@@ -927,6 +928,29 @@ public Rel visit(ExtensionMulti extensionMulti, EmptyVisitationContext context)
...
+   * Rejects {@link RelCommon} data on a relation whose protobuf message has no {@code common}
+   * field. {@code UpdateRel} is the only such message (spec v0.99.0), so an update relation cannot
+   * express an emit mapping, a hint, a rel anchor or a common advanced extension. The POJO model
+   * exposes those accessors on every {@link io.substrait.relation.Rel}, and an emit mapping
+   * additionally changes {@link io.substrait.relation.Rel#getRecordType()} — serializing such a
+   * relation would silently produce a plan with a different schema, so fail instead.
+   *
+   * @param rel the relation about to be serialized
+   */
+  private static void checkNoRelCommon(io.substrait.relation.Rel rel) {
```
alexandrefimov: `SubstraitBuilder.namedUpdate(…, Rel.Remap)` exists solely to put a remap on a `NamedUpdate`, so from here on every plan built through it throws at serialisation. Worth dropping or deprecating that overload in this PR — otherwise the DSL keeps handing out a relation whose only possible outcome is this exception.

core/src/test/java/io/substrait/type/proto/RelCommonRoundtripTest.java
```diff
@@ -0,0 +1,447 @@
...
+ * miss its {@code RelCommon} wiring.
+ */
+@TestInstance(TestInstance.Lifecycle.PER_CLASS)
+class RelCommonRoundtripTest extends TestBase {
+
+  /**
+   * Relations whose protobuf message has no {@code common} field and which therefore cannot carry
+   * any {@code RelCommon} data at all (spec v0.99.0).
+   */
+  static final java.util.Set<Class<? extends Rel>> WITHOUT_REL_COMMON =
```
alexandrefimov: `ReferenceRel` has no `common` field either — `javap` over protobuf 0.99.0 gives `UpdateRel`, `ReferenceRel` and the `PlanRel`/`Rel` wrappers. No POJO models it today, so nothing is broken, but "the only such message" (here and on `checkNoRelCommon`) is what whoever models `ReferenceRel` will read. "the only modelled one" stays true either way.

core/src/main/java/io/substrait/relation/RelProtoConverter.java
nielspardon: Deliberately not deprecating: substrait-io/substrait#1168 adds `RelCommon common = 6` to `UpdateRel`, so this overload becomes valid again once that lands and we bump `substrait-packaging`. Deprecating now would mean un-deprecating a release later.

Documented the limitation on the overload instead — it now says the returned relation can't be serialized while `UpdateRel` has no `common` field, and points at the 5-arg overload for anything that has to round-trip.

core/src/test/java/io/substrait/type/proto/RelCommonRoundtripTest.java
nielspardon: Right, fixed — reproduced the same three (`PlanRel`, `ReferenceRel`, `UpdateRel`) with `javap`. Now "the only *modeled* message without one" here, on `checkNoRelCommon`, in the `newNamedUpdate` comment and in AGENTS.md, each noting `ReferenceRel` has none either but isn't modeled.

On `checkNoRelCommon` I also carried over why substrait-io/substrait#1168 leaves `ReferenceRel` alone: it points at another subtree rather than producing output of its own, so the common fields don't mean the same thing there. That's the bit worth having if someone does model it.

core/src/main/java/io/substrait/relation/RelProtoConverter.java
alexandrefimov: That makes sense. The Javadoc now makes the current limitation clear, and avoiding a temporary deprecation is reasonable given substrait-io/substrait#1168. Thanks.

core/src/test/java/io/substrait/type/proto/RelCommonRoundtripTest.java
alexandrefimov: Thanks, the wording is accurate now, and the `ReferenceRel` distinction gives useful context.

alexandrefimov (approved):
LGTM.

One note for whoever merges: #1103 (spec v0.101.0) wires `UpdateRel.common`, which removes the need for the `NamedUpdate` serialization guard here — whichever of the two PRs lands second should drop the guard and the `checkNoRelCommon` carve-out in `RelCommonRoundtripTest`.

nielspardon:
Updated the PR after #1103 got merged to main.

alexandrefimov (approved):
LGTM. Re-checked after #1103: `NamedUpdate` now goes through the centralized `RelCommon` path in both directions, the obsolete guard and test carve-out are gone, and the round-trip test covers all four common fields. Current CI is green.

andrew-coleman (commented):
Nice cleanup — routing every `newXxx` through a single `applyRelCommon` is clearly the right shape, and I checked the mechanical side holds: all 35 `newXxx` methods whose proto message carries a `common` field go through it, every `RelProtoConverter.visit` sets `common(rel)`, no field was lost in the refactor, and `Cross` picking up `hint` is a genuine bug fix. The `Optional<? extends T>` signatures on the new `Rel` defaults do match what Immutables generates.

Three things worth addressing, one of which is a real behaviour issue:

1. **`applyRelCommon` isn't the no-op its javadoc claims for a delegating custom `Rel`, and can destroy a hint.** Gating each copy on *inequality* rather than on *presence in the proto* means a transparent wrapper that derives `getHint()`/`getRemap()` from its input gets asked to clear data the proto never said to clear. Suggestions inline.
2. **The `(R) result` cast is unsound outside Immutables** — worth a guard that blames the offending override rather than the caller.
3. **`SubstraitRepeatRel` has a double-remap** that the new `withRemap` override makes reachable, and the new test passes vacuously because its scan has one field.

I applied all of this on top of `7d69ab6` and verified it: both new guard tests fail on the current converter with `UnsupportedOperationException: class ...$DelegatingRel does not support setting a relation anchor` (confirming the old code really does try to clear a delegating wrapper's data), and the widened isthmus assertion fails with `expected: <[Str, I64]> but was: <[I64, Str]>` (confirming the double-remap). With the fixes: `:core:test` 659 tests and `:isthmus:test` 1056 tests pass, `spotlessCheck` and `:core:javadoc` clean.

Happy to push the patch as a branch you can merge if that's easier than applying six suggestions.

core/src/main/java/io/substrait/relation/ProtoRelConverter.java
```diff
@@ -1637,6 +1476,55 @@ protected AbstractExchangeRel.ExchangeTarget newExchangeTarget(
...
+   *
+   * <p>This is the proto → POJO counterpart of {@link RelProtoConverter}'s {@code
+   * common(io.substrait.relation.Rel)} and exists so that adding a {@code RelCommon} field means
+   * touching one method rather than every {@code newXxx} converter. Every {@code newXxx} method
+   * whose protobuf message carries a {@code common} field must route its result through here.
+   *
+   * <p>Each field is only copied when it actually differs from what {@code rel} already carries.
+   * That keeps this method a true no-op for a {@code common { direct {} }} message — including for
+   * custom, non-Immutables {@link Rel} implementations, which inherit {@code Rel}'s throwing {@code
+   * withXxx} defaults and would otherwise fail on a relation that carries no common data at all.
```
andrew-coleman: The "true no-op for a `common { direct {} }` message ... including for custom, non-Immutables `Rel` implementations" claim doesn't hold for a *delegating* custom `Rel` — the pattern this PR itself documents in `SubtraitRepeatRel`, which forwards `getHint()`/`getRemap()` to its input.

If the input carries a hint in its own `RelCommon` but this rel's own common is `direct {}`, then `optionalHint` is empty, `empty.equals(Optional.of(inputHint))` is false, and `withHint(Optional.empty())` fires — the wrapper forwards it to the input, destroying a hint that was read correctly. Same for `withRemap` (drops the input's emit mapping) and `withCommonExtension`.

Gating on presence in the proto first makes the property universal, since a freshly built rel has nothing to clear.

```suggestion
   * <p>Each field is only copied when the protobuf message actually carries it, and then only when
   * the converted value differs from what {@code rel} already holds. That keeps this method a true
   * no-op for a {@code common { direct {} }} message — including for custom, non-Immutables {@link
   * Rel} implementations, which inherit {@code Rel}'s throwing {@code withXxx} defaults and would
   * otherwise fail on a relation that carries no common data at all. Gating on presence rather than
   * on inequality alone also means a relation that derives one of these fields from its input (a
   * delegating wrapper) is never asked to clear it.
```

core/src/main/java/io/substrait/relation/ProtoRelConverter.java
```diff
@@ -1637,6 +1476,55 @@ protected AbstractExchangeRel.ExchangeTarget newExchangeTarget(
...
+      result = result.withRemap(remap);
+    }
+    Optional<AdvancedExtension> commonExtension = optionalAdvancedExtension(relCommon);
+    if (!commonExtension.equals(result.getCommonExtension())) {
+      result = result.withCommonExtension(commonExtension);
+    }
+    Optional<Hint> hint = optionalHint(relCommon);
+    if (!hint.equals(result.getHint())) {
+      result = result.withHint(hint);
+    }
```
andrew-coleman: Presence gating for the four fields. Keeping the equality check nested also avoids gratuitous `withXxx` calls on a wrapper that already agrees.

```suggestion
    if (relCommon.hasRelAnchor()) {
      Optional<Integer> relAnchor = optionalRelAnchor(relCommon);
      if (!relAnchor.equals(result.getRelAnchor())) {
        result = result.withRelAnchor(relAnchor);
      }
    }
    if (relCommon.hasEmit()) {
      Optional<Rel.Remap> remap = optionalRelmap(relCommon);
      if (!remap.equals(result.getRemap())) {
        result = result.withRemap(remap);
      }
    }
    if (relCommon.hasAdvancedExtension()) {
      Optional<AdvancedExtension> commonExtension = optionalAdvancedExtension(relCommon);
      if (!commonExtension.equals(result.getCommonExtension())) {
        result = result.withCommonExtension(commonExtension);
      }
    }
    if (relCommon.hasHint()) {
      Optional<Hint> hint = optionalHint(relCommon);
      if (!hint.equals(result.getHint())) {
        result = result.withHint(hint);
      }
    }
```

core/src/main/java/io/substrait/relation/ProtoRelConverter.java
```diff
@@ -1637,6 +1476,55 @@ protected AbstractExchangeRel.ExchangeTarget newExchangeTarget(
...
+    return (R) result;
```
andrew-coleman, line 1546: Minor, separate from the presence-gating comment above: the comment at the top of the method is accurate for Immutables, but `Rel.withRemap/withHint/withCommonExtension` are only declared to return `Rel`. Nothing stops a custom override from returning its delegate (`return input.withHint(hint);` instead of re-wrapping), which surfaces as a `ClassCastException` at the *caller's* assignment rather than at the offending override. A cheap guard names the culprit:

```suggestion
    // A custom Rel may return its delegate rather than a re-wrapped copy from one of the withXxx
    // methods above; blame that override rather than failing the caller's assignment with a bare
    // ClassCastException.
    if (!rel.getClass().isInstance(result)) {
      throw new IllegalStateException(
          String.format(
              "%s returned %s from a RelCommon copy method; withRelAnchor/withRemap/withHint/"
                  + "withCommonExtension must return the same relation type",
              rel.getClass().getName(), result.getClass().getName()));
    }
    return (R) result;
```

core/src/test/java/io/substrait/type/proto/RelCommonRoundtripTest.java
```diff
@@ -0,0 +1,418 @@
...
+    ApplyRelCommonConverter converter = new ApplyRelCommonConverter(functionCollector, extensions);
+    Rel custom = new PassThroughRel(left);
+
+    assertEquals(custom, converter.applyRelCommon(custom, RelCommon.getDefaultInstance()));
+    assertEquals(
+        custom,
+        converter.applyRelCommon(
+            custom,
+            RelCommon.newBuilder().setDirect(RelCommon.Direct.getDefaultInstance()).build()));
+  }
```
andrew-coleman, line 156: `PassThroughRel` returns `Optional.empty()` from all four accessors, so it's the one custom-`Rel` shape for which the no-op claim holds unconditionally — it can't catch the case where `applyRelCommon` decides a field "differs". A test for a *delegating* wrapper does guard the documented invariant (and fails on the current converter with `UnsupportedOperationException: ... does not support setting a relation anchor`):

```suggestion
  }

  @Test
  void applyRelCommonLeavesADelegatingCustomRelAlone() {
    // PassThroughRel reports Optional.empty() from every accessor, so it cannot catch the case
    // where applyRelCommon decides a field "differs". A custom Rel that derives its common data
    // from its input — the pattern documented by isthmus' SubstraitRepeatRel — reports data that
    // the enclosing direct{} message does not carry. applyRelCommon must still leave it alone
    // rather than trying to clear it, which for a delegating wrapper would destroy the input's own
    // common data (and here hits Rel's throwing withXxx defaults).
    ApplyRelCommonConverter converter = new ApplyRelCommonConverter(functionCollector, extensions);
    Rel custom = new DelegatingRel(withRelCommon(left));

    assertTrue(custom.getHint().isPresent());
    assertTrue(custom.getCommonExtension().isPresent());
    assertTrue(custom.getRemap().isPresent());
    assertTrue(custom.getRelAnchor().isPresent());

    assertEquals(custom, converter.applyRelCommon(custom, RelCommon.getDefaultInstance()));
    assertEquals(
        custom,
        converter.applyRelCommon(
            custom,
            RelCommon.newBuilder().setDirect(RelCommon.Direct.getDefaultInstance()).build()));
  }
```

That needs a `DelegatingRel` helper alongside `PassThroughRel` — same shape, but every accessor forwards to `input` and the throwing `withXxx` defaults are left inherited:

```java
  /**
   * A hand-written {@link Rel} that derives all of its {@code RelCommon} data from its input, as a
   * transparent wrapper would. It inherits {@code Rel}'s throwing {@code withXxx} defaults, so any
   * attempt to clear one of those fields fails loudly.
   */
  static final class DelegatingRel extends SingleInputRel {
    private final Rel input;

    DelegatingRel(Rel input) {
      this.input = input;
    }

    @Override
    public Rel getInput() {
      return input;
    }

    @Override
    protected Type.Struct deriveRecordType() {
      return input.getRecordType();
    }

    @Override
    public Optional<Rel.Remap> getRemap() {
      return input.getRemap();
    }

    @Override
    public Optional<AdvancedExtension> getCommonExtension() {
      return input.getCommonExtension();
    }

    @Override
    public Optional<Hint> getHint() {
      return input.getHint();
    }

    @Override
    public Optional<Integer> getRelAnchor() {
      return input.getRelAnchor();
    }

    @Override
    public <O, C extends VisitationContext, E extends Exception> O accept(
        RelVisitor<O, C, E> visitor, C context) {
      throw new UnsupportedOperationException("not visitable");
    }
  }
```

isthmus/src/test/java/io/substrait/isthmus/SubtraitRelVisitorExtensionTest.java
```diff
@@ -134,6 +134,21 @@ public Rel withRelAnchor(final Optional<Integer> relAnchor) {
       return new SubstraitRepeatRel(input.withRelAnchor(relAnchor), repeatCount);
     }
 
+    @Override
+    public Rel withRemap(final Optional<? extends Remap> remap) {
+      return new SubstraitRepeatRel(input.withRemap(remap), repeatCount);
+    }
```
andrew-coleman, line 178: Adding this override makes a latent double-remap in this example reachable. `deriveRecordType()` (line 102) returns `input.getRecordType()`, which is already the input's *emitted* (post-remap) type, while `getRemap()` (line 107) also delegates to the input — so `AbstractRel.getRecordType()` applies the same mapping a second time. With a 2-field input and remap `[1,0]`, `SubstraitRepeatRel.getRecordType()` yields `[T0,T1]` instead of `[T1,T0]`.

Since `deriveRecordType()` can't reach the input's pre-remap type from here, the fix is for the wrapper to hold its own remap rather than inherit one:

```suggestion
    @Override
    public Rel withRemap(final Optional<? extends Remap> remap) {
      // Held on this relation rather than pushed down to the input, mirroring getRemap().
      return new SubstraitRepeatRel(input, repeatCount, remap.map(Remap.class::cast));
    }
```

with a field and a private constructor (line 87), which the other three `withXxx` overrides then thread through unchanged:

```java
    private final Optional<Remap> remap;

    public SubstraitRepeatRel(final Rel input, final int repeatCount) {
      this(input, repeatCount, Optional.empty());
    }

    private SubstraitRepeatRel(
        final Rel input, final int repeatCount, final Optional<Remap> remap) {
      this.input = input;
      this.repeatCount = repeatCount;
      this.remap = remap;
    }
```

and `getRemap()` (line 106) no longer delegating:

```java
    @Override
    public Optional<Remap> getRemap() {
      // Deliberately not delegated to the input: deriveRecordType() already returns the input's
      // *emitted* type, so inheriting the input's remap would make AbstractRel.getRecordType()
      // apply that same mapping a second time.
      return remap;
    }
```

isthmus/src/test/java/io/substrait/isthmus/SubtraitRelVisitorExtensionTest.java
```diff
@@ -294,4 +309,26 @@ void customRelSupportsRelAnchor() {
...
+    // The remaining RelCommon copy methods are delegated the same way, so a custom Rel can be
+    // handed to code that stamps an emit mapping, a hint or a common extension onto an arbitrary
+    // relation without hitting Rel's throwing defaults.
+    final SubstraitBuilder sb = new SubstraitBuilder();
+    final Rel scan = sb.namedScan(List.of("t"), List.of("a"), List.of(TypeCreator.REQUIRED.I64));
+    final SubstraitRepeatRel repeat = new SubstraitRepeatRel(scan, 3);
+
+    final Rel remapped = repeat.withRemap(Optional.of(Rel.Remap.of(List.of(0))));
+    assertTrue(remapped instanceof SubstraitRepeatRel);
+    assertEquals(List.of(0), remapped.getRemap().orElseThrow(AssertionError::new).indices());
```
andrew-coleman: This passes vacuously today: the scan has a single field, so `Remap.of(List.of(0))` is the identity and applying it twice is indistinguishable from applying it once. A two-field reordering mapping makes the assertion actually exercise the pattern being demonstrated:

```suggestion
    final Rel scan =
        sb.namedScan(
            List.of("t"),
            List.of("a", "b"),
            List.of(TypeCreator.REQUIRED.I64, TypeCreator.REQUIRED.STRING));
    final SubstraitRepeatRel repeat = new SubstraitRepeatRel(scan, 3);

    // A reordering mapping, so that applying it twice would be observable in the record type.
    final Rel remapped = repeat.withRemap(Optional.of(Rel.Remap.of(List.of(1, 0))));
    assertTrue(remapped instanceof SubstraitRepeatRel);
    assertEquals(List.of(1, 0), remapped.getRemap().orElseThrow(AssertionError::new).indices());
    assertEquals(
        List.of(TypeCreator.REQUIRED.STRING, TypeCreator.REQUIRED.I64),
        remapped.getRecordType().fields());
```

Against the current delegating `getRemap()` this fails with:

```
expected: <[Str{nullable=false}, I64{nullable=false}]> but was: <[I64{nullable=false}, Str{nullable=false}]>
```

i.e. the mapping is applied twice and returns to identity.

core/src/test/java/io/substrait/type/proto/RelCommonRoundtripTest.java
nielspardon: Applied, with one change to the helper. As written, `deriveRecordType()` returns `input.getRecordType()` — the input's already-*emitted* type — while `getRemap()` forwards the input's mapping, which is the same double-remap you flagged on `SubstraitRepeatRel`. The input here is `withRelCommon(left)`, whose mapping is `[1,0]`, so the fixture would have shipped a demonstration of the bug the other comment fixes.

`DelegatingRel` now takes the input's pre-emit struct for `deriveRecordType()`, which is what makes delegating `getRemap()` sound, and the test asserts the wrapper and its input agree on record type. I also dropped the "the pattern documented by isthmus' `SubstraitRepeatRel`" clause from the comment — after the other fix that wrapper holds its own mapping rather than inheriting one, so a transparent wrapper is the honest framing.

The test does fail without the presence gating, as you predicted.

isthmus/src/test/java/io/substrait/isthmus/SubtraitRelVisitorExtensionTest.java
```diff
@@ -126,12 +136,28 @@ public Optional<Integer> getRelAnchor() {
...
+      return new SubstraitRepeatRel(input, repeatCount, remap.map(Remap.class::cast));
+    }
+
+    @Override
+    public Rel withCommonExtension(final Optional<? extends AdvancedExtension> commonExtension) {
+      return new SubstraitRepeatRel(input.withCommonExtension(commonExtension), repeatCount, remap);
+    }
+
+    @Override
+    public Rel withHint(final Optional<? extends Hint> hint) {
```
alexandrefimov, line 192: The emit mapping moved onto the wrapper, but `hint` and `commonExtension` still delegate both ways, and the write side has the same destructive shape the remap fix removed:

```
input hint before   = Optional[Hint{alias=input_alias, ...}]
wrapper hint after  = Optional[Hint{alias=wrapper_alias, ...}]
input hint after    = Optional[Hint{alias=wrapper_alias, ...}]
```

Stamping a hint on the wrapper replaces the input's own. Presence gating keeps `applyRelCommon` from reaching this on an empty message, but a proto that carries `hint` on the wrapper rel still does — and unlike `withRelAnchor`, where the delegation is deliberate and documented (id-based outer references bind to the input), nothing states why an alias, output names or row-count stats describing this relation belong on its input.

Same treatment as `remap` — hold them on the wrapper — would make the example self-consistent. Not blocking; it is a fixture, but the PR presents it as the reference pattern for custom `Rel` implementors.

core/src/main/java/io/substrait/relation/ProtoRelConverter.java
```diff
@@ -1637,6 +1476,76 @@ protected AbstractExchangeRel.ExchangeTarget newExchangeTarget(
...
+    if (relCommon.hasHint()) {
+      Optional<Hint> hint = optionalHint(relCommon);
+      if (!hint.equals(result.getHint())) {
+        result = result.withHint(hint);
+      }
+    }
+    // A custom Rel may return its delegate rather than a re-wrapped copy from one of the withXxx
+    // methods above; blame that override rather than failing the caller's assignment with a bare
+    // ClassCastException.
+    if (!rel.getClass().isInstance(result)) {
```
alexandrefimov, line 1539: Minor: nothing exercises this branch. `PassThroughRel` and `DelegatingRel` never reach it, so the message is unverified production code. A third fixture whose `withHint` returns `input.withHint(hint)` would pin it in a few lines.

alexandrefimov (commented):
Re-reviewed on ef0f136. Both regression tests are real — I ran them against 7d69ab6: the core one fails with `UnsupportedOperationException: class ...$DelegatingRel does not support setting a relation anchor`, the isthmus one with `expected: <[Str, I64]> but was: <[I64, Str]>`. Presence gating is symmetric with the writer (`RelProtoConverter.common` emits `direct {}` exactly when there is no remap), and the only `newXxx` that sets common data on the builder itself is `newLateralJoin`'s `relAnchor`, which the gated branch still compares and skips. `:core:test` 658 / `:isthmus:test` 1056 pass here.

My two earlier approvals missed the clearing bug, so I went back over the fixture as well — one thing left, inline.

andrew-coleman (approved):

PR #1074, opened 2026-08-06
feat(core)!: enforce Fetch offset/count expressions are integer-typed

benbellick:
Adds enforcement that FetchRel's offset_expr/count_expr resolve to an integer type (I8/I16/I32/I64), rejecting non-integer expressions at construction. The spec only recommends i64 for these fields:

https://github.com/substrait-io/substrait/blob/v0.99.0/proto/substrait/algebra.proto#L345-L366

BREAKING CHANGE: `Fetch` relation now enforces that `offset` and `count` are integer types (or `unbound`). Otherwise, an exception is thrown. 

---
Note: This PR was developed with AI assistance. All changes have been reviewed, and I take full responsibility for this contribution.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fetch operations now validate offset and count expressions and reject non-integer values.
  * Integer values of all supported widths and unbound expressions remain accepted.
  * Invalid Fetch expressions are also rejected during protocol conversion.

### DIFF
--- a/core/src/main/java/io/substrait/relation/Fetch.java
+++ b/core/src/main/java/io/substrait/relation/Fetch.java
@@ -29,6 +29,32 @@ public abstract class Fetch extends SingleInputRel implements HasExtension {
    */
   public abstract Optional<Expression> getCount();
 
+  @Value.Check
+  protected void check() {
+    getOffset()
+        .ifPresent(
+            offset -> {
+              Type type = offset.getType();
+              if (!type.isInteger()) {
+                throw new IllegalArgumentException(
+                    "Fetch offset expression must have an integer type "
+                        + "(I8, I16, I32 or I64), but got: "
+                        + type);
+              }
+            });
+    getCount()
+        .ifPresent(
+            count -> {
+              Type type = count.getType();
+              if (!type.isInteger()) {
+                throw new IllegalArgumentException(
+                    "Fetch count expression must have an integer type "
+                        + "(I8, I16, I32 or I64), but got: "
+                        + type);
+              }
+            });
+  }
+
   @Override
   protected Type.Struct deriveRecordType() {
     return getInput().getRecordType();
--- a/core/src/main/java/io/substrait/relation/Fetch.java
+++ b/core/src/main/java/io/substrait/relation/Fetch.java
@@ -1,6 +1,7 @@
 package io.substrait.relation;
 
 import io.substrait.expression.Expression;
+import io.substrait.type.StringTypeVisitor;
 import io.substrait.type.Type;
 import io.substrait.util.VisitationContext;
 import java.util.Optional;
@@ -29,30 +30,31 @@ public abstract class Fetch extends SingleInputRel implements HasExtension {
    */
   public abstract Optional<Expression> getCount();
 
+  /**
+   * Validates that the offset and count expressions are integer-typed. Both must evaluate to a
+   * non-negative integer; {@code i64} is recommended but not required, so any integer width is
+   * accepted (spec v0.99.0).
+   *
+   * @throws IllegalArgumentException if the offset or count expression is not integer-typed
+   */
   @Value.Check
   protected void check() {
-    getOffset()
-        .ifPresent(
-            offset -> {
-              Type type = offset.getType();
-              if (!type.isInteger()) {
-                throw new IllegalArgumentException(
-                    "Fetch offset expression must have an integer type "
-                        + "(I8, I16, I32 or I64), but got: "
-                        + type);
-              }
-            });
-    getCount()
-        .ifPresent(
-            count -> {
-              Type type = count.getType();
-              if (!type.isInteger()) {
-                throw new IllegalArgumentException(
-                    "Fetch count expression must have an integer type "
-                        + "(I8, I16, I32 or I64), but got: "
-                        + type);
-              }
-            });
+    requireIntegerType(getOffset(), "offset");
+    requireIntegerType(getCount(), "count");
+  }
+
+  private static void requireIntegerType(Optional<Expression> expression, String field) {
+    expression.ifPresent(
+        e -> {
+          Type type = e.getType();
+          if (!type.isInteger()) {
+            throw new IllegalArgumentException(
+                "Fetch "
+                    + field
+                    + " expression must have an integer type (i8, i16, i32 or i64), but got: "
+                    + type.accept(new StringTypeVisitor()));
+          }
+        });
   }
 
   @Override
--- a/core/src/main/java/io/substrait/type/Type.java
+++ b/core/src/main/java/io/substrait/type/Type.java
@@ -53,6 +53,18 @@ public interface Type extends TypeExpression, ParameterizedType, NullableType, F
     return TypeCreator.asNullable(this).equals(TypeCreator.asNullable(other));
   }
 
+  /**
+   * Returns whether this is one of the integer types {@link I8}, {@link I16}, {@link I32} or {@link
+   * I64}. The Substrait spec only "recommends" {@code i64} for contexts such as {@code FetchRel}'s
+   * {@code offset_expr}/{@code count_expr}, so callers that accept any integer width can use this
+   * to reject non-integer types.
+   *
+   * @return {@code true} if this is an integer type
+   */
+  default boolean isInteger() {
+    return this instanceof I8 || this instanceof I16 || this instanceof I32 || this instanceof I64;
+  }
+
   /** The boolean type. */
   @Value.Immutable
   abstract class Bool implements Type {
--- a/core/src/main/java/io/substrait/type/Type.java
+++ b/core/src/main/java/io/substrait/type/Type.java
@@ -54,10 +54,8 @@ public interface Type extends TypeExpression, ParameterizedType, NullableType, F
   }
 
   /**
-   * Returns whether this is one of the integer types {@link I8}, {@link I16}, {@link I32} or {@link
-   * I64}. The Substrait spec only "recommends" {@code i64} for contexts such as {@code FetchRel}'s
-   * {@code offset_expr}/{@code count_expr}, so callers that accept any integer width can use this
-   * to reject non-integer types.
+   * Returns whether this is one of the fixed-width signed integer types: {@link I8}, {@link I16},
+   * {@link I32} or {@link I64}.
    *
    * @return {@code true} if this is an integer type
    */
--- a/core/src/test/java/io/substrait/relation/FetchTest.java
+++ b/core/src/test/java/io/substrait/relation/FetchTest.java
@@ -1,0 +1,64 @@
+package io.substrait.relation;
+
+import static org.junit.jupiter.api.Assertions.assertThrows;
+
+import io.substrait.TestBase;
+import java.util.Arrays;
+import org.junit.jupiter.api.Test;
+
+/**
+ * Validation tests for {@link Fetch}, whose offset/count expressions must have an integer type.
+ * Round-trip coverage lives in the {@code io.substrait.type.proto} package.
+ */
+class FetchTest extends TestBase {
+
+  // Reuse the same schema shape as FetchRoundtripTest so the two files stay in sync.
+  final Rel table =
+      sb.namedScan(Arrays.asList("T"), Arrays.asList("a", "b"), Arrays.asList(R.I64, R.STRING));
+
+  /** Every integer width is an acceptable offset/count expression type. */
+  @Test
+  void integerWidthsAccepted() {
+    fetch().offset(sb.i8(1)).count(sb.i8(2)).build();
+    fetch().offset(sb.i16(1)).count(sb.i16(2)).build();
+    fetch().offset(sb.i32(1)).count(sb.i32(2)).build();
+    fetch().offset(sb.i64(1)).count(sb.i64(2)).build();
+  }
+
+  /** A non-integer offset expression is rejected at construction time. */
+  @Test
+  void nonIntegerOffsetRejected() {
+    assertThrows(IllegalArgumentException.class, () -> fetch().offset(sb.fp64(1.0)).build());
+  }
+
+  /** A non-integer count expression is rejected at construction time. */
+  @Test
+  void nonIntegerCountRejected() {
+    assertThrows(IllegalArgumentException.class, () -> fetch().count(sb.fp64(1.0)).build());
+  }
+
+  /** A non-integer type reaches the check via the proto conversion path too. */
+  @Test
+  void nonIntegerOffsetRejectedViaProto() {
+    // Build a FetchRel proto directly with a non-integer offset_expr so the Fetch POJO check is
+    // exercised by ProtoRelConverter rather than by direct construction.
+    io.substrait.proto.Rel inputProto = relProtoConverter.toProto(table);
+    io.substrait.proto.Expression fp64Expr =
+        io.substrait.proto.Expression.newBuilder()
+            .setLiteral(io.substrait.proto.Expression.Literal.newBuilder().setFp64(1.0).build())
+            .build();
+    io.substrait.proto.FetchRel fetchRel =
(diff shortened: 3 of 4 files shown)


nielspardon:
trying to trigger CI through close / reopen. there was a long GH actions outage yesterday and the checks were not triggered

nielspardon (commented):
**Should `unbound` be exempt?** `Type.Unbound` isn't an integer, so a `DynamicParameter` offset — or a field reference into an unbound column — now throws at construction. And since `ProtoRelConverter.from()` builds the POJO, such a plan fails to *import*, not merely to build locally. `Type.Unbound`'s own Javadoc says a partially bound plan "may be serialized and exchanged, but it must be bound to concrete types before execution", which is exactly what an import-time type check can't honour. This is also the first rel-level validation that inspects an *expression's* type, so it sets the precedent for later ones. My inline suggestion exempts it, with the one-line revert noted there.

Three smaller ones:

- **`TopN` has the same two fields and the same spec wording**, and is reachable via `ProtoRelConverter.newTopN` — validating one and not the other half-enforces the invariant. The same check with the message prefixed `TopN` compiles and keeps the suite green.
- **`FetchRoundtripTest` only ever uses `i64`.** One case would pin the widened acceptance through proto conversion: `offset(sb.i8(3)).count(sb.i32(7))`.
- **`from()` now throws on plans it previously accepted**, and the check isn't overridable, so consumers get no leniency seam. Worth deciding deliberately whether that carries a literal `BREAKING CHANGE:` footer — a heading alone produces no release-notes text.

Not for this PR: the spec's "evaluating to a negative integer should result in an error" is statically checkable for literals.

Suggestions below are all compiled against your head commit.

core/src/main/java/io/substrait/relation/Fetch.java
```diff
@@ -29,6 +29,32 @@ public abstract class Fetch extends SingleInputRel implements HasExtension {
...
+            count -> {
+              Type type = count.getType();
+              if (!type.isInteger()) {
+                throw new IllegalArgumentException(
+                    "Fetch count expression must have an integer type "
+                        + "(I8, I16, I32 or I64), but got: "
+                        + type);
+              }
+            });
+  }
```
nielspardon, line 58: Two things bundled here, since they touch the same lines.

**For discussion:** exempting `unbound` (see summary). If you'd rather not, drop `&& !(type instanceof Type.Unbound)` and the second Javadoc paragraph — everything else stands on its own.

The rest: Javadoc, which `-Xdoclint:all -Xwerror` requires on a `protected` member; collapsing the duplicated branches; and rendering the type through `StringTypeVisitor`, so the message reads `decimal<10,2>` instead of leaking Immutables `toString` (today it really does say `but got: Unbound{}`).

Needs one import outside this hunk: `import io.substrait.type.StringTypeVisitor;`

```suggestion
  /**
   * Validates that the offset and count expressions are integer-typed. Both must evaluate to a
   * non-negative integer; {@code i64} is recommended but not required, so any integer width is
   * accepted (spec v0.99.0).
   *
   * <p>The {@link Type.Unbound unbound} type is also accepted, so that a partially bound plan can
   * still be serialized and exchanged. The integer constraint can only be enforced once the type
   * has been bound.
   *
   * @throws IllegalArgumentException if the offset or count expression is neither integer-typed nor
   *     unbound
   */
  @Value.Check
  protected void check() {
    requireIntegerType(getOffset(), "offset");
    requireIntegerType(getCount(), "count");
  }

  private static void requireIntegerType(Optional<Expression> expression, String field) {
    expression.ifPresent(
        e -> {
          Type type = e.getType();
          if (!type.isInteger() && !(type instanceof Type.Unbound)) {
            throw new IllegalArgumentException(
                "Fetch "
                    + field
                    + " expression must have an integer type (i8, i16, i32 or i64), but got: "
                    + type.accept(new StringTypeVisitor()));
          }
        });
  }
```

core/src/main/java/io/substrait/type/Type.java
```diff
@@ -53,6 +53,18 @@ default boolean equalsIgnoringNullability(Type other) {
...
+   * Returns whether this is one of the integer types {@link I8}, {@link I16}, {@link I32} or {@link
+   * I64}. The Substrait spec only "recommends" {@code i64} for contexts such as {@code FetchRel}'s
+   * {@code offset_expr}/{@code count_expr}, so callers that accept any integer width can use this
+   * to reject non-integer types.
+   *
+   * @return {@code true} if this is an integer type
+   */
+  default boolean isInteger() {
+    return this instanceof I8 || this instanceof I16 || this instanceof I32 || this instanceof I64;
+  }
```
nielspardon, line 64: The predicate itself is good. I would decouple the Javadoc from its caller, though: this is a general predicate on the public `Type` interface, and documenting `FetchRel`'s recommendation here inverts the dependency - the type model should not need to know who is asking. The spec rationale belongs in `Fetch.check()`, where I have put it.

```suggestion
  /**
   * Returns whether this is one of the fixed-width signed integer types: {@link I8}, {@link I16},
   * {@link I32} or {@link I64}.
   *
   * @return {@code true} if this is an integer type
   */
  default boolean isInteger() {
    return this instanceof I8 || this instanceof I16 || this instanceof I32 || this instanceof I64;
  }
```

core/src/test/java/io/substrait/relation/FetchTest.java
```diff
@@ -0,0 +1,64 @@
...
+            .build();
+    io.substrait.proto.Rel protoRel =
+        io.substrait.proto.Rel.newBuilder().setFetch(fetchRel).build();
+    assertThrows(IllegalArgumentException.class, () -> protoRelConverter.from(protoRel));
+  }
+
+  private ImmutableFetch.Builder fetch() {
+    return Fetch.builder().input(table);
+  }
+}
```
nielspardon, line 64: Whole-file suggestion, since the pieces share imports. What changes:

- **Mixed widths** - `integerWidthsAccepted` only paired each width with itself, so a helper comparing the two types to *each other* rather than checking each independently would pass every current test and fail this one.
- **Nullable integers** - the bigger gap: the spec assigns meaning to null (`offset` null -> 0, `count` null -> ALL), and nothing pins that through the check today.
- **Non-literal expressions** - `SubstraitRelVisitor` deliberately passes arbitrary Calcite `RexNode`s through as offset/count, so a field reference and a cast belong here.
- **`unbound`** - encodes whatever we settle in the summary; delete if we go the other way.
- **`assertDoesNotThrow` + message assertions**, so a failure names the offending combination and `nonIntegerCountRejected` cannot pass on the *offset* branch firing.

`FetchRel` and `RelCommon` are importable; `io.substrait.proto.Expression` and `io.substrait.proto.Rel` have to stay qualified - they collide with the POJO types.

```suggestion
package io.substrait.relation;

import static org.junit.jupiter.api.Assertions.assertDoesNotThrow;
import static org.junit.jupiter.api.Assertions.assertThrows;
import static org.junit.jupiter.api.Assertions.assertTrue;

import io.substrait.TestBase;
import io.substrait.expression.Expression;
import io.substrait.expression.ExpressionCreator;
import io.substrait.proto.FetchRel;
import io.substrait.proto.RelCommon;
import io.substrait.type.StringTypeVisitor;
import io.substrait.type.Type;
import java.util.Arrays;
import java.util.List;
import org.junit.jupiter.api.Test;

/**
 * Validation tests for {@link Fetch}, whose offset/count expressions must have an integer type.
 * Round-trip coverage lives in the {@code io.substrait.type.proto} package.
 */
class FetchTest extends TestBase {

  final Rel table =
      sb.namedScan(Arrays.asList("T"), Arrays.asList("a", "b"), Arrays.asList(R.I64, R.STRING));

  /** Every integer width is acceptable, in every offset/count combination. */
  @Test
  void integerWidthsAccepted() {
    List<Expression> widths = Arrays.asList(sb.i8(1), sb.i16(1), sb.i32(1), sb.i64(1));
    for (Expression offset : widths) {
      for (Expression count : widths) {
        assertDoesNotThrow(
            () -> fetch().offset(offset).count(count).build(),
            () -> "offset " + render(offset) + " / count " + render(count));
      }
    }
  }

  /**
   * A nullable integer is accepted: the spec assigns meaning to a null offset (treated as 0) and a
   * null count (all remaining rows).
   */
  @Test
  void nullableIntegerAccepted() {
    Expression nullOffset = ExpressionCreator.typedNull(N.I64);
    Expression nullCount = ExpressionCreator.typedNull(N.I32);
    assertDoesNotThrow(() -> fetch().offset(nullOffset).count(nullCount).build());
  }

  /** The check inspects the expression's type, so non-literal integer expressions are accepted. */
  @Test
  void nonLiteralIntegerExpressionsAccepted() {
    // Column "a" is i64. Isthmus passes arbitrary Calcite RexNodes through as offset/count, so the
    // check has to hold for more than literals.
    assertDoesNotThrow(() -> fetch().count(sb.fieldReference(table, 0)).build());
    Expression cast =
        ExpressionCreator.cast(R.I64, sb.i32(1), Expression.FailureBehavior.THROW_EXCEPTION);
    assertDoesNotThrow(() -> fetch().count(cast).build());
  }

  /**
   * An unbound offset/count is accepted: a partially bound plan may be serialized and exchanged
   * before its types are known, so the integer constraint cannot be enforced yet.
   */
  @Test
  void unboundAccepted() {
    Expression unboundParameter =
        Expression.DynamicParameter.builder()
            .type(Type.Unbound.builder().build())
            .parameterReference(0)
            .build();
    assertDoesNotThrow(() -> fetch().offset(unboundParameter).count(unboundParameter).build());
  }

  /** A non-integer offset expression is rejected at construction time. */
  @Test
  void nonIntegerOffsetRejected() {
    IllegalArgumentException e =
        assertThrows(IllegalArgumentException.class, () -> fetch().offset(sb.fp64(1.0)).build());
    assertTrue(e.getMessage().contains("offset"), e.getMessage());
    assertTrue(e.getMessage().contains("fp64"), e.getMessage());
  }

  /** A non-integer count expression is rejected at construction time. */
  @Test
  void nonIntegerCountRejected() {
    IllegalArgumentException e =
        assertThrows(IllegalArgumentException.class, () -> fetch().count(sb.fp64(1.0)).build());
    assertTrue(e.getMessage().contains("count"), e.getMessage());
  }

  /** A non-integer type reaches the check via the proto conversion path too. */
  @Test
  void nonIntegerOffsetRejectedViaProto() {
    // Build a FetchRel proto directly with a non-integer offset_expr so the Fetch POJO check is
    // exercised by ProtoRelConverter rather than by direct construction.
    io.substrait.proto.Expression fp64Expr =
        io.substrait.proto.Expression.newBuilder()
            .setLiteral(io.substrait.proto.Expression.Literal.newBuilder().setFp64(1.0))
            .build();
    io.substrait.proto.Rel protoRel =
        io.substrait.proto.Rel.newBuilder()
            .setFetch(
                FetchRel.newBuilder()
                    .setCommon(RelCommon.newBuilder())
                    .setInput(relProtoConverter.toProto(table))
                    .setOffsetExpr(fp64Expr))
            .build();
    assertThrows(IllegalArgumentException.class, () -> protoRelConverter.from(protoRel));
  }

  private ImmutableFetch.Builder fetch() {
    return Fetch.builder().input(table);
  }

  private static String render(Expression expression) {
    return expression.getType().accept(new StringTypeVisitor());
  }
}
```

core/src/test/java/io/substrait/type/TypeTest.java
```diff
@@ -0,0 +1,61 @@
...
+    assertFalse(R.UUID.isInteger());
+  }
+
+  @Test
+  void compoundAndSpecialTypesAreNotIntegers() {
+    assertFalse(R.decimal(10, 2).isInteger());
+    assertFalse(R.struct(R.I64).isInteger());
+    assertFalse(R.list(R.I64).isInteger());
+    assertFalse(R.map(R.I64, R.I64).isInteger());
+  }
```
nielspardon, line 46: Worth pinning the two types where "not an integer" is a decision rather than an obvious fact - `unbound` especially, since it is what the summary's `unbound` question turns on.

(Separately: the two `equalsIgnoringNullability` tests below are unrelated to this PR. They are net-new coverage for a previously untested method so I am not going to argue for removing them, just noting they would read better as their own commit.)

```suggestion
    assertFalse(R.list(R.I64).isInteger());
    assertFalse(R.map(R.I64, R.I64).isInteger());
  }

  @Test
  void unboundAndUserDefinedAreNotIntegers() {
    assertFalse(Type.Unbound.builder().build().isInteger());
    assertFalse(R.userDefined("urn:test", "t").isInteger());
  }
```

core/src/main/java/io/substrait/relation/Fetch.java
alexandrefimov: Exempting `unbound` looks right to me, and the import half of the concern is concrete: proto `Type` has an `UNBOUND` kind and `ProtoTypeConverter` maps it to `Type.Unbound`, so an unbound-typed `DynamicParameter` offset/count is representable on the wire today, and with the check as-is `from()` starts rejecting it. `LIMIT ?` is the usual prepared-statement placeholder, and `SubstraitRelNodeConverter.visit(Fetch)` already passes non-literal offset/count through to Calcite for exactly that case. Bind time is where the integer constraint becomes checkable.

alexandrefimov:
On the `BREAKING CHANGE:` footer question — #1058 marked the assert-to-exception validation change as breaking, and `from()` newly rejecting previously-importable plans is the same category. I'd include the footer.

core/src/main/java/io/substrait/relation/Fetch.java
```diff
@@ -29,6 +30,33 @@ public abstract class Fetch extends SingleInputRel implements HasExtension {
...
+   * non-negative integer; {@code i64} is recommended but not required, so any integer width is
+   * accepted (spec v0.99.0).
+   *
+   * @throws IllegalArgumentException if the offset or count expression is not integer-typed
+   */
+  @Value.Check
+  protected void check() {
+    requireIntegerType(getOffset(), "offset");
+    requireIntegerType(getCount(), "count");
+  }
```
andrew-coleman, line 45: `TopN` has the same two fields — `getOffset()`/`getCount()` are the identical `Optional<Expression>` mapping of `TopNRel.offset`/`count`, under the same spec requirement — and no equivalent check, so the constraint is bypassable by emitting a `TopNRel` instead of a `FetchRel`.

Verified on both paths at `0f7daef4`: `ImmutableTopN.builder()...offset(sb.fp64(1.0)).count(sb.fp64(2.0)).build()` succeeds, and `ProtoRelConverter.from()` on a `TopNRel` whose `offset` is an fp64 literal returns a `TopN` with `offset` typed `FP64{nullable=false}`.

`TopN.java` isn't in this diff so this can't be a suggestion, but `requireIntegerType` probably wants to be reachable from `io.substrait.relation.physical` rather than `private` here, with a matching `@Value.Check` on `TopN`.

benbellick:
I appreciate everyone’s comments and the time spent reviewing this PR. That said, I’m finding the volume of AI-assisted feedback a little difficult to follow.

I believe this change is narrow and well-scoped. The spec defines an invariant for `FetchRel`, and this PR adds enforcement for it. I’ve updated the implementation to defer validation for `unbound` types. I deliberately left `TopN` out of scope and would prefer to address it in a follow-up to keep the PR small. 

If there are any blockers left, would someone be willing to provide a human summary of what the problem is? Otherwise, it would be great to get this one in. Thanks!

core/src/main/java/io/substrait/relation/Fetch.java
```diff
@@ -29,6 +30,34 @@ public abstract class Fetch extends SingleInputRel implements HasExtension {
...
+
+  private static void requireIntegerType(Optional<Expression> expression, String field) {
+    expression.ifPresent(
+        e -> {
+          Type type = e.getType();
+          if (!(type instanceof Type.Unbound) && !type.isInteger()) {
```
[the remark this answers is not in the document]
benbellick: yup, done

PR #1080, opened 2026-08-10
feat(isthmus): observe field reference types

alexandrefimov:
## Summary

- report Calcite input-schema types for root, outer and lambda field references through `TypeObserver`
- track input row types and lambda parameter types in the conversion context
- report a field reference converted with no input in scope as an inference failure instead of failing the conversion

## Motivation

#379 lists four conversion points where the Substrait-supplied type and the type Calcite would infer can diverge; scalar functions (#1015) and window functions (#1059) are already observed. Field references are the case the issue singles out as needing access to the underlying input schema, which is what the scope tracking here adds.

## Changes

Aggregate, sort and fetch open a conversion scope so their expressions can resolve input field types. Calcite's `Aggregate` and `Sort` carry no `variablesSet`, so an outer reference bound to one of those inputs has nowhere to be declared and is rejected — previously it failed anyway, in `requireScope`, because no scope existed at all. `NamedUpdate` applies its filter with the correlation ids resolved while converting the condition and the transformations.

Related to #379.

### DIFF
--- a/isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
@@ -52,6 +52,7 @@ import java.util.HashMap;
 import java.util.HashSet;
 import java.util.LinkedList;
 import java.util.List;
+import java.util.Locale;
 import java.util.Map;
 import java.util.Optional;
 import java.util.stream.Collectors;
@@ -384,7 +385,7 @@ public class SubstraitRelNodeConverter
       }
     }
 
-    exitUncorrelatedScope(context, "aggregate");
+    exitUncorrelatedScope(context, Aggregate.class);
 
     // RelBuilder deduplicates equal aggregate calls, and AggregateCall equality ignores the stored
     // type: two measures of the same function that differ only by their declared output type would
@@ -613,7 +614,7 @@ public class SubstraitRelNodeConverter
         sort.getSortFields().stream()
             .map(sortField -> directedRexNode(sortField, context))
             .collect(Collectors.toList());
-    exitUncorrelatedScope(context, "sort");
+    exitUncorrelatedScope(context, Sort.class);
     RelNode node = relBuilder.push(child).sort(sortExpressions).build();
     return applyRemap(node, sort.getRemap());
   }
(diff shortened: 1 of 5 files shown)


alexandrefimov:
Hi @vbarua, when you have a chance, could you take a look at the next #379 slice? It extends `TypeObserver` coverage from scalar and window functions to root, outer, and lambda field references. The main review point is whether the input and lambda scope tracking is the right boundary for observing Calcite-inferred field types. All 13 checks are green. Thanks!

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
```diff
@@ -383,6 +384,8 @@ public RelNode visit(Aggregate aggregate, Context context) throws RuntimeExcepti
       }
     }
 
+    exitUncorrelatedScope(context, "aggregate");
```
eatthepear: I wonder if we could use the Aggregate rel name for real rather than hardcoding a string here? (same thing for other spots)

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
```diff
@@ -705,6 +711,10 @@ public RelNode visit(NamedUpdate update, Context context) {
           transform.getTransformation().accept(expressionRexConverter, context));
     }
 
+    java.util.Set<CorrelationId> correlationIds = context.exitScope();
```
eatthepear: Question as I'm somewhat unfamiliar with what's going on here: what is this correlationIds thing, and why is it now being fed into the filter? I had thought the filter + build lines were moved because the type observer didn't want to see them or something?

eatthepear (commented):
Hi there, I work with @vbarua. It's been a while since I've looked at substrait-java, but I left a few comments of some things I noticed. Will leave the final approval for @vbarua

bvolpato (approved):
LGTM, will defer to Victor/Justin for a final word here though!

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
alexandrefimov: Yep, good call — switched all three call sites to use the relation classes.

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
alexandrefimov: You’re right — it doesn’t need to be passed here. This scope has no `rel_anchor`, so the set is always empty. Switched back to `filter(condition)`; the move is only to keep the input type in scope while converting the condition and transformations.

alexandrefimov:
Heads-up: the two OSV failures come from the existing Calcite/Avatica dependency chain (`httpcore5` 5.3.5), not from this change. The advisory was reviewed on August 12, so rerunning the jobs will not help. I’d keep the dependency update in a separate PR.

alexandrefimov:
The dependency fix is now in #1089. All three OSV checks are green there.

vbarua (approved):
Looks good, thanks everyone!

PR #1082, opened 2026-08-11
feat(core): update to substrait v0.100.0

### LINKED ISSUE #1076: Update to Substrait v0.100.0
# Release Notes

### ⚠ BREAKING CHANGES

* **extensions:** the two-argument `subtract:ptstz_iyear` overload,
`subtract(precision_timestamp_tz<P>, interval_year)`, is deprecated as of
0.100.0. Use the three-argument `subtract:ptstz_iyear_str` overload,
`subtract(precision_timestamp_tz<P>, interval_year, string)`, and pass an
explicit IANA timezone instead. `interval_year` carries a number of months,
and month arithmetic on an instant is only definable as field arithmetic on
a civil datetime; the timezone is what maps the instant to that calendar,
so the two-argument form has no well-defined result. This also aligns
`subtract` with `add`, which offers only the three-argument form for this
type combination. Nothing is removed by this change: the deprecation is
metadata only, consumers are not required to understand or validate it, and
no existing plan changes meaning. The overload is scheduled for removal in
a later release.

### Features

* add advanced_extension to ExpandRel ([#1167](https://github.com/substrait-io/substrait/issues/1167)) ([4d5d57a](https://github.com/substrait-io/substrait/commit/4d5d57a1918edaacef529845dcc1d9664aa6068d))
* **extensions:** add unsigned integer extension types (u8, u16, u32, u64) ([#953](https://github.com/substrait-io/substrait/issues/953)) ([f5e0715](https://github.com/substrait-io/substrait/commit/f5e0715454c4a7b54ed13efe018f590d1618623d))
* **extensions:** deprecate subtract(precision_timestamp_tz, interval_year) overload ([#1160](https://github.com/substrait-io/substrait/issues/1160)) ([993ad15](https://github.com/substrait-io/substrait/commit/993ad15ca6c8b0dbcafef7316f48d83e20902dd2))

### Bug Fixes

* remove ignored nullability markers from MIRROR declarations ([#1148](https://github.com/substrait-io/substrait/issues/1148)) ([f77d2d4](https://github.com/substrait-io/substrait/commit/f77d2d4b0e0a1640fa323c7370ccfab8c15877ef))

See https://github.com/substrait-io/substrait/releases/tag/v0.100.0

anasik:
- Closes #1076
- Bump substrait-packaging to 0.100.0
- Load unsigned_integers.yaml into the default extension catalog
- Add HasExtension/advanced_extension support to Expand, matching the new ExpandRel proto field, with round-trip test coverage

### DIFF
--- a/spark/spark_dialect.yaml
+++ b/spark/spark_dialect.yaml
@@ -127,7 +127,6 @@ supported_relations:
   - "UNION_ALL"
 dependencies:
   rounding: "extension:io.substrait:functions_rounding"
-  unsigned_integers: "extension:io.substrait:unsigned_integers"
   comparison: "extension:io.substrait:functions_comparison"
   logarithmic: "extension:io.substrait:functions_logarithmic"
   datetime: "extension:io.substrait:functions_datetime"
--- a/spark/spark_dialect.yaml
+++ b/spark/spark_dialect.yaml
@@ -127,6 +127,7 @@ supported_relations:
   - "UNION_ALL"
 dependencies:
   rounding: "extension:io.substrait:functions_rounding"
+  unsigned_integers: "extension:io.substrait:unsigned_integers"
   comparison: "extension:io.substrait:functions_comparison"
   logarithmic: "extension:io.substrait:functions_logarithmic"
   datetime: "extension:io.substrait:functions_datetime"
--- a/spark/src/main/scala/io/substrait/spark/utils/DialectGenerator.scala
+++ b/spark/src/main/scala/io/substrait/spark/utils/DialectGenerator.scala
@@ -63,6 +63,7 @@ class DialectGenerator {
     "extension:io.substrait:functions_rounding" -> "rounding",
     "extension:io.substrait:functions_rounding_decimal" -> "rounding_decimal",
     "extension:io.substrait:functions_string" -> "string",
+    "extension:io.substrait:unsigned_integers" -> "unsigned_integers",
     "extension:substrait:spark" -> "spark"
   )
 
--- a/spark/src/main/scala/io/substrait/spark/utils/DialectGenerator.scala
+++ b/spark/src/main/scala/io/substrait/spark/utils/DialectGenerator.scala
@@ -63,7 +63,6 @@ class DialectGenerator {
     "extension:io.substrait:functions_rounding" -> "rounding",
     "extension:io.substrait:functions_rounding_decimal" -> "rounding_decimal",
     "extension:io.substrait:functions_string" -> "string",
-    "extension:io.substrait:unsigned_integers" -> "unsigned_integers",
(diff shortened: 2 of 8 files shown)


nielspardon:
I approved the workflow runs yesterday and it looks like the Spark dialect file also needs to be updated in this PR.

The following command should regenerate it:
```
./gradlew dialect
```

anasik:
@nielspardon done

nielspardon:
We fixed a bug in the Spark dialect generation code in https://github.com/substrait-io/substrait-java/pull/1086. If you rebase we don't need the Spark dialect changes anymore.

anasik:
@nielspardon done

nielspardon:
I think you still would need to drop that second commit with the spark changes

anasik:
@nielspardon try now

spark/src/main/scala/io/substrait/spark/utils/DialectGenerator.scala
```diff
@@ -63,6 +63,7 @@ class DialectGenerator {
     "extension:io.substrait:functions_rounding" -> "rounding",
     "extension:io.substrait:functions_rounding_decimal" -> "rounding_decimal",
     "extension:io.substrait:functions_string" -> "string",
+    "extension:io.substrait:unsigned_integers" -> "unsigned_integers",
```
nielspardon: This map drives the `dependencies` block directly (`sourceURNs.map(_.swap)`), independently of which functions actually get emitted — so with no unsigned function emitted, the entry only produces an alias no `source` points at.

```suggestion
```

spark/spark_dialect.yaml
```diff
@@ -127,6 +127,7 @@ supported_relations:
   - "UNION_ALL"
 dependencies:
   rounding: "extension:io.substrait:functions_rounding"
+  unsigned_integers: "extension:io.substrait:unsigned_integers"
```
nielspardon: Consequence of the line above — `./gradlew dialect` stops emitting this once the `sourceURNs` entry is gone, so either apply both suggestions or apply that one and regenerate.

```suggestion
```

nielspardon (commented):
Now that the generator drops variants whose signature has no Spark sample value, no unsigned function is emitted (Spark has no unsigned type mapping), so this alias has nothing referencing it: the rebased dialect has zero functions with `source: "unsigned_integers"`. On main, declared and used aliases match exactly at 12/12, and `functions_list`/`functions_set` are precedent for a catalog extension that is simply not declared here. Suggestions below to drop both lines; they can come back together with the functions if Spark ever gains an unsigned type mapping.

nielspardon (approved):
LGTM, I think we can safely drop the two lines in the spark module

PR #1088, opened 2026-08-12
fix(core): make ProtoRelConverter.newTopN keep relAnchor on the way back from proto

### LINKED ISSUE #1085: ProtoRelConverter.newTopN drops relAnchor on proto to POJO
`newTopN` (`ProtoRelConverter.java:1063`) is the only proto to POJO reader whose common block omits `.relAnchor(optionalRelAnchor(rel.getCommon()))` — its chain ends at `.hint(optionalHint(rel.getCommon()))`. Every sibling has all four calls: `newProject`, `newExpand`, `newAggregate`, `newJoin`, and all five exchange readers.

The writer does emit it: `RelProtoConverter.visit(TopN)` goes through `common(topN)`, and `common()` (`RelProtoConverter.java:936`) does `rel.getRelAnchor().ifPresent(builder::setRelAnchor)`. So this is writer-writes / reader-drops, and POJO to proto to POJO is unequal for any `TopN` with an anchor set, silently.

Effect beyond round-trip fidelity: `relAnchor` registers the scope that outer references resolve against, so a lateral join whose right side outer-references a `TopN` below it loses that registration and the reference then resolves against the wrong record type or fails to resolve.

Fix is to add the missing `.relAnchor(optionalRelAnchor(rel.getCommon()))` to the chain.

anasik:
- Closes #1085
- newTopN dropped relAnchor converting proto to POJO while the writer emitted it, breaking round-trip fidelity and losing the anchor's outer-reference binding for any TopN with an anchor set
- Added relAnchorOnTopN as a regression test

### DIFF
--- a/core/src/main/java/io/substrait/relation/ProtoRelConverter.java
+++ b/core/src/main/java/io/substrait/relation/ProtoRelConverter.java
@@ -1082,7 +1082,8 @@ public class ProtoRelConverter {
     builder
         .commonExtension(optionalAdvancedExtension(rel.getCommon()))
         .remap(optionalRelmap(rel.getCommon()))
-        .hint(optionalHint(rel.getCommon()));
+        .hint(optionalHint(rel.getCommon()))
+        .relAnchor(optionalRelAnchor(rel.getCommon()));
     if (rel.hasAdvancedExtension()) {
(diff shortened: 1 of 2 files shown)


nielspardon (approved):
LGTM, thanks for fixing this

PR #1089, opened 2026-08-13
build(isthmus): update HttpCore for CVE-2026-54399

alexandrefimov:
Calcite 1.42.0 brings in HttpCore 5.3.x through Avatica and HttpClient. Those releases are affected by CVE-2026-54399.

Constrain both `httpcore5` and `httpcore5-h2` to 5.4.3 so the modules stay aligned on the patched release and the constraint is published to downstream `isthmus` consumers.

PR #1090, opened 2026-08-13
fix(isthmus): preserve field types in sort and aggregate references

### LINKED ISSUE #215: AggregateCall conversion ignores collations
The code that converts a Calcite AggregateCall to its Substrait equivalent does not use the collation on the AggregateCall to set the sortFields on the AggregateFunctionInvocation.

alexandrefimov:
Calcite collations were already propagated to Substrait sort fields, but those field references used the whole input struct as their type. The shared conversion is used by both aggregate sort arguments and Sort relations, so those plans were not round-trip stable. Aggregate pre-measure filters had the same problem.

Build each reference from the selected input field, preserving its exact type in Sort relations, aggregate sort arguments, and pre-measure filters.

Closes #215

### DIFF
--- a/isthmus/src/main/java/io/substrait/isthmus/SubstraitRelVisitor.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/SubstraitRelVisitor.java
@@ -507,12 +507,9 @@ public class SubstraitRelVisitor extends RelNodeVisitor<Rel, RuntimeException> {
   /**
    * Converts a Calcite {@link AggregateCall} to a Substrait {@link Aggregate.Measure}.
    *
-   * <p>This method handles the conversion of aggregate function calls from Calcite's representation
-   * to Substrait's format. For statistical aggregate functions (STDDEV_POP, STDDEV_SAMP, VAR_POP,
-   * VAR_SAMP), it automatically transforms the input relation by inserting a projection that casts
-   * the aggregate function's argument fields to DOUBLE (FP64) type, ensuring type compatibility
-   * with Substrait's statistical function requirements. Fields not referenced by the aggregate
-   * function are passed through unchanged.
+   * <p>Statistical aggregate arguments have already been rewritten by {@link
+   * #castStatisticalAggregatesToFloatingPoint(org.apache.calcite.rel.core.Aggregate)} before this
+   * method runs. That rewrite appends fp64 input columns and re-points the affected calls.
    *
    * <p>The method also processes optional filter arguments (FILTER clauses) by converting them to
    * Substrait's preMeasureFilter representation.
@@ -535,7 +532,8 @@ public class SubstraitRelVisitor extends RelNodeVisitor<Rel, RuntimeException> {
     }
     Builder builder = Aggregate.Measure.builder().function(invocation.get());
     if (call.filterArg != -1) {
-      builder.preMeasureFilter(FieldReference.newRootStructReference(call.filterArg, inputType));
+      builder.preMeasureFilter(
+          FieldReference.StructField.of(call.filterArg).constructOnRoot(inputType));
     }
     return builder.build();
   }
@@ -737,9 +735,7 @@ public class SubstraitRelVisitor extends RelNodeVisitor<Rel, RuntimeException> {
     Expression.SortDirection direction = asSortDirection(collation);
 
     return Expression.SortField.builder()
-        .expr(
-            FieldReference.newRootStructReference(
-                collation.getFieldIndex(), inputType.fields().get(collation.getFieldIndex())))
+        .expr(FieldReference.StructField.of(collation.getFieldIndex()).constructOnRoot(inputType))
         .direction(direction)
         .build();
   }
--- a/isthmus/src/main/java/io/substrait/isthmus/SubstraitRelVisitor.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/SubstraitRelVisitor.java
@@ -536,8 +536,6 @@ public class SubstraitRelVisitor extends RelNodeVisitor<Rel, RuntimeException> {
     if (call.filterArg != -1) {
       builder.preMeasureFilter(FieldReference.newRootStructReference(call.filterArg, inputType));
     }
-    // TODO: handle the collation on the AggregateCall
-    //   https://github.com/substrait-io/substrait-java/issues/215
     return builder.build();
   }
 
@@ -738,7 +736,9 @@ public class SubstraitRelVisitor extends RelNodeVisitor<Rel, RuntimeException> {
     Expression.SortDirection direction = asSortDirection(collation);
 
     return Expression.SortField.builder()
-        .expr(FieldReference.newRootStructReference(collation.getFieldIndex(), inputType))
+        .expr(
+            FieldReference.newRootStructReference(
+                collation.getFieldIndex(), inputType.fields().get(collation.getFieldIndex())))
         .direction(direction)
         .build();
   }
--- a/isthmus/src/test/java/io/substrait/isthmus/AggregationFunctionsTest.java
+++ b/isthmus/src/test/java/io/substrait/isthmus/AggregationFunctionsTest.java
@@ -138,8 +138,8 @@ class AggregationFunctionsTest extends PlanTestBase {
   void aggregateCallCollation() {
     builder.values(new String[] {"value", "first_sort", "second_sort"}, 1, 2, 3);
     RexNode value = builder.field("value");
-    RexNode firstSort = builder.nullsLast(builder.field("first_sort"));
-    RexNode secondSort = builder.nullsFirst(builder.desc(builder.field("second_sort")));
+    RexNode firstSort = builder.nullsFirst(builder.field("first_sort"));
+    RexNode secondSort = builder.nullsLast(builder.desc(builder.field("second_sort")));
 
     RelNode calcite =
         builder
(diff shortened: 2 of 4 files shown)


alexandrefimov:
Known CI dependency: the two OSV failures are from the existing HttpCore version and are fixed by #1089. No additional change is needed in this PR; I will update it from main after #1089 is merged.

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelVisitor.java
```diff
@@ -537,8 +537,6 @@ Aggregate.Measure fromAggCall(RelNode input, Type.Struct inputType, AggregateCal
     if (call.filterArg != -1) {
       builder.preMeasureFilter(FieldReference.newRootStructReference(call.filterArg, inputType));
```
nielspardon: The same defect this PR fixes is still here for `preMeasureFilter`: `inputType` is the whole struct, not the filter column's type, so `SUM(x) FILTER (WHERE flag)` isn't round-trip-stable either. The TODO removed just below suggested this method was done.

```suggestion
      builder.preMeasureFilter(
          FieldReference.StructField.of(call.filterArg).constructOnRoot(inputType));
```

A `roundTripPreMeasureFilter` next to your new `roundTripSortingArguments`, reusing the existing `withPreMeasureFilter` helper, catches it.

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelVisitor.java
```diff
@@ -739,7 +737,9 @@ public static Expression.SortField toSortField(
     Expression.SortDirection direction = asSortDirection(collation);
 
     return Expression.SortField.builder()
-        .expr(FieldReference.newRootStructReference(collation.getFieldIndex(), inputType))
+        .expr(
+            FieldReference.newRootStructReference(
+                collation.getFieldIndex(), inputType.fields().get(collation.getFieldIndex())))
```
nielspardon: `constructOnRoot` already is this lookup (`FieldReference.java:396-404`) plus a range check that throws `Field reference offset (%s) must be less than number of fields in struct (%s)` — and it's the same helper `ProtoExpressionConverter` uses on the way back, so calling it makes the two directions symmetric by construction instead of by coincidence. As written, an out-of-range collation on this `public static` method gets a bare `IndexOutOfBoundsException` from inside a stream `map`, and the index is read twice.

```suggestion
        .expr(FieldReference.StructField.of(collation.getFieldIndex()).constructOnRoot(inputType))
```

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelVisitor.java
```diff
@@ -739,7 +737,9 @@ public static Expression.SortField toSortField(
     Expression.SortDirection direction = asSortDirection(collation);
```
nielspardon, line 735: Unrelated to your change, but in the helper this calls: the ASC branch treats `NullDirection.UNSPECIFIED` as `FIRST`, while Calcite's `Direction.ASCENDING.defaultNullDirection()` is `LAST`. So a hand-built `new RelFieldCollation(i, ASCENDING, UNSPECIFIED)` becomes `ASC_NULLS_FIRST` and moves nulls to the front. The DESC branch is coincidentally right. Not reachable through `RelBuilder` (it pre-resolves via `Util.first(...)`), but `SubstraitRelVisitor.convert(RelNode, …)` is public.

One-line fix if you want it here: resolve `UNSPECIFIED` through `collation.direction.defaultNullDirection()` before comparing. Otherwise it's tracked separately — no need to grow this PR.

isthmus/src/test/java/io/substrait/isthmus/AggregationFunctionsTest.java
```diff
@@ -125,4 +133,32 @@ void withGrouping(String aggFunction) {
             numericTypesTable);
     assertFullRoundTrip(rel);
   }
+
+  @Test
+  void aggregateCallCollation() {
+    builder.values(new String[] {"value", "first_sort", "second_sort"}, 1, 2, 3);
+    RexNode value = builder.field("value");
+    RexNode firstSort = builder.nullsLast(builder.field("first_sort"));
+    RexNode secondSort = builder.nullsFirst(builder.desc(builder.field("second_sort")));
```
nielspardon: These two wrappers are no-ops: ASC already defaults to NULLS LAST and DESC to NULLS FIRST, so both collations come out as the bare forms and the PR body's "mixed … null-order collations" isn't actually covered. `ASC_NULLS_FIRST` and `DESC_NULLS_LAST` are untested in both directions here and in `roundTripSortingArguments`, so a regression that ignored `nullDirection` entirely and returned the per-direction default would keep both tests green.

```suggestion
    RexNode firstSort = builder.nullsFirst(builder.field("first_sort"));
    RexNode secondSort = builder.nullsLast(builder.desc(builder.field("second_sort")));
```

Pairs with the comment below — that one updates the expected directions to match.

isthmus/src/test/java/io/substrait/isthmus/AggregationFunctionsTest.java
```diff
@@ -125,4 +133,32 @@ void withGrouping(String aggFunction) {
...
+        List.of(
+            Expression.SortField.builder()
+                .expr(FieldReference.newRootStructReference(1, inputType.fields().get(1)))
+                .direction(Expression.SortDirection.ASC_NULLS_LAST)
+                .build(),
+            Expression.SortField.builder()
+                .expr(FieldReference.newRootStructReference(2, inputType.fields().get(2)))
+                .direction(Expression.SortDirection.DESC_NULLS_FIRST)
+                .build()),
+        converted.getMeasures().get(0).getFunction().sort());
```
nielspardon: `inputType` comes from the conversion's own output and `inputType.fields().get(i)` re-runs the production expression, so this can only distinguish struct-vs-field — if the Values to `VirtualTableScan` types regressed to nullable, expected would track actual. The columns are `INTEGER NOT NULL`, so the types can be stated literally, as elsewhere in this class.

```suggestion
    assertEquals(
        List.of(
            sb.sortField(
                FieldReference.newRootStructReference(1, R.I32),
                Expression.SortDirection.ASC_NULLS_FIRST),
            sb.sortField(
                FieldReference.newRootStructReference(2, R.I32),
                Expression.SortDirection.DESC_NULLS_LAST)),
        converted.getMeasures().get(0).getFunction().sort());
```

isthmus/src/test/java/io/substrait/isthmus/ComplexAggregateTest.java
```diff
@@ -131,6 +131,27 @@ void handleComplexSortingArguments() {
     validateAggregateTransformation(rel, expectedFinal);
   }
 
+  @Test
+  void roundTripSortingArguments() {
```
nielspardon, line 151: Non-blocking: every other test in this class asserts a `PreCalciteAggregateValidator` rewrite of a *complex* argument via `validateAggregateTransformation`; this one is `assertFullRoundTrip` over plain field references, so aggregate-collation coverage now lives in two files under two idioms. Reads more naturally next to `aggregateCallCollation` in `AggregationFunctionsTest`.

nielspardon (changes requested):
The type fix is the right one, and it's worth stating that it repairs more than the title says: `toSortField` is shared with `visit(Sort)`, so this also makes every Calcite-derived `Sort` rel round-trip-stable. I checked by reverting only the production line — `assertFullRoundTrip("SELECT L_ORDERKEY FROM LINEITEM ORDER BY L_ORDERKEY")` fails on `main` and passes with your change.

Three asks that have no line in the diff to hang off:

- The `fromAggCall` Javadoc (lines 509-514) attributes the fp64 rewrite to this method and describes it as casting the argument in place. It's actually `castStatisticalAggregatesToFloatingPoint`, called from `visit(Aggregate)`, and it *appends* fp64 columns and re-points the call. This diff removes the method's only other comment, so that paragraph is now its whole description.
- Please add one `assertFullRoundTrip` with an `ORDER BY` to pin the `Sort`-rel half. Nothing currently covers it: `ComplexSortTest` compares Calcite explain text, `SubstraitRelNodeConverterTest$Sort` asserts only `getRowType()`, and `simpleOrderByClause` uses `assertSqlSubstraitRelRoundTrip`, which compares two POJOs both produced by `SubstraitRelVisitor` and so is blind to a uniformly-wrong type.
- Since the squash message becomes the changelog entry, could the title name the real defect and the `Sort` scope rather than "preserve aggregate collations"? Collations have in fact been propagated since the initial commit (`AggregateFunctionConverter.sortFields`), which your body says correctly — the removed TODO was just stale, so #215's literal text was already satisfied and the type defect is the fix worth advertising. I'd also drop the "cover mixed direction and null-order collations…" sentence; the diff shows the tests.

Everything else I ran into is pre-existing and out of scope here, and has been filed as issues.

alexandrefimov:
Addressed the requested changes in 56106e7: fixed the pre-measure filter field type, switched sort references to constructOnRoot, strengthened the null-order expectations, corrected the fromAggCall Javadoc, and added a full ORDER BY round-trip. I also updated the title and description to describe the field-type defect and Sort scope. I left NullDirection.UNSPECIFIED and the optional test move out of this PR as suggested.

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelVisitor.java
alexandrefimov: Fixed in 56106e7. preMeasureFilter now uses constructOnRoot(inputType), with a focused full round-trip test.

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelVisitor.java
alexandrefimov: Changed to constructOnRoot(inputType) in 56106e7. This also gives the same range validation as the reverse conversion.

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelVisitor.java
alexandrefimov: Agreed this is separate from the field-type fix. I left it out of this PR as suggested.

isthmus/src/test/java/io/substrait/isthmus/AggregationFunctionsTest.java
alexandrefimov: Fixed in 56106e7. Both direct conversion and full round-trip tests now use non-default ASC NULLS FIRST and DESC NULLS LAST.

isthmus/src/test/java/io/substrait/isthmus/AggregationFunctionsTest.java
alexandrefimov: Fixed in 56106e7. The expectations now use explicit non-nullable R.I32 field types.

isthmus/src/test/java/io/substrait/isthmus/ComplexAggregateTest.java
alexandrefimov: Thanks. I kept the existing test location in this PR to avoid unrelated test movement; the behavior is covered by the focused round-trip.

nielspardon (approved):
All addressed — thanks. The `UNSPECIFIED` null-direction case you left out is tracked in #1092.

PR #1099, opened 2026-08-14
fix(isthmus)!: use a distinct datetime subtraction operator

### LINKED ISSUE #377: (isthmus) wrong Substrait to Calcite mapping for subtract:date_iday
When the function `subtract:date_iday` is mapped from Calcite to Substrait, if the Calcite type inference is used to derive the return type a java.lang.ArrayIndexOutOfBoundsException is thrown. This likely indicates that the function it is being mapped to in Calcite is not quite the right one.

This issue is sidestepped by using the return type as given by the scalar function invocation, see https://github.com/substrait-io/substrait-java/pull/355, but it _should_ be feasible to use the Calcite type inference and we should still investigate the underlying issue.

alexandrefimov:
`subtract:date_iday` was resolved to Calcite `MINUS_DATE`, whose return type inference expects a third operand. The earlier `MINUS_DATE2` approach removed the exception, but it still compared equal to `MINUS_DATE` and inferred `DATE`.

Use an Isthmus-owned `DATETIME_SUBTRACT` operator. It has a distinct Calcite identity, handles both two-operand datetime/interval and three-operand timezone forms, and infers `TIMESTAMP` at the configured maximum precision for `DATE - interval_day`. A dedicated reverse mapper keeps existing Calcite-to-Substrait conversion of the stock datetime subtraction operators working.

Regression coverage includes both the legacy `DATE` output shape and the spec `precision_timestamp<P>` shape, independent operator identity, TypeObserver inference, round-trip conversion, and the three-operand timezone digest path.

BREAKING CHANGE: Converted Substrait datetime subtraction expressions now use an Isthmus-owned operator without a Calcite Enumerable implementation. Consumers executing these plans through the Enumerable convention must provide an implementation or handle this operator before execution.

Closes #377

### DIFF
--- a/isthmus/src/main/java/io/substrait/isthmus/expression/FunctionMappings.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/expression/FunctionMappings.java
@@ -8,9 +8,13 @@ import java.util.Set;
 import org.apache.calcite.rel.type.RelDataType;
 import org.apache.calcite.rel.type.RelDataTypeFactory;
 import org.apache.calcite.sql.SqlBasicFunction;
+import org.apache.calcite.sql.SqlCall;
 import org.apache.calcite.sql.SqlFunction;
+import org.apache.calcite.sql.SqlFunctionCategory;
+import org.apache.calcite.sql.SqlKind;
 import org.apache.calcite.sql.SqlOperator;
 import org.apache.calcite.sql.SqlOperatorBinding;
+import org.apache.calcite.sql.SqlWriter;
 import org.apache.calcite.sql.fun.SqlLibraryOperators;
 import org.apache.calcite.sql.fun.SqlStdOperatorTable;
 import org.apache.calcite.sql.type.OperandTypes;
@@ -74,17 +78,7 @@ public class FunctionMappings {
    * two-operand {@code date - interval_day}, whose result is a precision timestamp. This distinct
    * operator keeps Calcite from treating those incompatible operators as equal.
    */
-  public static final SqlFunction DATETIME_SUBTRACT =
-      SqlBasicFunction.create(
-          "DATETIME_SUBTRACT",
-          FunctionMappings::inferDatetimeSubtractType,
-          OperandTypes.or(
-              OperandTypes.DATETIME_INTERVAL,
-              OperandTypes.sequence(
-                  "DATETIME_SUBTRACT(<DATETIME>, <INTERVAL>, <CHARACTER>)",
-                  OperandTypes.DATETIME,
-                  OperandTypes.INTERVAL,
-                  OperandTypes.CHARACTER)));
+  public static final SqlFunction DATETIME_SUBTRACT = new DatetimeSubtractionFunction();
 
   /** Scalar function mappings. */
   public static final ImmutableList<Sig> SCALAR_SIGS =
@@ -262,6 +256,31 @@ public class FunctionMappings {
       nullable |= binding.getOperandType(i).isNullable();
     }
     return typeFactory.createTypeWithNullability(resultType, nullable);
+  }
+
+  private static final class DatetimeSubtractionFunction extends SqlFunction {
+    private DatetimeSubtractionFunction() {
+      super(
+          "DATETIME_SUBTRACT",
+          SqlKind.OTHER_FUNCTION,
+          FunctionMappings::inferDatetimeSubtractType,
+          null,
+          OperandTypes.or(
+              OperandTypes.DATETIME_INTERVAL,
+              OperandTypes.sequence(
+                  "DATETIME_SUBTRACT(<DATETIME>, <INTERVAL>, <CHARACTER>)",
+                  OperandTypes.DATETIME,
+                  OperandTypes.INTERVAL,
+                  OperandTypes.CHARACTER)),
+          SqlFunctionCategory.NUMERIC);
+    }
+
+    @Override
+    public void unparse(SqlWriter writer, SqlCall call, int leftPrec, int rightPrec) {
+      writer
+          .getDialect()
+          .unparseSqlDatetimeArithmetic(writer, call, SqlKind.MINUS, leftPrec, rightPrec);
+    }
   }
 
   /**
--- a/isthmus/src/main/java/io/substrait/isthmus/expression/FunctionMappings.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/expression/FunctionMappings.java
@@ -248,10 +248,10 @@ public class FunctionMappings {
     RelDataType resultType = datetimeType;
     if (datetimeType.getSqlTypeName() == SqlTypeName.DATE
         && intervalType.getFamily() == SqlTypeFamily.INTERVAL_DAY_TIME) {
-      // Calcite's interval type does not retain Substrait's interval_day<P> precision, so the
-      // exact timestamp precision cannot be inferred here. Use the type system maximum as a
-      // best-effort inference; ExpressionRexConverter keeps the declared Substrait output type
-      // authoritative.
+      // Isthmus builds every interval_day with SubstraitTypeSystem.DAY_SECOND_INTERVAL, which
+      // pins the fractional-second precision at 6, so the operand no longer carries Substrait's
+      // interval_day<P>. Use the type system maximum as a best-effort inference;
+      // ExpressionRexConverter keeps the declared Substrait output type authoritative.
       int precision = typeFactory.getTypeSystem().getMaxPrecision(SqlTypeName.TIMESTAMP);
       resultType = typeFactory.createSqlType(SqlTypeName.TIMESTAMP, precision);
     }
--- a/isthmus/src/main/java/io/substrait/isthmus/expression/FunctionMappings.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/expression/FunctionMappings.java
@@ -77,6 +77,10 @@ public class FunctionMappings {
    * interval qualifier operand or return the first operand type. Neither matches Substrait's
    * two-operand {@code date - interval_day}, whose result is a precision timestamp. This distinct
    * operator keeps Calcite from treating those incompatible operators as equal.
+   *
+   * <p>Calcite's Enumerable convention cannot execute this custom operator because it has no {@code
+   * RexImpTable} implementor. It is currently used to preserve conversion semantics and operator
+   * identity.
    */
   public static final SqlFunction DATETIME_SUBTRACT = new DatetimeSubtractionFunction();
 
@@ -244,9 +248,10 @@ public class FunctionMappings {
     RelDataType resultType = datetimeType;
     if (datetimeType.getSqlTypeName() == SqlTypeName.DATE
         && intervalType.getFamily() == SqlTypeFamily.INTERVAL_DAY_TIME) {
-      // Calcite interval literals report their millisecond storage scale rather than the
-      // configured Substrait interval precision. Use the type system constraint so inference does
-      // not silently narrow a precision_timestamp result.
+      // Calcite's interval type does not retain Substrait's interval_day<P> precision, so the
+      // exact timestamp precision cannot be inferred here. Use the type system maximum as a
+      // best-effort inference; ExpressionRexConverter keeps the declared Substrait output type
+      // authoritative.
       int precision = typeFactory.getTypeSystem().getMaxPrecision(SqlTypeName.TIMESTAMP);
       resultType = typeFactory.createSqlType(SqlTypeName.TIMESTAMP, precision);
     }
@@ -272,11 +277,17 @@ public class FunctionMappings {
                   OperandTypes.DATETIME,
                   OperandTypes.INTERVAL,
                   OperandTypes.CHARACTER)),
-          SqlFunctionCategory.NUMERIC);
+          SqlFunctionCategory.TIMEDATE);
     }
 
     @Override
     public void unparse(SqlWriter writer, SqlCall call, int leftPrec, int rightPrec) {
+      if (call.operandCount() > 2) {
+        // unparseSqlDatetimeArithmetic treats the third operand as an interval qualifier and
+        // writes it after the closing parenthesis. Substrait uses it as a timezone argument.
+        super.unparse(writer, call, leftPrec, rightPrec);
+        return;
+      }
       writer
           .getDialect()
           .unparseSqlDatetimeArithmetic(writer, call, SqlKind.MINUS, leftPrec, rightPrec);
--- a/isthmus/src/main/java/io/substrait/isthmus/expression/FunctionMappings.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/expression/FunctionMappings.java
@@ -5,9 +5,12 @@ import io.substrait.isthmus.AggregateFunctions;
 import java.util.Locale;
 import java.util.Map;
 import java.util.Set;
+import org.apache.calcite.rel.type.RelDataType;
+import org.apache.calcite.rel.type.RelDataTypeFactory;
 import org.apache.calcite.sql.SqlBasicFunction;
 import org.apache.calcite.sql.SqlFunction;
 import org.apache.calcite.sql.SqlOperator;
+import org.apache.calcite.sql.SqlOperatorBinding;
 import org.apache.calcite.sql.fun.SqlLibraryOperators;
 import org.apache.calcite.sql.fun.SqlStdOperatorTable;
 import org.apache.calcite.sql.type.OperandTypes;
@@ -63,6 +66,26 @@ public class FunctionMappings {
           ReturnTypes.ARG0_NULLABLE,
           OperandTypes.family(SqlTypeFamily.INTEGER, SqlTypeFamily.INTEGER));
 
+  /**
+   * The Substrait datetime subtraction function.
+   *
+   * <p>Calcite's datetime subtraction operators either infer their return type from a third
+   * interval qualifier operand or return the first operand type. Neither matches Substrait's
+   * two-operand {@code date - interval_day}, whose result is a precision timestamp. This distinct
+   * operator keeps Calcite from treating those incompatible operators as equal.
+   */
+  public static final SqlFunction DATETIME_SUBTRACT =
+      SqlBasicFunction.create(
+          "DATETIME_SUBTRACT",
+          FunctionMappings::inferDatetimeSubtractType,
+          OperandTypes.or(
+              OperandTypes.DATETIME_INTERVAL,
+              OperandTypes.sequence(
+                  "DATETIME_SUBTRACT(<DATETIME>, <INTERVAL>, <CHARACTER>)",
+                  OperandTypes.DATETIME,
+                  OperandTypes.INTERVAL,
+                  OperandTypes.CHARACTER)));
+
   /** Scalar function mappings. */
   public static final ImmutableList<Sig> SCALAR_SIGS =
       ImmutableList.<Sig>builder()
@@ -98,7 +121,7 @@ public class FunctionMappings {
               s(SqlStdOperatorTable.IS_NULL, "is_null"),
               s(SqlStdOperatorTable.IS_NOT_NULL, "is_not_null"),
               s(SqlStdOperatorTable.NOT_EQUALS, "not_equal"),
-              s(SqlStdOperatorTable.MINUS_DATE, "subtract"),
+              s(DATETIME_SUBTRACT, "subtract"),
               s(SqlStdOperatorTable.DATETIME_PLUS, "add"),
               s(SqlStdOperatorTable.EXTRACT, "extract"),
               s(SqlStdOperatorTable.CEIL, "ceil"),
@@ -214,10 +237,32 @@ public class FunctionMappings {
           SqlStdOperatorTable.MINUS,
           resolver(
               SqlStdOperatorTable.MINUS, Set.of("i8", "i16", "i32", "i64", "fp32", "fp64", "dec")),
-          SqlStdOperatorTable.MINUS_DATE,
-          resolver(SqlStdOperatorTable.MINUS_DATE, Set.of("date", "ts", "tstz", "pts", "ptstz")),
+          DATETIME_SUBTRACT,
+          resolver(DATETIME_SUBTRACT, Set.of("date", "ts", "tstz", "pts", "ptstz")),
           SqlStdOperatorTable.BIT_LEFT_SHIFT,
           resolver(SqlStdOperatorTable.BIT_LEFT_SHIFT, Set.of("i8", "i16", "i32", "i64")));
+
+  private static RelDataType inferDatetimeSubtractType(SqlOperatorBinding binding) {
+    RelDataType datetimeType = binding.getOperandType(0);
+    RelDataType intervalType = binding.getOperandType(1);
+    RelDataTypeFactory typeFactory = binding.getTypeFactory();
+
+    RelDataType resultType = datetimeType;
+    if (datetimeType.getSqlTypeName() == SqlTypeName.DATE
+        && intervalType.getFamily() == SqlTypeFamily.INTERVAL_DAY_TIME) {
+      // Calcite interval literals report their millisecond storage scale rather than the
+      // configured Substrait interval precision. Use the type system constraint so inference does
+      // not silently narrow a precision_timestamp result.
+      int precision = typeFactory.getTypeSystem().getMaxPrecision(SqlTypeName.TIMESTAMP);
+      resultType = typeFactory.createSqlType(SqlTypeName.TIMESTAMP, precision);
+    }
+
+    boolean nullable = false;
+    for (int i = 0; i < binding.getOperandCount(); i++) {
+      nullable |= binding.getOperandType(i).isNullable();
+    }
+    return typeFactory.createTypeWithNullability(resultType, nullable);
+  }
 
   /**
    * Creates a signature mapping entry.
--- a/isthmus/src/main/java/io/substrait/isthmus/expression/ScalarFunctionConverter.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/expression/ScalarFunctionConverter.java
@@ -17,9 +17,6 @@ import org.apache.calcite.rel.type.RelDataType;
 import org.apache.calcite.rel.type.RelDataTypeFactory;
 import org.apache.calcite.rex.RexCall;
 import org.apache.calcite.rex.RexNode;
-import org.apache.calcite.sql.SqlOperator;
-import org.apache.calcite.sql.fun.SqlInternalOperators;
-import org.apache.calcite.sql.fun.SqlStdOperatorTable;
 
 /**
  * Converts Calcite {@link RexCall} scalar functions to Substrait {@link Expression} using known
@@ -68,6 +65,7 @@ public class ScalarFunctionConverter
 
     mappers =
         List.of(
+            new DatetimeSubtractionFunctionMapper(functions),
             new ConcatFunctionMapper(functions),
             new TrimFunctionMapper(functions),
             new SqrtFunctionMapper(functions),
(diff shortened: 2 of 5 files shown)


alexandrefimov:
The two failing OSV checks are unrelated to this change. The new httpclient5 advisory is fixed in #1100.

isthmus/src/main/java/io/substrait/isthmus/expression/ScalarFunctionConverter.java
```diff
@@ -85,6 +88,20 @@ protected ImmutableList<FunctionMappings.Sig> getSigs() {
...
+    return super.getSqlOperatorFromSubstraitFunc(key, outputType, arguments)
+        .map(
+            operator ->
+                // MINUS_DATE models datetime - datetime and derives its interval result from a
+                // third qualifier operand. Substrait datetime subtraction takes an interval as
+                // its second operand, which is the two-operand MINUS_DATE2 form in Calcite.
+                operator == SqlStdOperatorTable.MINUS_DATE
+                    ? SqlInternalOperators.MINUS_DATE2
+                    : operator);
+  }
```
nielspardon: This belongs in the registry rather than an override — `substraitFuncKeyToSqlOperatorMap` is built from `getSigs()`, so `FunctionMappings:101` does it declaratively:

```diff
-    s(SqlStdOperatorTable.MINUS_DATE, "subtract"),
+    s(SqlInternalOperators.MINUS_DATE2, "subtract"),
```

`MINUS_DATE` calls still resolve by equality, so this makes the `convert()` block below unnecessary too. Move the entry rather than adding one — `OPERATOR_RESOLVER` is a `Map.of` and an equal second key throws at class-init. Precedent: `ConcatFunctionMapper`.

The override is also on the 3-arg overload while callers use the 2-arg one, which is the form subclasses override (`SubstraitExpressionConverterTest:748`) — they'd silently lose the fix.

On the comment: `MINUS_DATE2` isn't a two-operand operator (both share `OperandTypes.MINUS_DATE_OPERATOR`, count range 3/3 — only the return-type inference differs), and a two-operand `MINUS_DATE` Rex call is what Calcite itself builds in `convertDatetimeMinus`. Keying on operator identity rather than "two args, second an interval" also swaps in `MINUS_DATE2` for the 3-arg timezone overloads, whose `toString()` throws.

isthmus/src/main/java/io/substrait/isthmus/expression/ScalarFunctionConverter.java
```diff
@@ -96,10 +113,16 @@ protected ImmutableList<FunctionMappings.Sig> getSigs() {
...
+                rexBuilder.makeCall(
+                    call.getType(), SqlStdOperatorTable.MINUS_DATE, call.getOperands())
+            : call;
     // If a mapping applies to this call, use it; otherwise default behavior.
-    return getMappingForCall(call)
-        .map(mapping -> mappedConvert(mapping, call, topLevelConverter))
-        .orElseGet(() -> defaultConvert(call, topLevelConverter));
+    return getMappingForCall(mappedCall)
+        .map(mapping -> mappedConvert(mapping, mappedCall, topLevelConverter))
+        .orElseGet(() -> defaultConvert(mappedCall, topLevelConverter));
```
nielspardon: Inert, per the operator equality — `signatures` is a plain `HashMap`, so this call already resolved before the change. Costs a throwaway `RexCall`, an unchecked downcast, and the parser position (`makeCall` here uses `SqlParserPos.ZERO`).

```suggestion
    // If a mapping applies to this call, use it; otherwise default behavior.
    return getMappingForCall(call)
        .map(mapping -> mappedConvert(mapping, call, topLevelConverter))
        .orElseGet(() -> defaultConvert(call, topLevelConverter));
```

Leaves `SqlStdOperatorTable` unused here if the override moves.

isthmus/src/test/java/io/substrait/isthmus/FunctionConversionTest.java
```diff
@@ -50,22 +58,33 @@ class FunctionConversionTest extends PlanTestBase {
...
-    // TODO: https://github.com/substrait-io/substrait-java/issues/377
+    AtomicReference<TypeObservation> observed = new AtomicReference<>();
+    ExpressionRexConverter observingConverter =
+        new ExpressionRexConverter(
+            typeFactory,
+            scalarFnConverter,
+            windowFnConverter,
+            TypeConverter.DEFAULT,
+            observed::set);
+    Type outputType = TypeCreator.REQUIRED.precisionTimestamp(6);
```
nielspardon: DATE wasn't invalid — it's what our own forward direction emits (`FunctionFinder` takes the output type from `call.getType()`; the declared `precision_timestamp<P>` is never enforced). Testing the spec shape is right, but the shape we actually produce now has no assertion anywhere, and it's the one that reproduces the AIOOBE if this regresses. A `@ParameterizedTest` over both would cover it.

isthmus/src/test/java/io/substrait/isthmus/FunctionConversionTest.java
```diff
@@ -50,22 +58,33 @@ class FunctionConversionTest extends PlanTestBase {
...
             ExpressionCreator.intervalDay(false, 120, 0, 0, 6));
 
-    RexNode calciteExpr = expr.accept(expressionRexConverter, Context.newContext());
+    RexNode calciteExpr = expr.accept(observingConverter, Context.newContext());
+    assertEquals(TypeConverter.DEFAULT.toCalcite(typeFactory, outputType), calciteExpr.getType());
+    assertSame(
+        SqlInternalOperators.MINUS_DATE2,
+        assertInstanceOf(RexCall.class, calciteExpr).getOperator());
+    assertTrue(observed.get().inferenceFailure().isEmpty());
+    RelDataType inferredType = observed.get().inferredType().orElseThrow();
```
nielspardon: `assertSame` fails with two identity hashes on same-named operators, and `assertTrue(…isEmpty())` duplicates the next line's `orElseThrow()` while discarding the exception the PR is about.

```suggestion
    // assertSame is required: SqlOperator.equals compares only (class, name, kind), so
    // assertEquals would pass for MINUS_DATE too.
    assertSame(
        SqlInternalOperators.MINUS_DATE2,
        assertInstanceOf(RexCall.class, calciteExpr).getOperator(),
        "expected the two-operand MINUS_DATE2 form, not MINUS_DATE");
    TypeObservation observation = observed.get();
    assertNotNull(observation, "no type observation was recorded");
    observation
        .inferenceFailure()
        .ifPresent(e -> fail("Calcite return-type inference failed", e));
    RelDataType inferredType = observation.inferredType().orElseThrow();
```

Needs `assertNotNull`/`fail` imported, drops `assertTrue`.

isthmus/src/test/java/io/substrait/isthmus/FunctionConversionTest.java
```diff
@@ -50,22 +58,33 @@ class FunctionConversionTest extends PlanTestBase {
...
     assertEquals(
-        TypeConverter.DEFAULT.toCalcite(typeFactory, TypeCreator.REQUIRED.DATE),
-        calciteExpr.getType());
+        TypeConverter.DEFAULT.toCalcite(typeFactory, TypeCreator.REQUIRED.DATE), inferredType);
+    assertNotEquals(calciteExpr.getType(), inferredType);
```
nielspardon: The `assertNotEquals` can't fail on its own — both operands are pinned exactly above — and it's what turns the type gap into a pinned expectation; the first commit asserted the equality.

```suggestion
    // Known gap: MINUS_DATE2 infers ARG0 (DATE) while the spec declares
    // subtract(date, interval_day<P>) -> precision_timestamp<P>. The supplied Substrait type
    // stays authoritative (asserted above); Calcite's inference does not yet agree.
    assertEquals(
        TypeConverter.DEFAULT.toCalcite(typeFactory, TypeCreator.REQUIRED.DATE), inferredType);
```

Also drops the `assertNotEquals` import. The DATE assertion pins a calcite-1.42 internal, so an upgrade fixing that inference turns it red with Isthmus unchanged.

nielspardon (commented):
The diagnosis is right — `MINUS_DATE`'s `ARG2_NULLABLE` is what fails on a two-operand Rex call — but I think the fix needs redirecting.

`MINUS_DATE2` is `equals` to `MINUS_DATE`: the `SqlDatetimeSubtractionOperator` constructor discards its `name` argument and calls `super("-", SqlKind.MINUS, …)`, and `SqlOperator.equals` compares only (class, name, kind). So half the diff is inert, the rest fits in one token in `FunctionMappings` (threads below), and `RexCall.equals` can't separate the two nodes either — CSE in `RelBuilder.project` can substitute `MINUS_DATE` back into a plan mixing converted and SQL-parsed subplans, reviving the exception this removes.

It also doesn't fix the type. `MINUS_DATE2` infers `ARG0`, but the spec declares `subtract(date, interval_day<P>) -> precision_timestamp<P>`, so inference now returns DATE where the node says TIMESTAMP(6) — a throw becomes a silently truncating re-derivation. #377 asks for agreement, so I'd rather not close it here. Calcite's `add` counterpart manages this via `SqlTimestampAddFunction.deduceType`; no stock minus operator does, which points at an Isthmus-owned operator (as `FunctionMappings.RIGHTSHIFT` already does) or a cast.

Also worth a look: `PrecisionTimestampDatetimeSubtractionTest`'s Javadoc still says `MINUS_DATE`, now true only of the forward leg; `TypeObserver` consumers silently change signal; and moving public output from `SqlStdOperatorTable` to `SqlInternalOperators` breaks downstream `==` checks while `.equals()` still passes — a plain `fix` puts nothing in the changelog for that.

isthmus/src/main/java/io/substrait/isthmus/expression/ScalarFunctionConverter.java
alexandrefimov: Addressed with a distinct `DATETIME_SUBTRACT` operator instead of `MINUS_DATE2`. Calcite considers the stock datetime-minus operators equal, so using either one still collides in the mapping. The custom operator now has independent identity and correct timestamp inference; the parameterized test covers both the legacy `DATE` output and the spec timestamp output, and the timezone form has a stable digest.

isthmus/src/main/java/io/substrait/isthmus/expression/FunctionMappings.java
```diff
@@ -214,11 +231,58 @@ public class FunctionMappings {
...
+                  OperandTypes.CHARACTER)),
+          SqlFunctionCategory.NUMERIC);
+    }
+
+    @Override
+    public void unparse(SqlWriter writer, SqlCall call, int leftPrec, int rightPrec) {
+      writer
+          .getDialect()
+          .unparseSqlDatetimeArithmetic(writer, call, SqlKind.MINUS, leftPrec, rightPrec);
+    }
```
nielspardon, line 294: `SqlDialect.unparseSqlDatetimeArithmetic` writes operand 2 bare and *outside* the closing paren — it is built for `MINUS_DATE`'s trailing `SqlIntervalQualifier` (`(a - b) DAY`). Given a timezone string you get invalid SQL, silently:

```
2-operand: (TIMESTAMP '1970-01-01 00:00:00.000' - INTERVAL '1' YEAR)
3-operand: (TIMESTAMP '1970-01-01 00:00:00.000' - INTERVAL '1' YEAR) 'UTC'
```

That is this operator through `PostgresqlSqlDialect.DEFAULT`, on the three-operand call `subtract:ptstz_iyear_str` actually produces. No infix form carries a timezone, so rendering it properly is its own change — but the delegation shouldn't emit broken SQL meanwhile:

```suggestion
    @Override
    public void unparse(SqlWriter writer, SqlCall call, int leftPrec, int rightPrec) {
      if (call.operandCount() > 2) {
        // unparseSqlDatetimeArithmetic is built for MINUS_DATE's trailing interval qualifier and
        // would emit the timezone operand bare, after the closing parenthesis.
        super.unparse(writer, call, leftPrec, rightPrec);
        return;
      }
      writer
          .getDialect()
          .unparseSqlDatetimeArithmetic(writer, call, SqlKind.MINUS, leftPrec, rightPrec);
    }
```

The commit message and description both claim the three-operand form is handled — that's the claim I'd either walk back or finish.

isthmus/src/main/java/io/substrait/isthmus/expression/FunctionMappings.java
```diff
@@ -214,11 +231,58 @@ public class FunctionMappings {
...
+          null,
+          OperandTypes.or(
+              OperandTypes.DATETIME_INTERVAL,
+              OperandTypes.sequence(
+                  "DATETIME_SUBTRACT(<DATETIME>, <INTERVAL>, <CHARACTER>)",
+                  OperandTypes.DATETIME,
+                  OperandTypes.INTERVAL,
+                  OperandTypes.CHARACTER)),
+          SqlFunctionCategory.NUMERIC);
+    }
```
nielspardon, line 281: Owning the operator costs executability: `RexImpTable.INSTANCE.get(...)` returns `DatetimeArithmeticImplementor` for `MINUS_DATE` and `MINUS_DATE2`, and `null` for this one — so a converted `RelNode` run through Calcite's enumerable convention can no longer translate datetime subtraction. Nothing here catches it; `PostgreSqlIntegrationTest` executes generated SQL, not Rex.

Against `main` that's arguably a fair trade (there, conversion succeeded and any Calcite re-derivation threw), but it's worth being deliberate: either register an implementor or state the limitation.

Separately, `SqlFunctionCategory.NUMERIC` on a datetime function looks like a slip — `TIMEDATE` fits.

isthmus/src/main/java/io/substrait/isthmus/expression/FunctionMappings.java
```diff
@@ -214,11 +231,58 @@ public class FunctionMappings {
...
+      // not silently narrow a precision_timestamp result.
+      int precision = typeFactory.getTypeSystem().getMaxPrecision(SqlTypeName.TIMESTAMP);
+      resultType = typeFactory.createSqlType(SqlTypeName.TIMESTAMP, precision);
+    }
+
+    boolean nullable = false;
+    for (int i = 0; i < binding.getOperandCount(); i++) {
+      nullable |= binding.getOperandType(i).isNullable();
+    }
+    return typeFactory.createTypeWithNullability(resultType, nullable);
```
nielspardon, line 263: Your reply says the operator now has correct timestamp inference. It does at precision 6 — pinning to `getMaxPrecision(TIMESTAMP)` leaves the disagreement in place everywhere else:

| declared output | node type | inferred |
| --- | --- | --- |
| `precision_timestamp<0>` | `TIMESTAMP(0)` | `TIMESTAMP(6)` |
| `precision_timestamp<3>` | `TIMESTAMP(3)` | `TIMESTAMP(6)` |
| `precision_timestamp<6>` | `TIMESTAMP(6)` | `TIMESTAMP(6)` |

It's also type-system dependent: 6 under `SubstraitTypeSystem`, 3 under Calcite's default `MAX_DATETIME_PRECISION`. A `SqlOperatorBinding` can't recover the interval's `P`, so exact inference may simply not be reachable — if that's the conclusion, I'd say so in the comment (the plan's type stays authoritative) rather than "does not silently narrow", and cover one non-max precision in the parameterized test.

Minor: the nullability loop is `SqlTypeTransforms.TO_NULLABLE`, which `SqlDatetimePlusOperator` composes via `.andThen(...)`.

isthmus/src/main/java/io/substrait/isthmus/expression/FunctionMappings.java
```diff
@@ -98,7 +115,7 @@ public class FunctionMappings {
               s(SqlStdOperatorTable.IS_NULL, "is_null"),
               s(SqlStdOperatorTable.IS_NOT_NULL, "is_not_null"),
               s(SqlStdOperatorTable.NOT_EQUALS, "not_equal"),
-              s(SqlStdOperatorTable.MINUS_DATE, "subtract"),
+              s(DATETIME_SUBTRACT, "subtract"),
```
nielspardon, line 122: The `add` line immediately below has the same gap, and this PR closes #377. Spec declares `add(date, interval_day<P>) -> precision_timestamp<P>`, and stock `DATETIME_PLUS` infers `DATE`: I get node `TIMESTAMP(6)` vs inferred `DATE`, with no exception — so it fails quietly rather than loudly, which is why it never showed up as a bug.

Fine to leave for a follow-up, but worth saying in the issue or description that only the subtract half is fixed. It's also what leaves the pair asymmetric — `add` keeps a stock, executable operator (thread 2).

isthmus/src/test/java/io/substrait/isthmus/FunctionConversionTest.java
```diff
@@ -48,29 +64,72 @@ class FunctionConversionTest extends PlanTestBase {
...
+            "subtract:ptstz_iyear_str",
+            TypeCreator.REQUIRED.precisionTimestampTZ(3),
+            ExpressionCreator.precisionTimestampTZ(false, 0, 3),
+            ExpressionCreator.intervalYear(false, 1, 0),
+            ExpressionCreator.string(false, "UTC"));
+
+    RexCall calciteExpr =
+        assertInstanceOf(RexCall.class, expr.accept(expressionRexConverter, Context.newContext()));
+    assertSame(FunctionMappings.DATETIME_SUBTRACT, calciteExpr.getOperator());
+    assertDoesNotThrow(calciteExpr::toString);
```
nielspardon: A stable digest is a real improvement over the old crash, but `assertDoesNotThrow(calciteExpr::toString)` is much weaker than "handles the three-operand timezone form" — it passes while the SQL for this exact call is malformed (thread 1).

Worth an assertion that runs this through `SubstraitToSql`. It can't be driven from SQL input the way `dateSubtractIntervalUnparsesAsSqlArithmetic` is, since no SQL parses to `subtract:ptstz_iyear_str` — it needs a hand-built plan wrapping this invocation.

nielspardon (commented):
The redirect landed. I verified `DATETIME_SUBTRACT.equals(MINUS_DATE)` and `.equals(MINUS_DATE2)` are false in both directions, so the `Map.of` collision, the CSE substitution and the reverse-lookup dedup are all gone — and since the two stock operators still compare equal to each other, the new mapper picks up `MINUS_DATE2` calls for free. Dead pre-map and the assertion issues are resolved too.

Correcting my earlier review: I pointed at `SqlTimestampAddFunction.deduceType` as the mechanism to reuse. It returns the datetime type unchanged for `(DAY, DATE)`, so it yields `DATE` — your special case does more than the stock operator, not less.

Five threads below, all verified at runtime against calcite-core 1.42.0. The three-operand SQL output is the one I'd want fixed before merge; the rest are judgement calls.

One request on the description rather than the code: make the `BREAKING CHANGE:` footer say what breaks for consumers (the executability point in thread 2) instead of the operator rename — that footer is what reaches `CHANGELOG.md`.

isthmus/src/main/java/io/substrait/isthmus/expression/FunctionMappings.java
alexandrefimov: Good catch. Fixed in 322c88e: calls with more than two operands now use the normal function-call unparse path, while two-operand calls keep dialect datetime arithmetic. I also added a SubstraitToSql regression test for subtract:ptstz_iyear_str that checks the timezone stays inside the function call.

isthmus/src/main/java/io/substrait/isthmus/expression/FunctionMappings.java
alexandrefimov: Agreed. In 322c88e I changed the category to TIMEDATE and documented that the custom operator has no RexImpTable implementor, so Enumerable cannot execute it. I left execution support outside this PR; the limitation is now explicit.

isthmus/src/main/java/io/substrait/isthmus/expression/FunctionMappings.java
alexandrefimov: Agreed. The exact P is not available from the Calcite interval type. I updated the comment to describe max precision as best-effort inference and the declared Substrait output as authoritative. The parameterized test now covers P=3 and asserts node TIMESTAMP(3) versus inferred TIMESTAMP(6).

isthmus/src/main/java/io/substrait/isthmus/expression/FunctionMappings.java
alexandrefimov: Agreed. I filed #1116 for the addition path and am keeping it outside this PR. This PR fixes only subtraction; addition still uses the stock, executable DATETIME_PLUS operator with the DATE inference mismatch.

isthmus/src/test/java/io/substrait/isthmus/FunctionConversionTest.java
alexandrefimov: Replaced the stable-digest test in 322c88e with a hand-built plan passed through SubstraitToSql. It checks that DATETIME_SUBTRACT keeps the UTC argument inside the function call and rejects the malformed closing-parenthesis followed by UTC form.

isthmus/src/main/java/io/substrait/isthmus/expression/FunctionMappings.java
```diff
@@ -214,11 +235,65 @@ public class FunctionMappings {
...
+    RelDataType intervalType = binding.getOperandType(1);
+    RelDataTypeFactory typeFactory = binding.getTypeFactory();
+
+    RelDataType resultType = datetimeType;
+    if (datetimeType.getSqlTypeName() == SqlTypeName.DATE
+        && intervalType.getFamily() == SqlTypeFamily.INTERVAL_DAY_TIME) {
+      // Calcite's interval type does not retain Substrait's interval_day<P> precision, so the
+      // exact timestamp precision cannot be inferred here. Use the type system maximum as a
+      // best-effort inference; ExpressionRexConverter keeps the declared Substrait output type
+      // authoritative.
```
nielspardon: The precision is not lost by Calcite — Isthmus drops it one layer up. `SubstraitTypeSystem.DAY_SECOND_INTERVAL` is `new SqlIntervalQualifier(DAY, -1, SECOND, 6, ZERO)`, pinning the fractional-second precision at 6, and `TypeConverter.visit(Type.IntervalDay)` uses that one constant for every `P`:

```
interval_day<0> -> INTERVAL DAY TO SECOND(6)  scale=6
interval_day<3> -> INTERVAL DAY TO SECOND(6)  scale=6
interval_day<6> -> INTERVAL DAY TO SECOND(6)  scale=6
```

Calcite keeps it when asked — `SqlIntervalQualifier(DAY, -1, SECOND, 3, ZERO)` gives `INTERVAL DAY TO SECOND(3)` with `scale=3` — and `TypeConverter` already reads it back in the other direction: `type.getScale()` at `TypeConverter.java:168` turns that type into `IntervalDay{precision=3}`.

That matters because it changes where the ceiling is. Exact inference is reachable; it needs the forward conversion to carry `P` into the qualifier, after which `intervalType.getScale()` here is exactly right (and equals `getMaxPrecision(TIMESTAMP)` today, since every interval Isthmus builds is scale 6). Keeping the max-precision fallback is fine — I would just not attribute it to Calcite:

```suggestion
      // Isthmus builds every interval_day with SubstraitTypeSystem.DAY_SECOND_INTERVAL, which
      // pins the fractional-second precision at 6, so the operand no longer carries Substrait's
      // interval_day<P>. Use the type system maximum as a best-effort inference;
      // ExpressionRexConverter keeps the declared Substrait output type authoritative.
```

isthmus/src/test/java/io/substrait/isthmus/FunctionConversionTest.java
```diff
@@ -48,29 +63,60 @@ class FunctionConversionTest extends PlanTestBase {
...
+  static Stream<Arguments> subtractDateIDayTypes() {
+    Type maxPrecisionTimestamp = TypeCreator.REQUIRED.precisionTimestamp(6);
+    return Stream.of(
+        Arguments.of(
+            Named.of("legacy DATE output", TypeCreator.REQUIRED.DATE), maxPrecisionTimestamp),
+        Arguments.of(
+            Named.of("spec max-precision output", maxPrecisionTimestamp), maxPrecisionTimestamp),
+        Arguments.of(
+            Named.of("spec non-max-precision output", TypeCreator.REQUIRED.precisionTimestamp(3)),
+            maxPrecisionTimestamp));
```
nielspardon, line 111: This case varies only the declared output — the argument on line 83 stays `interval_day<6>`, so the invocation is a shape the spec does not declare: `subtract(date, interval_day<P>) -> precision_timestamp<P>` ties the output precision to the argument's.

The honest `P=3` invocation, `intervalDay(false, 120, 0, 0, 3)` with `precisionTimestamp(3)`, does not survive the round trip:

```
node type      = TIMESTAMP(3)
original args  = [..., IntervalDayLiteral{..., precision=3}]
roundtrip args = [..., IntervalDayLiteral{..., precision=6}]
equal          = false
```

It fails `assertEquals(expr, reverse)` on the interval, for the reason in the thread above. So the real non-max-precision case can't be written until the forward conversion stops pinning 6 — which I think is worth saying out loud rather than having a case name imply it's covered:

```suggestion
        Arguments.of(
            Named.of("non-max declared output", TypeCreator.REQUIRED.precisionTimestamp(3)),
            maxPrecisionTimestamp));
```

nielspardon (commented):
Two accuracy points below — both about what the code and the test say they do, not about behavior.

One out-of-diff note, since this closes #377: the forward direction is still spec-noncompliant. `SELECT event_date - INTERVAL '5' DAY FROM events` produces `subtract:date_iday` with `declaredOutput=Date{nullable=true}`, where the YAML declares `precision_timestamp<P>`. So on plans Isthmus itself emits, the operator's inference (`TIMESTAMP(6)`) still disagrees with the node type — no longer as an exception, but the disagreement #377 describes hasn't gone away for that path. Worth capturing that half alongside #1116.

isthmus/src/main/java/io/substrait/isthmus/expression/FunctionMappings.java
alexandrefimov: Confirmed and applied in 9a87eb0 — `TypeConverter.visit(Type.IntervalDay)` ignores `expr.precision()` and hands `createSqlIntervalType` the shared `DAY_SECOND_INTERVAL` qualifier, so the operand arrives at scale 6 regardless of `P`. Took your comment as written.

Carrying `P` into the qualifier is the real fix and it belongs with #1114, which is the same pinning on the literal side.

isthmus/src/test/java/io/substrait/isthmus/FunctionConversionTest.java
alexandrefimov: Renamed to `non-max declared output` in 9a87eb0. Verified the honest invocation separately — `intervalDay(false, 120, 0, 0, 3)` with `precisionTimestamp(3)` gives node type `TIMESTAMP(3)` and comes back with `precision=6`, failing `assertEquals(expr, reverse)` on the interval, so the case cannot be written until the forward conversion stops pinning 6.

alexandrefimov:
Both threads applied in 9a87eb0.

On the out-of-diff point: reproduced it — `SELECT event_date - INTERVAL '5' DAY FROM events` gives `subtract:date_iday` with `declaredOutput=Date{nullable=true}`. The cause is one level below the operator: `FunctionFinder` sets `outputType` from `typeConverter.toSubstrait(call.getType())` (`FunctionConverter.java:744`) and never reads the matched variant's declared return type, so `add` is in the same state. Filed as #1117 rather than folding it into #1116, since that one is the Substrait → Calcite half and this is a plan Isthmus emits.

Also corrected #1116, which repeated the "Calcite does not retain interval_day<P>" attribution you disproved here.

nielspardon (approved):
The redirect holds up: distinct Calcite identity from both stock datetime-minus operators, the three-operand form no longer emits a bare timezone operand after the closing paren, and the executability trade-off is documented on the operator and in the `BREAKING CHANGE:` footer.

Two follow-ups filed rather than held against this PR — #1118 for the qualifier pinning your new comment points at (it is the type side of the opposite direction, so #1114 alone would not cover it), and #1119 for the Calcite → Substrait half of the type mismatch, which #377 does not cover and this PR does not change.

PR #1100, opened 2026-08-14
build(isthmus): update HttpClient for GHSA-hjcp-jmpx-g3qm

alexandrefimov:
Calcite 1.42.0 pulls Avatica 1.28.0, which resolves httpclient5 5.5. OSV now reports GHSA-hjcp-jmpx-g3qm for that version; the first fixed version is 5.6.3.

Add a dependency constraint so Isthmus and its CLI consumers select httpclient5 5.6.3.


nielspardon (approved):
LGTM

PR #1042, opened 2026-07-27
feat(core)!: bump substrait to v0.99.0

nielspardon:
## Summary

Bumps the `substrait-packaging` artifacts to [v0.99.0](https://github.com/substrait-io/substrait/releases/tag/v0.99.0) and implements its one feature — the new built-in `unbound` type ([substrait-io/substrait#1081](https://github.com/substrait-io/substrait/pull/1081)).

`unbound` is a placeholder for a type that has not yet been bound to a concrete type. It lets a producer serialize the shape of a plan — its relations and the names and ordering of its fields — before the concrete field types are known; a downstream binder fills them in before execution. A plan that contains an `unbound` type anywhere within it is *partially bound*.

## Changes

- **`substrait-packaging`** bumped `0.98.0` → `0.99.0` in `gradle/libs.versions.toml`. This pulls in the regenerated proto (adds `Type.Unbound`), the repackaged extension YAMLs (the old `extensions/unknown.yaml` placeholder is dropped upstream), and the dialect schema with `UNBOUND`. The spec version reported by `io.substrait.SubstraitVersion.VERSION` follows from the same catalog bump.
- New **`Type.Unbound`** POJO, wired through `TypeVisitor` and both proto converters (`TypeProtoConverter` / `ProtoTypeConverter`), plus the remaining direct `TypeVisitor` implementors (`StringTypeVisitor`, `NamedFieldCountingTypeVisitor`, Isthmus/Spark `IgnoreNullableAndParameters`, examples `TypeStringify`).
- **`TypeKind.UNBOUND`** added to the dialect model for parity with the updated dialect schema.
- **Tests**: a bare `Type.Unbound` proto round-trip, and a `NamedScan` whose schema carries unbound field types (the partially-bound-plan use case).

## A note on nullability

Unlike every other Substrait type, `unbound`'s proto message is empty: it carries no nullability and no parameters (both are determined only when it is bound). So `Type.Unbound.nullable()` throws `UnsupportedOperationException` and `withNullable()` is a no-op, rather than falsely reporting the type as non-nullable. Isthmus and Spark reject unbound types — their type visitors inherit the throwing default — matching the spec guidance that consumers which do not support partially bound plans should reject them.

Closes #1038

BREAKING CHANGE: adds `Type.Unbound` and a new `TypeVisitor.visit(Type.Unbound)` method that direct `TypeVisitor` implementors must handle.

🤖 Generated with AI

PR #1103, opened 2026-08-16
feat(core)!: update to Substrait v0.101.0

### LINKED ISSUE #1102: Update to Substrait v0.101.0
# Release Notes

### ⚠ BREAKING CHANGES

* **extensions:** `extensions/extension_types.yaml` and
`extensions/type_variations.yaml` are no longer part of the extension
catalog. They move to `site/examples/extensions/`, and their URNs change
from `extension:io.substrait:extension_types` and
`extension:io.substrait:type_variations` to
`extension:org.example:extension_types` and
`extension:org.example:type_variations`.

Consumers that resolve `extension:io.substrait:extension_types` — the
`point`/`line` user-defined types, which substrait-java and substrait-go
both register today — will no longer find it in the catalog, and should
either define those types locally or load the example explicitly under
its new URN.

The remaining example files keep their location but also move to the
`extension:org.example:` owner, including
`lambda_function_example.yaml`, which previously duplicated the official
`extension:io.substrait:functions_list` URN.

Nothing in the `extension:org.example:` namespace carries a
compatibility guarantee: those files exist to illustrate the extension
schema and to serve as parser fixtures, and may change without a
deprecation cycle.
* **protos:** `Expression.FieldReference.OuterReference.steps_out`
is deprecated in favor of `rel_reference`, which names the binding
relation via its plan-wide unique `RelCommon.rel_anchor` and therefore
resolves unambiguously in DAG-shaped plans with shared relations
(`ReferenceRel`), where counting subquery boundaries upward does not.
Consumers should add `rel_reference` support before producers switch to
it. Note that both fields are members of the same `outer_reference_type`
oneof, so a plan cannot carry both forms at once: the migration is
consumer-first rather than dual-write. Nothing is removed by this
change: the deprecation is metadata only, consumers are not required to
understand or validate it, and no existing plan changes meaning.
`steps_out` remains fully functional and is scheduled for removal in a
later release.

### Features

* **protos:** deprecate OuterReference.steps_out in favor of rel_reference ([#1132](https://github.com/substrait-io/substrait/issues/1132)) ([25940fb](https://github.com/substrait-io/substrait/commit/25940fb57d57dd8970112bc7d4b611012998eb52))

### Bug Fixes

* add missing RelCommon field to UpdateRel ([#1168](https://github.com/substrait-io/substrait/issues/1168)) ([cdf0102](https://github.com/substrait-io/substrait/commit/cdf0102f0d45e044cfa08a9a81174c5892f72f6a))

### Code Refactoring

* **extensions:** move example YAML files out of the extensions folder ([#1136](https://github.com/substrait-io/substrait/issues/1136)) ([c9f697e](https://github.com/substrait-io/substrait/commit/c9f697e6a1773e9d1248e87a2bd0fabe0c0e7906))

See https://github.com/substrait-io/substrait/releases/tag/v0.101.0

anasik:
- Closes #1102
- Bump substrait-packaging to spec v0.101.0                                                                                                                                                                     
- Remove `DefaultExtensionCatalog.EXTENSION_TYPES`: spec v0.101.0 relocates extension_types.yaml out of the default extension set into `substrait/examples/extensions/` under a new URN (`extension:org.example:extension_types`), demoting it to a documentation example
- Update the two affected literal round-trip tests: one moves to the existing local nested-types fixture, the other to `DefaultExtensionCatalog.UNSIGNED_INTEGERS/u8` to keep `DEFAULT_COLLECTION` UDT coverage
- Wire UpdateRel's new common field (rel_anchor/hint/remap/advanced_extension) into NamedUpdate conversion, matching every other rel type; add a round-trip test.
- Drop the now-dead type_variations.yaml entry from `DefaultExtensionCatalogTest.UNSUPPORTED_FILES` — also relocated out of `substrait/extensions/` in this release, so the entry was unreachable
- Update isthmus's NamedUpdate conversion to match: thread the update's rel_anchor and correlation scope through instead of discarding them, and reject an emit mapping on a `NamedUpdate` rather than silently dropping it, since TableModify's row type is a single ROWCOUNT column and has no columns to remap against
- Considered the spec's companion v0.101.0 change deprecating `OuterReference.steps_out` in favor of `rel_reference`; no code change is needed since `FieldReference` already models both forms.

BREAKING CHANGE: `DefaultExtensionCatalog.EXTENSION_TYPES` is removed and `extension:io.substrait:extension_types` is no longer loaded into `DEFAULT_COLLECTION`, so code referencing the constant no longer compiles and plans referencing that URN — including previously serialized ones — no longer import. To keep reading existing plans, declare the point and line types in your own YAML under the original URN.

### DIFF
--- a/core/src/main/java/io/substrait/relation/ProtoRelConverter.java
+++ b/core/src/main/java/io/substrait/relation/ProtoRelConverter.java
@@ -536,6 +536,8 @@ public class ProtoRelConverter {
     NamedStruct tableSchema = newNamedStruct(rel.getTableSchema());
     ProtoExpressionConverter converter =
         new ProtoExpressionConverter(lookup, extensions, tableSchema.struct(), this);
+    optionalRelAnchor(rel.getCommon())
+        .ifPresent(anchor -> anchorScopes.put(anchor, tableSchema.struct()));
     List<NamedUpdate.TransformExpression> transformations =
         new ArrayList<>(rel.getTransformationsCount());
     for (UpdateRel.TransformExpression transformation : rel.getTransformationsList()) {
--- a/core/src/main/java/io/substrait/relation/ProtoRelConverter.java
+++ b/core/src/main/java/io/substrait/relation/ProtoRelConverter.java
@@ -551,6 +551,11 @@ public class ProtoRelConverter {
             .tableSchema(tableSchema)
             .addAllTransformations(transformations)
             .condition(converter.from(rel.getCondition()));
+    builder
+        .commonExtension(optionalAdvancedExtension(rel.getCommon()))
+        .remap(optionalRelmap(rel.getCommon()))
+        .hint(optionalHint(rel.getCommon()))
+        .relAnchor(optionalRelAnchor(rel.getCommon()));
     if (rel.hasAdvancedExtension()) {
       builder.extension(protoExtensionConverter.fromProto(rel.getAdvancedExtension()));
     }
--- a/core/src/test/java/io/substrait/type/proto/LiteralRoundtripTest.java
+++ b/core/src/test/java/io/substrait/type/proto/LiteralRoundtripTest.java
@@ -79,7 +79,11 @@ class LiteralRoundtripTest extends TestBase {
 
     Expression.UserDefinedLiteral val =
         ExpressionCreator.userDefinedLiteralAny(
-            false, NESTED_TYPES_URN, "point", java.util.Collections.emptyList(), anyValue);
+            false,
+            DefaultExtensionCatalog.UNSIGNED_INTEGERS,
+            "u8",
+            Collections.emptyList(),
+            anyValue);
 
     verifyRoundTrip(val);
   }
--- a/core/src/test/java/io/substrait/type/proto/LiteralRoundtripTest.java
+++ b/core/src/test/java/io/substrait/type/proto/LiteralRoundtripTest.java
@@ -96,7 +96,7 @@ class LiteralRoundtripTest extends TestBase {
             ExpressionCreator.i32(false, 42), ExpressionCreator.i32(false, 100));
     Expression.UserDefinedLiteral val =
         ExpressionCreator.userDefinedLiteralStruct(
-            false, NESTED_TYPES_URN, "point", java.util.Collections.emptyList(), fields);
+            false, NESTED_TYPES_URN, "point", Collections.emptyList(), fields);
 
     verifyRoundTrip(val);
   }
--- a/core/src/test/java/io/substrait/type/proto/LiteralRoundtripTest.java
+++ b/core/src/test/java/io/substrait/type/proto/LiteralRoundtripTest.java
@@ -79,11 +79,7 @@ class LiteralRoundtripTest extends TestBase {
 
     Expression.UserDefinedLiteral val =
         ExpressionCreator.userDefinedLiteralAny(
-            false,
-            DefaultExtensionCatalog.EXTENSION_TYPES,
-            "point",
-            java.util.Collections.emptyList(),
-            anyValue);
+            false, NESTED_TYPES_URN, "point", java.util.Collections.emptyList(), anyValue);
 
     verifyRoundTrip(val);
   }
@@ -96,11 +92,7 @@ class LiteralRoundtripTest extends TestBase {
             ExpressionCreator.i32(false, 42), ExpressionCreator.i32(false, 100));
     Expression.UserDefinedLiteral val =
         ExpressionCreator.userDefinedLiteralStruct(
-            false,
-            DefaultExtensionCatalog.EXTENSION_TYPES,
-            "point",
-            java.util.Collections.emptyList(),
-            fields);
+            false, NESTED_TYPES_URN, "point", java.util.Collections.emptyList(), fields);
 
     verifyRoundTrip(val);
   }
--- a/core/src/main/java/io/substrait/extension/DefaultExtensionCatalog.java
+++ b/core/src/main/java/io/substrait/extension/DefaultExtensionCatalog.java
@@ -60,9 +60,6 @@ public class DefaultExtensionCatalog {
   /** Extension identifier for string functions. */
   public static final String FUNCTIONS_STRING = "extension:io.substrait:functions_string";
 
-  /** Identifier for extension types. */
-  public static final String EXTENSION_TYPES = "extension:io.substrait:extension_types";
-
   /** Extension identifier for unsigned integer types and functions. */
   public static final String UNSIGNED_INTEGERS = "extension:io.substrait:unsigned_integers";
 
@@ -95,7 +92,6 @@ public class DefaultExtensionCatalog {
             .map(c -> String.format("/substrait/extensions/functions_%s.yaml", c))
             .collect(Collectors.toList());
 
-    defaultFiles.add("/substrait/extensions/extension_types.yaml");
     defaultFiles.add("/substrait/extensions/unsigned_integers.yaml");
 
     return SimpleExtension.load(defaultFiles);
--- a/core/src/main/java/io/substrait/relation/RelProtoConverter.java
+++ b/core/src/main/java/io/substrait/relation/RelProtoConverter.java
@@ -615,6 +615,7 @@ public class RelProtoConverter
   public Rel visit(NamedUpdate update, EmptyVisitationContext context) throws RuntimeException {
     UpdateRel.Builder builder =
         UpdateRel.newBuilder()
+            .setCommon(common(update))
             .setNamedTable(NamedTable.newBuilder().addAllNames(update.getNames()))
             .setTableSchema(update.getTableSchema().toProto(typeProtoConverter))
             .addAllTransformations(
--- a/core/src/test/java/io/substrait/extension/DefaultExtensionCatalogTest.java
+++ b/core/src/test/java/io/substrait/extension/DefaultExtensionCatalogTest.java
@@ -36,10 +36,7 @@ class DefaultExtensionCatalogTest {
           "functions_aggregate_decimal_output.yaml",
           // functions_geometry.yaml defines user-defined types (u!geometry) that are not
           // supported by Calcite's type conversion in isthmus.
-          "functions_geometry.yaml",
-          // type_variations.yaml only defines type variations, which are not tracked by
-          // ExtensionCollection (no functions or types), so containsUrn cannot verify it.
-          "type_variations.yaml");
+          "functions_geometry.yaml");
 
   private static final ObjectMapper YAML_MAPPER = new ObjectMapper(new YAMLFactory());
 
--- a/core/src/test/java/io/substrait/type/proto/NestedOuterReferenceRoundtripTest.java
+++ b/core/src/test/java/io/substrait/type/proto/NestedOuterReferenceRoundtripTest.java
@@ -4,9 +4,12 @@ import io.substrait.TestBase;
 import io.substrait.expression.FieldReference;
 import io.substrait.relation.Join;
 import io.substrait.relation.LateralJoin;
+import io.substrait.relation.NamedUpdate;
 import io.substrait.relation.Rel;
 import io.substrait.relation.Rel.Remap;
+import io.substrait.type.NamedStruct;
 import java.util.Arrays;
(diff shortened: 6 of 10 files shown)


nielspardon:
> * Remove DefaultExtensionCatalog.EXTENSION_TYPES: extension_types.yaml no longer ships in the extensions artifact in this release.

We may need to announce this change of behavior as a breaking change just in case someone relied on this extension being loaded by default.

core/src/test/java/io/substrait/type/proto/LiteralRoundtripTest.java
```diff
@@ -104,13 +103,9 @@ void userDefinedLiteralWithAnyRepresentation() {
 
     Expression.UserDefinedLiteral val =
         ExpressionCreator.userDefinedLiteralAny(
-            false,
-            DefaultExtensionCatalog.EXTENSION_TYPES,
-            "point",
-            java.util.Collections.emptyList(),
-            anyValue);
+            false, NESTED_TYPES_URN, "point", java.util.Collections.emptyList(), anyValue);
```
nielspardon: I think switching to `NESTED_TYPES_URN` is fair here since it makes the testcase more consistent with the other tests in this class while accounting for the relocation of the `extension_types.yaml` file in the spec.

anasik: So no changes needed, right?

nielspardon: I did run a full review with AI help and it did identify a couple more places we need to update here.

core/src/test/java/io/substrait/type/proto/LiteralRoundtripTest.java
```diff
@@ -10,7 +10,6 @@
 import io.substrait.expression.ExpressionCreator;
 import io.substrait.expression.proto.ExpressionProtoConverter;
 import io.substrait.expression.proto.ProtoExpressionConverter;
-import io.substrait.extension.DefaultExtensionCatalog;
 import io.substrait.extension.ExtensionCollector;
```
nielspardon: Needs the import back for the suggestion below.

```suggestion
import io.substrait.extension.DefaultExtensionCatalog;
import io.substrait.extension.ExtensionCollector;
```

nielspardon (commented):
Thanks for the bump. A few things — the breaking-change wording I raised above, plus some isthmus work I think this PR has to carry.

**Breaking change.** Concrete text for the behaviour change I flagged, since the title and body become the squash-merge message:

```
feat(core)!: update to Substrait v0.101.0
```

```
BREAKING CHANGE: `DefaultExtensionCatalog.EXTENSION_TYPES` is removed and `extension:io.substrait:extension_types` is no longer loaded into `DEFAULT_COLLECTION`, so code referencing the constant no longer compiles and plans referencing that URN — including previously serialized ones — no longer import. Spec v0.101.0 demoted the extension to a documentation example under a new URN; to keep reading existing plans, declare the `point` and `line` types in your own YAML under the original URN.
```

Worth also correcting the bullet in the description: `extension_types.yaml` does still ship in 0.101.0, relocated to `substrait/examples/extensions/` with its URN changed to `extension:org.example:extension_types`. Dropping it from the default set is still right, just not for the reason given.

**isthmus needs three changes in this PR.** Wiring `UpdateRel.common` makes an emit mapping and a `rel_anchor` reachable from the wire for the first time, and `SubstraitRelNodeConverter.visit(NamedUpdate)` handles neither — it was correct by construction only while the field did not exist. I can't leave these inline since the file isn't in the diff:

- **line 700** — `AnchoredInput.of(Optional.empty(), …)` should pass `update.getRelAnchor()`, as every other scope entry does (189, 209, 241-242, 325, 613, 651). Without it, a plan carrying `rel_reference` into an update's condition throws from `requireScope`, and update anchors never reach `seenAnchors`, so the duplicate-anchor guard is blind to them.
- **lines 716-719** — the correlation ids from `exitScope()` are discarded; `relBuilder.filter(context.exitScope(), condition)` matches lines 191 and 222. Without this the `LogicalFilter` references an undeclared correlation variable, so it needs to land together with the anchor change.
- **lines 730-737** — this is the only supported relation that never calls `applyRemap`, so an emit mapping is silently dropped. I'd reject rather than remap here: `TableModify`'s row type is a single `ROWCOUNT` column, not the updated columns. #1069 rejected this at core serialisation time, and this PR removes the need for that guard, so isthmus is where it now belongs.

**Two smaller v0.101.0 leftovers**, both on files outside the diff:

- `core/src/test/java/io/substrait/extension/DefaultExtensionCatalogTest.java:40-42` — 0.101.0 also relocated `type_variations.yaml` out of `substrait/extensions/`, so that `UNSUPPORTED_FILES` entry is now unreachable; the loop only consults the set for files the scan finds. The guard also only checks shipped → loaded, never that every file it expects still ships, which is the direction that broke here.
- `core/src/main/java/io/substrait/expression/FieldReference.java:40-50` — the spec's other v0.101.0 change deprecates `OuterReference.steps_out` in favour of `rel_reference`, and nothing mirrors it, so callers get no signal. A `<p>` note on the accessor is enough; `@Deprecated` on it would spray warnings through `FieldReference`'s own `@Value.Check`, `isOuterReference()`, `OuterReferenceConverter` and `ExpressionProtoConverter`, all of which still have to read that encoding.

core/src/test/java/io/substrait/type/proto/LiteralRoundtripTest.java
```diff
@@ -104,13 +103,9 @@ void userDefinedLiteralWithAnyRepresentation() {
...
         ExpressionCreator.userDefinedLiteralAny(
-            false,
-            DefaultExtensionCatalog.EXTENSION_TYPES,
-            "point",
-            java.util.Collections.emptyList(),
-            anyValue);
+            false, NESTED_TYPES_URN, "point", java.util.Collections.emptyList(), anyValue);
 
-    verifyRoundTrip(val);
+    verifyNestedTypesRoundTrip(val);
```
nielspardon: Revising what I said on the thread below — moving both of these onto the local fixture leaves no test anywhere that resolves a user-defined type through `DEFAULT_COLLECTION`. Every other UDT test in the repo builds its own collection, and `UNSIGNED_INTEGERS` is referenced by no test at all, so that path goes from two tests to zero.

`unsigned_integers.yaml` still ships and still defines `u8`, so the coverage is preservable here; I checked that both representations round-trip through the inherited `verifyRoundTrip`.

```suggestion
    Expression.UserDefinedLiteral val =
        ExpressionCreator.userDefinedLiteralAny(
            false,
            DefaultExtensionCatalog.UNSIGNED_INTEGERS,
            "u8",
            Collections.emptyList(),
            anyValue);

    verifyRoundTrip(val);
```

core/src/test/java/io/substrait/type/proto/LiteralRoundtripTest.java
```diff
@@ -121,13 +116,9 @@ void userDefinedLiteralWithStructRepresentation() {
...
         ExpressionCreator.userDefinedLiteralStruct(
-            false,
-            DefaultExtensionCatalog.EXTENSION_TYPES,
-            "point",
-            java.util.Collections.emptyList(),
-            fields);
+            false, NESTED_TYPES_URN, "point", java.util.Collections.emptyList(), fields);
 
-    verifyRoundTrip(val);
+    verifyNestedTypesRoundTrip(val);
```
nielspardon: `Collections` is imported and used unqualified nine times elsewhere in this file.

```suggestion
    Expression.UserDefinedLiteral val =
        ExpressionCreator.userDefinedLiteralStruct(
            false, NESTED_TYPES_URN, "point", Collections.emptyList(), fields);

    verifyNestedTypesRoundTrip(val);
```

core/src/main/java/io/substrait/relation/ProtoRelConverter.java
```diff
@@ -551,6 +551,11 @@ protected Rel newNamedUpdate(UpdateRel rel) {
             .tableSchema(tableSchema)
             .addAllTransformations(transformations)
             .condition(converter.from(rel.getCondition()));
+    builder
+        .commonExtension(optionalAdvancedExtension(rel.getCommon()))
+        .remap(optionalRelmap(rel.getCommon()))
+        .hint(optionalHint(rel.getCommon()))
+        .relAnchor(optionalRelAnchor(rel.getCommon()));
```
nielspardon, line 560: The four setters are right, but the anchor is read here while the condition (line 553) and the transformations (541-547) were already converted above. So a nested `rel_reference` bound to this update's own anchor isn't in `anchorScopes` yet, and `outerScopeForRelReference` silently falls back to the *subquery's* root type — the outer reference ends up typed against the wrong struct, with nothing thrown.

`newLateralJoin` (1142-1148) has the identical shape and handles it by registering the scope before converting the input that hosts the references. Same fix here, just after the `converter` is built on line 538:

```java
    optionalRelAnchor(rel.getCommon())
        .ifPresent(anchor -> anchorScopes.put(anchor, tableSchema.struct()));
```

`tableSchema.struct()` rather than `getRecordType()`, since that's what the update's own expressions are typed against. Note `from(Rel)` at 185-187 then overwrites the entry with the remapped record type, so references inside the update see the table schema while references outside see the emit output — probably the right split, but currently implicit.

Separately: should `outerScopeForRelReference` fail on an unknown anchor rather than falling back? isthmus's `Context.requireScope` throws, and that asymmetry is why this one is invisible instead of a test failure.

core/src/test/java/io/substrait/type/proto/LiteralRoundtripTest.java
nielspardon: Actually one change after all — sorry, reversing myself here. Switching to `NESTED_TYPES_URN` is consistent with the siblings, but these two were the last tests resolving a user-defined type through `DEFAULT_COLLECTION`, so that path drops to zero coverage. I left a suggestion on this hunk pointing this one at `UNSIGNED_INTEGERS`/`u8` instead, which keeps the default-catalog path exercised. The struct variant below can stay on the local fixture.

alexandrefimov (commented):
1. This bump should be marked breaking, the way the v0.99.0 one was (`feat(core)!: bump substrait to v0.99.0`, #1042): `DefaultExtensionCatalog.EXTENSION_TYPES` is removed public API, and plans resolving `extension:io.substrait:extension_types` through `DEFAULT_COLLECTION` stop resolving — the spec release notes name substrait-java as an affected consumer. Suggest `feat(core)!: update to Substrait v0.101.0` plus a closing `BREAKING CHANGE:` footer telling consumers what to do instead: define `point`/`line` locally, or vendor the example YAML that now lives in the spec repo under `site/examples/extensions/` with URN `extension:org.example:extension_types`.

2. `DefaultExtensionCatalogTest.UNSUPPORTED_FILES` still lists `type_variations.yaml`. `allExtensionYamlFilesAccountedFor` iterates only the files it finds on the classpath, so once the file stops shipping the entry is never consulted — the test stays green with a dead entry and comment. Since this is the bump that removes the file from the artifact, dropping the entry belongs here.

The `steps_out` deprecation half of v0.101.0 needs no code changes: `FieldReference` already models both `outerReferenceStepsOut` and `outerReferenceRelReference`. Worth a line in the PR description so the record shows it was considered.

alexandrefimov:
Thanks, the implementation changes now address the review points. I checked head `62afed68` against current main. Two things remain before this is ready:

1. CI currently fails in `:core:compileTestJava`: `LiteralRoundtripTest` still calls `verifyNestedTypesRoundTrip`, which was removed by #1111. After updating from main, this should use `verifyRoundTrip(val)`.
2. Please add focused Isthmus coverage for the new `NamedUpdate` behavior: anchor/correlation handling in the condition, and rejection of an emit mapping. The current `relAnchorOnNamedUpdate` test only proves that the anchor field round-trips; it does not exercise these conversion paths.

Everything else from the review looks addressed.

nielspardon (approved):
LGTM, thanks @anasik

alexandrefimov (approved):
LGTM. The v0.101.0 catalog changes and breaking-change documentation match the spec release. UpdateRel.common now round-trips in both directions, and the NamedUpdate anchor, correlation, and remap behavior is covered.

PR #1110, opened 2026-08-17
fix(isthmus)!: preserve literal nullability converting to Calcite

### LINKED ISSUE #1066: isthmus: literal nullability is dropped converting to Calcite, producing non-conformant plans
A Substrait literal's `nullable` flag does not survive a trip through Calcite. Most literals come back non-nullable, which is a silent type change, and in a nested container it turns a conformant input plan into a non-conformant one that substrait-java then rejects.

Verified on `main` at 33e2f6a6.

## Minimal reproduction

Round-tripping a single nullable literal through `SubstraitToCalcite` and back (`Project` over a `namedScan`, remapped so the projection survives):

| input | comes back as |
| --- | --- |
| `ExpressionCreator.i32(true, 5)` | `I32Literal{nullable=false, value=5}` |
| `ExpressionCreator.bool(true, true)` | `BoolLiteral{nullable=false, value=true}` |
| `ExpressionCreator.fp64(true, 1.5)` | `FP64Literal{nullable=false, value=1.5}` |
| `ExpressionCreator.string(true, "x")` | `Cast{type=Str{nullable=true}, input=StrLiteral{nullable=false, value=x}}` |

Only the character path keeps the nullable type, and it does so by picking up a `Cast` wrapper rather than staying a nullable literal.

## Why it matters: nested containers stop round-tripping

`Expression.NestedList` and `Expression.NestedMap` require homogeneously-typed elements, matching the spec (`Nested.List` is "A homogeneously-typed list of one or more expressions"). Calcite drops nullability from literals but preserves it on everything else, so a container that was homogeneous on input comes back mixed — and the resulting plan is rejected:

```java
// both values are i32? on input, so this plan is conformant
Expression.NestedList.builder()
    .addValues(ExpressionCreator.i32(true, 5))    // literal      -> returns as i32
    .addValues(sb.fieldReference(table, 0))       // nullable i32 -> returns as i32?
    .build();
```

- `NestedList` → `AssertionError: All values in NestedList must have the same type`
- `NestedMap` → `IllegalArgumentException: All values in a NestedMap must have the same type`
- `NestedStruct` → no error (structs are heterogeneous by design), but the round trip is not identity: the literal field silently loses its nullability

The nested-list case reproduces on `main` today. The map and struct cases become reachable with #1062 / #1063, which add `NestedMap` and the Calcite conversions for nested structs and maps — the underlying defect is the same and predates them.

Note the inconsistency in how the two containers reject it: `NestedList.check()` uses `assert`, so outside a `-ea` JVM it does not fire at all and `getType()` later reads `values().get(0)` instead. See #1047 / #1058.

## Root cause

`ExpressionRexConverter`'s literal visitors hand the nullable type to `RexBuilder.makeLiteral(value, type)`, but Calcite gives every non-null literal a NOT NULL type, so the nullability is discarded at that point:

```java
// isthmus/src/main/java/io/substrait/isthmus/expression/ExpressionRexConverter.java:204-207
public RexNode visit(Expression.I32Literal expr, Context context) {
  return rexBuilder.makeLiteral(expr.value(), typeConverter.toCalcite(typeFactory, expr.getType()));
}
```

Coming back, `LiteralConverter` faithfully reads the nullability off the `RexLiteral`'s type (`LiteralConverter.java:110-124`), which by then is non-nullable. Nothing on the return path is wrong; the information is already gone.

## Possible directions

1. **Represent nullable literals the way Calcite does** — wrap them in a `CAST` to the nullable type on the way out, which is what already happens incidentally for `CHAR`/`VARCHAR`, and unwrap that shape on the way back so it does not surface as a `Expression.Cast`. Fixes the root cause for every literal, but touches every literal visitor and needs care to avoid a cast appearing in plans that previously had none.
2. **Compare element types ignoring nullability** in `NestedList`/`NestedMap`, deriving the container's element type as the least-restrictive of its entries. Much smaller, and it stops the hard failure, but it relaxes a check that mirrors the spec and leaves the underlying nullability loss in place.

(1) is the real fix; (2) would be a targeted mitigation for the container case if (1) is too broad to take on now.

alexandrefimov:
Substrait literal nullability did not survive conversion to Calcite: `RexBuilder` gives every non-null literal a NOT NULL type, so nullable literals came back non-nullable, and a homogeneous `NestedList` holding a nullable literal next to a nullable field reference came back mixed-typed and was rejected. Only the character/binary paths kept the type, incidentally, by picking up a cast wrapper.

This makes that shape the contract. Scalar literal visitors request their exact Substrait type (`makeLiteral(value, type, allowCast = true)`); the `makeTimeLiteral` / `makeTimestampLiteral` / `makeIntervalLiteral` paths wrap through `preserveNullability`. Interval-day no longer needs it: #1120 moved that visit onto the type converter, which already carries nullability. On the way back, `CallConverters.CAST` folds a cast over a non-null literal that changes nothing but nullability into the literal itself, carrying the cast's nullability, instead of surfacing an `Expression.Cast`. `LogicalValues` tuples keep bare literals — the row type already declares field nullability there, so that one cast is unwrapped, and only that one: a cast doing anything else is doing work, and dropping it would leave the literal under a row type that does not describe it.

Found along the way and left as is, tracked separately: the Calcite→Substrait path normalizes time/timestamp literals to microsecond precision (#1114) and a `precision_timestamp_tz` literal returns as a plain `precision_timestamp` (#1113).

Closes #1066

BREAKING CHANGE: Substrait→Calcite conversion now represents a nullable literal as a cast of the literal to its nullable type where it previously produced a bare literal with a non-nullable type, and converted row types become nullable accordingly. Calcite→Substrait conversion folds nullability-only casts over literals into nullable literals instead of Expression.Cast.

### DIFF
--- a/isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
@@ -87,6 +87,7 @@ import org.apache.calcite.sql.SqlAggFunction;
 import org.apache.calcite.sql.SqlKind;
 import org.apache.calcite.sql.SqlOperator;
 import org.apache.calcite.sql.fun.SqlStdOperatorTable;
+import org.apache.calcite.sql.type.SqlTypeUtil;
 import org.apache.calcite.tools.Frameworks;
 import org.apache.calcite.tools.RelBuilder;
 
@@ -815,10 +816,16 @@ public class SubstraitRelNodeConverter
           RexNode rexNode = literal.accept(expressionRexConverter, context);
           // A nullable literal converts as CAST(literal AS nullable type). LogicalValues tuples
           // hold bare literals, and the field's nullability is already declared in the row type,
-          // so unwrap the cast here.
-          if (rexNode.isA(SqlKind.CAST)
-              && ((RexCall) rexNode).getOperands().get(0) instanceof RexLiteral) {
-            rexNode = ((RexCall) rexNode).getOperands().get(0);
+          // so unwrap that cast here. Only that one: a cast that changes anything else is doing
+          // work, and dropping it would leave the literal under a row type that does not describe
+          // it. Such a tuple is malformed whatever happens to the cast, and the cast below is what
+          // reports it.
+          if (rexNode.isA(SqlKind.CAST)) {
+            RexNode castOperand = ((RexCall) rexNode).getOperands().get(0);
+            if (castOperand instanceof RexLiteral
+                && SqlTypeUtil.equalSansNullability(rexNode.getType(), castOperand.getType())) {
+              rexNode = castOperand;
+            }
           }
           tupleBuilder.add((RexLiteral) rexNode);
         }
--- a/isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
@@ -78,11 +78,13 @@ import org.apache.calcite.rel.type.RelDataType;
 import org.apache.calcite.rel.type.RelDataTypeFactory;
 import org.apache.calcite.rel.type.RelDataTypeField;
 import org.apache.calcite.rex.RexBuilder;
+import org.apache.calcite.rex.RexCall;
 import org.apache.calcite.rex.RexInputRef;
 import org.apache.calcite.rex.RexLiteral;
 import org.apache.calcite.rex.RexNode;
 import org.apache.calcite.rex.RexSlot;
 import org.apache.calcite.sql.SqlAggFunction;
+import org.apache.calcite.sql.SqlKind;
 import org.apache.calcite.sql.SqlOperator;
 import org.apache.calcite.sql.fun.SqlStdOperatorTable;
 import org.apache.calcite.tools.Frameworks;
@@ -810,8 +812,15 @@ public class SubstraitRelNodeConverter
         ImmutableList.Builder<RexLiteral> tupleBuilder = ImmutableList.builder();
         for (Expression expr : rowExpr.fields()) {
           final Expression.Literal literal = (Expression.Literal) expr;
-          final RexLiteral rexNode = (RexLiteral) literal.accept(expressionRexConverter, context);
-          tupleBuilder.add(rexNode);
+          RexNode rexNode = literal.accept(expressionRexConverter, context);
+          // A nullable literal converts as CAST(literal AS nullable type). LogicalValues tuples
+          // hold bare literals, and the field's nullability is already declared in the row type,
+          // so unwrap the cast here.
+          if (rexNode.isA(SqlKind.CAST)
+              && ((RexCall) rexNode).getOperands().get(0) instanceof RexLiteral) {
+            rexNode = ((RexCall) rexNode).getOperands().get(0);
+          }
+          tupleBuilder.add((RexLiteral) rexNode);
         }
         tuplesBuilder.add(tupleBuilder.build());
       }
--- a/isthmus/src/main/java/io/substrait/isthmus/expression/CallConverters.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/expression/CallConverters.java
@@ -66,7 +66,10 @@ public class CallConverters {
           RexNode operand = call.getOperands().get(0);
           if (operand instanceof RexLiteral && !((RexLiteral) operand).isNull()) {
             RexLiteral literal = (RexLiteral) operand;
-            if (type.equalsIgnoringNullability(typeConverter.toSubstrait(literal.getType()))) {
+            Type literalType = typeConverter.toSubstrait(literal.getType());
+            // Nullability only: an exact no-op cast is not this representation and stays an
+            // Expression.Cast, so a plan that carries one keeps carrying it.
+            if (!type.equals(literalType) && type.equalsIgnoringNullability(literalType)) {
               return literalConverter.convert(literal, call.getType());
             }
           }
--- a/isthmus/src/main/java/io/substrait/isthmus/expression/CallConverters.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/expression/CallConverters.java
@@ -15,6 +15,7 @@ import java.util.stream.Collectors;
 import org.apache.calcite.rel.type.RelDataTypeField;
 import org.apache.calcite.rex.RexBuilder;
 import org.apache.calcite.rex.RexCall;
+import org.apache.calcite.rex.RexLiteral;
 import org.apache.calcite.rex.RexNode;
 import org.apache.calcite.rex.RexProgram;
 import org.apache.calcite.rex.RexUtil;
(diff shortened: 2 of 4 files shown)


nielspardon:
With #1120 merged you can update this PR and we can check what needs fixing.

alexandrefimov:
Rebased. The interval-day half went away rather than merged: #1120 moved `visit(IntervalDayLiteral)` onto `makeLiteral(value, typeConverter.toCalcite(...), true)`, so the type already carries the nullability there and the `preserveNullability` wrapper had nothing left to do. What remains is `makeTimeLiteral`, `makeTimestampLiteral` and the year-interval `makeIntervalLiteral`.

The round-trip table keeps its `interval_day` row. It passes because of #1120 now rather than because of this PR, but it seemed worth keeping as coverage rather than deleting.

Also checked this against the neighbours: #1110, #1121 and #1127 all touch the same visits, and a tree with all of them merged onto `main` builds. Not something CI can tell us, since it only ever sees them one at a time.

isthmus/src/test/java/io/substrait/isthmus/LiteralNullabilityRoundtripTest.java
```diff
@@ -0,0 +1,114 @@
...
+  }
+
+  /** A non-nullable literal keeps converting to a bare Calcite literal, with no cast wrapper. */
+  @Test
+  void nonNullableLiteralStaysBare() {
+    Project project =
+        sb.project(input -> List.of(ExpressionCreator.i32(false, 5)), Remap.of(List.of(1)), table);
+    RelNode calcite = substraitToCalcite.convert(project);
+    RexNode rex = ((LogicalProject) calcite).getProjects().get(0);
+    assertEquals(SqlKind.LITERAL, rex.getKind());
```
nielspardon, line 162: The footer makes the cast shape the contract, and both the fold in `CallConverters` and the unwrap in `SubstraitRelNodeConverter` depend on it — but nothing asserts it, and the round-trip tests pass either way. Worth a mirror of this test (needs `assertTrue` and `RexCall` imports):

```java
/** A nullable literal converts to a cast of the literal to its nullable type. */
@Test
void nullableLiteralBecomesNullabilityCast() {
  Project project =
      sb.project(input -> List.of(ExpressionCreator.i32(true, 5)), Remap.of(List.of(1)), table);
  RelNode calcite = substraitToCalcite.convert(project);
  RexNode rex = ((LogicalProject) calcite).getProjects().get(0);
  assertEquals(SqlKind.CAST, rex.getKind());
  assertTrue(rex.getType().isNullable());
  assertEquals(SqlKind.LITERAL, ((RexCall) rex).getOperands().get(0).getKind());
}
```

nielspardon (commented):
Rebase reasoning checks out — `visit(IntervalDayLiteral)` on main does go through the type converter now, so that half really is subsumed rather than merged. Design verified against Calcite 1.42: `makeAbstractCast` does no literal folding, so the nullable shape is stable and matches what the scalar visitors produce. Three small things inline.

Separately, the `precision_timestamp_tz` degradation and the microsecond normalization are recorded only in a test comment — worth issues so they don't get lost.

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
```diff
@@ -810,8 +812,15 @@ public RelNode visit(VirtualTableScan virtualTableScan, Context context) {
...
-          final RexLiteral rexNode = (RexLiteral) literal.accept(expressionRexConverter, context);
-          tupleBuilder.add(rexNode);
+          RexNode rexNode = literal.accept(expressionRexConverter, context);
+          // A nullable literal converts as CAST(literal AS nullable type). LogicalValues tuples
+          // hold bare literals, and the field's nullability is already declared in the row type,
+          // so unwrap the cast here.
+          if (rexNode.isA(SqlKind.CAST)
+              && ((RexCall) rexNode).getOperands().get(0) instanceof RexLiteral) {
+            rexNode = ((RexCall) rexNode).getOperands().get(0);
+          }
```
nielspardon, line 829: This strips any cast over a literal, not only a nullability one. The reachable non-nullability case is a non-nullable varchar/varbinary literal longer than its declared precision: `canRemoveCastFromLiteral` returns false there, so a real truncating cast survives and gets unwrapped into a wider literal under a narrower row type. Malformed input, but it used to fail loudly on the `(RexLiteral)` cast. Requiring the cast to differ in nullability alone would keep it loud. (The nullable variant of the same input still throws, since the operand is then a `RexCall`, not a `RexLiteral`.)

isthmus/src/main/java/io/substrait/isthmus/expression/CallConverters.java
```diff
@@ -38,28 +39,41 @@ public class CallConverters {
    * <p>On SAFE_CAST, sets {@link Expression.FailureBehavior#RETURN_NULL}; otherwise
    * THROW_EXCEPTION.
    *
+   * <p>A cast that changes nothing but nullability over a non-null literal is how {@link
+   * ExpressionRexConverter} represents a nullable Substrait literal (Calcite types every non-null
+   * {@link RexLiteral} NOT NULL), so it converts back to a literal carrying the cast's nullability
+   * rather than to an {@link Expression.Cast}.
```
nielspardon, line 45: The guard is `equalsIgnoringNullability`, so this also folds an exact no-op cast — `CAST(1 AS INTEGER)` written in SQL now converts to a literal rather than an `Expression.Cast`. Equivalent semantically, but this paragraph and the breaking-change footer both say nullability-only.

isthmus/src/test/java/io/substrait/isthmus/LiteralNullabilityRoundtripTest.java
alexandrefimov: Taken as written. You are right that nothing pinned it — on `main` the rex there is `kind=LITERAL` with `nullable=false`, so the test fails without the change and passes with it.

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
alexandrefimov: Reproduced before touching it. A `varchar<3>` virtual table holding `'abcdef'` gives `LogicalValues(tuples=[[{ 'abcdef' }]])` under `RecordType(VARCHAR(3) A)` on this branch, and `ClassCastException: RexCall cannot be cast to RexLiteral` on `main` — exactly the trade you describe, and mine to have introduced.

The unwrap now requires the cast and its operand to be equal sans nullability, so that input is loud again, with a test on it.

isthmus/src/main/java/io/substrait/isthmus/expression/CallConverters.java
alexandrefimov: Tightened — the fold now requires the two types to differ, so an exact cast stays an `Expression.Cast` and the code says what the footer says.

Your example does not reproduce, though, and I could not find one that does. `SELECT CAST(1 AS INTEGER)` emits `I32Literal` on `main` as well: Calcite folds the cast during SQL-to-Rel, so no CAST node reaches the converter. It cannot be reached from the other side either — a Substrait `cast(i32, i32-literal)` comes out as a bare `RexLiteral`, since Calcite drops an exact cast over a literal on the way out. So the discrepancy was in the wording rather than in behaviour. The test builds the node directly for that reason, rather than pretending a plan can produce one.

alexandrefimov:
Rebased. The core commit is gone: #1136 landed the same one-line fix, and its `LiteralTypeDerivationTest` covers what the test in that commit covered plus `interval_day` and `interval_compound`, so keeping mine would only have duplicated it. The description and the breaking footer no longer mention it.

On the two things recorded only in a test comment — both are filed already, and both are closed by #1121: #1114 for the microsecond normalization, whose interval third went with #1120, and #1113 for the `precision_timestamp_tz` degradation. The description points at them now instead of leaving them in a comment.

nielspardon (approved):
LGTM

Issue #1114, opened 2026-08-17
isthmus: Calcite→Substrait normalizes time/timestamp/interval-day literal precision to microseconds


alexandrefimov:
Verified on main at 683ca77. `LiteralConverter` converts temporal literal values at a fixed precision 6 instead of the literal's own: `TIME` → micros, `TIMESTAMP` / `TIMESTAMP_WITH_LOCAL_TIME_ZONE` → epoch micros, `INTERVAL_DAY` → subseconds scaled to micros. A Substrait → Calcite → Substrait round trip is value-preserving but not identity:

```
PrecisionTimeLiteral{value=1000, precision=3}                    → PrecisionTimeLiteral{value=1000000, precision=6}
IntervalDayLiteral{days=1, seconds=2, subseconds=3, precision=3} → IntervalDayLiteral{days=1, seconds=2, subseconds=3000, precision=6}
```

Scaling millisecond-precision literals up is not forced — the literal's declared precision is available on the Calcite side, and the type converter already reads it. Structurally-changed literals make POJO-equality plan comparison see a change where none happened, and hand a producer back a different precision than it chose.

There is a real constraint underneath, but it is not the same one for all three types. `TIME` and `TIMESTAMP` values are capped at microseconds because `SubstraitTypeSystem` sets the datetime maximum to 6 (#995). An `INTERVAL_DAY` value is coarser still: Calcite carries day-time interval literals in **milliseconds** whatever the type's scale says, so anything finer is lost regardless of the declared precision. Neither constraint forces the *declared* precision to be rewritten, which is what this issue is about.

At minimum the normalization deserves documenting; preserving source precision would make the round trip identity for precisions Calcite can represent. (Round-trip tests added in #1110 pin the current normalized behavior and can be tightened if this is fixed.)

alexandrefimov:
The interval third is in #1120. Root cause turned out to be one level above `LiteralConverter`: `TypeConverter` built every interval from the shared `DAY_SECOND_INTERVAL` qualifier at precision 6 while `ExpressionRexConverter` had a second private one pinned at 3, so `P` was gone before the literal converter saw it. Building the qualifier from `P` makes the round trip identity for the interval case.

`TIME` and `TIMESTAMP` are a different fix — their values are converted to micros as well as their precision pinned, so honouring the declared precision means rescaling the value. Leaving this open for those.

alexandrefimov:
[...]

Issue #1113, opened 2026-08-17
isthmus: precision_timestamp_tz literals cap at precision 3 converting to Calcite and come back as plain precision_timestamp


alexandrefimov:
Verified on main at 683ca77. `precision_timestamp_tz` literals are broken in both conversion directions, in ways the corresponding type conversion is not.

## Substrait → Calcite: wrong `SqlTypeName` in the max-precision lookup

`ExpressionRexConverter.visit(PrecisionTimestampTZLiteral)` validates precision via `getMaxPrecision(SqlTypeName.TIMESTAMP_TZ)`. `SubstraitTypeSystem` overrides `TIMESTAMP_WITH_LOCAL_TIME_ZONE` (returning 6) but not `TIMESTAMP_TZ`, so the lookup falls through to Calcite's default of 3, and a microsecond-precision TZ literal is rejected:

```
unsupported precision_timestamp precision 6, max precision in Calcite type system is set to 3
```

Everything else TZ-related uses `TIMESTAMP_WITH_LOCAL_TIME_ZONE`: `TypeConverter.ToRelDataType` checks that name, so a `precision_timestamp_tz<6>` type converts fine — only the literal is rejected. This is the codebase's single `TIMESTAMP_TZ` lookup. (The message also says `precision_timestamp` for what is a `precision_timestamp_tz` literal.)

## Calcite → Substrait: the TZ literal comes back TZ-less

`LiteralConverter.convert` handles `TIMESTAMP` and `TIMESTAMP_WITH_LOCAL_TIME_ZONE` in one branch returning `ExpressionCreator.precisionTimestamp(...)`, while `TypeConverter.toSubstrait` maps the same `SqlTypeName` to `precisionTimestampTZ` — the literal converter disagrees with the type converter about what the value is.

## Repro

Round-tripping through `SubstraitToCalcite` and back (`Project` over a `namedScan`, remapped so the projection survives):

```
PrecisionTimestampTZLiteral{nullable=false, value=1000000, precision=3}
  → PrecisionTimestampLiteral{nullable=false, value=1000000000, precision=6}
```

At precision 6 the forward conversion throws before the reverse defect is reachable.

nielspardon:
[...]

Issue #995, opened 2026-07-09
isthmus: Substrait→Calcite conversion rejects precision_time / precision_timestamp / precision_timestamp_tz precision > 6


nielspardon:
## Description

Isthmus's Substrait → Calcite conversion rejects any `precision_time`, `precision_timestamp`, or `precision_timestamp_tz` whose precision is greater than **6**, even though the Substrait spec allows sub-second precision from **0 to 12** (0 = seconds, 3 = milliseconds, 6 = microseconds, 9 = nanoseconds, 12 = picoseconds — see [`type.proto`](https://github.com/substrait-io/substrait/blob/v0.88.0/proto/substrait/type.proto#L172-L191)).

Converting a plan that uses e.g. nanosecond precision fails with:

```
java.lang.IllegalArgumentException: unsupported precision_timestamp precision 9, max precision in Calcite type system is set to 6
```

The cap comes from `SubstraitTypeSystem.getMaxPrecision`, which returns `6` for `TIME`, `TIMESTAMP`, and `TIMESTAMP_WITH_LOCAL_TIME_ZONE`. The precision is then validated against that maximum in two places:

- `TypeConverter.ToRelDataType#visit(Type.PrecisionTime | PrecisionTimestamp | PrecisionTimestampTZ)`
- `ExpressionRexConverter` (the corresponding literal-conversion paths)

The Substrait POJO model itself does not cap the precision, so such a plan is perfectly representable in `substrait-java` and can be produced by other engines; only the conversion to Calcite fails.

## Relationship to #93

This is the **reverse direction** of #93. That issue reported that the Calcite → Substrait direction rejected valid `TIME(6)`/`TIMESTAMP(6)` inputs. It was resolved by the migration to `PrecisionTime`/`PrecisionTimestamp` (#721) and the removal of the deprecated `Time`/`Timestamp`/`TimestampTZ` types (#950): the Calcite → Substrait direction now passes the precision straight through without a cap. The Substrait → Calcite direction still caps at 6, which this issue tracks.

## Notes / open questions

- Current Calcite dependency is `1.42.0`. It may be possible to raise `SubstraitTypeSystem.getMaxPrecision` for `TIMESTAMP`/`TIME` beyond 6 (worth checking what this Calcite version actually supports).
[...]

PR #1120, opened 2026-08-18
fix(isthmus)!: preserve interval_day precision in both conversion directions

### LINKED ISSUE #1118: isthmus: Substrait→Calcite type conversion pins interval_day<P> to microsecond precision
Verified on #1099's head (9a87eb0). `TypeConverter.visit(Type.IntervalDay)` ignores `expr.precision()` and hands `createSqlIntervalType` the shared `SubstraitTypeSystem.DAY_SECOND_INTERVAL` qualifier, which fixes the fractional-second precision at 6:

```
interval_day<0> -> INTERVAL DAY TO SECOND(6)  scale=6
interval_day<3> -> INTERVAL DAY TO SECOND(6)  scale=6
interval_day<6> -> INTERVAL DAY TO SECOND(6)  scale=6
```

Calcite retains the precision when it is given one — `new SqlIntervalQualifier(DAY, -1, SECOND, 3, ZERO)` produces `INTERVAL DAY TO SECOND(3)` with `scale=3` — and the reverse direction already reads it back, `type.getScale()` at `TypeConverter.java:168` turning that type into `IntervalDay{nullable=false, precision=3}`. The loss is one-directional and comes from the shared qualifier constant, not from Calcite.

Two consequences. First, an `interval_day<P>` type does not survive Substrait → Calcite → Substrait for `P != 6`: `interval_day<3>` becomes `INTERVAL DAY TO SECOND(6)` and comes back as `interval_day<6>`. This is type conversion only, no literal involved.

Second, it is what blocks exact return-type inference for `subtract(date, interval_day<P>) -> precision_timestamp<P>`. `FunctionMappings.inferDatetimeSubtractType`, added in #1099, has to fall back to `getMaxPrecision(TIMESTAMP)` because the operand no longer carries `P`; once the qualifier carries it, `intervalType.getScale()` is exactly right and the fallback can go.

Related to #1114, which is the same microsecond pinning on the literal-value side (`LiteralConverter`, Calcite → Substrait). This is the type side of the opposite direction, so fixing #1114 alone would not cover it.

alexandrefimov:
Substrait's `interval_day<P>` carries a fractional-second precision, and every path across the Calcite boundary discarded it. `TypeConverter` built every interval from the shared `SubstraitTypeSystem.DAY_SECOND_INTERVAL` qualifier at precision 6, `ExpressionRexConverter` had a second private qualifier pinned at 3 — so the two directions of the same conversion disagreed about the type of the same value — and coming back, `LiteralConverter` reported a fixed 6 whatever the Calcite literal's type said:

```
interval_day<0> -> INTERVAL DAY TO SECOND(6)  scale=6
interval_day<3> -> INTERVAL DAY TO SECOND(6)  scale=6
```

Build the qualifier from `P` instead, in both places, and read the declared precision back off the type when converting a Calcite interval literal. Calcite keeps a qualifier's precision as the type's scale, and `TypeConverter.toSubstrait` already read that scale, so the round trip becomes identity:

```
IntervalDayLiteral{days=1, seconds=2, subseconds=3, precision=3}
  -> 86402003:INTERVAL DAY TO SECOND(3)
  -> IntervalDayLiteral{days=1, seconds=2, subseconds=3, precision=3}
```

An interval written without an explicit precision is unaffected — Calcite gives it scale 6, which is exactly what was hardcoded. One written as `DAY TO SECOND(P)` parses to that `P` and now keeps it.

The value stays in milliseconds, which is what Calcite's own interval literals carry. A sub-millisecond component is therefore still truncated; what this changes is that the declared precision is no longer overwritten with a fixed 6.

With `P` available on the operand, `DATETIME_SUBTRACT` now infers `precision_timestamp<P>` exactly, as `functions_datetime.yaml` declares, instead of falling back to the type system maximum — the gap #1099 had to leave as a best-effort comment. The parameterized test it left behind, which could only vary the declared output, now varies the argument precision too.

Closes #1118, which reports the type-conversion half of this and was filed just before this PR went up.

It also removes the hardcoded 6 on the literal side — `LiteralConverter` reported `intervalDay(..., 6)` whatever scale the Calcite literal carried — so the interval third of #1114 goes with it. The `TIME` / `TIMESTAMP` halves of #1114 are a separate value rescale, in #1121.

BREAKING CHANGE: an `interval_day` precision outside 0 to 9 is now rejected instead of silently narrowed to 6. The bound is `getMaxScale(INTERVAL_DAY_SECOND)`, the same one Calcite validates an interval qualifier against, so nothing that parses from SQL is refused. Converted Calcite types and generated SQL now carry the interval's declared precision, so `INTERVAL ... DAY TO SECOND(P)` replaces the previous fixed `DAY TO SECOND(3)`. A Calcite interval literal whose declared scale is below 3 now converts to a Substrait `interval_day` at that scale, dropping the millisecond remainder towards zero, where it previously reported `interval_day<6>` and kept it.

### DIFF
--- a/isthmus/src/main/java/io/substrait/isthmus/SubstraitTypeSystem.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/SubstraitTypeSystem.java
@@ -28,9 +28,30 @@ public class SubstraitTypeSystem extends RelDataTypeSystemImpl {
   public static final SqlIntervalQualifier YEAR_MONTH_INTERVAL =
       new SqlIntervalQualifier(TimeUnit.YEAR, TimeUnit.MONTH, SqlParserPos.ZERO);
 
-  /** Interval qualifier from day to fractional second at microsecond precision. */
-  public static final SqlIntervalQualifier DAY_SECOND_INTERVAL =
-      new SqlIntervalQualifier(TimeUnit.DAY, -1, TimeUnit.SECOND, 6, SqlParserPos.ZERO);
+  /**
+   * Interval qualifier from day to fractional second at microsecond precision.
+   *
+   * <p>Conversion of a Substrait {@code interval_day<P>} builds its qualifier from {@code P} via
+   * {@link #daySecondInterval(int)} instead; sharing one constant is what used to drop the
+   * precision.
+   */
+  public static final SqlIntervalQualifier DAY_SECOND_INTERVAL = daySecondInterval(6);
+
+  /**
+   * Returns an interval qualifier from day to fractional second at the given precision.
+   *
+   * <p>Substrait's {@code interval_day<P>} carries its own fractional-second precision, and Calcite
+   * keeps it: the qualifier's precision becomes the scale of the resulting {@code INTERVAL DAY TO
+   * SECOND} type. Building the qualifier per value rather than sharing one constant is what lets
+   * {@code P} survive the conversion.
+   *
+   * @param precision the fractional-second precision
+   * @return the interval qualifier
+   */
+  public static SqlIntervalQualifier daySecondInterval(final int precision) {
+    return new SqlIntervalQualifier(
+        TimeUnit.DAY, -1, TimeUnit.SECOND, precision, SqlParserPos.ZERO);
+  }
 
   /**
    * Public no-argument constructor.
--- a/isthmus/src/main/java/io/substrait/isthmus/SubstraitTypeSystem.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/SubstraitTypeSystem.java
@@ -29,21 +29,10 @@ public class SubstraitTypeSystem extends RelDataTypeSystemImpl {
       new SqlIntervalQualifier(TimeUnit.YEAR, TimeUnit.MONTH, SqlParserPos.ZERO);
 
   /**
-   * Interval qualifier from day to fractional second at microsecond precision.
-   *
-   * <p>Conversion of a Substrait {@code interval_day<P>} builds its qualifier from {@code P} via
-   * {@link #daySecondInterval(int)} instead; sharing one constant is what used to drop the
-   * precision.
-   */
-  public static final SqlIntervalQualifier DAY_SECOND_INTERVAL = daySecondInterval(6);
-
-  /**
    * Returns an interval qualifier from day to fractional second at the given precision.
    *
-   * <p>Substrait's {@code interval_day<P>} carries its own fractional-second precision, and Calcite
-   * keeps it: the qualifier's precision becomes the scale of the resulting {@code INTERVAL DAY TO
-   * SECOND} type. Building the qualifier per value rather than sharing one constant is what lets
-   * {@code P} survive the conversion.
+   * <p>The qualifier's precision becomes the scale of the resulting {@code INTERVAL DAY TO SECOND}
+   * type, which is how a Substrait {@code interval_day<P>} carries {@code P} into Calcite.
    *
    * @param precision the fractional-second precision
    * @return the interval qualifier
--- a/isthmus/src/main/java/io/substrait/isthmus/TypeConverter.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/TypeConverter.java
@@ -1,6 +1,5 @@
 package io.substrait.isthmus;
 
-import static io.substrait.isthmus.SubstraitTypeSystem.DAY_SECOND_INTERVAL;
 import static io.substrait.isthmus.SubstraitTypeSystem.YEAR_MONTH_INTERVAL;
 
 import io.substrait.function.NullableType;
@@ -342,8 +341,17 @@ public class TypeConverter {
 
     @Override
     public RelDataType visit(Type.IntervalDay expr) {
+      int maxPrecision = typeFactory.getTypeSystem().getMaxPrecision(SqlTypeName.INTERVAL_DAY);
+      if (expr.precision() > maxPrecision) {
+        throw new IllegalArgumentException(
+            String.format(
+                "unsupported interval_day precision %s, max precision in Calcite type system is set to %s",
+                expr.precision(), maxPrecision));
+      }
       return typeFactory.createTypeWithNullability(
-          typeFactory.createSqlIntervalType(DAY_SECOND_INTERVAL), n(expr));
+          typeFactory.createSqlIntervalType(
+              SubstraitTypeSystem.daySecondInterval(expr.precision())),
+          n(expr));
     }
 
     @Override
--- a/isthmus/src/main/java/io/substrait/isthmus/TypeConverter.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/TypeConverter.java
@@ -341,11 +341,15 @@ public class TypeConverter {
 
     @Override
     public RelDataType visit(Type.IntervalDay expr) {
-      int maxPrecision = typeFactory.getTypeSystem().getMaxPrecision(SqlTypeName.INTERVAL_DAY);
-      if (expr.precision() > maxPrecision) {
+      // getMaxScale, not getMaxPrecision: for an interval the latter bounds the leading field,
+      // while the fractional-second bound is the scale of the INTERVAL_DAY_SECOND type produced
+      // here. That is also the knob SqlValidatorImpl checks an interval qualifier against, so this
+      // accepts exactly what SQL written as DAY TO SECOND(P) parses to.
+      int maxPrecision = typeFactory.getTypeSystem().getMaxScale(SqlTypeName.INTERVAL_DAY_SECOND);
+      if (expr.precision() < 0 || expr.precision() > maxPrecision) {
         throw new IllegalArgumentException(
             String.format(
-                "unsupported interval_day precision %s, max precision in Calcite type system is set to %s",
+                "unsupported interval_day precision %s, Calcite type system allows 0 to %s",
                 expr.precision(), maxPrecision));
       }
       return typeFactory.createTypeWithNullability(
--- a/isthmus/src/main/java/io/substrait/isthmus/expression/ExpressionRexConverter.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/expression/ExpressionRexConverter.java
@@ -19,7 +19,6 @@ import io.substrait.expression.WindowBound;
 import io.substrait.extension.SimpleExtension;
 import io.substrait.isthmus.SubstraitRelNodeConverter;
 import io.substrait.isthmus.SubstraitRelNodeConverter.Context;
-import io.substrait.isthmus.SubstraitTypeSystem;
 import io.substrait.isthmus.TypeConverter;
 import io.substrait.type.StringTypeVisitor;
 import io.substrait.type.Type;
@@ -408,11 +407,14 @@ public class ExpressionRexConverter
             ? (expr.subseconds() / (int) Math.pow(10, expr.precision() - 3))
             : (expr.subseconds() * (int) Math.pow(10, 3 - expr.precision()));
     // The value is in milliseconds, which is what Calcite's own interval literals carry, while the
-    // qualifier keeps the declared interval_day<P> so the precision survives the round trip. A
+    // type keeps the declared interval_day<P> so the precision survives the round trip. A
     // sub-millisecond component therefore does not: it is truncated by the conversion above.
-    return rexBuilder.makeIntervalLiteral(
+    // Taking the type from the converter rather than building a qualifier here is what applies the
+    // precision bound and the literal's nullability, as every other literal visit does.
+    return rexBuilder.makeLiteral(
         new BigDecimal((expr.days() * MILLIS_IN_DAY + expr.seconds() * 1_000L + milliseconds)),
-        SubstraitTypeSystem.daySecondInterval(expr.precision()));
+        typeConverter.toCalcite(typeFactory, expr.getType()),
+        true);
   }
 
   @Override
--- a/isthmus/src/main/java/io/substrait/isthmus/expression/ExpressionRexConverter.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/expression/ExpressionRexConverter.java
@@ -19,6 +19,7 @@ import io.substrait.expression.WindowBound;
 import io.substrait.extension.SimpleExtension;
 import io.substrait.isthmus.SubstraitRelNodeConverter;
 import io.substrait.isthmus.SubstraitRelNodeConverter.Context;
+import io.substrait.isthmus.SubstraitTypeSystem;
 import io.substrait.isthmus.TypeConverter;
 import io.substrait.type.StringTypeVisitor;
 import io.substrait.type.Type;
@@ -72,14 +73,6 @@ public class ExpressionRexConverter
           -1,
           org.apache.calcite.avatica.util.TimeUnit.MONTH,
           -1,
-          SqlParserPos.QUOTED_ZERO);
-
-  private static final SqlIntervalQualifier DAY_SECOND_INTERVAL =
-      new SqlIntervalQualifier(
-          org.apache.calcite.avatica.util.TimeUnit.DAY,
-          -1,
-          org.apache.calcite.avatica.util.TimeUnit.SECOND,
-          3, // Calcite only supports millisecond at the moment
           SqlParserPos.QUOTED_ZERO);
 
   private static final long MILLIS_IN_DAY = TimeUnit.DAYS.toMillis(1);
@@ -414,10 +407,12 @@ public class ExpressionRexConverter
         expr.precision() > 3
             ? (expr.subseconds() / (int) Math.pow(10, expr.precision() - 3))
             : (expr.subseconds() * (int) Math.pow(10, 3 - expr.precision()));
+    // The value is in milliseconds, which is what Calcite's own interval literals carry, while the
+    // qualifier keeps the declared interval_day<P> so the precision survives the round trip. A
+    // sub-millisecond component therefore does not: it is truncated by the conversion above.
     return rexBuilder.makeIntervalLiteral(
-        // Current Calcite behavior is to store milliseconds since Epoch
         new BigDecimal((expr.days() * MILLIS_IN_DAY + expr.seconds() * 1_000L + milliseconds)),
-        DAY_SECOND_INTERVAL);
+        SubstraitTypeSystem.daySecondInterval(expr.precision()));
   }
 
   @Override
--- a/isthmus/src/main/java/io/substrait/isthmus/expression/FunctionMappings.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/expression/FunctionMappings.java
@@ -249,11 +249,10 @@ public class FunctionMappings {
     if (datetimeType.getSqlTypeName() == SqlTypeName.DATE
         && intervalType.getFamily() == SqlTypeFamily.INTERVAL_DAY_TIME) {
       // The spec declares subtract(date, interval_day<P>) -> precision_timestamp<P> (spec
-      // v0.101.0), and the interval operand carries P as its fractional-second scale. The result is
-      // capped at the timestamp maximum, which is lower than the interval one, so an
-      // interval_day<9>
-      // still yields precision_timestamp<6> until the datetime ceiling itself moves (see #995's
-      // subject: TypeConverter rejects precision_timestamp above 6).
+      // v0.101.0), and the interval operand carries P as its fractional-second scale. The result
+      // is capped at the timestamp maximum, which is lower than the interval one, so an
+      // interval_day<9> still yields precision_timestamp<6> until the Substrait->Calcite
+      // conversion accepts a precision_timestamp above 6.
       int precision =
           Math.max(
               0,
--- a/isthmus/src/main/java/io/substrait/isthmus/expression/FunctionMappings.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/expression/FunctionMappings.java
@@ -248,13 +248,18 @@ public class FunctionMappings {
     RelDataType resultType = datetimeType;
     if (datetimeType.getSqlTypeName() == SqlTypeName.DATE
         && intervalType.getFamily() == SqlTypeFamily.INTERVAL_DAY_TIME) {
-      // The spec declares subtract(date, interval_day<P>) -> precision_timestamp<P>, and the
-      // interval operand carries P as its fractional-second scale, so the result precision follows
-      // from the argument. Clamped because the two type-system maxima are configured separately.
+      // The spec declares subtract(date, interval_day<P>) -> precision_timestamp<P> (spec
+      // v0.101.0), and the interval operand carries P as its fractional-second scale. The result is
+      // capped at the timestamp maximum, which is lower than the interval one, so an
+      // interval_day<9>
+      // still yields precision_timestamp<6> until the datetime ceiling itself moves (see #995's
+      // subject: TypeConverter rejects precision_timestamp above 6).
       int precision =
-          Math.min(
-              intervalType.getScale(),
-              typeFactory.getTypeSystem().getMaxPrecision(SqlTypeName.TIMESTAMP));
+          Math.max(
+              0,
+              Math.min(
+                  intervalType.getScale(),
+                  typeFactory.getTypeSystem().getMaxPrecision(SqlTypeName.TIMESTAMP)));
       resultType = typeFactory.createSqlType(SqlTypeName.TIMESTAMP, precision);
     }
 
--- a/isthmus/src/main/java/io/substrait/isthmus/expression/FunctionMappings.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/expression/FunctionMappings.java
@@ -248,11 +248,13 @@ public class FunctionMappings {
     RelDataType resultType = datetimeType;
     if (datetimeType.getSqlTypeName() == SqlTypeName.DATE
         && intervalType.getFamily() == SqlTypeFamily.INTERVAL_DAY_TIME) {
-      // Isthmus builds every interval_day with SubstraitTypeSystem.DAY_SECOND_INTERVAL, which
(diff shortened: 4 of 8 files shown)


alexandrefimov:
Heads-up on ordering: this conflicts textually with #1110, which wraps the same `makeIntervalLiteral` call in its new `preserveNullability(...)` while this changes the qualifier argument. The two compose — the resolution is `preserveNullability(rexBuilder.makeIntervalLiteral(value, SubstraitTypeSystem.daySecondInterval(expr.precision())), expr.getType())` — so whichever lands second just needs that one hunk resolved. No behavioural interaction between them.

alexandrefimov:
Checked the interaction with the other two rather than assuming: merged #1110, this, and #1121 together on top of `main`, resolved the one hunk as described above, and the combined tree builds with `:core:test` 710 and `:isthmus:test` 1078 green. #1121 merges cleanly on top with no resolution needed.

The resolution, for the record:

```java
    return preserveNullability(
        rexBuilder.makeIntervalLiteral(
            new BigDecimal((expr.days() * MILLIS_IN_DAY + expr.seconds() * 1_000L + milliseconds)),
            SubstraitTypeSystem.daySecondInterval(expr.precision())),
        expr.getType());
```

isthmus/src/main/java/io/substrait/isthmus/expression/LiteralConverter.java
```diff
@@ -234,15 +234,24 @@ public Expression.Literal convert(RexLiteral literal, RelDataType resultType) {
...
+          int precision = resultType.getScale();
+          long millisPart = interval.toMillisPart();
+          int subseconds =
+              precision > 3
+                  ? (int) (millisPart * (long) Math.pow(10, precision - 3))
+                  : (int) (millisPart / (long) Math.pow(10, 3 - precision));
 
-          return ExpressionCreator.intervalDay(nullable, (int) days, (int) seconds, micros, 6);
+          return ExpressionCreator.intervalDay(
+              nullable, (int) days, (int) seconds, subseconds, precision);
```
nielspardon, line 260: Below scale 3 the division drops the millisecond remainder the fixed 6 always kept: an `INTERVAL DAY TO SECOND(0)` literal holding 1500 ms returns 1000 ms, and since the arithmetic floors, −1500 ms returns −2000 ms. `interval_day<0>` is what `TypeConverter` now emits for the spec-legal type, and no test covers a scale below 3 with a fractional part.

Narrowing silently, rejecting, or reporting at `max(P, 3)` are all defensible — your call, but with a test on it. Two things worth doing either way: `ExpressionCreator.intervalDay` takes a `long`, so the `int` narrowing is needless (unreachable once comment 2 bounds the guard, but it costs nothing to drop), and the `double` `Math.pow` can be exact.

```suggestion
          int precision = resultType.getScale();
          long millisPart = interval.toMillisPart();
          long subseconds =
              precision > 3
                  ? millisPart * LongMath.pow(10, precision - 3)
                  : millisPart / LongMath.pow(10, 3 - precision);

          return ExpressionCreator.intervalDay(
              nullable, (int) days, (int) seconds, subseconds, precision);
```

Needs `import com.google.common.math.LongMath;`. The inverse at `ExpressionRexConverter:406-409` and a third copy in the new test have to stay exact inverses; `core`'s `DecimalUtil.POWER_OF_10` would give all three one home — best done in #1121, which needs the same rescale for `TIME`/`TIMESTAMP`.

isthmus/src/main/java/io/substrait/isthmus/TypeConverter.java
```diff
@@ -342,8 +341,17 @@ public RelDataType visit(Type.IntervalYear expr) {
 
     @Override
     public RelDataType visit(Type.IntervalDay expr) {
+      int maxPrecision = typeFactory.getTypeSystem().getMaxPrecision(SqlTypeName.INTERVAL_DAY);
+      if (expr.precision() > maxPrecision) {
+        throw new IllegalArgumentException(
+            String.format(
+                "unsupported interval_day precision %s, max precision in Calcite type system is set to %s",
+                expr.precision(), maxPrecision));
+      }
```
nielspardon: For intervals `getMaxPrecision` is the leading-field bound; the fractional-second bound is `getMaxScale`, and the type produced here carries `INTERVAL_DAY_SECOND`, which `SubstraitTypeSystem` doesn't override. `SqlValidatorImpl` validates an interval qualifier's fractional-second precision against `getMaxScale(qualifier.typeName())`, so Calcite's inbound ceiling is already 9: SQL written as `... DAY TO SECOND(7)` parses to `interval_day<7>`, and this guard then refuses to convert that plan back. It's also one-sided, and −1 is `PRECISION_NOT_SPECIFIED`, so `interval_day<-1>` silently round-trips as `interval_day<6>`.

```suggestion
      int maxPrecision = typeFactory.getTypeSystem().getMaxScale(SqlTypeName.INTERVAL_DAY_SECOND);
      if (expr.precision() < 0 || expr.precision() > maxPrecision) {
        throw new IllegalArgumentException(
            String.format(
                "unsupported interval_day precision %s, Calcite type system allows 0 to %s",
                expr.precision(), maxPrecision));
      }
```

That raises the ceiling from 6 to 9, so the BREAKING CHANGE footer wants an edit — but it makes the break smaller and keeps the two directions agreeing, where holding at 6 would mean overriding `getMaxScale` for the interval names and newly rejecting SQL that parses today. It does mean the PR body's "SQL-parsed intervals are unaffected" only holds for intervals written without an explicit precision.

This is the fourth copy of these seven lines in this file and the seventh across isthmus, but #1113 is one of them already drifted — I'd extract a shared `checkPrecision(...)` once that's fixed rather than here. #1124 needs the interval knob to be the scale one either way.

isthmus/src/main/java/io/substrait/isthmus/expression/ExpressionRexConverter.java
```diff
@@ -414,10 +407,12 @@ public RexNode visit(Expression.IntervalDayLiteral expr, Context context)
...
             ? (expr.subseconds() / (int) Math.pow(10, expr.precision() - 3))
             : (expr.subseconds() * (int) Math.pow(10, 3 - expr.precision()));
+    // The value is in milliseconds, which is what Calcite's own interval literals carry, while the
+    // qualifier keeps the declared interval_day<P> so the precision survives the round trip. A
+    // sub-millisecond component therefore does not: it is truncated by the conversion above.
     return rexBuilder.makeIntervalLiteral(
-        // Current Calcite behavior is to store milliseconds since Epoch
         new BigDecimal((expr.days() * MILLIS_IN_DAY + expr.seconds() * 1_000L + milliseconds)),
-        DAY_SECOND_INTERVAL);
+        SubstraitTypeSystem.daySecondInterval(expr.precision()));
```
nielspardon: This is the only literal visit that doesn't take its type from `typeConverter.toCalcite(...)`, which costs it twice: it skips comment 2's guard, so `interval_day<9>` converts here but throws on the type path in the same plan, and `createSqlIntervalType` hardcodes a non-nullable type, so a nullable interval literal gets a NOT NULL Rex type — #1110's bug, in the one visitor #1110 special-cases.

```suggestion
    return rexBuilder.makeLiteral(
        new BigDecimal((expr.days() * MILLIS_IN_DAY + expr.seconds() * 1_000L + milliseconds)),
        typeConverter.toCalcite(typeFactory, expr.getType()),
        true);
```

This also replaces the merge resolution posted above, which keeps both problems.

isthmus/src/main/java/io/substrait/isthmus/expression/LiteralConverter.java
```diff
@@ -234,15 +234,24 @@ public Expression.Literal convert(RexLiteral literal, RelDataType resultType) {
...
+          // Calcite represents day/time intervals in milliseconds, whatever the type's scale says.
           Long totalMillis = Objects.requireNonNull(literal.getValueAs(Long.class));
           Duration interval = Duration.ofMillis(totalMillis);
 
           long days = interval.toDays();
           long seconds = interval.minusDays(days).toSeconds();
-          int micros = interval.toMillisPart() * 1000;
+          // Report the interval at its declared precision rather than at a fixed 6, so that a
+          // Substrait interval_day<P> keeps its P across a round trip. The value itself carries no
+          // more than milliseconds, so a precision above 3 only widens the unit, never the detail.
```
nielspardon: Calcite day-time interval values may carry fractional milliseconds — `RexLiteral` says so, and line 238's `getValueAs(Long.class)` is what discards them. So a precision above 3 can correspond to real detail this drops, and the result declares `interval_day<6>` on a millisecond value; a precision below 3 narrows the detail outright (comment 1).

```suggestion
          // Report the interval at the precision the Calcite type declares, so a Substrait
          // interval_day<P> keeps its P across a round trip. The value is narrowed to whole
          // milliseconds here, and P only sets the unit the subseconds component is expressed in.
```

The mirrored comment at `ExpressionRexConverter:410-412` has the same defect.

isthmus/src/main/java/io/substrait/isthmus/expression/FunctionMappings.java
```diff
@@ -248,11 +248,13 @@ private static RelDataType inferDatetimeSubtractType(SqlOperatorBinding binding)
...
-      // ExpressionRexConverter keeps the declared Substrait output type authoritative.
-      int precision = typeFactory.getTypeSystem().getMaxPrecision(SqlTypeName.TIMESTAMP);
+      // The spec declares subtract(date, interval_day<P>) -> precision_timestamp<P>, and the
+      // interval operand carries P as its fractional-second scale, so the result precision follows
+      // from the argument. Clamped because the two type-system maxima are configured separately.
+      int precision =
+          Math.min(
+              intervalType.getScale(),
+              typeFactory.getTypeSystem().getMaxPrecision(SqlTypeName.TIMESTAMP));
       resultType = typeFactory.createSqlType(SqlTypeName.TIMESTAMP, precision);
```
nielspardon: The clamp only bounds the top, so a negative scale reaches `createSqlType(SqlTypeName.TIMESTAMP, -2)`, which Calcite guards only with an `assert`. With the ceiling at 9 it also starts binding: `subtract(date, interval_day<9>)` yields `precision_timestamp<6>` because the datetime ceiling is 6 (#995), not because the result precision follows from the argument — so the comment should say it's capped there and point at #995 rather than claiming the argument decides it. Line 251 also wants the spec version per AGENTS.md.

isthmus/src/test/java/io/substrait/isthmus/CalciteLiteralTest.java
```diff
@@ -213,30 +215,42 @@ void tIntervalYearMonthWithPrecision() {
...
+    // Calcite carries an interval literal's value in milliseconds whatever the type says, so a
+    // half-second is all the detail every precision under test can hold. What varies is the unit
+    // the subseconds component is expressed in, which is what interval_day<P> declares.
+    long millis =
+        TimeUnit.DAYS.toMillis(3)
+            + TimeUnit.HOURS.toMillis(5)
+            + TimeUnit.MINUTES.toMillis(7)
+            + TimeUnit.SECONDS.toMillis(9)
+            + (precision == 0 ? 0 : 500);
+    int subseconds = precision == 0 ? 0 : (int) (500 * Math.pow(10, precision - 3));
```
nielspardon: `+ (precision == 0 ? 0 : 500)` removes the fractional part exactly when precision is 0 — the case comment 1 breaks. Line 230 then derives the expectation from the production formula, so an inverted exponent would compute a matching wrong expectation and still pass. A `@CsvSource` table fixes both and lets precisions 1, 2, 4 and 5 in.

isthmus/src/test/java/io/substrait/isthmus/FunctionConversionTest.java
```diff
@@ -103,12 +103,15 @@ static Stream<Arguments> subtractDateIDayTypes() {
...
+            Named.of("legacy DATE output", TypeCreator.REQUIRED.DATE), 6, maxPrecisionTimestamp),
         Arguments.of(
-            Named.of("spec max-precision output", maxPrecisionTimestamp), maxPrecisionTimestamp),
+            Named.of("spec max-precision output", maxPrecisionTimestamp), 6, maxPrecisionTimestamp),
+        // subtract(date, interval_day<P>) -> precision_timestamp<P>: the inferred precision follows
+        // the argument, so this case varies both rather than only the declared output.
         Arguments.of(
-            Named.of("non-max declared output", TypeCreator.REQUIRED.precisionTimestamp(3)),
-            maxPrecisionTimestamp));
+            Named.of("spec non-max precision", TypeCreator.REQUIRED.precisionTimestamp(3)),
```
nielspardon, line 112: With declared and inferred precision both 3, this no longer guards that the declared output type wins over Calcite's inference — the behaviour #1099 added. A fourth case with declared `precisionTimestamp(3)` against an interval precision of 6 restores it.

isthmus/src/main/java/io/substrait/isthmus/SubstraitTypeSystem.java
```diff
@@ -28,9 +28,30 @@ public class SubstraitTypeSystem extends RelDataTypeSystemImpl {
...
-  public static final SqlIntervalQualifier DAY_SECOND_INTERVAL =
-      new SqlIntervalQualifier(TimeUnit.DAY, -1, TimeUnit.SECOND, 6, SqlParserPos.ZERO);
+  /**
+   * Interval qualifier from day to fractional second at microsecond precision.
+   *
+   * <p>Conversion of a Substrait {@code interval_day<P>} builds its qualifier from {@code P} via
+   * {@link #daySecondInterval(int)} instead; sharing one constant is what used to drop the
+   * precision.
+   */
+  public static final SqlIntervalQualifier DAY_SECOND_INTERVAL = daySecondInterval(6);
```
nielspardon: `DAY_SECOND_INTERVAL` has no references left after this PR, so the new Javadoc documents a public constant that pins P=6 and tells readers not to use it. Delete it, or `@Deprecated` — the commit is already breaking. The history narration here and on line 45 belongs in the commit body rather than the Javadoc.

nielspardon (changes requested):
Three blocking things: a declared scale below 3 now loses value where the fixed 6 didn't (1), the new guard reads Calcite's leading-field bound instead of the fractional-second one (2), and the literal path skips it entirely (3). Comment 1 is the one I'd want your call on rather than a fix.

Separately, Spark rejects any `interval_day` precision but 6 (`Util.assertMicroseconds`), so isthmus-produced plans will hit it more often after this — a follow-up issue, not this PR.

isthmus/src/main/java/io/substrait/isthmus/expression/LiteralConverter.java
alexandrefimov: Went with honouring the declared precision, i.e. your first option — but the rounding was a defect on top of it, not part of the choice, so it is fixed either way.

`Duration.ofMillis(-1500)` normalises to seconds=-2 with a positive 500 ms part, so dropping the part rounded away from zero. Scaling the whole value before taking it apart makes it truncate towards zero instead: -1500 ms at scale 0 is now -1000 ms rather than -2000 ms.

I did consider reporting at `max(P, 3)` or picking the precision from the value, and dropped both — this overload is handed `resultType` precisely so the literal matches the type the containing schema requires (its javadoc says as much), and returning a precision the caller did not ask for is the #1064 shape of bug. Rejecting felt too much for an inconsistency Calcite itself permits.

The table now covers 0 through 6 with a fractional part, both signs, plus a multi-day value so the decomposition is actually exercised. `LongMath.pow` and the `long` subseconds are in.

isthmus/src/main/java/io/substrait/isthmus/TypeConverter.java
alexandrefimov: Verified and applied, including the ceiling change. `SubstraitTypeSystem` reports `maxPrecision=6, maxScale=9` for `INTERVAL_DAY`, and `INTERVAL_DAY_SECOND` — the name the produced type actually carries — is `maxPrecision=10, maxScale=9`, unoverridden. And SQL confirms your point directly:

```
INTERVAL '1 00:00:00.1234567' DAY TO SECOND(7)  -> INTERVAL DAY TO SECOND(7) scale=7
INTERVAL '1 00:00:00.123456789' DAY TO SECOND(9) -> INTERVAL DAY TO SECOND(9) scale=9
```

so the old guard would have refused plans Isthmus can produce itself. Lower bound added too — `daySecondInterval(-1)` does come back as scale 6.

`BREAKING CHANGE` footer and the "SQL-parsed intervals are unaffected" claim both corrected; the latter only held for intervals written without an explicit precision.

Leaving the shared `checkPrecision(...)` for after #1113 as you suggest.

isthmus/src/main/java/io/substrait/isthmus/expression/ExpressionRexConverter.java
alexandrefimov: Applied. Both problems were real: with a qualifier-built type the literal skipped the bound entirely, and `createSqlIntervalType` returns a NOT NULL type, so a nullable interval literal came out non-nullable. Added a test for the nullability half.

This also retracts the merge resolution I posted in the PR comment above — it preserved both. Since the type now comes from the converter, the hunk that conflicted with #1110 is gone, and the two should merge without it.

isthmus/src/main/java/io/substrait/isthmus/expression/LiteralConverter.java
alexandrefimov: Taken as written. The comment now says the value is narrowed to whole milliseconds by `getValueAs(Long.class)` and that P only sets the unit, rather than claiming Calcite has nothing finer to give. Fixed the mirrored one at the `ExpressionRexConverter` end too.

isthmus/src/main/java/io/substrait/isthmus/expression/FunctionMappings.java
alexandrefimov: Both applied — lower bound on the clamp, and the comment now says the result is capped at the datetime ceiling and points at #995, instead of implying the argument settles it. Spec version added per AGENTS.md.

isthmus/src/test/java/io/substrait/isthmus/CalciteLiteralTest.java
alexandrefimov: You are right that it dodged its own case. Replaced with a `@CsvSource` table: precisions 0 through 6, a fractional part in every row, both signs where the sign matters, expectations written out rather than derived, and a separate column for the millisecond value that should come back — so an inverted exponent shows up on one side or the other.

isthmus/src/test/java/io/substrait/isthmus/FunctionConversionTest.java
alexandrefimov: Restored as a fourth case: declared `precisionTimestamp(3)` against an interval precision of 6, so the assertion that the supplied Substrait type beats Calcite's inference is exercised again. Thanks for catching that — I had replaced the guard rather than added to it.

isthmus/src/main/java/io/substrait/isthmus/SubstraitTypeSystem.java
alexandrefimov: Deleted. Nothing referenced it after this change, and a public constant whose javadoc explains why not to use it is worse than its absence. The history narration is in the commit body instead.

alexandrefimov:
Thanks for this one — the `getMaxPrecision` / `getMaxScale` split is not something I would have found on my own, and it turned a guard that looked like a safety net into one that would have rejected plans Isthmus itself emits. I checked every claim before touching anything and they all held, including the `-2000 ms` I doubted: `Duration` normalises `-1500 ms` to seconds=-2 with a positive 500 ms part, so the rounding really does run away from zero.

All eight addressed in 941059b, replies in the threads. The one you left to me went with honouring the declared precision, with the rounding fixed as a separate defect — reasoning in the first thread, happy to be argued out of it.

Filed #1125 for the Spark side.

isthmus/src/main/java/io/substrait/isthmus/expression/LiteralConverter.java
```diff
@@ -234,15 +234,30 @@ public Expression.Literal convert(RexLiteral literal, RelDataType resultType) {
...
+                  ? totalMillis * LongMath.pow(10, precision - 3)
+                  : totalMillis / LongMath.pow(10, 3 - precision);
+          long unitsPerSecond = LongMath.pow(10, precision);
+          long unitsPerDay = TimeUnit.DAYS.toSeconds(1) * unitsPerSecond;
 
-          return ExpressionCreator.intervalDay(nullable, (int) days, (int) seconds, micros, 6);
+          long days = total / unitsPerDay;
+          long remainder = total - days * unitsPerDay;
+          long seconds = remainder / unitsPerSecond;
+          long subseconds = remainder - seconds * unitsPerSecond;
```
nielspardon: Scaling the whole value first trades the sign bug for an overflow, and the spec's range puts it well in reach: `interval_day` allows ±3,650,000 days, but `totalMillis * 10^6` passes a `long` above 106,752 days, so at P=9 anything beyond ~292 years wraps — a 300-year interval comes back as `-8,979,464,073,709,551,616`. P=8 wraps above ~2,923 years. `LongMath.pow` is exact, but the multiply after it isn't checked, so it's silent. Raising the ceiling to 9 — which I suggested on the guard — is what brought this inside the representable range; at 6 the threshold sits near 292,000 years.

Decomposing in milliseconds and scaling only the remainder keeps the same towards-zero semantics with nothing larger than 999 to scale:

```suggestion
          // Decompose in milliseconds and scale only the sub-second remainder. Scaling the whole
          // value first overflows a long past ~292 years at P=9, well inside the spec's
          // [-3,650,000..3,650,000] day range. Integer division truncates towards zero and the
          // remainder keeps the dividend's sign, so a negative interval still narrows towards zero.
          long millisPerDay = TimeUnit.DAYS.toMillis(1);
          long days = totalMillis / millisPerDay;
          long remainder = totalMillis - days * millisPerDay;
          long seconds = remainder / 1_000L;
          long millisPart = remainder - seconds * 1_000L;
          long subseconds =
              precision > 3
                  ? millisPart * LongMath.pow(10, precision - 3)
                  : millisPart / LongMath.pow(10, 3 - precision);
```

That reproduces every row in the new `@CsvSource` table, including both negatives and the day-spanning pair, so only a large-day row at P=9 needs adding. The sub-3 truncation question from the thread above is settled by it either way.

Separately, nothing in `core` enforces the spec's per-component bound, so `3650001d -86400s` is constructible — filed as #1129, not this PR's problem.

isthmus/src/main/java/io/substrait/isthmus/expression/FunctionMappings.java
```diff
@@ -248,11 +248,18 @@ private static RelDataType inferDatetimeSubtractType(SqlOperatorBinding binding)
...
-      // pins the fractional-second precision at 6, so the operand no longer carries Substrait's
-      // interval_day<P>. Use the type system maximum as a best-effort inference;
-      // ExpressionRexConverter keeps the declared Substrait output type authoritative.
-      int precision = typeFactory.getTypeSystem().getMaxPrecision(SqlTypeName.TIMESTAMP);
+      // The spec declares subtract(date, interval_day<P>) -> precision_timestamp<P> (spec
+      // v0.101.0), and the interval operand carries P as its fractional-second scale. The result is
+      // capped at the timestamp maximum, which is lower than the interval one, so an
+      // interval_day<9>
+      // still yields precision_timestamp<6> until the datetime ceiling itself moves (see #995's
+      // subject: TypeConverter rejects precision_timestamp above 6).
```
nielspardon: My asking you to point at #995 was wrong — AGENTS.md keeps GitHub references out of source, describing behaviour and spec version instead. Naming the cap does the same job:

```suggestion
      // The spec declares subtract(date, interval_day<P>) -> precision_timestamp<P> (spec
      // v0.101.0), and the interval operand carries P as its fractional-second scale. The result
      // is capped at the timestamp maximum, which is lower than the interval one, so an
      // interval_day<9> still yields precision_timestamp<6> until the Substrait->Calcite
      // conversion accepts a precision_timestamp above 6.
```

That also unwraps `interval_day<9>` from its own line.

nielspardon (commented):
Two things left: the rewritten arithmetic overflows inside the spec's day range, and the `#995` reference is mine to take back.

isthmus/src/main/java/io/substrait/isthmus/expression/LiteralConverter.java
alexandrefimov: Applied as suggested, in 8fb856c. I checked the overflow rather than taking it: at P=9 the old code turns 3,650,000 days into 20,432 days 25,947 s — the wrap is exactly where you said it is.

The two `P=9` rows are in, one per sign, at the spec's maximum day count plus 500 ms:

```
"9, 315360000000500, 3650000, 0,  500000000, 315360000000500",
"9,-315360000000500,-3650000, 0, -500000000,-315360000000500",
```

Both fail on the previous commit with the values above and pass on this one. The other twelve rows are unchanged and still pass, so the sub-3 truncation is settled by the same code as you said.

I reworded the comment above the `P=0` rows too — it justified the truncation by "the whole value is scaled before it is taken apart", which is no longer what happens.

isthmus/src/main/java/io/substrait/isthmus/expression/FunctionMappings.java
alexandrefimov: Taken as written. Your wording also fixes the wrapping I left — `interval_day<9>` was sitting on a line of its own.

isthmus/src/main/java/io/substrait/isthmus/expression/LiteralConverter.java
alexandrefimov: One update here, since my explanation above went stale with the rewrite: truncation towards zero no longer comes from scaling the whole value first, it comes from the millisecond remainder keeping the dividend's sign. The choice and the outcome are unchanged — `-1500` ms at scale 0 is still `-1000` ms, asserted by the same row.

nielspardon (approved):
LGTM

PR #1121, opened 2026-08-18
fix(isthmus)!: convert temporal literals at their declared type

### LINKED ISSUE #1113: isthmus: precision_timestamp_tz literals cap at precision 3 converting to Calcite and come back as plain precision_timestamp

alexandrefimov:
`LiteralConverter.convert` computes the Substrait type of the literal it is handed on its first line, then does not use it in the temporal cases. `TIME` and `TIMESTAMP` were reported at a fixed precision 6 with the value rescaled to microseconds, and `TIMESTAMP_WITH_LOCAL_TIME_ZONE` shared the `TIMESTAMP` branch and came back as a plain `precision_timestamp`. `TypeConverter.toSubstrait` maps those same Calcite types to `precision_time<P>`, `precision_timestamp<P>` and `precision_timestamp_tz<P>`, so the two halves of one conversion disagreed about the same literal.

That is not only a round trip losing the producer's choice — SQL literals carry their own precision, so it is what Isthmus emits:

```
TIMESTAMP '2024-01-01 00:00:00'      -> Calcite TIMESTAMP(0) -> precision_timestamp<6>
TIME '12:30:45'                      -> Calcite TIME(0)      -> precision_time<6>
```

Take the precision from the result type and express the value at it, and route `TIMESTAMP_WITH_LOCAL_TIME_ZONE` to `precisionTimestampTZ`. At precision 6 the arithmetic is unchanged; other precisions round-trip instead of being widened.

In the other direction, the maximum-precision check for a `precision_timestamp_tz` literal looked up `SqlTypeName.TIMESTAMP_TZ`. `SubstraitTypeSystem` does not configure that name — it overrides `TIMESTAMP_WITH_LOCAL_TIME_ZONE` — so the lookup fell through to Calcite's default of 3 and rejected a microsecond TZ literal, while the corresponding *type* conversion accepted it. It now checks the name everything else TZ-related uses.

Emitting the declared precision only works if the reverse direction takes it back, and it did not: `getTimestampString` and `createTimeString` handled 0, 3, 6 and 9 and threw for everything else, so `TIMESTAMP '2024-01-01 00:00:00.12'` — precision 2, from the fractional digits written — produced a plan Isthmus could not read. `algebra.proto` documents what 0, 3, 6, 9 and 12 mean and leaves the field an unbounded `int32`, so the unit is now 10^-precision at any precision a `TimeString` or `TimestampString` carries. Precisions 1, 2, 4 and 5 were listed in `ExpressionConvertabilityTest` as unsupported for both timestamp literals; they moved to the supported side.

Two smaller things. A temporal type built from a Java class rather than a SQL type name carries `PRECISION_NOT_SPECIFIED`, which nothing read: a `java.sql.Time` field of a reflective schema became `precision_time<-1>` with the value divided by 10^10, and the timestamp side failed inside Guava. Both the type and the literal now read the sentinel as precision 0, the way Calcite's own `RexBuilder.clean` does. And the precision check, written out six times and identical apart from a `SqlTypeName` and a hand-typed noun, is now `SubstraitTypeSystem.requireSupportedPrecision` — a static, since `TypeConverter` and `ExpressionRexConverter` have no common ancestor to hang an instance method on.

Closes #1113. Finishes #1114, whose interval third is in #1120.

BREAKING CHANGE: a Calcite `TIME` or `TIMESTAMP` literal now converts to a Substrait literal carrying its own precision rather than always precision 6, so a plan built from `TIMESTAMP '2024-01-01 00:00:00'` declares `precision_timestamp<0>` with a value in seconds where it previously declared `precision_timestamp<6>` with a value in microseconds. A `TIMESTAMP WITH LOCAL TIME ZONE` literal converts to `precision_timestamp_tz` instead of `precision_timestamp`.

### DIFF
--- a/isthmus/src/main/java/io/substrait/isthmus/expression/ExpressionRexConverter.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/expression/ExpressionRexConverter.java
@@ -1,6 +1,7 @@
 package io.substrait.isthmus.expression;
 
 import com.google.common.collect.ImmutableList;
+import com.google.common.math.LongMath;
 import io.substrait.expression.AbstractExpressionVisitor;
 import io.substrait.expression.EnumArg;
 import io.substrait.expression.Expression;
@@ -19,6 +20,7 @@ import io.substrait.expression.WindowBound;
 import io.substrait.extension.SimpleExtension;
 import io.substrait.isthmus.SubstraitRelNodeConverter;
 import io.substrait.isthmus.SubstraitRelNodeConverter.Context;
+import io.substrait.isthmus.SubstraitTypeSystem;
 import io.substrait.isthmus.TypeConverter;
 import io.substrait.type.StringTypeVisitor;
 import io.substrait.type.Type;
@@ -273,13 +275,8 @@ public class ExpressionRexConverter
 
   @Override
   public RexNode visit(PrecisionTimeLiteral expr, Context context) throws RuntimeException {
-    int maxPrecision = typeFactory.getTypeSystem().getMaxPrecision(SqlTypeName.TIME);
-    if (expr.precision() > maxPrecision) {
-      throw new IllegalArgumentException(
-          String.format(
-              "unsupported precision_time precision %s, max precision in Calcite type system is set to %s",
-              expr.precision(), maxPrecision));
-    }
+    SubstraitTypeSystem.requireSupportedPrecision(
+        typeFactory.getTypeSystem(), SqlTypeName.TIME, "precision_time", expr.precision());
 
     return preserveNullability(
         rexBuilder.makeTimeLiteral(
@@ -290,15 +287,9 @@ public class ExpressionRexConverter
   /**
    * Creates a TimeString from a time value with the specified precision.
    *
-   * <p>The time unit for the value is dictated by the supplied precision, with the following valid
-   * values:
-   *
-   * <ul>
-   *   <li>0 - seconds
-   *   <li>3 - milliseconds
-   *   <li>6 - microseconds
-   *   <li>9 - nanoseconds
-   * </ul>
+   * <p>The time unit for the value is dictated by the supplied precision: the value counts
+   * 10^-precision seconds, so precision 0 is seconds, 3 milliseconds, 6 microseconds and 9
+   * nanoseconds, and every precision in between is a unit of its own.
    *
    * @param value the time value in the appropriate unit based on precision
    * @param precision the precision level
@@ -333,13 +324,11 @@ public class ExpressionRexConverter
 
   @Override
   public RexNode visit(PrecisionTimestampLiteral expr, Context context) throws RuntimeException {
-    int maxPrecision = typeFactory.getTypeSystem().getMaxPrecision(SqlTypeName.TIMESTAMP);
-    if (expr.precision() > maxPrecision) {
-      throw new IllegalArgumentException(
-          String.format(
-              "unsupported precision_timestamp precision %s, max precision in Calcite type system is set to %s",
-              expr.precision(), maxPrecision));
-    }
+    SubstraitTypeSystem.requireSupportedPrecision(
+        typeFactory.getTypeSystem(),
+        SqlTypeName.TIMESTAMP,
+        "precision_timestamp",
+        expr.precision());
 
     return preserveNullability(
         rexBuilder.makeTimestampLiteral(
@@ -350,15 +339,13 @@ public class ExpressionRexConverter
   @Override
   public RexNode visit(PrecisionTimestampTZLiteral expr, Context context) throws RuntimeException {
     // TIMESTAMP_WITH_LOCAL_TIME_ZONE, not TIMESTAMP_TZ: that is the name TypeConverter maps
-    // precision_timestamp_tz to, and the only one SubstraitTypeSystem configures a maximum for.
-    int maxPrecision =
-        typeFactory.getTypeSystem().getMaxPrecision(SqlTypeName.TIMESTAMP_WITH_LOCAL_TIME_ZONE);
-    if (expr.precision() > maxPrecision) {
-      throw new IllegalArgumentException(
-          String.format(
-              "unsupported precision_timestamp_tz precision %s, max precision in Calcite type system is set to %s",
-              expr.precision(), maxPrecision));
-    }
+    // precision_timestamp_tz to, so the bound checked here is the one the converted type is
+    // built against.
+    SubstraitTypeSystem.requireSupportedPrecision(
+        typeFactory.getTypeSystem(),
+        SqlTypeName.TIMESTAMP_WITH_LOCAL_TIME_ZONE,
+        "precision_timestamp_tz",
+        expr.precision());
 
     return rexBuilder.makeLiteral(
         getTimestampString(expr.value(), expr.precision()),
@@ -376,25 +363,23 @@ public class ExpressionRexConverter
   /**
    * Returns how many units of a temporal value at the given precision make up one second.
    *
+   * <p>Every precision from 0 to 9 is a unit, not just the multiples of three: {@code
+   * algebra.proto} documents what 0, 3, 6, 9 and 12 mean and leaves the field an unbounded {@code
+   * int32}, and Calcite types a literal by the fractional digits actually written, so {@code
+   * TIMESTAMP '2024-01-01 00:00:00.12'} is precision 2.
+   *
    * @param precision the fractional-second precision
    * @param typeName the Substrait type being converted, for the failure message
    * @return the number of units per second
-   * @throws IllegalArgumentException if the precision is not one Calcite can be given
+   * @throws IllegalArgumentException if the precision is finer than nanoseconds or negative
    */
   private static long unitsPerSecond(int precision, String typeName) {
-    switch (precision) {
-      case 0:
-        return 1L;
-      case 3:
-        return 1_000L;
-      case 6:
-        return 1_000_000L;
-      case 9:
-        return 1_000_000_000L;
-      default:
-        throw new IllegalArgumentException(
-            String.format("Cannot handle %s with precision %d.", typeName, precision));
+    if (precision < 0 || precision > 9) {
+      // A nanosecond is the finest unit a TimeString or TimestampString carries.
+      throw new IllegalArgumentException(
+          String.format("Cannot handle %s with precision %d.", typeName, precision));
     }
+    return LongMath.pow(10, precision);
   }
 
   /**
--- a/isthmus/src/main/java/io/substrait/isthmus/expression/ExpressionRexConverter.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/expression/ExpressionRexConverter.java
@@ -349,11 +349,14 @@ public class ExpressionRexConverter
 
   @Override
   public RexNode visit(PrecisionTimestampTZLiteral expr, Context context) throws RuntimeException {
-    int maxPrecision = typeFactory.getTypeSystem().getMaxPrecision(SqlTypeName.TIMESTAMP_TZ);
+    // TIMESTAMP_WITH_LOCAL_TIME_ZONE, not TIMESTAMP_TZ: that is the name TypeConverter maps
+    // precision_timestamp_tz to, and the only one SubstraitTypeSystem configures a maximum for.
+    int maxPrecision =
+        typeFactory.getTypeSystem().getMaxPrecision(SqlTypeName.TIMESTAMP_WITH_LOCAL_TIME_ZONE);
     if (expr.precision() > maxPrecision) {
       throw new IllegalArgumentException(
           String.format(
-              "unsupported precision_timestamp precision %s, max precision in Calcite type system is set to %s",
+              "unsupported precision_timestamp_tz precision %s, max precision in Calcite type system is set to %s",
               expr.precision(), maxPrecision));
     }
 
--- a/isthmus/src/main/java/io/substrait/isthmus/expression/LiteralConverter.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/expression/LiteralConverter.java
@@ -199,7 +199,7 @@ public class LiteralConverter {
           LocalTime localTime = LocalTime.parse(time.toString(), CALCITE_LOCAL_TIME_FORMATTER);
           int precision = TypeConverter.precisionOf(resultType);
           return ExpressionCreator.precisionTime(
-              nullable, subsecondsOf(localTime.toNanoOfDay(), precision), precision);
+              nullable, rescaleNanos(localTime.toNanoOfDay(), precision), precision);
         }
       case TIMESTAMP:
       case TIMESTAMP_WITH_LOCAL_TIME_ZONE:
@@ -210,7 +210,7 @@ public class LiteralConverter {
           int precision = TypeConverter.precisionOf(resultType);
           long value =
               localDateTime.toEpochSecond(ZoneOffset.UTC) * LongMath.pow(10, precision)
-                  + subsecondsOf(localDateTime.getNano(), precision);
+                  + rescaleNanos(localDateTime.getNano(), precision);
           // toEpochSecond floors, and the nanosecond part it leaves behind is always positive, so
           // a pre-epoch timestamp narrowed to a coarser precision moves back in time rather than
           // towards the epoch. That is what a timestamp wants — 1969-12-31 23:59:59.5 at second
@@ -317,12 +317,13 @@ public class LiteralConverter {
    * Rescales a nanosecond count to the fractional-second unit a Substrait temporal literal of the
    * given precision is expressed in.
    *
-   * @param nanos the sub-second component in nanoseconds, which {@code LocalTime} and {@code
-   *     LocalDateTime} both report as a non-negative value
+   * @param nanos a non-negative nanosecond count: a whole time of day for a {@code precision_time},
+   *     the sub-second remainder for a {@code precision_timestamp}. Non-negativity is the
+   *     precondition the division relies on, since it truncates towards zero rather than flooring.
    * @param precision the fractional-second precision of the target literal
-   * @return the sub-second component in units of 10^-precision seconds
+   * @return {@code nanos} in units of 10^-precision seconds
    */
-  private static long subsecondsOf(long nanos, int precision) {
+  private static long rescaleNanos(long nanos, int precision) {
     return precision >= 9
         ? nanos * LongMath.pow(10, precision - 9)
         : nanos / LongMath.pow(10, 9 - precision);
--- a/isthmus/src/main/java/io/substrait/isthmus/expression/LiteralConverter.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/expression/LiteralConverter.java
@@ -24,6 +24,7 @@ import java.util.stream.Collectors;
 import java.util.stream.IntStream;
 import org.apache.calcite.rel.type.RelDataType;
 import org.apache.calcite.rex.RexLiteral;
+import org.apache.calcite.sql.type.SqlTypeName;
 import org.apache.calcite.util.DateString;
 import org.apache.calcite.util.NlsString;
 import org.apache.calcite.util.TimeString;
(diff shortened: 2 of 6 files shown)


alexandrefimov:
Two things carried over from the #1120 review, so they do not have to be said twice.

The rescale now uses `LongMath.pow` instead of a hand-rolled loop, matching what #1120 ended up with. A shared home for it — `DecimalUtil.POWER_OF_10` is `private` today — belongs with the `checkPrecision(...)` extraction you suggested doing once this lands, since this PR is what un-drifts the TZ copy.

I also checked whether the rounding defect you found in #1120 exists here. It does not, and for a reason worth stating: Calcite hands the literal over already at its declared precision, so there is no narrowing decision on this side. Where narrowing does show, a timestamp floors rather than truncating towards the epoch — `1969-12-31 23:59:59.5` at second precision is `23:59:59`, not the epoch — which is the opposite of the interval case, where the value is a duration and narrowing shortens it. That is now a comment and a test rather than something a reader has to reconstruct.

While checking it I hit an unrelated, pre-existing bug: a pre-epoch `precision_timestamp` with a fractional second throws on the way to Calcite, because `getTimestampString` truncates towards zero and then hands `TimestampString.withNanos` a negative value. Reproduces on `main` at precision 6 too, so it is not from this change — filed as #1126.

alexandrefimov:
Self-review pass before this sits waiting, in the shape of the #1120 one.

Checked at runtime rather than by reading, and two of the three came back clean:

- `resultType.getPrecision()` cannot exceed 6 here. `createSqlType(TIMESTAMP, 9)` returns `TIMESTAMP(6)` under `SubstraitTypeSystem` — Calcite clamps to the type-system maximum — so the new `epochSeconds * 10^P` cannot overflow any further than the old `epochSeconds * 10^6` did.
- Calcite reports `TIMESTAMP(0)` / `TIME(0)` for a type built without an explicit precision, not `PRECISION_NOT_SPECIFIED`, so there is no `-1` reaching the rescale. If one ever did, `LongMath.pow` rejects a negative exponent rather than computing something.
- The gap: the `BREAKING CHANGE` footer is about the plans Isthmus emits from SQL, and nothing asserted that half. Added in 76d6af8 — `TIMESTAMP '2024-01-01 00:00:00'` now yields `precision_timestamp<0>` with a value in seconds, `TIMESTAMP '...00:00:00.123'` yields `<3>`, `TIME '12:30:45'` yields `precision_time<0>`. Reverting `LiteralConverter` to its `main` version fails it with `precision=6` on both sides, so it is pinning the change rather than restating it.

Also noted while there: #1127 fixes a pre-epoch throw in the same file, and the two do not overlap — that one is inside `getTimestampString` / `createTimeString`, this one is in the callers' counterpart on the other side.

isthmus/src/main/java/io/substrait/isthmus/expression/ExpressionRexConverter.java
```diff
@@ -346,11 +346,14 @@ public RexNode visit(PrecisionTimestampLiteral expr, Context context) throws Run
 
   @Override
   public RexNode visit(PrecisionTimestampTZLiteral expr, Context context) throws RuntimeException {
-    int maxPrecision = typeFactory.getTypeSystem().getMaxPrecision(SqlTypeName.TIMESTAMP_TZ);
+    // TIMESTAMP_WITH_LOCAL_TIME_ZONE, not TIMESTAMP_TZ: that is the name TypeConverter maps
```
nielspardon, line 341: The two `{0,3,6,9}` switches are what block arbitrary precision, and generalizing them is one rescale — sketch, untested:

```java
private TimestampString getTimestampString(long value, int precision) {
  long unitsPerSecond = LongMath.pow(10, precision);
  long seconds = Math.floorDiv(value, unitsPerSecond);
  TimestampString timestamp =
      TimestampString.fromMillisSinceEpoch(TimeUnit.SECONDS.toMillis(seconds));
  return precision == 0
      ? timestamp
      : timestamp.withNanos(
          (int) (Math.floorMod(value, unitsPerSecond) * LongMath.pow(10, 9 - precision)));
}
```

`floorDiv`/`floorMod` also fix a pre-epoch bug in the current code: `TimeUnit.MILLISECONDS.toSeconds(-500)` truncates to 0, leaving a negative nanosecond that `TimestampString.withNanos` rejects, so every sub-second pre-epoch timestamp throws today at precision 3 and 6 — including the `precisionTimestamp(false, -500, 3)` the new test asserts. `createTimeString` gets the same treatment.

While you're here: this guard is written out six times (here, 257, 335, and `TypeConverter` 301/313/326), identical apart from a `SqlTypeName` and a hand-typed name — which is what produced the bug you're fixing, since the copy carried both the wrong `SqlTypeName` and the wrong noun. One `requireSupportedPrecision(SqlTypeName, String, int)` would leave a single place that decides what is supported. Minor: the new comment says this is the only name `SubstraitTypeSystem` configures a maximum for, but there are eight, including the `TIMESTAMP` the guard above depends on.

isthmus/src/main/java/io/substrait/isthmus/expression/LiteralConverter.java
```diff
@@ -196,23 +197,31 @@ public Expression.Literal convert(RexLiteral literal, RelDataType resultType) {
...
           TimestampString timestamp = literal.getValueAs(TimestampString.class);
           LocalDateTime localDateTime =
               LocalDateTime.parse(timestamp.toString(), CALCITE_LOCAL_DATETIME_FORMATTER);
-          // Calcite supports up to microsecond precision (6)
-          long epochSeconds = localDateTime.toEpochSecond(ZoneOffset.UTC);
-          long epochMicros =
-              TimeUnit.SECONDS.toMicros(epochSeconds)
-                  + TimeUnit.NANOSECONDS.toMicros(localDateTime.getNano());
-          return ExpressionCreator.precisionTimestamp(nullable, epochMicros, 6);
+          int precision = resultType.getPrecision();
```
nielspardon: Just to record what the above currently blocks, since it isn't obvious from this diff: `TIMESTAMP '2024-01-01 00:00:00.12'` fails the reverse conversion with `Cannot handle PrecisionTimestamp with precision 2.`, as do 1, 4 and 5 fractional digits and `CAST(x AS TIMESTAMP(2))`. Your own pre-epoch example, `TIMESTAMP '1969-12-31 23:59:59.5'`, is precision 1. The TIME branch at line 200 hits `createTimeString` the same way.

isthmus/src/main/java/io/substrait/isthmus/expression/LiteralConverter.java
```diff
@@ -196,23 +197,31 @@ public Expression.Literal convert(RexLiteral literal, RelDataType resultType) {
         {
           TimeString time = literal.getValueAs(TimeString.class);
           LocalTime localTime = LocalTime.parse(time.toString(), CALCITE_LOCAL_TIME_FORMATTER);
-          // Calcite supports up to microsecond precision (6), convert nanoseconds to microseconds
-          long nanos = localTime.toNanoOfDay();
-          long micros = TimeUnit.NANOSECONDS.toMicros(nanos);
-          return ExpressionCreator.precisionTime(nullable, micros, 6);
+          int precision = resultType.getPrecision();
```
nielspardon: Neither temporal branch guards `RelDataType.PRECISION_NOT_SPECIFIED` (-1), which the `VARCHAR` case above does. TIME is the bad half — `subsecondsOf(nanos, -1)` divides by 10^10, so a `java.sql.Time`-typed schema field yields `precision_time<-1>` with value 4504 instead of 45045 and nothing downstream rejects it; TIMESTAMP instead dies as `exponent (-1) must be >= 0` from inside Guava. Reachable via `visit(Values)` over a `ReflectiveSchema`-style struct. Calcite's `RexBuilder.clean` treats the sentinel as precision 0, which would be the smallest fix.

isthmus/src/test/java/io/substrait/isthmus/CalciteLiteralTest.java
```diff
@@ -172,12 +177,100 @@ void tTimestampWithMilliMicroSeconds() {
...
+        ExpressionCreator.precisionTimestampTZ(false, timestampValue(precision), precision);
+
+    RexNode converted = timestampTz.accept(expressionRexConverter, Context.newContext());
+    assertEquals(SqlTypeName.TIMESTAMP_WITH_LOCAL_TIME_ZONE, converted.getType().getSqlTypeName());
+    assertEquals(precision, converted.getType().getPrecision());
+    assertEquals(timestampTz, converted.accept(rexExpressionConverter));
+  }
+
+  @Test
+  void tPrecisionTimestampBeforeTheEpochFloors() {
```
nielspardon, line 308: Once the reverse direction is general, the three `@ValueSource(ints = {0, 3, 6})` lists want to be `0`–`6` — that's the regression guard for all of the above. `tPrecisionTimestampBeforeTheEpochFloors` also can't currently fail: `RexBuilder` applies `TimestampString.round(0)` at construction so the precision-0 leg has `getNano() == 0`, and the precision-3 leg's `.5` is exactly 500 — no remainder either way, and swapping the floor for truncate-toward-zero leaves `:isthmus:test` green. Precision 1, where `.5` survives rounding, would make it bite, and it can round-trip both ways once `floorDiv` lands.

nielspardon (changes requested):
The `TIMESTAMP_TZ` diagnosis is right and the fix is real: `getMaxPrecision(TIMESTAMP_TZ)` is 3 while the type path uses `TIMESTAMP_WITH_LOCAL_TIME_ZONE` at 6, so `precision_timestamp_tz<6>` converted as a type but was rejected as a literal. Honouring the declared precision also fixes a real failure — `SELECT * FROM (VALUES (TIMESTAMP '2024-01-01 00:00:00'))` threw before this change and passes after it.

Emitting whatever precision Calcite declares is the right call, so the gap this opens is on the other side. `getTimestampString` and `createTimeString` only handle 0/3/6/9 and throw otherwise, so `TIMESTAMP '2024-01-01 00:00:00.12'` — which Calcite types `TIMESTAMP(2)` from the fractional digits written — now produces a plan Isthmus cannot read back. Nothing restricts a Substrait precision to multiples of three: `algebra.proto` only enumerates 0/3/6/9/12 as examples of what the number means, and this repo's own Spark path already rescales any precision 0–6 since #1128. So those two helpers should generalize rather than the forward direction narrowing back.

Worth rebasing onto main first — #1128, #1136 and #1110 landed after you branched and resolve several things this diff otherwise runs into. Also: the PR body's merge-order note and the sentence narrating the `tTimestampTZ` change are review notes rather than commit body.

isthmus/src/main/java/io/substrait/isthmus/expression/ExpressionRexConverter.java
nielspardon: One thing on the rebase ask in the review body — for this method in particular it's load-bearing rather than tidying. #1110 restructured the same visits: on main, `visit(PrecisionTimeLiteral)` and `visit(PrecisionTimestampLiteral)` are wrapped in a new `preserveNullability(...)` helper, and this TZ visit's `makeLiteral` call now takes a third argument (`allowCast = true`). `git merge-tree` reports no conflict, but that's textual only — #1110 changes literal nullability where this changes literal type and precision, in the same methods — so it's worth rebasing before reworking the switches rather than after.

alexandrefimov: Done as sketched: the switch is `LongMath.pow(10, precision)` now, bounded at 9 because that is the finest unit a `TimeString` or `TimestampString` carries. The `floorDiv`/`floorMod` half arrived with #1127, which this now sits on top of.

The guard is `SubstraitTypeSystem.requireSupportedPrecision(typeSystem, sqlTypeName, substraitTypeName, precision)` -- a static there rather than an instance method, since `TypeConverter` and `ExpressionRexConverter` have no common ancestor to hang one on. Six copies gone, and the wrong-noun message with them.

The comment is corrected too: the point is that `TIMESTAMP_WITH_LOCAL_TIME_ZONE` is the name `TypeConverter` maps `precision_timestamp_tz` to, not that it is the only name with a maximum -- you are right that there are eight.

One thing widening turned up: `ExpressionConvertabilityTest` listed precisions 1, 2, 4 and 5 as unsupported for both timestamp literals, so they were asserting the switch rather than the bound. They moved to the supported side.

isthmus/src/main/java/io/substrait/isthmus/expression/LiteralConverter.java
alexandrefimov: Both are covered now. Precision 0 through 6 round-trips in the three parameterised tests, and `TIMESTAMP '2024-01-01 00:00:00.12'` and `TIME '12:30:45.4'` are in the SQL test as the plans the parser actually produces -- the first is precision 2, the second precision 1.

isthmus/src/main/java/io/substrait/isthmus/expression/LiteralConverter.java
alexandrefimov: Right, and the type side has the same hole: `TypeConverter.toSubstrait` fed the sentinel straight into `precisionTime`/`precisionTimestamp`/`precisionTimestampTZ`, so a `precision_time<-1>` column would sit next to the literal. Both go through `TypeConverter.precisionOf` now, which reads it as 0 the way `RexBuilder.clean` does.

The test builds the types with `JavaTypeFactory.createJavaType(java.sql.Time.class)`. Checked what it gives without the fix: `PrecisionTimeLiteral{nullable=true, value=4504, precision=-1}`, exactly as you described.

isthmus/src/test/java/io/substrait/isthmus/CalciteLiteralTest.java
alexandrefimov: All three `@ValueSource` lists are 0-6, and the sub-second helper now takes the first `precision` digits of `.123456`, so 1, 2, 4 and 5 carry a fraction instead of testing a whole second. Confirmed they bite: with the old switch put back, each of those four fails in each of the three tests.

`tPrecisionTimestampBeforeTheEpochFloors` has precision 1 now, in both directions -- `.5` survives `round(1)`, so truncation towards zero would give 0 there and read back as the epoch itself.

alexandrefimov:
Rebased onto #1127 as you suggested, so the first two commits here are that PR and the generalization sits on top of its `floorDiv`/`floorMod` split.

The reverse direction takes any precision 0-9 now, `PRECISION_NOT_SPECIFIED` reads as 0 on both the type and the literal, and the six-times-over guard is one `SubstraitTypeSystem.requireSupportedPrecision`. The description lost the merge-order note and the `tTimestampTZ` narration.

One correction to something I then wrote myself: `algebra.proto` and `type.proto` say what 0, 3, 6, 9 and 12 mean and leave the field an unbounded `int32` — they do not say those are examples, that is a reading, so the comment and the commit body now claim only the former.

nielspardon:
I merged #1127 just now. Can you merge main into this PR branch and resolve the conflicts? Thanks.

alexandrefimov:
Rebased rather than merged: the branch was sitting on #1127's own two commits, and main now carries them as one squash, so a merge would have left both copies in the history here. Dropped the two, replayed the four on top of `fcfc36c`.

There was nothing to resolve in the end — `isthmus/` at the old base is identical to the squashed commit, so the four replayed unchanged and the diff against main is byte-identical to what the PR had before. The description lost its line about the merge order.

Happy to redo it as a merge if you'd rather have the merge commit in the history.

isthmus/src/test/java/io/substrait/isthmus/CalciteLiteralTest.java
```diff
@@ -232,12 +238,149 @@ void tTimestampWithMilliMicroSeconds() {
...
+                    .accept(expressionRexConverter, Context.newContext()))
+            .getValueAs(TimestampString.class));
+  }
+
+  @Test
+  void tSqlTimestampLiteralKeepsTheWrittenPrecision() throws Exception {
+    // The behaviour the breaking-change note is about: a literal written without a fractional part
+    // is TIMESTAMP(0) on the Calcite side, so the plan Isthmus emits declares
+    // precision_timestamp<0>
+    // with a value in seconds rather than precision_timestamp<6> with a value in microseconds.
```
nielspardon: "the breaking-change note" is a commit-message artifact, and this test outlives it: once the PR squashes, the note is a line in `CHANGELOG.md` that a reader here has no pointer to, and it describes a one-time transition where the test describes permanent behaviour. The rest of the comment already stands on its own, so dropping the first clause is enough (also fixes the stray wrap on `precision_timestamp<0>`):

```suggestion
    // A literal written without a fractional part is TIMESTAMP(0) on the Calcite side, so the
    // plan Isthmus emits declares precision_timestamp<0> with a value in seconds rather than
    // precision_timestamp<6> with a value in microseconds.
```

isthmus/src/main/java/io/substrait/isthmus/expression/LiteralConverter.java
```diff
@@ -304,6 +313,21 @@ public static byte[] padRightIfNeeded(
     return padRightIfNeeded(bytes.getBytes(), length);
   }
 
+  /**
+   * Rescales a nanosecond count to the fractional-second unit a Substrait temporal literal of the
+   * given precision is expressed in.
+   *
+   * @param nanos the sub-second component in nanoseconds, which {@code LocalTime} and {@code
```
nielspardon: This `@param` is narrower than one of the two call sites: the TIME branch passes `localTime.toNanoOfDay()`, a whole time of day, not a sub-second component. The arithmetic doesn't care, but as written it invites a future `checkArgument(nanos < 1_000_000_000)` that would quietly break TIME — non-negativity is the precondition the division actually relies on.

nielspardon (changes requested):
Two comment-wording asks inline. Both are about comments that lean on prose outside the codebase, so neither affects behaviour.

isthmus/src/test/java/io/substrait/isthmus/CalciteLiteralTest.java
alexandrefimov: Applied as suggested.

isthmus/src/main/java/io/substrait/isthmus/expression/LiteralConverter.java
alexandrefimov: Reworded — non-negativity is what the division relies on, since it truncates towards zero rather than flooring.

Went one step past the comment, though: `subsecondsOf` makes the same claim the `@param` did, and a reader meets the name before the doc, so it is `rescaleNanos` now. Happy to put the name back if you would rather keep this round to the wording.

nielspardon (approved):
LGTM, thanks

PR #1127, opened 2026-08-18
fix(isthmus): convert pre-epoch temporal literals instead of throwing

### LINKED ISSUE #1126: isthmus: a pre-epoch precision_timestamp with a fractional second throws converting to Calcite
Verified on `main` at 6b5ba4e. A `precision_timestamp` before the epoch with a fractional-second part throws on the way to Calcite:

```
PrecisionTimestampLiteral{value=-500, precision=3}    -> IllegalArgumentException
PrecisionTimestampLiteral{value=-500000, precision=6} -> IllegalArgumentException
```

Those are both `1969-12-31 23:59:59.5`. A whole-second value converts fine (`value=-1, precision=0`), so it is the fractional part that breaks it.

`ExpressionRexConverter.getTimestampString` splits the value into whole seconds and a nanosecond remainder:

```java
      case 3:
        {
          long seconds = TimeUnit.MILLISECONDS.toSeconds(value);
          int fracSecondsInNano =
              (int) (TimeUnit.MILLISECONDS.toNanos(value) - TimeUnit.SECONDS.toNanos(seconds));
          return TimestampString.fromMillisSinceEpoch(TimeUnit.SECONDS.toMillis(seconds))
              .withNanos(fracSecondsInNano);
        }
```

`TimeUnit.MILLISECONDS.toSeconds(-500)` truncates towards zero and gives `0`, so `fracSecondsInNano` is `-500_000_000` and `TimestampString.withNanos` rejects a negative argument. The same shape is in `createTimeString`, though a `precision_time` is a time of day and cannot be negative in practice.

The split has to floor rather than truncate for this to work: `-500 ms` is second `-1` plus `+500_000_000 ns`, which is what `TimestampString` expects. `Math.floorDiv` on the value before the split would do it.

The exception carries no message, so it also does not say what was rejected.

Found while working on #1121. Not introduced by it — the same throw happens at precision 6, which is what every literal used before that change.

alexandrefimov:
`getTimestampString` and `createTimeString` split a temporal value into whole seconds and a nanosecond remainder, once per supported precision, with truncating division:

```java
          long seconds = TimeUnit.MILLISECONDS.toSeconds(value);
          int fracSecondsInNano =
              (int) (TimeUnit.MILLISECONDS.toNanos(value) - TimeUnit.SECONDS.toNanos(seconds));
```

Before the epoch the value is negative and truncation rounds towards zero, so the remainder comes out negative — and `TimestampString.withNanos` rejects that. A `precision_timestamp` of `-500` at precision 3, which is `1969-12-31 23:59:59.5`, did not convert at all; it threw an `IllegalArgumentException` carrying no message.

Split with `floorDiv` / `floorMod` instead, so the remainder is never negative, and share the split between the two methods rather than repeating it four times each. The set of supported precisions (0, 3, 6, 9) is unchanged.

The second commit adds a rejection path `precision_time` did not have: it is a time of day, so a value below zero or a day or more is reported by value. Both ends read worse before — a negative value came back from `TimeString.fromMillisOfDay` as a corrupt time string, and a large one as either Guava's `Hour out of range: [24]` or, once the millisecond count overflows an int, a silently wrong time (4346734 seconds converted to 14:22:46). The bound is our reading of what a `precision_time` is; `algebra.proto` declares the field a signed `int64` with no stated range.

The new tests fail on `main` — the pre-epoch one with exactly the reported exception, the range one because nothing is thrown at all.

Closes #1126.

### DIFF
--- a/isthmus/src/main/java/io/substrait/isthmus/expression/ExpressionRexConverter.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/expression/ExpressionRexConverter.java
@@ -306,6 +306,12 @@ public class ExpressionRexConverter
    */
   protected TimeString createTimeString(long value, int precision) {
     long unitsPerSecond = unitsPerSecond(precision, "PrecisionTime");
+    if (value < 0 || value >= 86_400L * unitsPerSecond) {
+      // A precision_time is a time of day. Without this an out-of-range value reaches
+      // TimeString.fromMillisOfDay, which reports a corrupt time string rather than the value.
+      throw new IllegalArgumentException(
+          String.format("Cannot handle PrecisionTime with out-of-range value %d.", value));
+    }
     return TimeString.fromMillisOfDay(
             (int) TimeUnit.SECONDS.toMillis(secondsOf(value, unitsPerSecond)))
         .withNanos(nanosOf(value, unitsPerSecond));
--- a/isthmus/src/main/java/io/substrait/isthmus/expression/ExpressionRexConverter.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/expression/ExpressionRexConverter.java
@@ -305,36 +305,10 @@ public class ExpressionRexConverter
    * @return a TimeString representing the time value
    */
   protected TimeString createTimeString(long value, int precision) {
-    switch (precision) {
-      case 0:
-        return TimeString.fromMillisOfDay((int) TimeUnit.SECONDS.toMillis(value));
-      case 3:
-        {
-          long seconds = TimeUnit.MILLISECONDS.toSeconds(value);
-          int fracSecondsInNano =
-              (int) (TimeUnit.MILLISECONDS.toNanos(value) - TimeUnit.SECONDS.toNanos(seconds));
-          return TimeString.fromMillisOfDay((int) TimeUnit.SECONDS.toMillis(seconds))
-              .withNanos(fracSecondsInNano);
-        }
-      case 6:
-        {
-          long seconds = TimeUnit.MICROSECONDS.toSeconds(value);
-          int fracSecondsInNano =
-              (int) (TimeUnit.MICROSECONDS.toNanos(value) - TimeUnit.SECONDS.toNanos(seconds));
-          return TimeString.fromMillisOfDay((int) TimeUnit.SECONDS.toMillis(seconds))
-              .withNanos(fracSecondsInNano);
-        }
-      case 9:
-        {
-          long seconds = TimeUnit.NANOSECONDS.toSeconds(value);
-          int fracSecondsInNano = (int) (value - TimeUnit.SECONDS.toNanos(seconds));
-          return TimeString.fromMillisOfDay((int) TimeUnit.SECONDS.toMillis(seconds))
-              .withNanos(fracSecondsInNano);
-        }
-      default:
-        throw new IllegalArgumentException(
-            String.format("Cannot handle PrecisionTime with precision %d.", precision));
-    }
+    long unitsPerSecond = unitsPerSecond(precision, "PrecisionTime");
+    return TimeString.fromMillisOfDay(
+            (int) TimeUnit.SECONDS.toMillis(secondsOf(value, unitsPerSecond)))
+        .withNanos(nanosOf(value, unitsPerSecond));
   }
 
   @Override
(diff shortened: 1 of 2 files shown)


alexandrefimov:
Self-review turned up one thing in my own change, fixed in 0f6168d.

Flooring the split means a negative value no longer reaches `withNanos`, but for `createTimeString` it reaches `TimeString.fromMillisOfDay` instead, which reports `Invalid time format: [00:00:0/]` — a corrupt string rather than the value that was wrong. A `precision_time` is a time of day and cannot be negative, so it is now rejected by value with a message naming it. It threw before this PR too, with the bare `IllegalArgumentException` from `withNanos`, so this is a message rather than a behaviour change.

Two other checks came back clean: both `Cannot handle PrecisionTime/PrecisionTimestamp with precision %d.` messages are byte-identical to before, and nothing outside `ExpressionRexConverter` calls either method, so the shared split has no other callers to surprise.

One thing worth knowing for whoever reads the test: precision 9 is unreachable through this path even though the switch accepts it — `toCalcite` rejects `precision_timestamp<9>` first, per #995. So the test covers 0, 3 and 6.

isthmus/src/main/java/io/substrait/isthmus/expression/ExpressionRexConverter.java
```diff
@@ -284,36 +284,16 @@ public RexNode visit(PrecisionTimeLiteral expr, Context context) throws RuntimeE
...
-        }
-      default:
-        throw new IllegalArgumentException(
-            String.format("Cannot handle PrecisionTime with precision %d.", precision));
+    long unitsPerSecond = unitsPerSecond(precision, "PrecisionTime");
+    if (value < 0) {
+      // A precision_time is a time of day. Without this the negative value reaches
+      // TimeString.fromMillisOfDay, which reports a corrupt time string rather than the value.
+      throw new IllegalArgumentException(
+          String.format("Cannot handle PrecisionTime with negative value %d.", value));
```
nielspardon: This covers one end of the domain the comment invokes — a `precision_time` of a day or more isn't a time of day either, and the `(int)` narrowing below makes that end fail silently: precision 0 / `4_346_734` converts to `14:22:46`. A full range check covers both ends and makes the cast provably safe, and carries over unchanged once #1121 generalizes the unit.

```suggestion
    if (value < 0 || value >= 86_400L * unitsPerSecond) {
      // A precision_time is a time of day. Without this an out-of-range value reaches
      // TimeString.fromMillisOfDay, which reports a corrupt time string rather than the value.
      throw new IllegalArgumentException(
          String.format("Cannot handle PrecisionTime with out-of-range value %d.", value));
```

`tPrecisionTimeRejectsANegativeValue` needs the same wording. The bound is our reading, not the spec's — `algebra.proto` declares the field a signed `int64` with no stated lower bound.

isthmus/src/test/java/io/substrait/isthmus/CalciteLiteralTest.java
```diff
@@ -141,6 +142,52 @@ void tTimeWithNanoSecond() {
         rex.makeTimeLiteral(new TimeString("14:22:47.123456"), 6));
   }
 
+  @Test
+  void tPrecisionTimestampBeforeTheEpoch() {
+    // 1969-12-31 23:59:59.5 at each precision Calcite can be given. The value is negative and the
+    // sub-second part is not, so splitting it has to floor rather than truncate towards zero:
+    // TimestampString.withNanos rejects the negative remainder truncation leaves behind.
+    assertEquals(new TimestampString("1969-12-31 23:59:59"), timestampStringOf(-1L, 0));
```
nielspardon: Three of these five assertions already pass on `main` — only the `-500`/`-500_000` ones are regression coverage. This one can't exercise the change at all: `floorDiv(v, 1)` is the identity, and `RexBuilder` applies `round(0)` anyway. Precision 1 is the case that would bite, once #1121 makes it reachable.

nielspardon (changes requested):
The `floorDiv`/`floorMod` split is the right fix, and sharing it between the two methods is a real improvement.

This overlaps #1121 — both rewrite `getTimestampString` and `createTimeString` with the same split, so whichever lands second will conflict. I'd take this one first and let #1121's generalization sit on top; no need to act on the `{0,3,6,9}` Javadoc here, since #1121 replaces that switch outright.

One for the description, which becomes the squash-merge commit message: "both failure messages are unchanged" is no longer true — the second commit adds a third, and a rejection path `precision_time` didn't have.

isthmus/src/main/java/io/substrait/isthmus/expression/ExpressionRexConverter.java
alexandrefimov: Taken as written. One correction to the diagnosis, from pinning the test: the upper end is only silent past the int overflow. Below it `TimeString.fromMillisOfDay` still throws — `86_400` at precision 0 comes back as Guava's `Hour out of range: [24]` — and it is from `2_147_484` seconds on, where the millisecond count no longer fits an int, that the narrowing wraps and your `4_346_734` converts to `14:22:46`. Both ends are in the test now, with the first and last second that are still a time of day either side of the bound.

The reading-not-the-spec point went into the PR body rather than the code comment; happy to move it next to the check if you would rather have it there.

isthmus/src/test/java/io/substrait/isthmus/CalciteLiteralTest.java
alexandrefimov: Dropped the precision-0 assertion — between `floorDiv(v, 1)` and `round(0)` it asserted nothing about the change. The two pre-epoch ones stay as the regression guard, the epoch/positive pair as sanity that the shared split left the ordinary path alone, and precision 1 comes with #1121 once it is reachable.

alexandrefimov:
Rebased onto main and reworked per the review: the guard is a range check now, the test covers both ends of the domain with the boundary second either side, and the description no longer claims the failure messages are unchanged. Taking this one first, with #1121 to follow on top.

Unrelated to this PR, found while checking the branch: `./gradlew :isthmus:build` fails on a clean `origin/main` with every isthmus test throwing `NoClassDefFoundError: org/antlr/v4/runtime/tree/ParseTreeVisitor`. `:core:jar` and `:core:shadowJar` share an output file and there the plain one lands last. Filed as #1142.

nielspardon (approved):
LGTM

Issue #1125, opened 2026-08-18
spark: a column typed at a precision other than microseconds does not convert


alexandrefimov:
`Util.assertMicroseconds` rejects any precision but 6, and `ToSubstraitType.visit(Type.IntervalDay)` calls it before mapping to `DayTimeIntervalType.DEFAULT`:

```scala
  def assertMicroseconds(precision: Int): Unit = {
    if (precision != MICROSECOND_PRECISION) {
      throw new UnsupportedOperationException(
        s"Unsupported precision: $precision. Only microsecond precision ($MICROSECOND_PRECISION) is supported")
    }
  }
```

`ToSparkExpression` does the same for the literal. So an `interval_day<3>` plan does not convert to Spark at all, even though Spark's own `DayTimeIntervalType` is microsecond-based and could represent the value — `interval_day<3>` is a *coarser* precision, not a finer one, so the value always fits.

This is reachable today from any producer that emits a precision other than 6, and it becomes more common once isthmus stops normalizing: before #1120 isthmus reported every interval as `interval_day<6>` regardless of what the Calcite type said, so Spark saw 6 from that path whatever the plan carried. After it, an `interval_day<3>` travels through as `interval_day<3>`.

The same `assertMicroseconds` guards `precision_timestamp` and `precision_timestamp_tz`, where the argument is the same for any precision below 6 and different above it (a precision of 9 genuinely does not fit).

Widening the check to accept anything Spark can hold without loss — scaling the value to microseconds rather than refusing — would cover the coarser precisions in both directions. Precisions finer than microseconds still have to be rejected, and the message should say which side the limit comes from.

Raised by @nielspardon while reviewing #1120.

alexandrefimov:
#1128 covered the literal half of this: a coarse precision on a literal is rescaled into Spark's microseconds. The column half is untouched — a plan whose *column* type is `precision_timestamp<3>` is still rejected, because a Spark type carries no precision of its own and a type conversion has no value to rescale. Making that work means rescaling the data, which is a normalization pass over the plan rather than a change to the type mapping.

So this can either stay open for that half, or close with the normalization filed as its own issue. Your call — I left the `Closes` keyword off #1128 so both stayed available.

alexandrefimov:
[...]

PR #1128, opened 2026-08-18
fix(spark): accept Substrait precisions coarser than microseconds

alexandrefimov:
Spark stores timestamps and day-time intervals as a microseconds `Long`, and `Util.assertMicroseconds` refused any Substrait precision but 6. So a plan carrying `interval_day<3>` or `precision_timestamp<0>` did not convert to Spark at all, even though a coarser precision carries fewer digits than microseconds, not different ones, and fits that representation exactly.

The fix is a rescale rather than a relaxed guard, and it applies only where a value is in hand. `Util.toMicroseconds` scales a coarse value into microseconds and the three temporal literal conversions use it; the type conversions keep requiring exactly microseconds.

That split is the whole point. A Spark type carries no precision of its own, so mapping `precision_timestamp<3>` onto `TimestampNTZType` says nothing about the millisecond counts it describes, and a type conversion has no value at hand to rescale — a cast to a coarser precision becomes a cast between two `TimestampNTZType`s and truncates nothing, and a column declared at one has its data read as microseconds. Both previously failed loudly on the exact-6 check and still do. What changes is that a literal, which does carry its value, converts.

So this covers the literal half of #1125 and leaves the column half where it was: a plan whose *column* is `precision_timestamp<3>` is still rejected. Making that work means rescaling data rather than mapping a type — a normalization pass over the plan — which is worth deciding on deliberately rather than acquiring as a side effect here. The `Closes` keyword is off this description for that reason.

A precision finer than microseconds is still rejected everywhere: it does not fit, and rounding it would be the silent kind of loss this is avoiding.

The reverse direction needs nothing, and the asymmetry mostly closes on its own: a coarse precision now enters only on a literal, is rescaled there, and leaves as 6, which is lossless; a column type at a coarse precision does not enter at all, so nothing widens.

Two things came along that are not strictly the fix:

- The rescale multiplies, so it uses `Math.multiplyExact`, and the interval literal's `days * 86400 * 1e6` does too. Neither overflows within the spec's ranges, but `Long.MAX_VALUE` at precision 0 used to wrap to `-1000000` — one second before the epoch — rather than fail.
- `DialectGenerator` declared `max_precision = 9` for `PRECISION_TIMESTAMP`, `PRECISION_TIMESTAMP_TZ` and `INTERVAL_DAY`, which was already wrong before this PR, since the guard has always required 6. It reads `Util.MICROSECOND_PRECISION` now, and `spark_dialect.yaml` is regenerated. `core`'s `SparkDialectParseTest` pins that number off the copied file, so it moves with it.

Part of #1125. Found by @nielspardon while reviewing #1120.

### DIFF
--- a/spark/src/main/scala/io/substrait/spark/ToSubstraitType.scala
+++ b/spark/src/main/scala/io/substrait/spark/ToSubstraitType.scala
@@ -60,16 +60,18 @@ private class ToSparkType(dfsNames: Seq[String])
   override def visit(expr: Type.Bool): BooleanType = BooleanType
 
   override def visit(expr: Type.PrecisionTimestamp): TimestampNTZType = {
-    Util.assertMicroseconds(expr.precision())
+    // Spark has one timestamp type, stored as microseconds. A coarser Substrait precision maps
+    // onto it without loss; the literal converter scales the value.
+    Util.assertRepresentableAsMicroseconds(expr.precision())
     TimestampNTZType
   }
   override def visit(expr: Type.PrecisionTimestampTZ): TimestampType = {
-    Util.assertMicroseconds(expr.precision())
+    Util.assertRepresentableAsMicroseconds(expr.precision())
     TimestampType
   }
 
   override def visit(expr: Type.IntervalDay): DayTimeIntervalType = {
-    Util.assertMicroseconds(expr.precision())
+    Util.assertRepresentableAsMicroseconds(expr.precision())
     DayTimeIntervalType.DEFAULT
   }
 
--- a/spark/src/main/scala/io/substrait/spark/ToSubstraitType.scala
+++ b/spark/src/main/scala/io/substrait/spark/ToSubstraitType.scala
@@ -59,19 +59,21 @@ private class ToSparkType(dfsNames: Seq[String])
 
   override def visit(expr: Type.Bool): BooleanType = BooleanType
 
+  // These three keep rejecting anything but microsecond precision. Spark's types carry no
+  // precision of their own, so mapping a coarser one onto them would silently reinterpret the
+  // values it describes — a column of millisecond counts read as microsecond counts. The literal
+  // conversions rescale instead, which they can do because they hold the value.
   override def visit(expr: Type.PrecisionTimestamp): TimestampNTZType = {
-    // Spark has one timestamp type, stored as microseconds. A coarser Substrait precision maps
-    // onto it without loss; the literal converter scales the value.
-    Util.assertRepresentableAsMicroseconds(expr.precision())
+    Util.assertMicroseconds(expr.precision())
     TimestampNTZType
   }
   override def visit(expr: Type.PrecisionTimestampTZ): TimestampType = {
-    Util.assertRepresentableAsMicroseconds(expr.precision())
+    Util.assertMicroseconds(expr.precision())
     TimestampType
   }
 
   override def visit(expr: Type.IntervalDay): DayTimeIntervalType = {
-    Util.assertRepresentableAsMicroseconds(expr.precision())
+    Util.assertMicroseconds(expr.precision())
     DayTimeIntervalType.DEFAULT
   }
 
--- a/spark/src/main/scala/io/substrait/spark/utils/Util.scala
+++ b/spark/src/main/scala/io/substrait/spark/utils/Util.scala
@@ -25,12 +25,38 @@ object Util {
   val MICROS_PER_SECOND: Long = 1000 * 1000
   val MICROSECOND_PRECISION = 6 // for PrecisionTimestamp(TZ) and IntervalDay types
 
-  def assertMicroseconds(precision: Int): Unit = {
-    // Spark uses microseconds as a Long value as the "physical" type for most time things
-    if (precision != MICROSECOND_PRECISION) {
+  /**
+   * Checks that a Substrait fractional-second precision is one Spark's microsecond representation
+   * can hold. Coarser precisions carry fewer digits, not different ones, so they fit.
+   */
+  def assertRepresentableAsMicroseconds(precision: Int): Unit = {
+    if (precision < 0 || precision > MICROSECOND_PRECISION) {
       throw new UnsupportedOperationException(
-        s"Unsupported precision: $precision. Only microsecond precision ($MICROSECOND_PRECISION) is supported")
+        s"Unsupported precision: $precision. Spark stores time values as microseconds " +
+          s"($MICROSECOND_PRECISION), which cannot represent a finer precision")
     }
+  }
+
+  /**
+   * Rescales a sub-second value from the given Substrait precision to the microseconds Spark uses
+   * as its physical representation.
+   *
+   * A precision coarser than microseconds carries fewer digits, not different ones, so the value
+   * is exact after scaling. A finer one does not fit and is rejected rather than rounded.
+   */
+  def toMicroseconds(value: Long, precision: Int): Long = {
+    assertRepresentableAsMicroseconds(precision)
+    value * powerOfTen(MICROSECOND_PRECISION - precision)
+  }
+
+  private def powerOfTen(exponent: Int): Long = {
+    var result = 1L
+    var i = 0
+    while (i < exponent) {
+      result *= 10
+      i += 1
+    }
+    result
   }
 
   /**
--- a/spark/src/main/scala/io/substrait/spark/utils/Util.scala
+++ b/spark/src/main/scala/io/substrait/spark/utils/Util.scala
@@ -25,15 +25,24 @@ object Util {
   val MICROS_PER_SECOND: Long = 1000 * 1000
   val MICROSECOND_PRECISION = 6 // for PrecisionTimestamp(TZ) and IntervalDay types
 
+  /** Indexed by exponent, which [[toMicroseconds]] bounds to 0..MICROSECOND_PRECISION. */
+  private val POWERS_OF_TEN: Array[Long] =
+    Array(1L, 10L, 100L, 1000L, 10000L, 100000L, 1000000L)
+
   /**
-   * Checks that a Substrait fractional-second precision is one Spark's microsecond representation
-   * can hold. Coarser precisions carry fewer digits, not different ones, so they fit.
+   * Checks that a Substrait fractional-second precision is one a Spark type can carry.
+   *
+   * Spark has a single microsecond-based representation for each of these types, so the precision
+   * has to be exactly microseconds. A coarser one describes values in different units, and a type
+   * conversion has no values in hand to rescale — see [[toMicroseconds]], which is what the literal
+   * conversions use where a value is available.
    */
-  def assertRepresentableAsMicroseconds(precision: Int): Unit = {
-    if (precision < 0 || precision > MICROSECOND_PRECISION) {
+  def assertMicroseconds(precision: Int): Unit = {
+    if (precision != MICROSECOND_PRECISION) {
       throw new UnsupportedOperationException(
-        s"Unsupported precision: $precision. Spark stores time values as microseconds " +
-          s"($MICROSECOND_PRECISION), which cannot represent a finer precision")
+        s"Unsupported precision: $precision. Spark stores time values as microseconds, so a type " +
+          s"must declare a precision of exactly $MICROSECOND_PRECISION; a value at a coarser " +
+          s"precision is converted by rescaling it")
     }
   }
 
@@ -41,22 +50,16 @@ object Util {
    * Rescales a sub-second value from the given Substrait precision to the microseconds Spark uses
    * as its physical representation.
    *
-   * A precision coarser than microseconds carries fewer digits, not different ones, so the value
-   * is exact after scaling. A finer one does not fit and is rejected rather than rounded.
+   * A precision coarser than microseconds carries fewer digits, not different ones, so the value is
+   * exact after scaling. A finer one does not fit and is rejected rather than rounded.
    */
   def toMicroseconds(value: Long, precision: Int): Long = {
-    assertRepresentableAsMicroseconds(precision)
-    value * powerOfTen(MICROSECOND_PRECISION - precision)
-  }
-
-  private def powerOfTen(exponent: Int): Long = {
-    var result = 1L
-    var i = 0
-    while (i < exponent) {
-      result *= 10
-      i += 1
+    if (precision < 0 || precision > MICROSECOND_PRECISION) {
+      throw new UnsupportedOperationException(
+        s"Unsupported precision: $precision. Spark stores time values as microseconds, " +
+          s"so the precision must be between 0 and $MICROSECOND_PRECISION")
     }
-    result
+    Math.multiplyExact(value, POWERS_OF_TEN(MICROSECOND_PRECISION - precision))
   }
 
   /**
--- a/spark/src/test/scala/io/substrait/spark/TypesAndLiteralsSuite.scala
+++ b/spark/src/test/scala/io/substrait/spark/TypesAndLiteralsSuite.scala
@@ -1,6 +1,7 @@
 package io.substrait.spark
 
 import io.substrait.spark.expression.{ToSparkExpression, ToSubstraitLiteral}
+import io.substrait.spark.utils.Util
 
 import org.apache.spark.SparkFunSuite
 import org.apache.spark.sql.Row
@@ -10,9 +11,8 @@ import org.apache.spark.sql.types._
 import org.apache.spark.substrait.SparkTypeUtil
 import org.apache.spark.unsafe.types.UTF8String
 
-import io.substrait.expression.ExpressionCreator
-import io.substrait.spark.utils.Util
 import io.substrait.`type`.TypeCreator
+import io.substrait.expression.{Expression => SExpression, ExpressionCreator}
 import io.substrait.util.EmptyVisitationContext
 
 import java.time.{Duration, Instant, LocalDate, LocalDateTime, Period}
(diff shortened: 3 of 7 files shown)


spark/src/main/scala/io/substrait/spark/ToSubstraitType.scala
```diff
@@ -60,16 +60,18 @@ private class ToSparkType(dfsNames: Seq[String])
   override def visit(expr: Type.Bool): BooleanType = BooleanType
 
   override def visit(expr: Type.PrecisionTimestamp): TimestampNTZType = {
-    Util.assertMicroseconds(expr.precision())
+    // Spark has one timestamp type, stored as microseconds. A coarser Substrait precision maps
+    // onto it without loss; the literal converter scales the value.
+    Util.assertRepresentableAsMicroseconds(expr.precision())
```
nielspardon: `assertRepresentableAsMicroseconds` here relaxes the guard for **every** consumer of `ToSparkType`, but the rescale this PR adds lives only in the three literal visits. So the comment above holds for literals and not for the rest:

- **Cast targets** (`ToSparkExpression.scala:190`): `CAST(x AS precision_timestamp<3>)` becomes `Cast(x, TimestampNTZType)`, which truncates nothing. Checked on `spark-3.5_2.12`: casting `precisionTimestamp(1234567, 6)` to `precision_timestamp<3>` evaluates to `1234567` micros where the plan asks for `1234000`. Reachable from plain SQL — `TypeConverter.java:151` emits `precisionTimestamp(type.getPrecision())` and Calcite's default `TIMESTAMP` precision is 0, so `CAST(x AS TIMESTAMP)` produces `precision_timestamp<0>`.
- **Schema and column types** via `toStructType`/`toAttributeSeq`, reached from `ToLogicalPlan.scala` 432 (LocalFiles read schema), 520 (write), 580 (DDL) and 676 (root record type): a column declared `precision_timestamp<3>` becomes `TimestampNTZType` and its millisecond values are then read as microseconds.

Both previously failed loudly on the exact-6 check, so as written this trades a hard error for the same silent 1000x error the PR exists to fix.

Would it work to keep the guard strict in `ToSparkType` and move the relaxation into the literal visits, where a value is actually in hand? That keeps the #1125 fix and leaves the other paths rejecting. The alternative shape would be a plan-level normalization that rescales values and rewrites precisions to 6 before conversion — more work, but it would cover column data too. Your call which fits better.

spark/src/main/scala/io/substrait/spark/utils/Util.scala
```diff
@@ -25,12 +25,38 @@ object Util {
...
+  }
+
+  /**
+   * Rescales a sub-second value from the given Substrait precision to the microseconds Spark uses
+   * as its physical representation.
+   *
+   * A precision coarser than microseconds carries fewer digits, not different ones, so the value
+   * is exact after scaling. A finer one does not fit and is rejected rather than rounded.
+   */
+  def toMicroseconds(value: Long, precision: Int): Long = {
```
nielspardon, line 56: Worth deciding deliberately before precision 0 becomes valid: `Literal.PrecisionTimestamp.precision` and `Literal.IntervalDayToSecond.precision` are plain (non-`optional`) `int32` in `algebra.proto`, and `ProtoExpressionConverter` reads them with a bare `getPrecision()` — so a producer that omits the field is indistinguishable from one that means seconds. That previously failed loudly here; now the value is multiplied by 10^6.

Core guards the analogous *type* field explicitly (`ProtoTypeConverter.java:58`, "Interval day type must have its precision set"), but the literal has no equivalent, so this check was the last backstop.

spark/src/main/scala/io/substrait/spark/utils/Util.scala
```diff
@@ -25,12 +25,38 @@ object Util {
...
+    assertRepresentableAsMicroseconds(precision)
+    value * powerOfTen(MICROSECOND_PRECISION - precision)
```
nielspardon: `value * powerOfTen(...)` wraps silently, which contradicts the Scaladoc's "the value is exact after scaling" just above: `toMicroseconds(1000000000000000L, 0)` returns `3875820019684212736` with no error.

The overflow region sits above Spark's own maximum timestamp, so no representable instant is corrupted — but sentinel values land badly: `Long.MAX_VALUE` at precision 0 yields exactly `-1000000` micros (1969-12-31T23:59:59) and `Long.MIN_VALUE` yields exactly `0`.

```suggestion
    Math.multiplyExact(value, powerOfTen(MICROSECOND_PRECISION - precision))
```

The same reasoning applies to the interval expression you rewrote at `ToSparkExpression.scala:129`, where `days * 86400 * 1e6` wraps above 106,751,991 days. That one predates this PR, so entirely optional — but the rewrite is a natural moment for `Math.addExact`/`multiplyExact` there too.

nielspardon (commented):
The fix for #1125 is right in substance — a coarse precision does fit Spark's microsecond representation, and scaling the literal value is the correct reading. My concern is placement: the guard you relaxed sat in the *type* visitor, which serves far more than literals, while the rescale only landed in the literal visits. That turns two previously-loud paths into silently-wrong ones. Details inline.

Two things that can't be anchored to the diff:

- `DialectGenerator.supportedTypes()` still declares `max_precision = Some(9)` for `PRECISION_TIMESTAMP`, `PRECISION_TIMESTAMP_TZ` and `INTERVAL_DAY`, and the checked-in `spark/spark_dialect.yaml` agrees. This PR is what decides the answer is 0..6, so a producer reading the dialect to decide what to emit is told 9 is fine and then hits the rejection your new test asserts. `DialectSuite` can't catch the divergence — it only compares the generator against the file, and both hold the same number. Feeding these from `Util.MICROSECOND_PRECISION` keeps them in step.
- The Spark→Substrait direction still hard-pins precision 6 (`ToSubstraitType.scala:154-156`, `ToSubstraitLiteral.scala:38,103,105`), so a coarse precision that now enters comes back out as 6. For a literal that's value-lossless since it was rescaled; for a column type the type widens while the data doesn't. Worth a deliberate decision rather than a side effect.

Unrelated to the code: per AGENTS.md ("Keep PR descriptions high-signal"), the closing paragraph of the description — "Verified on all three variants ... 62 tests each" — is worth removing, since the title and body become the squash-merge commit that semantic-release turns into CHANGELOG.md and CI already reports what passed. The rationale paragraphs above it are exactly what that guidance asks for.

spark/src/main/scala/io/substrait/spark/utils/Util.scala
```diff
@@ -25,12 +25,38 @@ object Util {
...
+   * Checks that a Substrait fractional-second precision is one Spark's microsecond representation
+   * can hold. Coarser precisions carry fewer digits, not different ones, so they fit.
+   */
+  def assertRepresentableAsMicroseconds(precision: Int): Unit = {
+    if (precision < 0 || precision > MICROSECOND_PRECISION) {
       throw new UnsupportedOperationException(
-        s"Unsupported precision: $precision. Only microsecond precision ($MICROSECOND_PRECISION) is supported")
+        s"Unsupported precision: $precision. Spark stores time values as microseconds " +
+          s"($MICROSECOND_PRECISION), which cannot represent a finer precision")
+    }
```
nielspardon: Small thing: the `precision < 0` branch reports the opposite of the problem. `assertRepresentableAsMicroseconds(-1)` throws "...which cannot represent a finer precision", which sends the reader looking for a nanosecond issue that isn't there. Stating the accepted range covers both branches:

```suggestion
    if (precision < 0 || precision > MICROSECOND_PRECISION) {
      throw new UnsupportedOperationException(
        s"Unsupported precision: $precision. Spark stores time values as microseconds, " +
          s"so the precision must be between 0 and $MICROSECOND_PRECISION")
    }
```

spark/src/test/scala/io/substrait/spark/TypesAndLiteralsSuite.scala
```diff
@@ -237,4 +240,51 @@ class TypesAndLiteralsSuite extends SparkFunSuite {
         .sorted
         .sameElements(originalValues))
   }
+
+  test("coarser Substrait precisions convert to Spark microseconds") {
```
nielspardon: Two gaps in the new coverage:

- The assertions compare only `.value`, which is typed `Any`, so they pin the magnitude but not the type. `assert(x.value === 1000000L)` would still pass if the conversion returned an `Int`, or wrapped the right number in `TimestampType` where `TimestampNTZType` was meant — and the two visits at `ToSparkExpression.scala:114`/`:121` differ only in that. Asserting `.dataType` alongside the value closes it.
- The rejection test below covers only `precisionTimestamp`, though the relaxation touched three carriers. Nothing asserts that `precisionTimestampTZ(..., 9)` or `intervalDay(..., 9)` is still refused, at either the literal or the type level — so dropping the guard from `ToSubstraitType.scala:69` or `:74` later would leave this suite green while nanosecond values are silently truncated. `interval_day` also appears only at precision 3 here, as a single case outside the loop.

spark/src/test/scala/io/substrait/spark/TypesAndLiteralsSuite.scala
```diff
@@ -10,6 +10,9 @@ import org.apache.spark.sql.types._
 import org.apache.spark.substrait.SparkTypeUtil
 import org.apache.spark.unsafe.types.UTF8String
 
+import io.substrait.expression.ExpressionCreator
+import io.substrait.spark.utils.Util
```
nielspardon: Nit: `spark/.scalafmt.conf` puts `io.substrait.spark.*` in the first import group and sets `contiguousGroups = no`, so this one belongs beside line 3 rather than in the catch-all block — running the formatter will relocate it and produce an unrelated diff later.

Worth fixing by hand, because nothing in the build currently wires scalafmt in, so CI won't flag it. I filed #1130 for that gap.

spark/src/main/scala/io/substrait/spark/utils/Util.scala
```diff
@@ -25,12 +25,38 @@ object Util {
...
+   *
+   * A precision coarser than microseconds carries fewer digits, not different ones, so the value
+   * is exact after scaling. A finer one does not fit and is rejected rather than rounded.
+   */
+  def toMicroseconds(value: Long, precision: Int): Long = {
+    assertRepresentableAsMicroseconds(precision)
+    value * powerOfTen(MICROSECOND_PRECISION - precision)
+  }
+
+  private def powerOfTen(exponent: Int): Long = {
```
nielspardon: Optional: `assertRepresentableAsMicroseconds` bounds the exponent to 0..6, so `powerOfTen` can only ever return one of seven values. A 7-entry `Array[Long]` indexed by `MICROSECOND_PRECISION - precision` would make the legal domain visible at a glance and remove the loop.

Worth noting the loop currently returns `1` for a negative exponent rather than failing, so its correctness depends entirely on the caller's guard — a lookup can't be wrong that way.

spark/src/test/scala/io/substrait/spark/TypesAndLiteralsSuite.scala
nielspardon: Update: #1130 is fixed — #1132 landed the Scala Spotless step on the parent `:spark` project, so `./gradlew :spark:spotlessApply` now works and `:spark:spotlessCheck` gates the build. Which means the opposite of what I said above: CI **will** flag this once you rebase on main.

It also wants two changes here rather than just the one I described. #1132 reformatted the files that previously had import sort-order drift, so that ordering is enforced now too — under `rewrite.imports.sort = scalastyle` the backtick sorts before letters:

```
 import io.substrait.spark.expression.{ToSparkExpression, ToSubstraitLiteral}
+import io.substrait.spark.utils.Util
 ...
-import io.substrait.expression.ExpressionCreator
-import io.substrait.spark.utils.Util
 import io.substrait.`type`.TypeCreator
+import io.substrait.expression.ExpressionCreator
```

Rebasing on main and running `./gradlew :spark:spotlessApply` handles both, so nothing to do by hand.

spark/src/main/scala/io/substrait/spark/ToSubstraitType.scala
alexandrefimov: Verified the cast before changing anything, and it reproduces exactly: `precisionTimestamp(1234567, 6)` cast to `precision_timestamp<3>` becomes a `Cast(_, TimestampNTZType)` that evaluates to `1234567`, where the plan asks for `1234000`.

Went with your first shape. `ToSparkType` requires exactly microseconds again and the three literal visits rescale. They no longer convert the type they declare — `ToSparkType.convert` would reject it now — so they build the microsecond type instead, which is what the rescaled value lands in.

That leaves the column half of #1125 where it was, and I think that is right here: making it work means rescaling data, not mapping a type. I have left the `Closes` keyword off the description rather than decide that unilaterally — happy either to close #1125 and file the normalization pass separately, or to leave it open for that half.

The cast case has a test of its own now, so relaxing the type guard later fails instead of truncating silently.

spark/src/main/scala/io/substrait/spark/utils/Util.scala
alexandrefimov: Right that this was the last backstop, and it is gone: precision 0 is a valid literal precision now and gets multiplied by 10^6.

I do not think it can go back here, though. "0" and "unset" are indistinguishable on this side of the wire — `getPrecision()` returns the same `0` either way — so a check here guesses about the producer rather than validating the plan. Where it can be done is core, which already guards the analogous type field, and #1129 is open on exactly that gap for the literals. Better landing there once than approximated per consumer.

Worth noting the old failure was not really loud either: a producer that omitted the field got "Unsupported precision: 0", which reads as "your precision is wrong" rather than "your field is missing".

spark/src/main/scala/io/substrait/spark/utils/Util.scala
alexandrefimov: Taken. `Math.multiplyExact` in the rescale, and `Math.addExact` / `Math.multiplyExact` in the interval literal.

Reproduced both sentinels before changing anything: `toMicroseconds(Long.MaxValue, 0)` returned `-1000000` and `Long.MinValue` returned `0`. There is a test on both now, and on the interval's day multiply at `Int.MaxValue` days.

spark/src/main/scala/io/substrait/spark/utils/Util.scala
alexandrefimov: Taken, close to verbatim — the range message lives on `toMicroseconds`, which is where the 0..6 check now is. The type guard kept a message of its own, since it wants exactly 6 and should say why a literal at a coarser precision still converts.

There is a test on the negative case pinning both the value and the range in the message, since that was the half that read backwards.

spark/src/test/scala/io/substrait/spark/TypesAndLiteralsSuite.scala
alexandrefimov: Both taken. `.dataType` is asserted alongside every value now, which does separate the NTZ and TZ visits. The rejection test covers all three carriers at both the literal and the type level, and `interval_day` runs across the same precision table as the timestamps rather than appearing once outside the loop.

One deliberate difference in that table: `interval_day` uses `subseconds = 0` at precision 0. At P=0 there is no room for a fractional part, so a non-zero value there is out of spec — which is the second gap in #1129.

spark/src/test/scala/io/substrait/spark/TypesAndLiteralsSuite.scala
alexandrefimov: Rebased and ran `./gradlew :spark:spotlessApply`. It made exactly the two changes you described — `io.substrait.`type`.TypeCreator` ahead of `io.substrait.expression....`, plus two Scaladoc rewraps I had not spotted — and `:spark:spotlessCheck` is green.

The part I had already done by hand, moving `io.substrait.spark.utils.Util` into the first group, survived the formatter.

spark/src/main/scala/io/substrait/spark/utils/Util.scala
alexandrefimov: Taken. Seven-entry `Array[Long]` indexed by the exponent, declared beside the constants so the domain is visible where the bound is.

You are right that the loop's negative-exponent behaviour was the worse half: it returned `1`, so a missing guard would have produced an unscaled value rather than failing.

alexandrefimov:
All taken except one, with the details in the threads. Every claim in the review reproduced before I touched anything — the cast evaluating to `1234567` where the plan asks for `1234000`, both overflow sentinels, and the backwards message on a negative precision.

Two notes on the pair that could not be anchored to the diff.

`DialectGenerator` reads `Util.MICROSECOND_PRECISION` now and `spark_dialect.yaml` is regenerated to 6. One consequence worth flagging: `core`'s `SparkDialectParseTest.parsesPrecisionTypes` asserts that number off the copied file, so it moves with it and this PR now touches `core/src/test`. Say the word if you would rather have the dialect change stand on its own. Separately from the framing — the 9 was already wrong before this PR, since the guard has always required exactly 6, so it is a divergence this fixes rather than one it creates.

On the reverse direction: with the type conversions back to requiring exactly 6, the asymmetry mostly closes by itself. A coarse precision now enters only on a literal, is rescaled there, and leaves as 6, which is value-lossless; a column type at a coarse precision does not enter at all, so nothing widens. The decision that remains is the one in the first thread — whether the column case should be made to work through a normalization pass.

Description trimmed, and the `Closes` keyword taken off it for the reason in that thread.

spark/src/main/scala/io/substrait/spark/utils/Util.scala
nielspardon: Agreed with where you landed. Precision 0 is a legitimate value per spec, and the unset-vs-explicit-0 ambiguity is a proto field-presence problem that the adapter has no way to see, so rejecting 0 here would trade a rare silent error for a common false one. The test naming the wire reachability is the right amount of record to leave.

The residue belongs in `core`, near #1129 — guarding the literal the way `ProtoTypeConverter` already guards the type. Not this PR.

nielspardon:
Re-reviewed at `2ef26890` — all threads resolved.

Two of the fixes are better than what I suggested: taking `Math.multiplyExact` to the pre-existing day multiply as well, and dropping `expr.getType()` from the literal visits entirely rather than keeping it. The latter also sidesteps #1134, where `PrecisionTimeLiteral.getType()` derives the wrong type class. The precision-6 round-trip asymmetry I raised in the review body is now moot by construction rather than patched, since nothing coarser than 6 can enter as a type.

Checked locally as well as in CI, since the shared Scala source compiles against two Scala versions: the suite is green on both `spark-3.5_2.12` and `spark-4.0_2.13`, `DialectSuite` confirms the generated dialect still matches the checked-in yaml, and `:spark:spotlessCheck` is clean.

nielspardon (approved):
LGTM

PR #1140, opened 2026-08-20
fix(spark): report interval_year overflow instead of wrapping

### LINKED ISSUE #1138: spark: interval_year literal conversion silently wraps on Int overflow
`ToSparkExpression.visit(SExpression.IntervalYearLiteral, ...)` (spark/src/main/scala/io/substrait/spark/expression/ToSparkExpression.scala:152) flattens years and months with unchecked `Int` arithmetic:

```scala
val months = expr.years() * 12 + expr.months()
```

`Expression.IntervalYearLiteral` declares `int years()` and `int months()` (core/src/main/java/io/substrait/expression/Expression.java:580, 587) and the proto fields are `int32`, so there is no `Long` operand to force widening — unlike the day-time case, this really is `Int` arithmetic throughout. The product overflows above `years = 178,956,970`; at `178,956,971` it wraps to `-2,147,483,644` months, turning a large positive interval into a large negative one. The wrapped result is still a valid `Int`, so nothing downstream can tell it from an intended value.

## Why it is worth a look now

#1128 hardened the sibling path eight lines above. `visit(SExpression.IntervalDayLiteral, ...)` now reads:

```scala
val seconds =
  Math.addExact(Math.multiplyExact(expr.days().toLong, Util.SECONDS_PER_DAY), expr.seconds().toLong)
```

So the two adjacent literal conversions in the same visitor now take opposite stances on the same class of arithmetic. That inconsistency is the substance of this issue more than the overflow itself.

## Honest severity

This is not reachable from a spec-conformant plan. The spec bounds `interval_year` to "[-10,000..10,000] years with month precision (= [-120,000..120,000] months)" and constrains each component independently, as quoted in #1129. Overflow needs roughly 17,900x the legal maximum, so it takes a producer that is already well outside the spec.

That makes #1129 the primary fix: a `@Value.Check` enforcing the per-component 10,000-year bound in `core` would make this unreachable rather than merely loud. This issue is defense-in-depth for the case where a plan reaches the converter unvalidated, and it is worth filing separately mainly because #1129's own open questions flag it as wanting a breaking release, so it may not land soon.

If you would rather carry this under #1129 than track it on its own, closing it as covered there is reasonable.

## Fix

```scala
val months = Math.addExact(Math.multiplyExact(expr.years(), 12), expr.months())
```

`Math.multiplyExact(int, int)` and `addExact(int, int)` both have `Int` overloads, so no widening is needed and the literal stays an `Int` as `YearMonthIntervalType` requires. A test asserting `ArithmeticException` for `intervalYear(false, Int.MaxValue, 0)` would match the pattern #1128 established for the day-time path.

alexandrefimov:
`ToSparkExpression.visit(IntervalYearLiteral)` flattened years and months with unchecked `Int` arithmetic, and Spark's physical type for `YearMonthInterval` is a single months `Int`. A large year count outran that carrier silently: 178,956,971 years came out as −2,147,483,644 months, and `Int.MinValue` years as exactly 0 — a zero-length interval rather than a sign flip. It throws `ArithmeticException` now.

Only the total is significant, so the check sits on it rather than on the intermediate product: 178,956,971 years with −12 months is 2,147,483,640 months and still converts. The bound is the carrier's, not the spec's much tighter 10,000-year one, so reaching it takes a producer already far past that.

Closes #1138.

### DIFF
--- a/spark/src/main/scala/io/substrait/spark/expression/ToSparkExpression.scala
+++ b/spark/src/main/scala/io/substrait/spark/expression/ToSparkExpression.scala
@@ -148,10 +148,12 @@ class ToSparkExpression(
   override def visit(
       expr: SExpression.IntervalYearLiteral,
       context: EmptyVisitationContext): Literal = {
-    // Spark uses a single months Int as the "physical" type for YearMonthInterval. Both operands
-    // are int32 on the wire, so unlike the day-time case nothing widens the arithmetic and a large
-    // year count wraps a positive interval into a negative one; report it instead.
-    val months = Math.addExact(Math.multiplyExact(expr.years(), 12), expr.months())
+    // Spark uses a single months Int as the "physical" type for YearMonthInterval, and both
+    // components are int32, so the flattened total can outrun that carrier: a large year count
+    // used to wrap a positive interval into a negative one. Only the total is significant, so
+    // the check belongs on it rather than on the intermediate product.
+    val months =
+      Math.toIntExact(expr.years().toLong * Util.MONTHS_PER_YEAR + expr.months())
     Literal(months, ToSparkType.convert(expr.getType))
   }
 
--- a/spark/src/main/scala/io/substrait/spark/expression/ToSparkExpression.scala
+++ b/spark/src/main/scala/io/substrait/spark/expression/ToSparkExpression.scala
@@ -152,8 +152,7 @@ class ToSparkExpression(
     // components are int32, so the flattened total can outrun that carrier: a large year count
     // used to wrap a positive interval into a negative one. Only the total is significant, so
     // the check belongs on it rather than on the intermediate product.
-    val months =
-      Math.toIntExact(expr.years().toLong * Util.MONTHS_PER_YEAR + expr.months())
+    val months = Math.toIntExact(expr.years() * Util.MONTHS_PER_YEAR + expr.months())
     Literal(months, ToSparkType.convert(expr.getType))
   }
 
--- a/spark/src/main/scala/io/substrait/spark/expression/ToSparkExpression.scala
+++ b/spark/src/main/scala/io/substrait/spark/expression/ToSparkExpression.scala
@@ -148,8 +148,10 @@ class ToSparkExpression(
   override def visit(
       expr: SExpression.IntervalYearLiteral,
       context: EmptyVisitationContext): Literal = {
-    // Spark uses a single months Int as the "physical" type for YearMonthInterval
-    val months = expr.years() * 12 + expr.months()
+    // Spark uses a single months Int as the "physical" type for YearMonthInterval. Both operands
+    // are int32 on the wire, so unlike the day-time case nothing widens the arithmetic and a large
+    // year count wraps a positive interval into a negative one; report it instead.
+    val months = Math.addExact(Math.multiplyExact(expr.years(), 12), expr.months())
     Literal(months, ToSparkType.convert(expr.getType))
   }
 
--- a/spark/src/main/scala/io/substrait/spark/utils/Util.scala
+++ b/spark/src/main/scala/io/substrait/spark/utils/Util.scala
@@ -23,6 +23,7 @@ object Util {
 
   val SECONDS_PER_DAY: Long = 24 * 60 * 60
   val MICROS_PER_SECOND: Long = 1000 * 1000
+  val MONTHS_PER_YEAR: Int = 12
   val MICROSECOND_PRECISION = 6 // for PrecisionTimestamp(TZ) and IntervalDay types
 
   /** Indexed by exponent, which [[toMicroseconds]] bounds to 0..MICROSECOND_PRECISION. */
--- a/spark/src/main/scala/io/substrait/spark/utils/Util.scala
+++ b/spark/src/main/scala/io/substrait/spark/utils/Util.scala
@@ -23,7 +23,7 @@ object Util {
 
   val SECONDS_PER_DAY: Long = 24 * 60 * 60
   val MICROS_PER_SECOND: Long = 1000 * 1000
-  val MONTHS_PER_YEAR: Int = 12
+  val MONTHS_PER_YEAR: Long = 12
   val MICROSECOND_PRECISION = 6 // for PrecisionTimestamp(TZ) and IntervalDay types
 
   /** Indexed by exponent, which [[toMicroseconds]] bounds to 0..MICROSECOND_PRECISION. */
--- a/spark/src/test/scala/io/substrait/spark/TypesAndLiteralsSuite.scala
+++ b/spark/src/test/scala/io/substrait/spark/TypesAndLiteralsSuite.scala
@@ -308,17 +308,23 @@ class TypesAndLiteralsSuite extends SparkFunSuite {
   }
 
   test("a year-month interval reports overflow instead of wrapping") {
-    // years and months are int32 on the wire and Spark's physical type is a months Int, so there
-    // is no Long operand to widen this the way the day-time case is widened. Far outside the
-    // spec's 10,000-year bound, but it used to turn a large positive interval into a negative one:
-    // 178,956,971 years came out as -2,147,483,644 months, and Int.MaxValue years as -12.
+    // years and months are both int32 and Spark's physical type is a months Int, so the flattened
+    // total can outrun the carrier: 178,956,971 years used to come out as -2,147,483,644 months,
+    // and Int.MinValue years as exactly 0 — a zero-length interval. This is the carrier's bound,
+    // not the spec's much tighter 10,000-year one, so it takes a producer already far past that.
     intercept[ArithmeticException](
       sparkLiteral(ExpressionCreator.intervalYear(false, 178956971, 0)))
     intercept[ArithmeticException](
-      sparkLiteral(ExpressionCreator.intervalYear(false, Int.MaxValue, 0)))
+      sparkLiteral(ExpressionCreator.intervalYear(false, 178956970, 8)))
+    intercept[ArithmeticException](
+      sparkLiteral(ExpressionCreator.intervalYear(false, Int.MinValue, 0)))
 
-    // The spec's maximum still converts, and the months carry through.
-    val atTheBound = sparkLiteral(ExpressionCreator.intervalYear(false, 10000, 0))
+    // Only the total is significant, so an intermediate past Int.MaxValue is not itself an error.
+    val mixedSigns = sparkLiteral(ExpressionCreator.intervalYear(false, 178956971, -12))
+    assert(mixedSigns.value === 2147483640)
+
+    // The spec's maximum still converts, and the months component carries through.
+    val atTheBound = sparkLiteral(ExpressionCreator.intervalYear(false, 9999, 12))
     assert(atTheBound.value === 120000)
     assert(atTheBound.dataType === YearMonthIntervalType.DEFAULT)
   }
--- a/spark/src/test/scala/io/substrait/spark/TypesAndLiteralsSuite.scala
+++ b/spark/src/test/scala/io/substrait/spark/TypesAndLiteralsSuite.scala
@@ -307,6 +307,22 @@ class TypesAndLiteralsSuite extends SparkFunSuite {
       sparkLiteral(ExpressionCreator.intervalDay(false, Int.MaxValue, 0, 0L, 6)))
   }
 
+  test("a year-month interval reports overflow instead of wrapping") {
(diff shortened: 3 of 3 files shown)


nielspardon:
I started looking at this one already. Just trying to make sense of all findings Claude produced before confronting you with them.

spark/src/main/scala/io/substrait/spark/expression/ToSparkExpression.scala
```diff
@@ -148,8 +148,10 @@ class ToSparkExpression(
   override def visit(
       expr: SExpression.IntervalYearLiteral,
       context: EmptyVisitationContext): Literal = {
-    // Spark uses a single months Int as the "physical" type for YearMonthInterval
-    val months = expr.years() * 12 + expr.months()
+    // Spark uses a single months Int as the "physical" type for YearMonthInterval. Both operands
+    // are int32 on the wire, so unlike the day-time case nothing widens the arithmetic and a large
+    // year count wraps a positive interval into a negative one; report it instead.
+    val months = Math.addExact(Math.multiplyExact(expr.years(), 12), expr.months())
```
nielspardon: Widen once and check at the carrier boundary, which is where the value actually has to fit.

This also drops the `unlike the day-time case` contrast, which does not hold: `IntervalDayLiteral.days()` and `.seconds()` are `int` too, and that is precisely why the path above writes `.toLong`. The real difference is the destination physical type, not the wire.

```suggestion
    // Spark uses a single months Int as the "physical" type for YearMonthInterval, and both
    // components are int32, so the flattened total can outrun that carrier: a large year count
    // used to wrap a positive interval into a negative one. Only the total is significant, so
    // the check belongs on it rather than on the intermediate product.
    val months = Math.toIntExact(expr.years().toLong * 12 + expr.months())
```

spark/src/test/scala/io/substrait/spark/TypesAndLiteralsSuite.scala
```diff
@@ -307,6 +307,22 @@ class TypesAndLiteralsSuite extends SparkFunSuite {
...
 
+  test("a year-month interval reports overflow instead of wrapping") {
+    // years and months are int32 on the wire and Spark's physical type is a months Int, so there
+    // is no Long operand to widen this the way the day-time case is widened. Far outside the
+    // spec's 10,000-year bound, but it used to turn a large positive interval into a negative one:
+    // 178,956,971 years came out as -2,147,483,644 months, and Int.MaxValue years as -12.
+    intercept[ArithmeticException](
+      sparkLiteral(ExpressionCreator.intervalYear(false, 178956971, 0)))
+    intercept[ArithmeticException](
+      sparkLiteral(ExpressionCreator.intervalYear(false, Int.MaxValue, 0)))
```
nielspardon: Swap the `Int.MaxValue` case, which trips the same multiply as the one above it, for the boundary only the addition crosses, and pin the mixed-sign total that must still convert.

`Int.MinValue` years is worth its own case because it wrapped to exactly `0` — a zero-length interval, which is harder to spot than a sign flip.

```suggestion
    intercept[ArithmeticException](
      sparkLiteral(ExpressionCreator.intervalYear(false, 178956971, 0)))
    intercept[ArithmeticException](
      sparkLiteral(ExpressionCreator.intervalYear(false, 178956970, 8)))
    intercept[ArithmeticException](
      sparkLiteral(ExpressionCreator.intervalYear(false, Int.MinValue, 0)))

    // Only the total is significant, so an intermediate past Int.MaxValue is not itself an error.
    val mixedSigns = sparkLiteral(ExpressionCreator.intervalYear(false, 178956971, -12))
    assert(mixedSigns.value === 2147483640)
```

spark/src/test/scala/io/substrait/spark/TypesAndLiteralsSuite.scala
```diff
@@ -307,6 +307,22 @@ class TypesAndLiteralsSuite extends SparkFunSuite {
       sparkLiteral(ExpressionCreator.intervalDay(false, Int.MaxValue, 0, 0L, 6)))
   }
 
+  test("a year-month interval reports overflow instead of wrapping") {
+    // years and months are int32 on the wire and Spark's physical type is a months Int, so there
+    // is no Long operand to widen this the way the day-time case is widened. Far outside the
+    // spec's 10,000-year bound, but it used to turn a large positive interval into a negative one:
+    // 178,956,971 years came out as -2,147,483,644 months, and Int.MaxValue years as -12.
```
nielspardon: Say that the bound being enforced is the carrier's, not the spec's. As written, `Far outside the spec's 10,000-year bound` alongside `atTheBound` below reads as if the spec's limit were what fires, but the throw only starts at ~178.9M years, so `intervalYear(false, 100000, 500000)` still converts silently.

```suggestion
    // years and months are both int32 and Spark's physical type is a months Int, so the flattened
    // total can outrun the carrier: 178,956,971 years used to come out as -2,147,483,644 months,
    // and Int.MinValue years as exactly 0 — a zero-length interval. This is the carrier's bound,
    // not the spec's much tighter 10,000-year one, so it takes a producer already far past that.
```

spark/src/test/scala/io/substrait/spark/TypesAndLiteralsSuite.scala
```diff
@@ -307,6 +307,22 @@ class TypesAndLiteralsSuite extends SparkFunSuite {
...
+    intercept[ArithmeticException](
+      sparkLiteral(ExpressionCreator.intervalYear(false, 178956971, 0)))
+    intercept[ArithmeticException](
+      sparkLiteral(ExpressionCreator.intervalYear(false, Int.MaxValue, 0)))
+
+    // The spec's maximum still converts, and the months carry through.
+    val atTheBound = sparkLiteral(ExpressionCreator.intervalYear(false, 10000, 0))
+    assert(atTheBound.value === 120000)
+    assert(atTheBound.dataType === YearMonthIntervalType.DEFAULT)
```
nielspardon: Spell the bound with a nonzero months component so the assertion earns the comment — `10000, 0` reaches 120000 from the years term alone, so deleting `+ expr.months()` entirely would still pass all three lines.

```suggestion
    // The spec's maximum still converts, and the months component carries through.
    val atTheBound = sparkLiteral(ExpressionCreator.intervalYear(false, 9999, 12))
    assert(atTheBound.value === 120000)
    assert(atTheBound.dataType === YearMonthIntervalType.DEFAULT)
```

spark/src/main/scala/io/substrait/spark/expression/ToSparkExpression.scala
nielspardon: Non-blocking, and both points are inherited rather than introduced here: the escaping exception says only `integer overflow`, where this module's other refusals name the value and the legal range (`Util.toMicroseconds`), and the bare `12` sits next to a day-time sibling that names `Util.SECONDS_PER_DAY` and `Util.MICROS_PER_SECOND`. Spark exports `DateTimeConstants.MONTHS_PER_YEAR` if you would rather not add a `Util` constant for it. The day-time path already throws bare, so this is a raise-the-bar note.

nielspardon (commented):
Guard the flattened total rather than the intermediate product: `Math.multiplyExact(expr.years(), 12)` throws on year/month pairs whose total months does fit an `Int`, and for `interval_year` the spec makes only the total significant — it treats `1001y -12000m` as a legal spelling of `1y 0m`. `(178956971, -12)`, `(178956971, -5)` and `(-178956971, 12)` all have totals inside `Int` range and converted correctly before this change; they now throw.

The first two suggestions are one edit — the second needs the first; the rest is comment and coverage polish. Filed #1235, #1236, #1237 and #1238 for siblings I hit while reading this, the first of which is the same wrap still live in isthmus.

spark/src/main/scala/io/substrait/spark/expression/ToSparkExpression.scala
alexandrefimov: Taken as written.

spark/src/test/scala/io/substrait/spark/TypesAndLiteralsSuite.scala
alexandrefimov: Taken as written.

spark/src/test/scala/io/substrait/spark/TypesAndLiteralsSuite.scala
alexandrefimov: Taken as written.

spark/src/test/scala/io/substrait/spark/TypesAndLiteralsSuite.scala
alexandrefimov: Taken as written.

spark/src/main/scala/io/substrait/spark/expression/ToSparkExpression.scala
alexandrefimov: Took the constant, as `Util.MONTHS_PER_YEAR` beside `SECONDS_PER_DAY` and `MICROS_PER_SECOND` — the module has no other `DateTimeConstants` use, and those two are where the day-time path reads its own.

Left the message alone. Naming the value and the range here and not on the day-time path would split the two literals again on a second axis, which is the thing this PR is closing; worth raising for both together.

spark/src/main/scala/io/substrait/spark/utils/Util.scala
```diff
@@ -23,6 +23,7 @@ object Util {
 
   val SECONDS_PER_DAY: Long = 24 * 60 * 60
   val MICROS_PER_SECOND: Long = 1000 * 1000
+  val MONTHS_PER_YEAR: Int = 12
```
nielspardon: Make it `Long`, like the two constants above it. Then `expr.years() * Util.MONTHS_PER_YEAR` is `Long` arithmetic unconditionally and the call site has nothing to remember; as `Int`, the guard is one dropped `.toLong` away from wrapping again. The new test catches that edit today — the type stops it being expressible.

```suggestion
  val MONTHS_PER_YEAR: Long = 12
```

spark/src/main/scala/io/substrait/spark/expression/ToSparkExpression.scala
```diff
@@ -148,8 +148,12 @@ class ToSparkExpression(
...
       expr: SExpression.IntervalYearLiteral,
       context: EmptyVisitationContext): Literal = {
-    // Spark uses a single months Int as the "physical" type for YearMonthInterval
-    val months = expr.years() * 12 + expr.months()
+    // Spark uses a single months Int as the "physical" type for YearMonthInterval, and both
+    // components are int32, so the flattened total can outrun that carrier: a large year count
+    // used to wrap a positive interval into a negative one. Only the total is significant, so
+    // the check belongs on it rather than on the intermediate product.
+    val months =
+      Math.toIntExact(expr.years().toLong * Util.MONTHS_PER_YEAR + expr.months())
```
nielspardon: The companion edit. At 89 columns scalafmt keeps it on one line.

```suggestion
    val months = Math.toIntExact(expr.years() * Util.MONTHS_PER_YEAR + expr.months())
```

nielspardon (commented):
All five taken, and the check now sits where the value actually has to fit. The mutations that survived round 1 no longer do — dropping the months term, dropping the widening, or relaxing the exactness each fails the suite now, on all three variants.

Agreed on leaving both overflow messages bare: naming the value and the range on the year-month path alone would re-split the two literals on a second axis, which is the thing this PR closes. Filed #1271 for the pair.

One non-blocking suggestion left, on the type of the new constant.

spark/src/main/scala/io/substrait/spark/expression/ToSparkExpression.scala
nielspardon: Agreed, and I filed #1271 so the two paths get it together — that also covers the day-time side, which overflows from about 106.7M days, inside `int32` and so reachable in its own right.

`Util.MONTHS_PER_YEAR` is the right home for the constant.

spark/src/main/scala/io/substrait/spark/utils/Util.scala
alexandrefimov: Taken as written.

spark/src/main/scala/io/substrait/spark/expression/ToSparkExpression.scala
alexandrefimov: Taken as written.

nielspardon (approved):

Issue #1139, opened 2026-08-20
core: return-type derivation supports only decimal and wildcard shapes, so resolveType throws for 86 variants


alexandrefimov:
`SimpleExtension.Function.resolveType` evaluates a declaration's return expression against the actual argument types and applies its nullability policy. `FunctionBindingResolver.deriveOutputType` and `validateOutputType` are built on it, and it is what a consumer would use to check a plan's declared output type against the function it invokes. It cannot evaluate most declared return expressions: `TypeExpressionEvaluator` binds two things — a numbered wildcard and `Decimal<P,S>` — and evaluates the same two, and every other shape reaches the throwing base of `TypeExpressionVisitor` as `Cannot evaluate return-type expression: …`. I would like to take this.

Running `resolveType` over every declared variant whose arguments are all value arguments, against the extensions pinned in `gradle/libs.versions.toml` (spec v0.101.0): 306 of 377 scalar variants evaluate, 70 of 83 aggregate, 81 of 94 window. 86 of the 97 failures are unsupported return *shapes*; the other 11 are an artefact of the census, which passes a placeholder `i64` for every argument and so cannot bind a wildcard.

| shape | count | example |
| --- | --- | --- |
| `varchar<L1>` | 24 | `concat:vchar` |
| return program | 15 | `add:dec_dec` |
| `precision_timestamp<P>` | 11 | `add:pts_iyear` |
| `precision_timestamp_tz<P>` | 10 | `add:ptstz_iyear_str` |
| `interval_day<P>` | 9 | `multiply:i8_iday` |
| `fixedchar<L1>` | 7 | `reverse:fchar` |
| `list<any2>` | 6 | `transform:list_func` |
| `precision_time<P>` | 4 | `min:pt` |

`add:dec_dec` is worth singling out: decimal addition, the case the `Decimal<P,S>` support exists for, declares its return as a return program and so is not derivable either.

## Why it matters

`ScalarFunctionVariant.validateOutputType` is a `return;` with a TODO, so nothing checks a plan's declared output type against its declaration today. Turning it on needs a derivation that covers more than two shapes.

And #1117: Isthmus takes a function's Substrait output type from the Calcite node rather than from the declaration, so it emits `subtract(date, interval_day<P>)` with a `date` output where the spec declares `precision_timestamp<P>`. Consulting the declaration is the fix, and it throws for exactly that signature.

## Scope

[...]

PR #1141, opened 2026-08-20
feat(core): derive return types parameterized by a length or precision

alexandrefimov:
`TypeExpressionEvaluator` bound and evaluated two shapes — a numbered wildcard and `DECIMAL<P,S>` — and every other parameterized class reached the throwing base of `TypeExpressionVisitor`. So `resolveType`, which `FunctionBindingResolver.deriveOutputType` and `validateOutputType` are built on, could not derive the output type of 94 of the 602 variants the standard extensions ship: 320 of 396 scalar answered, 86 of 95 aggregate, 102 of 111 window. 71 of those derive now, and 23 are left, every one of them a `list` return or a multi-line return program.

The remaining classes whose parameter is an integer to substitute now bind and evaluate the way `DECIMAL` already did: `varchar`, `fixedchar`, `fixedbinary`, the three `precision_*` types, `interval_day` and `interval_compound`. The last two are there for symmetry; no standard extension declares them parameterized at all, as an argument or as a return.

What this changes in practice is `AggregateConversion.FunctionBindingValidation.EXTENSION_DECLARATION`, whose own documentation said it was not adoptable for plans using these functions. It still is not adoptable for the two groups below, and that Javadoc was wrong twice over: it filed `assume_timezone` and the `strptime_*` family under the parameterized classes when they fail on a return program, and everything else it listed was a scalar declaration the aggregate path never sees. It names the two standard aggregates that do reach that mode now — `quantile` and `avg` over a decimal — since that doc exists to tell a reader whether the mode is adoptable. Which shipped variants carry the shapes that stay underivable is left to `ParameterizedReturnTypeTest` instead, so no Javadoc goes stale on a `substrait-packaging` bump.

`list` returns and return programs are unchanged, and are worth their own issues rather than being folded in here: #1241 and #1242. A `list` return fails whatever its element, because the evaluator does not descend into a container: an element parameter it would otherwise substitute, as in `list<varchar<L1>>`, is as far out of reach as an element type to evaluate.

Isthmus derives the same declared return a second time, in `SimpleExtensionToSqlOperator`, from the first operand; core throwing for these classes is why the two were never compared, and with core answering they differ for eight scalar variants in the catalog — `multiply:i8_iday` derives `interval_day<P>` where the first operand is `i8`. Nothing goes red: under `PLAN_OUTPUT` that inference only decides whether the plan's type travels on a wrapper. #1240 has the rest.

Part of #1139. The inconsistent-`P` question raised there is not settled by this: binding now makes `bindType`'s existing `Inconsistent binding` reachable for a shared parameter, which is what the tests pin, but nothing turns validation on by default. Binding also fails closed on a declared and actual pair whose shapes do not match, throwing an `UnsupportedOperationException` naming both rather than binding nothing and saying nothing: otherwise `concat:vchar(varchar(20), string)` would have started deriving `varchar(20)` here, since isthmus lets a `string` operand through against a declared `varchar<L1>`, where main throws. List, map, struct and function declarations are exempt, because binding never descends into them and refusing them would reject calls that resolve today — `cardinality:list` and `all_match:list_func` among them.

### DIFF
--- a/core/src/main/java/io/substrait/type/TypeExpressionEvaluator.java
+++ b/core/src/main/java/io/substrait/type/TypeExpressionEvaluator.java
@@ -40,7 +40,7 @@ import java.util.OptionalInt;
  * assume_timezone:date_str_i8} and the {@code strptime_*} family. Among the standard aggregates,
  * {@code quantile} cannot be derived at all: its declared return {@code LIST?<any>} uses a plain
  * {@code any}, which carries no identity to bind. Both lists are pinned against the catalog by
- * {@code ParameterizedReturnTypeTest}, which is spec v0.101.0 today.
+ * {@code ParameterizedReturnTypeTest}.
  */
 public class TypeExpressionEvaluator {
 
@@ -229,7 +229,31 @@ public class TypeExpressionEvaluator {
             ((ParameterizedType.IntervalCompound) declared).precision().value(),
             ((Type.IntervalCompound) actual).precision(),
             bindNames);
+      } else if (!(declared instanceof Type) && !isContainer(declared)) {
+        // A shape one of the arms above should have taken: the declaration carries a parameter and
+        // the actual type is not the class that would bind it. Binding nothing here would enforce
+        // the shared-parameter rule for some calls and skip it for others.
+        throw new UnsupportedOperationException(
+            String.format(
+                "Cannot bind parameters from declared argument type %s to actual type %s",
+                declared, actual));
       }
+    }
+
+    /**
+     * Whether the declared type holds other types rather than an integer parameter. Binding does
+     * not descend into these, so their parameters bind nothing and a mismatch cannot be told from a
+     * shape this method simply does not reach yet -- unlike the classes above, refusing here would
+     * reject declarations that resolve today without binding anything, such as a {@code list<any1>}
+     * argument to a function returning a concrete type.
+     *
+     * @param declared the declared argument type
+     * @return {@code true} if the type is a list, map or struct declaration
+     */
+    private boolean isContainer(ParameterizedType declared) {
+      return declared instanceof ParameterizedType.ListType
+          || declared instanceof ParameterizedType.Map
+          || declared instanceof ParameterizedType.Struct;
     }
 
     private void bindType(String name, Type actual) {
--- a/core/src/main/java/io/substrait/type/TypeExpressionEvaluator.java
+++ b/core/src/main/java/io/substrait/type/TypeExpressionEvaluator.java
@@ -27,20 +27,21 @@ import java.util.OptionalInt;
  * type classes whose parameter is an integer to substitute: {@code DECIMAL<P,S>}, {@code
  * varchar<L1>}, {@code fixedchar<L1>}, {@code fixedbinary<L1>}, {@code precision_time<P>}, {@code
  * precision_timestamp<P>}, {@code precision_timestamp_tz<P>}, {@code interval_day<P>} and {@code
- * interval_compound<P>}. No standard extension returns a parameterized {@code fixedbinary} or
- * {@code interval_compound} -- those two are supported for symmetry, and pinned against
- * hand-written declarations rather than the catalog.
+ * interval_compound<P>}. No standard extension declares a parameterized {@code fixedbinary} or
+ * {@code interval_compound} at all, as an argument or as a return -- those two are supported for
+ * symmetry, and pinned against hand-written declarations rather than the catalog.
  *
- * <p>What still fails on the standard extension catalog is a return of {@code list<anyN>}, whose
- * parameter is a type to evaluate rather than an integer to substitute ({@code filter}, {@code
- * sort}, {@code transform}, {@code string_split}, {@code regexp_string_split}, {@code
- * regexp_match_substring_all}), and a multi-line return program -- the decimal variants of {@code
- * add}, {@code subtract}, {@code multiply}, {@code divide}, {@code modulus}, {@code ceil}, {@code
- * floor}, {@code round} and the {@code bitwise_*} family, together with {@code
- * assume_timezone:date_str_i8} and the {@code strptime_*} family. Among the standard aggregates,
- * {@code quantile} cannot be derived at all: its declared return {@code LIST?<any>} uses a plain
- * {@code any}, which carries no identity to bind. Both lists are pinned against the catalog by
- * {@code ParameterizedReturnTypeTest}.
+ * <p>A {@code list} return still fails whatever its element, because the evaluator does not descend
+ * into a container -- so an element parameter it would otherwise substitute, as in {@code
+ * list<varchar<L1>>}, is out of reach just as an element type to evaluate is. A multi-line return
+ * program still fails because evaluating one needs integer arithmetic over the bound parameters
+ * rather than substitution. And a plain {@code any} cannot be derived at all: unlike {@code any1}
+ * it names nothing, so there is no identity to bind.
+ *
+ * <p>Which shipped variants those cover is pinned by {@code ParameterizedReturnTypeTest} against
+ * the declarations the catalog ships, and deliberately not repeated here -- the catalog is owned
+ * upstream, so a list of names in this Javadoc would go stale on a {@code substrait-packaging} bump
+ * with nothing to catch it.
  */
 public class TypeExpressionEvaluator {
 
@@ -248,12 +249,13 @@ public class TypeExpressionEvaluator {
      * argument to a function returning a concrete type.
      *
      * @param declared the declared argument type
-     * @return {@code true} if the type is a list, map or struct declaration
+     * @return {@code true} if the type is a list, map, struct or function declaration
      */
     private boolean isContainer(ParameterizedType declared) {
       return declared instanceof ParameterizedType.ListType
           || declared instanceof ParameterizedType.Map
-          || declared instanceof ParameterizedType.Struct;
+          || declared instanceof ParameterizedType.Struct
+          || declared instanceof ParameterizedType.Func;
     }
 
     private void bindType(String name, Type actual) {
--- a/core/src/main/java/io/substrait/type/TypeExpressionEvaluator.java
+++ b/core/src/main/java/io/substrait/type/TypeExpressionEvaluator.java
@@ -27,8 +27,9 @@ import java.util.OptionalInt;
  * type classes whose parameter is an integer to substitute: {@code DECIMAL<P,S>}, {@code
  * varchar<L1>}, {@code fixedchar<L1>}, {@code fixedbinary<L1>}, {@code precision_time<P>}, {@code
  * precision_timestamp<P>}, {@code precision_timestamp_tz<P>}, {@code interval_day<P>} and {@code
- * interval_compound<P>}. The last two of those are supported for symmetry; no standard extension
- * declares them parameterized today.
+ * interval_compound<P>}. No standard extension returns a parameterized {@code fixedbinary} or
+ * {@code interval_compound} -- those two are supported for symmetry, and pinned against
+ * hand-written declarations rather than the catalog.
  *
  * <p>What still fails on the standard extension catalog is a return of {@code list<anyN>}, whose
  * parameter is a type to evaluate rather than an integer to substitute ({@code filter}, {@code
--- a/core/src/main/java/io/substrait/type/TypeExpressionEvaluator.java
+++ b/core/src/main/java/io/substrait/type/TypeExpressionEvaluator.java
@@ -33,11 +33,13 @@ import java.util.OptionalInt;
  * <p>What still fails on the standard extension catalog is a return of {@code list<anyN>}, whose
  * parameter is a type to evaluate rather than an integer to substitute ({@code filter}, {@code
  * sort}, {@code transform}, {@code string_split}, {@code regexp_string_split}, {@code
- * regexp_match_substring_all}), and a multi-line return program ({@code add}, {@code subtract},
- * {@code multiply}, {@code divide}, {@code modulus}, {@code ceil}, {@code floor}, {@code round},
- * the {@code bitwise_*} family, {@code assume_timezone} and {@code strptime_*}). Among the standard
- * aggregates, {@code quantile} cannot be derived at all: its declared return {@code LIST?<any>}
- * uses a plain {@code any}, which carries no identity to bind (spec v0.101.0).
+ * regexp_match_substring_all}), and a multi-line return program -- the decimal variants of {@code
+ * add}, {@code subtract}, {@code multiply}, {@code divide}, {@code modulus}, {@code ceil}, {@code
+ * floor}, {@code round} and the {@code bitwise_*} family, together with {@code
+ * assume_timezone:date_str_i8} and the {@code strptime_*} family. Among the standard aggregates,
+ * {@code quantile} cannot be derived at all: its declared return {@code LIST?<any>} uses a plain
+ * {@code any}, which carries no identity to bind. Both lists are pinned against the catalog by
+ * {@code ParameterizedReturnTypeTest}, which is spec v0.101.0 today.
  */
 public class TypeExpressionEvaluator {
 
--- a/core/src/main/java/io/substrait/type/TypeExpressionEvaluator.java
+++ b/core/src/main/java/io/substrait/type/TypeExpressionEvaluator.java
@@ -23,15 +23,21 @@ import java.util.OptionalInt;
  * UnsupportedOperationException}. The evaluator never falls back to a caller-supplied type: an
  * unresolved expression is an error, not a default.
  *
- * <p>Supported shapes are concrete types, numbered wildcards ({@code any1}) and parameterized
- * decimals ({@code DECIMAL<P,S>}). What actually fails on the standard extension catalog is the
- * other parameterized type classes — {@code varchar<L1>}, {@code fixedchar<L1>}, {@code
- * precision_time<P>}, {@code precision_timestamp<P>}, {@code precision_timestamp_tz<P>}, {@code
- * interval_day<P>}, {@code list<anyN>}, parameterized structs — and multi-line return programs;
- * {@code concat}, {@code concat_ws}, {@code assume_timezone} and the {@code strptime_*} family are
- * all rejected today. Among the standard aggregates, {@code quantile} is the one whose output type
- * cannot be derived at all: its declared return {@code LIST?<any>} uses a plain {@code any}, which
- * carries no identity to bind (spec v0.99.0).
+ * <p>Supported shapes are concrete types, numbered wildcards ({@code any1}), and the parameterized
+ * type classes whose parameter is an integer to substitute: {@code DECIMAL<P,S>}, {@code
+ * varchar<L1>}, {@code fixedchar<L1>}, {@code fixedbinary<L1>}, {@code precision_time<P>}, {@code
+ * precision_timestamp<P>}, {@code precision_timestamp_tz<P>}, {@code interval_day<P>} and {@code
+ * interval_compound<P>}. The last two of those are supported for symmetry; no standard extension
+ * declares them parameterized today.
+ *
+ * <p>What still fails on the standard extension catalog is a return of {@code list<anyN>}, whose
+ * parameter is a type to evaluate rather than an integer to substitute ({@code filter}, {@code
+ * sort}, {@code transform}, {@code string_split}, {@code regexp_string_split}, {@code
+ * regexp_match_substring_all}), and a multi-line return program ({@code add}, {@code subtract},
+ * {@code multiply}, {@code divide}, {@code modulus}, {@code ceil}, {@code floor}, {@code round},
+ * the {@code bitwise_*} family, {@code assume_timezone} and {@code strptime_*}). Among the standard
+ * aggregates, {@code quantile} cannot be derived at all: its declared return {@code LIST?<any>}
+ * uses a plain {@code any}, which carries no identity to bind (spec v0.101.0).
  */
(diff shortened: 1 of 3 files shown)


nielspardon:
I started looking at this one already. Just trying to make sense of all findings Claude produced before confronting you with them.

alexandrefimov:
This one is `core` rather than isthmus, so if it would help to bring in another pair of eyes, that works for me -- happy to leave it with you otherwise. No urgency; I've slowed down on opening new PRs anyway.

isthmus/src/main/java/io/substrait/isthmus/AggregateConversion.java
```diff
@@ -32,14 +32,14 @@ public enum FunctionBindingValidation {
      *
      * <p>Type derivation is fail-closed: a function whose return expression the derivation does not
      * yet support is rejected rather than assumed valid, so this mode is not adoptable for plans
-     * that use such functions. On the standard extension catalog the unsupported shapes are the
-     * parameterized type classes other than decimal — {@code varchar<L1>}, {@code fixedchar<L1>},
-     * {@code precision_time<P>}, {@code precision_timestamp<P>}, {@code precision_timestamp_tz<P>},
-     * {@code interval_day<P>}, {@code list<anyN>}, parameterized structs — and multi-line return
-     * programs; for example {@code concat}, {@code concat_ws}, {@code assume_timezone} and the
-     * {@code strptime_*} family are rejected today. {@code quantile}'s output type cannot be
+     * that use such functions. On the standard extension catalog what remains unsupported is a
```
nielspardon: Keep a parameterized struct in this list — `avg:dec` still can't derive. Its `intermediate: STRUCT<DECIMAL<38,S>,i64>` parses to a `ParameterizedType.Struct`, which has no `visit` override, and because `consumesIntermediateState()` includes `UNSPECIFIED` an ordinary decimal `avg` measure is still rejected in this mode — so of everything the list names, that is the one case that can actually reach it. The other names are all `scalar_functions` declarations, which `FunctionBindingResolver.validate(ResolvedAggregateBinding, …)` never sees. Filed as #1239.

core/src/main/java/io/substrait/type/TypeExpressionEvaluator.java
```diff
@@ -173,6 +182,53 @@ private void bind(ParameterizedType declared, Type actual, boolean bindNames) {
...
+        bindInteger(
+            ((ParameterizedType.IntervalDay) declared).precision().value(),
+            ((Type.IntervalDay) actual).precision(),
+            bindNames);
+      } else if (declared instanceof ParameterizedType.IntervalCompound
+          && actual instanceof Type.IntervalCompound) {
+        bindInteger(
+            ((ParameterizedType.IntervalCompound) declared).precision().value(),
+            ((Type.IntervalCompound) actual).precision(),
+            bindNames);
```
nielspardon, line 232: Consider a terminal `else` here: when the declared and actual classes don't match, `bind` binds nothing and says nothing, so the shared-parameter rule this PR extends is enforced for some calls and skipped for others. `concat:vchar(varchar(20), string)` now derives `varchar(20)` where main threw, and isthmus reaches it — `IgnoreNullableAndParameters.visit(Type.Str)` deliberately accepts a `string` operand against a declared `varchar<L1>`. `FunctionBindingResolver.typeMatches` fails closed on a shape it doesn't handle; this is the one link in the chain that doesn't.

core/src/test/java/io/substrait/type/ParameterizedReturnTypeTest.java
```diff
@@ -0,0 +1,179 @@
...
+  @Test
+  void oneParameterSharedByTwoArgumentsHasToAgree() {
+    assertEquals(
+        R.intervalDay(3), resolve("add_intervals:iday_iday", R.intervalDay(3), R.intervalDay(3)));
+
+    UnsupportedOperationException e =
+        assertThrows(
+            UnsupportedOperationException.class,
+            () -> resolve("add_intervals:iday_iday", R.intervalDay(3), R.intervalDay(6)));
+    assertTrue(e.getMessage().contains("P"), e.getMessage());
```
nielspardon: Assert on the message this test is named for — `contains("P")` also passes on main, where the same call throws `Cannot evaluate return-type expression: IntervalDay{… value=P}`, so it can't distinguish an inconsistent binding from an unsupported shape.

```suggestion
    assertTrue(
        e.getMessage().contains("Inconsistent binding for type parameter 'P'"), e.getMessage());
```

nielspardon (commented):
Add the argument-side failure to this mode's contract and pin it: derivation can now reject a call whose return needs no derivation at all, because a shared integer parameter binds inconsistently. `lt(precision_timestamp(3), precision_timestamp(6))` and `strpos(varchar(20), varchar(3))` derived `boolean`/`i64` on main and now throw — which is what the spec requires, so the behavior should stay, but 12 of the 13 affected variants are the temporal comparisons that are still an open upstream question (substrait-io/substrait#926, swept in substrait-io/substrait#1185), and `strpos` is only fixed by the v0.103.0 bump in #1191. Worth a line in the PR body too, since nothing in the diff says derivation can now fail on the arguments rather than on the return shape.

core/src/test/java/io/substrait/type/ParameterizedReturnTypeTest.java
```diff
@@ -0,0 +1,179 @@
...
+
+  @Test
+  void intervalDayCarriesItsPrecision() {
+    assertEquals(R.intervalDay(9), resolve("multiply:i8_iday", R.I8, R.intervalDay(9)));
+  }
+
+  @Test
+  void dateMinusIntervalDayDerivesAPrecisionTimestamp() {
+    // The signature #1117 is about: the spec declares precision_timestamp<P>, where P is the
+    // interval's, and nothing but the interval carries it.
```
nielspardon: Drop the issue number — AGENTS.md keeps GitHub references out of source. Naming the variant the assertions already use reads better anyway.

```suggestion
    // subtract(date, interval_day<P>) declares precision_timestamp<P>, and nothing but the
    // interval carries P.
```

isthmus/src/main/java/io/substrait/isthmus/AggregateConversion.java
```diff
@@ -32,14 +32,14 @@ public enum FunctionBindingValidation {
...
+     * return of {@code list<anyN>} ({@code filter}, {@code sort}, {@code transform}, {@code
+     * string_split}, {@code regexp_string_split}, {@code regexp_match_substring_all}) and a
+     * multi-line return program ({@code add}, {@code subtract}, {@code multiply}, {@code divide},
+     * {@code modulus}, {@code ceil}, {@code floor}, {@code round}, the {@code bitwise_*} family,
+     * {@code assume_timezone} and {@code strptime_*}). {@code quantile}'s output type cannot be
      * derived at all: its declared return {@code LIST?<any>} uses a plain {@code any}, which
-     * carries no identity to bind (spec v0.99.0).
+     * carries no identity to bind (spec v0.101.0).
```
nielspardon: Drop this marker rather than bumping it: `quantile`'s `LIST?<any>` has been in the catalog since spec v0.14.0, so neither v0.99.0 nor v0.101.0 was ever a since-claim. `TypeExpressionEvaluator.java:43` needs the same cut for a different reason — main has since moved the pin to 0.102.0, so "which is spec v0.101.0 today" is already stale on the merge result; without the clause the version lives only in `gradle/libs.versions.toml`, where it stays in lockstep. Replacement there is `* {@code ParameterizedReturnTypeTest}.`

```suggestion
     * carries no identity to bind.
```

isthmus/src/main/java/io/substrait/isthmus/AggregateConversion.java
alexandrefimov: Taken. The Javadoc names `avg` over a decimal and `quantile` now, and the scalar declarations are out of it; the catalog-wide list stays on `TypeExpressionEvaluator`, where `theReturnShapesThatAreNotDerivedYet` pins it against the shipped declarations.

core/src/main/java/io/substrait/type/TypeExpressionEvaluator.java
alexandrefimov: Added, with your `concat:vchar` pair as the test. It refuses on the declared type's kind, and list, map and struct are exempt there: refusing those too took `cardinality:list` and `index_in:any_list` with it, the only two casualties in the standard catalog, both declaring a concrete `i64` return that needs no binding at all. Leaving this open -- say the word and it goes blanket, with those two for #1241.

core/src/test/java/io/substrait/type/ParameterizedReturnTypeTest.java
alexandrefimov: Taken as written.

core/src/test/java/io/substrait/type/ParameterizedReturnTypeTest.java
alexandrefimov: Taken as written.

isthmus/src/main/java/io/substrait/isthmus/AggregateConversion.java
alexandrefimov: Both cut, with your replacement line on `TypeExpressionEvaluator`.

core/src/main/java/io/substrait/type/TypeExpressionEvaluator.java
```diff
@@ -173,9 +182,80 @@ private void bind(ParameterizedType declared, Type actual, boolean bindNames) {
...
+     * argument to a function returning a concrete type.
+     *
+     * @param declared the declared argument type
+     * @return {@code true} if the type is a list, map or struct declaration
+     */
+    private boolean isContainer(ParameterizedType declared) {
+      return declared instanceof ParameterizedType.ListType
+          || declared instanceof ParameterizedType.Map
+          || declared instanceof ParameterizedType.Struct;
+    }
```
nielspardon: Add `ParameterizedType.Func` here. `all_match:list_func` and `any_match:list_func` declare a concrete `boolean?` return and resolve today; with the new terminal `else` they throw instead, and the PR body's census does not show it because it excludes exactly the variants that take a function argument. `filter`/`transform` still fail on their `list` return, but now report a binding error that would survive the container work -- and a `Func` plainly holds other types rather than an integer parameter, which is the rationale this method already states.

```suggestion
     * @return {@code true} if the type is a list, map, struct or function declaration
     */
    private boolean isContainer(ParameterizedType declared) {
      return declared instanceof ParameterizedType.ListType
          || declared instanceof ParameterizedType.Map
          || declared instanceof ParameterizedType.Struct
          || declared instanceof ParameterizedType.Func;
    }
```

core/src/main/java/io/substrait/type/TypeExpressionEvaluator.java
```diff
@@ -23,15 +23,24 @@
...
+ * <p>What still fails on the standard extension catalog is a return of {@code list<anyN>}, whose
+ * parameter is a type to evaluate rather than an integer to substitute ({@code filter}, {@code
+ * sort}, {@code transform}, {@code string_split}, {@code regexp_string_split}, {@code
+ * regexp_match_substring_all}), and a multi-line return program -- the decimal variants of {@code
+ * add}, {@code subtract}, {@code multiply}, {@code divide}, {@code modulus}, {@code ceil}, {@code
+ * floor}, {@code round} and the {@code bitwise_*} family, together with {@code
+ * assume_timezone:date_str_i8} and the {@code strptime_*} family. Among the standard aggregates,
+ * {@code quantile} cannot be derived at all: its declared return {@code LIST?<any>} uses a plain
+ * {@code any}, which carries no identity to bind. Both lists are pinned against the catalog by
+ * {@code ParameterizedReturnTypeTest}.
```
nielspardon: Describe the failing shapes here and let the test own the census of names -- the catalog is owned upstream, so a variant list in this Javadoc goes stale on a `substrait-packaging` bump with nothing to catch it. The `list<anyN>` reason is also wrong: three of the six list-returning variants declare `list<varchar<L1>>`, an integer element parameter, so the blocker is not descending into a container rather than the element being a type to evaluate. This inverts the split you described, but `AggregateConversion`'s two aggregate names are worth keeping -- that doc exists to tell a reader whether the mode is adoptable.

```suggestion
 * interval_compound<P>}. No standard extension declares a parameterized {@code fixedbinary} or
 * {@code interval_compound} at all, as an argument or as a return -- those two are supported for
 * symmetry, and pinned against hand-written declarations rather than the catalog.
 *
 * <p>A {@code list} return still fails whatever its element, because the evaluator does not descend
 * into a container -- so an element parameter it would otherwise substitute, as in {@code
 * list<varchar<L1>>}, is out of reach just as an element type to evaluate is. A multi-line return
 * program still fails because evaluating one needs integer arithmetic over the bound parameters
 * rather than substitution. And a plain {@code any} cannot be derived at all: unlike {@code any1}
 * it names nothing, so there is no identity to bind.
 *
 * <p>Which shipped variants those cover is pinned by {@code ParameterizedReturnTypeTest} against
 * the declarations the catalog ships, and deliberately not repeated here -- the catalog is owned
 * upstream, so a list of names in this Javadoc would go stale on a {@code substrait-packaging} bump
 * with nothing to catch it.
```

core/src/test/java/io/substrait/type/ParameterizedReturnTypeTest.java
```diff
@@ -0,0 +1,220 @@
...
+            ParameterizedType.IntervalCompound.builder()
+                .nullable(false)
+                .precision(parameter("P"))
+                .build(),
+            R.intervalCompound(3)));
+  }
+
+  private static ParameterizedType.StringLiteral parameter(String name) {
+    return ParameterizedType.StringLiteral.builder().nullable(false).value(name).build();
+  }
```
nielspardon, line 142: Use `ParameterizedTypeCreator.REQUIRED`, which already has `fixedBinaryE`, `intervalCompoundE` and `parameter` -- the local `parameter()` helper duplicates the last of those and has no other caller. Worth widening the claim while you are here: neither shape appears anywhere in the catalog, as an argument or as a return, which is stronger than "as a return". Needs `import io.substrait.function.ParameterizedTypeCreator;`.

```suggestion
  /**
   * fixedbinary and interval_compound are the two parameterized shapes no standard extension
   * declares at all, as an argument or as a return, so they are pinned against a hand-written
   * declaration instead of the catalog.
   */
  @Test
  void theShapesTheCatalogDoesNotDeclareDeriveToo() {
    ParameterizedTypeCreator P = ParameterizedTypeCreator.REQUIRED;
    assertEquals(R.fixedBinary(9), derive(P.fixedBinaryE("L1"), R.fixedBinary(9)));
    assertEquals(R.intervalCompound(3), derive(P.intervalCompoundE("P"), R.intervalCompound(3)));
  }
```

core/src/test/java/io/substrait/type/ParameterizedReturnTypeTest.java
```diff
@@ -0,0 +1,220 @@
...
+            "divide:dec_dec",
+            "floor:dec",
+            "modulus:dec_dec",
+            "multiply:dec_dec",
+            "round:dec_i32",
+            "strptime_time:str_str_i8",
+            "strptime_timestamp:str_str_i8",
+            "strptime_timestamp:str_str_str_i8",
+            "subtract:dec_dec"),
+        variantsReturning(TypeExpression.ReturnProgram.class));
```
nielspardon: Assert that the two shapes actually throw, and drop the sentence pointing at a list that no longer exists after the change above. The census pins return classes only, so making one of these shapes derivable leaves this test green while the prose it guards goes stale.

```suggestion
   * The census of what the evaluator does not derive: a {@code list} return is the first shape, a
   * multi-line return program the second. {@link TypeExpressionEvaluator}'s Javadoc describes those
   * shapes and points here rather than naming variants, so this test is the only place a {@code
   * substrait-packaging} bump can make the two disagree.
   */
  @Test
  void theReturnShapesThatAreNotDerivedYet() {
    assertEquals(
        List.of(
            "filter:list_func",
            "quantile:req_req_i64_any",
            "regexp_match_substring_all:vchar_vchar_i64_i64",
            "regexp_string_split:vchar_vchar",
            "sort:list",
            "string_split:vchar_vchar",
            "transform:list_func"),
        variantsReturning(ParameterizedType.ListType.class));

    assertEquals(
        List.of(
            "add:dec_dec",
            "assume_timezone:date_str_i8",
            "bitwise_and:dec_dec",
            "bitwise_or:dec_dec",
            "bitwise_xor:dec_dec",
            "ceil:dec",
            "divide:dec_dec",
            "floor:dec",
            "modulus:dec_dec",
            "multiply:dec_dec",
            "round:dec_i32",
            "strptime_time:str_str_i8",
            "strptime_timestamp:str_str_i8",
            "strptime_timestamp:str_str_str_i8",
            "subtract:dec_dec"),
        variantsReturning(TypeExpression.ReturnProgram.class));

    // The lists above pin which variants carry each shape; these pin that the shapes actually fail,
    // so making one derivable cannot leave the census passing and the Javadoc stale.
    assertThrows(
        UnsupportedOperationException.class,
        () -> resolve("string_split:vchar_vchar", R.varChar(20), R.varChar(20)));
    assertThrows(
        UnsupportedOperationException.class,
        () -> resolve("add:dec_dec", R.decimal(10, 2), R.decimal(10, 2)));
```

nielspardon (commented):
The terminal `else` reads well, but `isContainer` omits `ParameterizedType.Func`, so it now refuses four `list_func` variants: `all_match` and `any_match` declare a concrete `boolean?` return and regress from resolving to throwing, while `filter` and `transform` still fail on their `list` return but report a binding error instead of the return-type one. A catalog sweep with a well-typed call per variant puts it at 540 resolved before the terminal `else`, 538 after, and 540 again with `Func` exempted, all tests still green.

The rest below is the documentation and test-fixture tail from round 1; the one worth a decision is whether a catalog-wide list of variant names belongs in Javadoc at all.

core/src/main/java/io/substrait/type/TypeExpressionEvaluator.java
alexandrefimov: Taken as written, and the census is fixed rather than trusted: it builds a call for a func-declared argument now instead of dropping the variant, which is what hid this. Widened, it puts the catalog at 579 resolved of 602 before the terminal `else`, 577 after, and 579 again with `Func` exempted -- your delta and your two names. My base is higher than your 540 because I substitute a concrete type under every wildcard independently, which well-types more variants than a stricter call would. `aContainerDeclarationIsNotRefusedForABindingItNeverMakes` covers all four now and fails if `Func` comes back out.

core/src/main/java/io/substrait/type/TypeExpressionEvaluator.java
alexandrefimov: Taken as written. Checked the correction before taking it: `string_split:vchar_vchar` returns `List<varchar<L1>>`, so the blocker is the container and not the element, and both the Javadoc and the PR body had it the other way round.

core/src/test/java/io/substrait/type/ParameterizedReturnTypeTest.java
alexandrefimov: Taken as written; the local `parameter` helper went with its last caller.

core/src/test/java/io/substrait/type/ParameterizedReturnTypeTest.java
alexandrefimov: Taken as written.

core/src/main/java/io/substrait/type/TypeExpressionEvaluator.java
alexandrefimov: Answered by the round above: not blanket. `Func` joins the exemption, and the two `list_func` variants that regressed are pinned alongside `cardinality:list`.

core/src/test/java/io/substrait/type/ParameterizedReturnTypeTest.java
```diff
@@ -0,0 +1,218 @@
...
+    assertEquals(N.precisionTime(3), resolve("min:pt", R.precisionTime(3)));
+    assertEquals(
+        R.precisionTimestampTZ(9),
+        resolve("add:ptstz_iyear_str", R.precisionTimestampTZ(9), R.INTERVAL_YEAR, R.STRING));
+  }
+
+  /**
+   * fixedbinary and interval_compound are the two parameterized shapes no standard extension
+   * declares as a return, so they are pinned against a hand-written declaration instead of the
+   * catalog.
```
nielspardon: This is the half of that ask that didn't land. The evaluator's Javadoc now says "at all, as an argument or as a return", so leaving this one at "as a return" reads as though the two shapes might still turn up as arguments somewhere in the catalog -- they don't.

```suggestion
   * fixedbinary and interval_compound are the two parameterized shapes no standard extension
   * declares at all, as an argument or as a return, so they are pinned against a hand-written
   * declaration instead of the catalog.
```

nielspardon (commented):
All four taken, and the two `all_match`/`any_match` assertions pin the regression better than my ask did. One residual below; the only other thing left on my list is the three-list `Stream.of` that `variant` and `variantsReturning` now each build separately, which is worth a shared helper if you touch the file again but not a round of its own.

core/src/test/java/io/substrait/type/ParameterizedReturnTypeTest.java
alexandrefimov: Taken as written. That was the half of your earlier suggestion I dropped -- I took the code out of it and left the sentence behind. The `Stream.of` triple stays for now, on your read that it is not worth a round of its own.

nielspardon (approved):
LGTM

PR #1150, opened 2026-08-24
refactor(isthmus): drop the unused TIME_WITH_LOCAL_TIME_ZONE precision override

### LINKED ISSUE #1149: SubstraitTypeSystem sets a max precision for TIME_WITH_LOCAL_TIME_ZONE, which nothing converts
`SubstraitTypeSystem.getMaxPrecision` raises the maximum precision to 6 for `SqlTypeName.TIME_WITH_LOCAL_TIME_ZONE`, alongside `TIME`, `TIMESTAMP` and `TIMESTAMP_WITH_LOCAL_TIME_ZONE`. Nothing converts that type in either direction: it is absent from `TypeConverter.toSubstrait`'s switch, and no Substrait type maps to it on the way back, because Substrait's type set has `precision_timestamp_tz` but no time-with-timezone counterpart. The `case TIME_WITH_LOCAL_TIME_ZONE:` in `getMaxPrecision` is the only mention of the name anywhere in the repository.

The entry's one observable effect is on Calcite's validator, and it makes the failure worse rather than better. Calcite's default maximum for the type is 3, so raising it to 6 lets a precision the converter cannot handle get further than it otherwise would:

```
maxPrecision TIME_WITH_LOCAL_TIME_ZONE   substrait=6  calciteDefault=3

SELECT CAST(NULL AS TIME(4) WITH LOCAL TIME ZONE) ...
  -> passes validation (rejected at precision > 3 without the override)
  -> UnsupportedOperationException: Unable to convert the type TIME_WITH_LOCAL_TIME_ZONE(4)
```

`TypeConverter.DEFAULT.toSubstrait` on a directly constructed `TIME_WITH_LOCAL_TIME_ZONE(4)` fails the same way. So no capability is gained or lost either way — precisions 4 to 6 currently fail late with a conversion error where they could fail during validation, and precision 0 to 3 fails at conversion regardless.

Dropping the case looks like the right fix, since the type is unconvertible by construction rather than merely unimplemented. Worth confirming the override was not added deliberately to keep some validation path working before removing it.

Noticed while reviewing #1121, which consolidated the six duplicated precision checks into `SubstraitTypeSystem.requireSupportedPrecision` and made this entry stand out. Not caused by that PR and out of scope for it.

alexandrefimov:
`getMaxPrecision` raised the maximum precision to 6 for `TIME_WITH_LOCAL_TIME_ZONE` alongside the types Isthmus actually converts. Substrait has `precision_timestamp_tz` but no time-with-timezone counterpart, so the type appears in neither direction of `TypeConverter`, and the `case` was the only mention of the name in the repository. It has been there since the initial commit rather than added to keep a validation path working.

Removing it changes nothing observable. Calcite's type factory clamps a precision above the maximum instead of rejecting it, so the entry was not letting a bad precision get further than it otherwise would: with it `CAST(NULL AS TIME(4) WITH LOCAL TIME ZONE)` reaches the converter as `TIME_WITH_LOCAL_TIME_ZONE(4)`, without it as `TIME_WITH_LOCAL_TIME_ZONE(3)`, and both fail with the same `UnsupportedOperationException`. Precisions 0 through 6 failed at conversion before and still do.

The test pins the premise as well as the entry: if the type ever does convert, the assertion that `toSubstrait` rejects it fails and points at the maximum that would then need setting.

Closes #1149.

### DIFF
--- a/isthmus/src/main/java/io/substrait/isthmus/SubstraitTypeSystem.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/SubstraitTypeSystem.java
@@ -110,7 +110,6 @@ public class SubstraitTypeSystem extends RelDataTypeSystemImpl {
       case INTERVAL_YEAR:
       case INTERVAL_YEAR_MONTH:
       case TIME:
-      case TIME_WITH_LOCAL_TIME_ZONE:
       case TIMESTAMP:
       case TIMESTAMP_WITH_LOCAL_TIME_ZONE:
         return 6;
--- a/isthmus/src/test/java/io/substrait/isthmus/SubstraitTypeSystemTest.java
+++ b/isthmus/src/test/java/io/substrait/isthmus/SubstraitTypeSystemTest.java
@@ -162,6 +162,20 @@ class SubstraitTypeSystemTest {
   }
 
   @Test
+  void timeWithLocalTimeZoneHasNoSubstraitMaxPrecision() {
+    // Substrait has precision_timestamp_tz but no time-with-timezone counterpart, so this type
+    // converts in neither direction and there is no Substrait maximum to report for it.
+    assertThrows(
+        UnsupportedOperationException.class,
+        () ->
+            TypeConverter.DEFAULT.toSubstrait(
+                TYPE_FACTORY.createSqlType(SqlTypeName.TIME_WITH_LOCAL_TIME_ZONE, 3)));
+    assertEquals(
+        RelDataTypeSystem.DEFAULT.getMaxPrecision(SqlTypeName.TIME_WITH_LOCAL_TIME_ZONE),
+        typeSystem.getMaxPrecision(SqlTypeName.TIME_WITH_LOCAL_TIME_ZONE));
+  }
+
+  @Test
   void canCreateDecimalWithMaxPrecision() {
     RelDataType decimalType = TYPE_FACTORY.createSqlType(SqlTypeName.DECIMAL, 38, 10);
     assertEquals(38, decimalType.getPrecision());


isthmus/src/main/java/io/substrait/isthmus/SubstraitTypeSystem.java
```diff
@@ -110,7 +110,6 @@ public int getMaxPrecision(final SqlTypeName typeName) {
       case INTERVAL_YEAR:
       case INTERVAL_YEAR_MONTH:
       case TIME:
```
nielspardon, line 112: `TIME WITH LOCAL TIME ZONE` is exactly the kind of type a consumer supplies a `UserTypeMapper` for, and `toSubstrait` consults that mapper before this switch — so a mapper carrying `type.getPrecision()` now records 3 in the plan where the DDL declared 6, with no error. That `TYPE_FACTORY.createSqlType(TIME_WITH_LOCAL_TIME_ZONE, 6)` answers 3 instead of 6 is the same silent narrowing #1169 marked breaking for the character and binary types. Keep the case, or mark the commit `!` with a footer and scope "changes nothing observable" to the SQL→Substrait path, where it is accurate.

isthmus/src/test/java/io/substrait/isthmus/SubstraitTypeSystemTest.java
```diff
@@ -161,6 +161,20 @@ void aRaggedBinaryUnionKeepsTheWidestWidth() {
...
+    // converts in neither direction and there is no Substrait maximum to report for it.
+    assertThrows(
+        UnsupportedOperationException.class,
+        () ->
+            TypeConverter.DEFAULT.toSubstrait(
+                TYPE_FACTORY.createSqlType(SqlTypeName.TIME_WITH_LOCAL_TIME_ZONE, 3)));
+    assertEquals(
+        RelDataTypeSystem.DEFAULT.getMaxPrecision(SqlTypeName.TIME_WITH_LOCAL_TIME_ZONE),
+        typeSystem.getMaxPrecision(SqlTypeName.TIME_WITH_LOCAL_TIME_ZONE));
+  }
```
nielspardon, line 176: Pass `6` rather than `3` and assert the message, so both halves fail if the case comes back — at precision 3 the type is unclamped under the old maximum too, so that assertion passes either way today. `assertAll` also makes good on the commit body's promise to point at the maximum that would need setting; as written the `assertThrows` fails first and no maximum is ever reported.

```suggestion
  /**
   * Substrait's only time type is {@code precision_time<P>}, which is timezone-free by
   * construction, so this type converts in neither direction and the type system has no maximum of
   * its own to report for it.
   */
  @Test
  void timeWithLocalTimeZoneKeepsCalcitesMaxPrecision() {
    assertAll(
        () -> assertEquals(3, typeSystem.getMaxPrecision(SqlTypeName.TIME_WITH_LOCAL_TIME_ZONE)),
        () -> {
          UnsupportedOperationException e =
              assertThrows(
                  UnsupportedOperationException.class,
                  () ->
                      TypeConverter.DEFAULT.toSubstrait(
                          TYPE_FACTORY.createSqlType(SqlTypeName.TIME_WITH_LOCAL_TIME_ZONE, 6)));
          assertTrue(e.getMessage().contains("TIME_WITH_LOCAL_TIME_ZONE(3)"), e.getMessage());
        });
  }
```

isthmus/src/main/java/io/substrait/isthmus/SubstraitTypeSystem.java
```diff
@@ -110,7 +110,6 @@ public int getMaxPrecision(final SqlTypeName typeName) {
...
-      case TIME_WITH_LOCAL_TIME_ZONE:
       case TIMESTAMP:
       case TIMESTAMP_WITH_LOCAL_TIME_ZONE:
         return 6;
```
nielspardon, line 115: Add one Javadoc sentence naming why this group returns 6 — these are the Calcite names `TypeConverter` passes to `requireSupportedPrecision`, and 6 is deliberately *above* Calcite's datetime default of 3. Without it the next reader sees `TIME(4) WITH LOCAL TIME ZONE` capped at 3 beside a timestamp sibling capped at 6 and re-adds the label. (The three `INTERVAL_*` cases don't fit that criterion — that's #1221.)

nielspardon (commented):
Either keep `case TIME_WITH_LOCAL_TIME_ZONE:` or mark this `refactor(isthmus)!:` with a `BREAKING CHANGE:` footer — a `refactor` with no `!` publishes no changelog entry at all, and the narrowing it ships is silent. The removal itself is right, and the spec argument behind it holds: Substrait's only time type is `precision_time<P>`, timezone-free by construction.

Filed #1243 and #1244 for two adjacent things I hit while reading this; both are out of scope here.

alexandrefimov:
Closing this: the case is not unused. I measured the `UserTypeMapper` path you named — a `TIME_WITH_LOCAL_TIME_ZONE` created at precision 6 reaches a mapper as 6 with the case present and as 3 without it. The retraction of "changes nothing observable" is in #1149.

Your other two suggestions apply to the file as it stands. Happy to send them as a small PR: the Javadoc sentence on that group, and a test at precision 6 pinning both the message and the precision a mapper is handed.

PR #1151, opened 2026-08-24
fix(isthmus)!: round-trip a virtual table whose rows are not all literals

### LINKED ISSUE #631: Flatten Expanded Calcite Representation of Virtual Tables w/ Complex Expressions when Converting to Substrait

alexandrefimov:
A `VirtualTableScan` whose rows are not all literals was expanded into a `UNION ALL` of one single-row projection per row, each over an empty `Values`. Nothing in that shape says it was a table, so the plan came back from Calcite as a `Set` of `Project`s -- a different relation than the one that went in -- and the schema's names went with it.

Emit an isthmus-owned relation instead. `VirtualTable` holds the rows as `RexNode`s and the schema as its row type, and `visitOther` recognises it by type. `SubstraitRelToSqlConverter` generates SQL for it, reaching a table inside a `RexSubQuery` that a pass over the tree cannot. `VirtualTableExpansionRule` still builds the union for a consumer whose planner knows only Calcite's own relations, one way: converting that expansion back to Substrait gives the union it is.

Coming back from Calcite, a value that is not a literal carries the type its own expressions give it. For a computed value inside a nullable struct that is not the type the column declares -- Calcite makes a nullable struct's fields nullable while the expression keeps its own -- and the conversion is now refused rather than cast, since casting would put an expression in the output the input did not have. Whether the spec lets a row differ from its column at all is open: substrait-io/substrait#1211 has the plans and what four implementations do with them.

Closes #631, #1227, #1228

BREAKING CHANGE: two changes, one per direction.

Substrait to Calcite: a virtual table whose rows are not all literals now converts to `io.substrait.isthmus.calcite.rel.VirtualTable` instead of a `UNION ALL` of single-row projections over an empty table. Code that hands the tree to something knowing only Calcite's own relations can expand it with `io.substrait.isthmus.calcite.rel.rules.VirtualTableExpansionRule`, which is one-way: converting the expansion back to Substrait gives that union, not the virtual table.

Calcite to Substrait: where a row's value carries a type its column does not declare -- a computed value inside a nullable struct, Calcite having made the struct's fields nullable while the expression keeps its own -- converting the table back now throws an `UnsupportedOperationException` naming the value and both types. It used to produce the `UNION ALL` above, which is the shape this release stops producing, so there is no way to keep the old output; whether the spec permits the mismatch at all is being settled in substrait-io/substrait#1211.

### DIFF
--- a/isthmus/src/main/java/io/substrait/isthmus/OuterReferenceResolver.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/OuterReferenceResolver.java
@@ -118,8 +118,10 @@ public class OuterReferenceResolver extends RelNodeVisitor<RelNode, RuntimeExcep
     // A relation's own expressions can hold a subquery binding outer references, and a subquery's
     // relation is not an input, so walking inputs never reaches it. Filter and Project scan theirs
     // before they get here; a virtual table's rows, and a join, calc or sort condition, arrive
-    // here. AbstractRelNode.accept(RexShuttle) returns the relation itself, so the result is the
-    // one already in the tree.
+    // here. The rewrite is an identity -- RexVisitor.visitSubQuery returns the subquery it was
+    // given -- so the relation the tree holds is unchanged and the result is discarded. Relations
+    // that do not override accept(RexShuttle) at all -- TableModify, Window, Match, Aggregate --
+    // keep their expressions to themselves; a correlation among those is still unresolved.
     other.accept(rexVisitor);
     for (RelNode child : other.getInputs()) {
       reverseAccept(child);
--- a/isthmus/src/main/java/io/substrait/isthmus/OuterReferenceResolver.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/OuterReferenceResolver.java
@@ -1,6 +1,5 @@
 package io.substrait.isthmus;
 
-import io.substrait.isthmus.calcite.rel.VirtualTable;
 import java.util.HashMap;
 import java.util.IdentityHashMap;
 import java.util.Map;
@@ -116,12 +115,12 @@ public class OuterReferenceResolver extends RelNodeVisitor<RelNode, RuntimeExcep
 
   @Override
   public RelNode visitOther(RelNode other) throws RuntimeException {
-    if (other instanceof VirtualTable) {
-      // A virtual table's rows are expressions, and a subquery among them binds outer references
-      // like one anywhere else. A subquery's relation is not an input, so walking inputs never
-      // reaches it; Filter and Project scan their own before they get here.
-      other.accept(rexVisitor);
-    }
+    // A relation's own expressions can hold a subquery binding outer references, and a subquery's
+    // relation is not an input, so walking inputs never reaches it. Filter and Project scan theirs
+    // before they get here; a virtual table's rows, and a join, calc or sort condition, arrive
+    // here. AbstractRelNode.accept(RexShuttle) returns the relation itself, so the result is the
+    // one already in the tree.
+    other.accept(rexVisitor);
     for (RelNode child : other.getInputs()) {
       reverseAccept(child);
     }
--- a/isthmus/src/main/java/io/substrait/isthmus/OuterReferenceResolver.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/OuterReferenceResolver.java
@@ -1,5 +1,6 @@
 package io.substrait.isthmus;
 
+import io.substrait.isthmus.calcite.rel.VirtualTable;
 import java.util.HashMap;
 import java.util.IdentityHashMap;
 import java.util.Map;
@@ -115,6 +116,12 @@ public class OuterReferenceResolver extends RelNodeVisitor<RelNode, RuntimeExcep
 
   @Override
   public RelNode visitOther(RelNode other) throws RuntimeException {
+    if (other instanceof VirtualTable) {
+      // A virtual table's rows are expressions, and a subquery among them binds outer references
+      // like one anywhere else. A subquery's relation is not an input, so walking inputs never
+      // reaches it; Filter and Project scan their own before they get here.
+      other.accept(rexVisitor);
+    }
     for (RelNode child : other.getInputs()) {
       reverseAccept(child);
     }
--- a/isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
@@ -11,6 +11,7 @@ import io.substrait.extension.SimpleExtension;
 import io.substrait.hint.Hint;
 import io.substrait.isthmus.calcite.rel.CreateTable;
 import io.substrait.isthmus.calcite.rel.CreateView;
+import io.substrait.isthmus.calcite.rel.VirtualTable;
 import io.substrait.isthmus.expression.AggregateFunctionConverter;
 import io.substrait.isthmus.expression.ExpressionRexConverter;
 import io.substrait.isthmus.expression.ScalarFunctionConverter;
@@ -63,6 +64,7 @@ import java.util.Optional;
 import java.util.stream.Collectors;
 import java.util.stream.IntStream;
 import java.util.stream.Stream;
+import org.apache.calcite.plan.Convention;
 import org.apache.calcite.plan.RelOptSchema;
 import org.apache.calcite.plan.RelOptTable;
 import org.apache.calcite.plan.RelOptUtil;
@@ -78,7 +80,6 @@ import org.apache.calcite.rel.core.JoinRelType;
 import org.apache.calcite.rel.core.TableModify;
 import org.apache.calcite.rel.logical.LogicalProject;
 import org.apache.calcite.rel.logical.LogicalTableModify;
-import org.apache.calcite.rel.logical.LogicalUnion;
 import org.apache.calcite.rel.logical.LogicalValues;
 import org.apache.calcite.rel.type.RelDataType;
 import org.apache.calcite.rel.type.RelDataTypeFactory;
@@ -924,45 +925,18 @@ public class SubstraitRelNodeConverter
           LogicalValues.create(relBuilder.getCluster(), rowType, tuplesBuilder.build()),
           virtualTableScan);
     } else {
-      // A row that does not fit a LogicalValues tuple is computed instead: we create a
-      // LogicalProject for each row to compute its values, and combine them together using a
-      // LogicalUnion. For example the following:
-      //
-      //   VirtualTable
-      //     (e1, e2)
-      //     (e3, e4)
-      //
-      //  Becomes:
-      //
-      //   LogicalUnion(all=[true])
-      //     LogicalProject(exprs=[e1, e2])
-      //       <Empty Row>
-      //     LogicalProject(exprs=[e3, e4])
-      //       <Empty Row>
-      //
-
-      RelDataType emptyRowType = typeFactory.createStructType(List.of(), List.of());
-      ImmutableList<ImmutableList<RexLiteral>> emptyRowValue = ImmutableList.of(ImmutableList.of());
-
-      List<RelNode> projects = new ArrayList<>();
-      for (final List<RexNode> rexRow : convertedRows) {
-        RelNode values = LogicalValues.create(relBuilder.getCluster(), emptyRowType, emptyRowValue);
-        RelNode project =
-            LogicalProject.create(
-                values, Collections.emptyList(), rexRow, rowType, Collections.emptySet());
-        projects.add(project);
-      }
-      RelNode union = LogicalUnion.create(projects, true);
-
-      // Apply a final LogicalProject on top to capture the field names from the VirtualTable
-      List<RexNode> topProjectExprs = new ArrayList<>();
-      for (int i = 0; i < rowType.getFieldCount(); i++) {
-        topProjectExprs.add(rexBuilder.makeInputRef(union, i));
-      }
-      RelNode topProject =
-          LogicalProject.create(
-              union, Collections.emptyList(), topProjectExprs, rowType, Collections.emptySet());
-      return applyRelCommon(topProject, virtualTableScan, topProject);
+      // A row that does not fit a LogicalValues tuple keeps its expressions, in a relation of our
+      // own: Calcite has none that holds them, and expanding the table into a projection per row
+      // does not come back -- the projection is what converts back, and the table is gone. A
+      // consumer whose planner only knows Calcite's own relations can expand it with
+      // VirtualTableExpansionRule.
+      return applyRelCommon(
+          new VirtualTable(
+              relBuilder.getCluster(),
+              relBuilder.getCluster().traitSetOf(Convention.NONE),
+              rowType,
+              convertedRows),
+          virtualTableScan);
     }
   }
 
--- a/isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
@@ -64,7 +64,6 @@ import java.util.Optional;
 import java.util.stream.Collectors;
 import java.util.stream.IntStream;
 import java.util.stream.Stream;
-import org.apache.calcite.plan.Convention;
 import org.apache.calcite.plan.RelOptSchema;
 import org.apache.calcite.plan.RelOptTable;
 import org.apache.calcite.plan.RelOptUtil;
@@ -931,12 +930,7 @@ public class SubstraitRelNodeConverter
       // consumer whose planner only knows Calcite's own relations can expand it with
       // VirtualTableExpansionRule.
       return applyRelCommon(
-          new VirtualTable(
-              relBuilder.getCluster(),
-              relBuilder.getCluster().traitSetOf(Convention.NONE),
-              rowType,
-              convertedRows),
-          virtualTableScan);
+          VirtualTable.create(relBuilder.getCluster(), rowType, convertedRows), virtualTableScan);
     }
   }
 
--- a/isthmus/src/main/java/io/substrait/isthmus/SubstraitRelVisitor.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/SubstraitRelVisitor.java
@@ -62,6 +62,7 @@ import org.apache.calcite.rel.type.RelDataTypeFactory;
 import org.apache.calcite.rel.type.RelDataTypeField;
 import org.apache.calcite.rex.RexBuilder;
 import org.apache.calcite.rex.RexInputRef;
+import org.apache.calcite.rex.RexLiteral;
 import org.apache.calcite.rex.RexNode;
 import org.apache.calcite.rex.RexUtil;
 import org.apache.calcite.sql.SqlKind;
@@ -972,17 +973,30 @@ public class SubstraitRelVisitor extends RelNodeVisitor<Rel, RuntimeException> {
   }
 
   /**
-   * Handles the isthmus {@link VirtualTable}, which is what a virtual table whose rows are not all
+   * Converts the isthmus {@link VirtualTable}, which is what a virtual table whose rows are not all
    * literals converts to.
    *
    * @param virtualTable Calcite virtual table
    * @return Substrait virtual table scan
    */
-  public Rel handleVirtualTable(VirtualTable virtualTable) {
+  @Override
+  public Rel visit(VirtualTable virtualTable) {
+    // At the row type's field types rather than the values' own, as visit(Values) does: a literal
+    // narrower than its column -- Calcite infers one for a tuple value, and pushes a struct's
+    // nullability down into its fields -- would otherwise disagree with the schema built from the
+    // same row type, and VirtualTableScan rejects the relation on that.
+    List<RelDataTypeField> rowFields = virtualTable.getRowType().getFieldList();
+    LiteralConverter literalConverter = new LiteralConverter(typeConverter);
     List<Expression.NestedStruct> rows = new ArrayList<>(virtualTable.getRows().size());
     for (List<RexNode> row : virtualTable.getRows()) {
-      List<Expression> fields =
-          row.stream().map(this::toExpression).collect(Collectors.toUnmodifiableList());
+      List<Expression> fields = new ArrayList<>(row.size());
+      for (int column = 0; column < row.size(); column++) {
+        RexNode value = row.get(column);
+        fields.add(
+            value instanceof RexLiteral
+                ? literalConverter.convert((RexLiteral) value, rowFields.get(column).getType())
+                : toExpression(value));
+      }
       rows.add(ExpressionCreator.nestedStruct(false, fields));
     }
     return VirtualTableScan.builder()
@@ -1005,9 +1019,6 @@ public class SubstraitRelVisitor extends RelNodeVisitor<Rel, RuntimeException> {
 
     } else if (other instanceof CreateView) {
       return handleCreateView((CreateView) other);
-
-    } else if (other instanceof VirtualTable) {
-      return handleVirtualTable((VirtualTable) other);
     }
     throw new UnsupportedOperationException("Unable to handle node: " + other);
   }
--- a/isthmus/src/main/java/io/substrait/isthmus/SubstraitRelVisitor.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/SubstraitRelVisitor.java
@@ -992,10 +992,24 @@ public class SubstraitRelVisitor extends RelNodeVisitor<Rel, RuntimeException> {
       List<Expression> fields = new ArrayList<>(row.size());
       for (int column = 0; column < row.size(); column++) {
         RexNode value = row.get(column);
-        fields.add(
+        RelDataType declaredType = rowFields.get(column).getType();
+        Expression converted =
             value instanceof RexLiteral
-                ? literalConverter.convert((RexLiteral) value, rowFields.get(column).getType())
-                : toExpression(value));
+                ? literalConverter.convert((RexLiteral) value, declaredType)
+                : toExpression(value);
+        // A value that is not a literal is converted from the expressions it is built of and
+        // takes its type from them, which the declared type cannot be put back on: casting at it
+        // would put an expression in the output the input did not have. Refused here rather than
+        // left to VirtualTableScan, whose check compares the two types without promoting either.
+        Type declared = typeConverter.toSubstrait(declaredType);
+        if (!converted.getType().equals(declared)) {
+          throw new UnsupportedOperationException(
+              String.format(
+                  "A virtual table's value %s converts to %s where its column is declared %s: "
+                      + "isthmus cannot convert a value that does not carry its column's type",
+                  value, converted.getType(), declared));
+        }
+        fields.add(converted);
       }
       rows.add(ExpressionCreator.nestedStruct(false, fields));
     }
--- a/isthmus/src/main/java/io/substrait/isthmus/SubstraitRelVisitor.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/SubstraitRelVisitor.java
@@ -8,6 +8,7 @@ import io.substrait.expression.FieldReference;
 import io.substrait.extension.SimpleExtension;
 import io.substrait.isthmus.calcite.rel.CreateTable;
 import io.substrait.isthmus.calcite.rel.CreateView;
+import io.substrait.isthmus.calcite.rel.VirtualTable;
 import io.substrait.isthmus.expression.AggregateFunctionConverter;
 import io.substrait.isthmus.expression.LiteralConverter;
 import io.substrait.isthmus.expression.RexExpressionConverter;
@@ -971,6 +972,26 @@ public class SubstraitRelVisitor extends RelNodeVisitor<Rel, RuntimeException> {
   }
 
   /**
+   * Handles the isthmus {@link VirtualTable}, which is what a virtual table whose rows are not all
+   * literals converts to.
+   *
+   * @param virtualTable Calcite virtual table
+   * @return Substrait virtual table scan
+   */
+  public Rel handleVirtualTable(VirtualTable virtualTable) {
+    List<Expression.NestedStruct> rows = new ArrayList<>(virtualTable.getRows().size());
+    for (List<RexNode> row : virtualTable.getRows()) {
+      List<Expression> fields =
+          row.stream().map(this::toExpression).collect(Collectors.toUnmodifiableList());
+      rows.add(ExpressionCreator.nestedStruct(false, fields));
+    }
+    return VirtualTableScan.builder()
+        .initialSchema(typeConverter.toNamedStruct(virtualTable.getRowType()))
+        .addAllRows(rows)
+        .build();
+  }
+
+  /**
    * Visits other Calcite nodes (e.g., DDL wrappers).
    *
    * @param other Calcite node
@@ -984,6 +1005,9 @@ public class SubstraitRelVisitor extends RelNodeVisitor<Rel, RuntimeException> {
 
     } else if (other instanceof CreateView) {
       return handleCreateView((CreateView) other);
+
+    } else if (other instanceof VirtualTable) {
+      return handleVirtualTable((VirtualTable) other);
     }
     throw new UnsupportedOperationException("Unable to handle node: " + other);
   }
--- a/isthmus/src/main/java/io/substrait/isthmus/calcite/rel/VirtualTable.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/calcite/rel/VirtualTable.java
@@ -131,6 +131,10 @@ public class VirtualTable extends AbstractRelNode {
    * join, and by {@code RelDecorrelator}: a consumer's planner leaves it unexpanded, and the
    * variables it resolves against have to travel with the relation that holds it.
    *
+   * <p>Only a consumer populates it. A conversion never does: an id is bound to a relation whose
+   * fields the reference names, and a leaf with no inputs has none, so an outer reference in a row
+   * belongs to the relation around the table the way one in a projection's expression does.
+   *
    * @return the correlation variables
    */
   @Override
--- a/isthmus/src/main/java/io/substrait/isthmus/calcite/rel/VirtualTable.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/calcite/rel/VirtualTable.java
@@ -1,19 +1,27 @@
 package io.substrait.isthmus.calcite.rel;
 
 import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableSet;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Set;
 import java.util.stream.Collectors;
+import org.apache.calcite.plan.Convention;
 import org.apache.calcite.plan.RelOptCluster;
+import org.apache.calcite.plan.RelOptCost;
+import org.apache.calcite.plan.RelOptPlanner;
 import org.apache.calcite.plan.RelTraitSet;
 import org.apache.calcite.rel.AbstractRelNode;
 import org.apache.calcite.rel.RelNode;
 import org.apache.calcite.rel.RelWriter;
+import org.apache.calcite.rel.core.CorrelationId;
 import org.apache.calcite.rel.metadata.RelMetadataQuery;
 import org.apache.calcite.rel.type.RelDataType;
 import org.apache.calcite.rex.RexNode;
 import org.apache.calcite.rex.RexShuttle;
+import org.apache.calcite.rex.RexUtil;
 import org.apache.calcite.sql.SqlExplainLevel;
+import org.apache.calcite.util.Litmus;
 
 /**
  * A table of rows given as expressions, which is what a Substrait virtual table is and what Calcite
@@ -24,15 +32,15 @@ import org.apache.calcite.sql.SqlExplainLevel;
  * rows into stock Calcite relations -- a projection per row, unioned -- loses that: the shape a
  * planner leaves behind is a projection over an empty table, which is what converts back. A
  * consumer that needs the expansion can ask for it with {@link
- * io.substrait.isthmus.calcite.rel.rules.VirtualTableExpansionRule}, knowing it is one-way; nothing
- * in isthmus runs that rule.
+ * io.substrait.isthmus.calcite.rel.rules.VirtualTableExpansionRule}, knowing it is one-way.
  *
  * <p>The rows are held at the row type's own field types, names included, so that the expansion has
  * nothing left to derive.
  */
 public class VirtualTable extends AbstractRelNode {
 
-  private final ImmutableList<ImmutableList<RexNode>> rows;
+  private final ImmutableList<List<RexNode>> rows;
+  private final ImmutableSet<CorrelationId> variablesSet;
 
   /**
    * VirtualTable constructor.
@@ -40,20 +48,70 @@ public class VirtualTable extends AbstractRelNode {
    * @param cluster the cluster this relation belongs to
    * @param traitSet the relation's traits
    * @param rowType the table's row type, carrying the schema's field names
+   * @param variablesSet the correlation variables the rows resolve against
    * @param rows one list of values per row, each value at the type its column is declared at
+   * @throws IllegalArgumentException if a row does not fit the row type
    */
   public VirtualTable(
       RelOptCluster cluster,
       RelTraitSet traitSet,
       RelDataType rowType,
+      Set<CorrelationId> variablesSet,
       List<? extends List<RexNode>> rows) {
     super(cluster, traitSet);
     this.rowType = rowType;
-    ImmutableList.Builder<ImmutableList<RexNode>> builder = ImmutableList.builder();
+    this.variablesSet = ImmutableSet.copyOf(variablesSet);
+    ImmutableList.Builder<List<RexNode>> builder = ImmutableList.builder();
     for (List<RexNode> row : rows) {
+      // Nothing else checks this: the deleted LogicalProject got it from RexUtil.compatibleTypes,
+      // and AbstractRelNode.isValid succeeds unconditionally.
+      if (row.size() != rowType.getFieldCount()) {
+        throw new IllegalArgumentException(
+            String.format(
+                "A virtual table's row has %d values where its type declares %d columns: %s",
+                row.size(), rowType.getFieldCount(), row));
+      }
+      if (!RexUtil.compatibleTypes(row, rowType, Litmus.IGNORE)) {
+        throw new IllegalArgumentException(
+            String.format(
+                "A virtual table's row %s does not fit the type %s its columns are declared at",
+                row, rowType.getFullTypeString()));
+      }
       builder.add(ImmutableList.copyOf(row));
     }
     this.rows = builder.build();
+  }
+
+  /**
+   * Creates a virtual table with no correlation variables, in the convention every relation this
+   * conversion builds is in.
+   *
+   * @param cluster the cluster this relation belongs to
+   * @param rowType the table's row type, carrying the schema's field names
+   * @param rows one list of values per row
+   * @return the virtual table
+   */
+  public static VirtualTable create(
+      RelOptCluster cluster, RelDataType rowType, List<? extends List<RexNode>> rows) {
+    return create(cluster, rowType, ImmutableSet.of(), rows);
+  }
+
+  /**
+   * Creates a virtual table whose rows resolve against the given correlation variables.
+   *
+   * @param cluster the cluster this relation belongs to
+   * @param rowType the table's row type, carrying the schema's field names
+   * @param variablesSet the correlation variables the rows resolve against
+   * @param rows one list of values per row
+   * @return the virtual table
+   */
+  public static VirtualTable create(
+      RelOptCluster cluster,
+      RelDataType rowType,
+      Set<CorrelationId> variablesSet,
+      List<? extends List<RexNode>> rows) {
+    return new VirtualTable(
+        cluster, cluster.traitSetOf(Convention.NONE), rowType, variablesSet, rows);
   }
 
   /**
@@ -61,18 +119,50 @@ public class VirtualTable extends AbstractRelNode {
    *
    * @return one list of values per row
    */
-  public List<? extends List<RexNode>> getRows() {
+  public List<List<RexNode>> getRows() {
     return rows;
   }
 
+  /**
+   * Returns the correlation variables the rows resolve against.
+   *
+   * <p>A row holding a {@link org.apache.calcite.rex.RexSubQuery} that binds an outer reference is
+   * unreachable by {@code SubQueryRemoveRule}, whose operands are a projection, a filter and a
+   * join, and by {@code RelDecorrelator}: a consumer's planner leaves it unexpanded, and the
+   * variables it resolves against have to travel with the relation that holds it.
+   *
+   * @return the correlation variables
+   */
   @Override
-  protected RelDataType deriveRowType() {
-    return rowType;
+  public Set<CorrelationId> getVariablesSet() {
+    return variablesSet;
   }
 
   @Override
   public double estimateRowCount(RelMetadataQuery mq) {
     return rows.size();
+  }
+
+  /**
+   * Returns the cost of this relation, which is the cost of what it expands into.
+   *
+   * <p>The inherited cost is a row count alone, which is cheaper than the projection per row the
+   * expansion builds -- a cost-based planner would fire {@link
+   * io.substrait.isthmus.calcite.rel.rules.VirtualTableExpansionRule} and then keep the unexpanded
+   * relation it started from.
+   *
+   * @param planner the planner asking
+   * @param mq the metadata query
+   * @return the cost of this relation
+   */
+  @Override
+  public RelOptCost computeSelfCost(RelOptPlanner planner, RelMetadataQuery mq) {
+    // Three relations per row -- an empty table, the projection over it, and the row's share of the
+    // union -- each costing its own row and the values it computes.
+    double relations = 3 * rows.size();
+    return planner
+        .getCostFactory()
+        .makeCost(relations, relations * (rowType.getFieldCount() + 1), 0);
   }
 
   /**
@@ -82,23 +172,32 @@ public class VirtualTable extends AbstractRelNode {
    * not pass one on keeps its expressions out of everything built on that -- finding the subqueries
    * that bind outer references among them.
    *
+   * <p>A rewrite that retypes a value retypes the column it stands in, so the row type is rebuilt
+   * from the rewritten rows where they no longer carry the declared types. The names of a rebuilt
+   * column are the expression's own below the top level, which is the most a relation can say
+   * without the schema that named them.
+   *
    * @param shuttle the rewrite to apply
    * @return this table with the rewritten rows, or itself where nothing changed
    */
   @Override
   public RelNode accept(RexShuttle shuttle) {
-    boolean changed = false;
     List<List<RexNode>> rewritten = new ArrayList<>(rows.size());
+    boolean changed = false;
     for (List<RexNode> row : rows) {
-      List<RexNode> rewrittenRow = new ArrayList<>(row.size());
-      for (RexNode value : row) {
-        RexNode visited = value.accept(shuttle);
-        changed |= visited != value;
-        rewrittenRow.add(visited);
-      }
+      List<RexNode> rewrittenRow = shuttle.apply(row);
+      changed |= rewrittenRow != row;
       rewritten.add(rewrittenRow);
     }
-    return changed ? new VirtualTable(getCluster(), getTraitSet(), rowType, rewritten) : this;
+    if (!changed) {
+      return this;
+    }
+    RelDataType rewrittenType =
+        rewritten.stream().allMatch(row -> RexUtil.compatibleTypes(row, rowType, Litmus.IGNORE))
+            ? rowType
+            : RexUtil.createStructType(
+                getCluster().getTypeFactory(), rewritten.get(0), rowType.getFieldNames(), null);
+    return new VirtualTable(getCluster(), getTraitSet(), rewrittenType, variablesSet, rewritten);
   }
 
   /**
@@ -111,6 +210,8 @@ public class VirtualTable extends AbstractRelNode {
   public RelWriter explainTerms(RelWriter pw) {
     return super.explainTerms(pw)
         .itemIf("type", rowType, pw.getDetailLevel() == SqlExplainLevel.DIGEST_ATTRIBUTES)
+        .itemIf("type", rowType.getFieldList(), pw.nest())
+        .itemIf("variablesSet", variablesSet, !variablesSet.isEmpty())
         .item(
             "rows",
             rows.stream()
@@ -135,6 +236,6 @@ public class VirtualTable extends AbstractRelNode {
     if (!inputs.isEmpty()) {
       throw new IllegalArgumentException("VirtualTable takes no inputs, but got " + inputs.size());
     }
-    return new VirtualTable(getCluster(), traitSet, rowType, rows);
+    return new VirtualTable(getCluster(), traitSet, rowType, variablesSet, rows);
   }
 }
--- a/isthmus/src/main/java/io/substrait/isthmus/calcite/rel/VirtualTable.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/calcite/rel/VirtualTable.java
@@ -175,7 +175,9 @@ public class VirtualTable extends AbstractRelNode {
       }
       RelDataType columnType = typeFactory.leastRestrictive(valueTypes);
       // Nothing in Calcite's type system unifies every pair -- a rewrite that leaves two rows with
-      // no common type keeps the declared one, and the constructor is left to report the row.
+      // no common type keeps the declared one, which then fits neither. The constructor checks the
+      // count and not the types, so the expansion is where that surfaces: Project.isValid with
+      // assertions on, and a projection whose declared type does not fit its expressions without.
       columnTypes.add(
           columnType == null ? rowType.getFieldList().get(column).getType() : columnType);
     }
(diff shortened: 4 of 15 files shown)


nielspardon:
Thanks for digging into this. I'd like to propose a different approach before you put more into this one, because I think the shape matcher will keep costing us.

What I'd like to try instead: an isthmus-owned Calcite relation carrying the virtual table's rows as `RexNode`s, emitted by `SubstraitRelNodeConverter` and recognized by `instanceof` in `visitOther` — the same pair `CreateTable`/`CreateView` already form. Recognition becomes identity, and the schema rides on the node instead of being rebuilt from a derived row type.

Then, for consumers who need stock Calcite, a `HepPlanner` rule that expands the node into `LogicalValues` / `LogicalUnion` of `LogicalProject`s — opt-in, never run by isthmus itself, and explicitly one-way: expand and convert back and you get the `Project` over an empty table again. Most of the forward-conversion work in this PR moves into that rule rather than being thrown away.

The reason I'd rather not refine the matcher: it accepts a strictly larger set of unions than the emitter produces, and that gap is doing real damage — arms differing only in nullability or width now throw `IllegalArgumentException` out of `visit(Union)` where they used to convert to a `Set`, which `Set.deriveRecordType` explicitly tolerates. `handWrittenUnionOfSingleRowProjectionsBecomesAVirtualTable` doesn't catch that because it never reaches the new code (literal-only arms get folded into one `LogicalValues`, so the reverse goes through `visit(Values)` — reverting both files leaves it green). And the recognition doesn't survive planning: `UNION_REMOVE` deletes the one-input union a single-row table emits, restoring the exact shape this fixes, while `UNION_MERGE` rewrites *into* the matching shape. Carrying the schema on the node also reaches the nested-struct case #631 is actually about, which the current `NamedStruct.names()`-against-top-level-types pairing can't express in either direction.

alexandrefimov:
All three check out; measurements below so nobody has to redo them.

The test is vacuous as you say. Reverting both main files leaves `handWrittenUnionOfSingleRowProjectionsBecomesAVirtualTable` green, and the reason is the one you gave — the hand-written union converts to a single `LogicalValues(tuples=[[{ 1 }, { 2 }]])`, so the reverse goes through `visit(Values)` and never reaches the matcher. `expressionVirtualTableUnderAProject` and `expressionContainingVirtualTable` do fail on revert, so the PR is not uncovered, just not where that test claims it is.

The nullability gap is a live regression rather than a corner. A union of two single-row projections differing only in nullability:

```
origin/main:   Set
this PR:       IllegalArgumentException: Row field type (I32{nullable=false})
               does not match schema field type (I32{nullable=true})
```

The matcher takes the schema from `union.getRowType()` and the rows from the arms' own expressions, so anything the least-restrictive type widened trips `VirtualTableScan`'s check.

And it does not survive planning. A single-row virtual table emits a one-input union:

```
LogicalUnion(all=[true])
  LogicalProject(c=[CAST(7:BIGINT):INTEGER NOT NULL])
    LogicalValues(tuples=[[{  }]])
```

`UNION_REMOVE` strips it, and what is left converts back to a `Project` — the shape this PR set out to fix.

So I would rather rework it your way than refine the matcher: recognition by identity cannot over-match, a rule cannot erase it, and carrying the schema on the node is the part that reaches the nested-struct case. I will take that on.

One thing I would like your read on before I start. #1153 changes `SubstraitRelNodeConverter.visit(VirtualTableScan)` — the same method the new emitter would replace — to fix a bare `AssertionError` on a virtual table with a struct column. Does that still stand on its own against the new shape, or should it fold into the rework?

nielspardon:
> One thing I would like your read on before I start. #1153 changes `SubstraitRelNodeConverter.visit(VirtualTableScan)` — the same method the new emitter would replace — to fix a bare `AssertionError` on a virtual table with a struct column. Does that still stand on its own against the new shape, or should it fold into the rework?

I did post review comments on #1153 just now. We can tackle that on its own.

alexandrefimov:
Reworked the way you proposed and pushed. The branch sits on top of #1153, so the first two commits here are that one.

Two things the description does not carry. One cost the shape match never had: a relation holding `RexNode`s has to pass a `RexShuttle` on, or every expression rewrite walks past the rows -- including the scan that finds the subqueries binding outer references, which the old expansion got for free from `visit(Project)`. A shuttle applied to the node found zero subqueries and rebuilt nothing before I added it. `VirtualTable.accept(RexShuttle)` and a case in `OuterReferenceResolver` cover it, pinned by `correlationInsideASubqueryInAVirtualTableRow`.

The other is a gain: the struct-column round trip comes back here. #1153 has to give it up, since a row literal cannot sit in a Values tuple and the column takes the projection encoding there, while on this branch `structColumnConverts` and `severalStructColumnsConvert` are back on `assertFullRoundTrip`.

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
```diff
@@ -850,43 +851,16 @@ public RelNode visit(VirtualTableScan virtualTableScan, Context context) {
...
-        topProjectExprs.add(rexBuilder.makeInputRef(union, i));
-      }
-      return LogicalProject.create(
-          union, Collections.emptyList(), topProjectExprs, rowType, Collections.emptySet());
+      // A row that does not fit a LogicalValues tuple keeps its expressions, in a relation of our
+      // own: Calcite has none that holds them, and expanding the table into a projection per row
+      // does not come back -- the projection is what converts back, and the table is gone. A
+      // consumer whose planner only knows Calcite's own relations can expand it with
+      // VirtualTableExpansionRule.
+      return new VirtualTable(
```
nielspardon: Register the rule in `SubstraitToSql.convert` and `SubstraitSqlDialect.toSql`, and say in the BREAKING CHANGE note that SQL generation needs it. `RelToSqlConverter.visit(RelNode)` throws unconditionally, so `-da` doesn't help, and `PlanTestBase.toSql` plus the four `examples/isthmus-api` entry points all route through those two methods.

nielspardon: Give `VirtualTable` a `variablesSet` and pass it through in `expand` instead of `Collections.emptySet()`. A `RexSubQuery` in a row is unreachable by `SubQueryRemoveRule` (Project/Filter/Join operands only) and by `RelDecorrelator`, so a consumer's planner leaves it unexpanded.

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelVisitor.java
```diff
@@ -931,6 +932,26 @@ public Rel handleCreateView(CreateView createView) {
...
+   */
+  public Rel handleVirtualTable(VirtualTable virtualTable) {
+    List<Expression.NestedStruct> rows = new ArrayList<>(virtualTable.getRows().size());
+    for (List<RexNode> row : virtualTable.getRows()) {
+      List<Expression> fields =
+          row.stream().map(this::toExpression).collect(Collectors.toUnmodifiableList());
+      rows.add(ExpressionCreator.nestedStruct(false, fields));
+    }
+    return VirtualTableScan.builder()
+        .initialSchema(typeConverter.toNamedStruct(virtualTable.getRowType()))
```
nielspardon, line 1017: Convert each row value at `virtualTable.getRowType().getFieldList().get(i).getType()`, as `visit(Values)` does. Taking the schema from the row type and the field types from the expressions reintroduces #1064 on this path: a computed field inside a nullable struct fails `VirtualTableScan.check`. A computed variant of the nullable-struct test would pin it.

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelVisitor.java
```diff
@@ -945,6 +966,9 @@ public Rel visitOther(RelNode other) {
 
     } else if (other instanceof CreateView) {
       return handleCreateView((CreateView) other);
+
+    } else if (other instanceof VirtualTable) {
```
nielspardon: Add a `visit(VirtualTable)` to `RelNodeVisitor.reverseAccept` alongside the existing overloads. `visitOther` is also the documented consumer extension hook, and the sanctioned pattern in `SubtraitRelVisitorExtensionTest` throws rather than calling `super` — so a consumer copying it now fails on a plain computed-VALUES plan with no compile-time signal.

isthmus/src/main/java/io/substrait/isthmus/calcite/rel/VirtualTable.java
```diff
@@ -0,0 +1,140 @@
...
+    for (List<RexNode> row : rows) {
+      List<RexNode> rewrittenRow = new ArrayList<>(row.size());
+      for (RexNode value : row) {
+        RexNode visited = value.accept(shuttle);
+        changed |= visited != value;
+        rewrittenRow.add(visited);
+      }
+      rewritten.add(rewrittenRow);
+    }
+    return changed ? new VirtualTable(getCluster(), getTraitSet(), rowType, rewritten) : this;
```
nielspardon: Re-derive the row type from the rewritten rows, preserving the schema's depth-first names. `Project.accept` does this and keeps only the field names, but a literal copy would drop the nested names #1153 added. `aShuttleReachesTheRows` substitutes a same-typed literal, so it can't catch this.

isthmus/src/main/java/io/substrait/isthmus/calcite/rel/VirtualTable.java
```diff
@@ -0,0 +1,140 @@
...
+    return rows;
+  }
+
+  @Override
+  protected RelDataType deriveRowType() {
+    return rowType;
+  }
+
+  @Override
+  public double estimateRowCount(RelMetadataQuery mq) {
```
nielspardon, line 98: Override `computeSelfCost` as well — the inherited `makeCost(rowCount, rowCount, 0)` makes the unexpanded relation cheaper than its own expansion, so a cost-based planner fires the rule and then discards the result.

isthmus/src/main/java/io/substrait/isthmus/calcite/rel/VirtualTable.java
```diff
@@ -0,0 +1,140 @@
...
+
+  /**
+   * VirtualTable constructor.
+   *
+   * @param cluster the cluster this relation belongs to
+   * @param traitSet the relation's traits
+   * @param rowType the table's row type, carrying the schema's field names
+   * @param rows one list of values per row, each value at the type its column is declared at
+   */
+  public VirtualTable(
```
nielspardon, line 51: Validate the rows against `rowType` here and throw `IllegalArgumentException` — arity, per-column type, distinct names. The deleted `LogicalProject.create` gave us `RexUtil.compatibleTypes` for free, and `AbstractRelNode.isValid` returns `litmus.succeed()`, so nothing checks it now. Same call as #1047.

isthmus/src/main/java/io/substrait/isthmus/calcite/rel/rules/VirtualTableExpansionRule.java
```diff
@@ -0,0 +1,137 @@
...
+    }
+
+    RelDataType emptyRowType = cluster.getTypeFactory().createStructType(List.of(), List.of());
+    ImmutableList<ImmutableList<RexLiteral>> singleEmptyRow = ImmutableList.of(ImmutableList.of());
+
+    List<RelNode> rowProjects = new ArrayList<>();
+    for (List<RexNode> row : virtualTable.getRows()) {
+      RelNode emptyRow = LogicalValues.create(cluster, emptyRowType, singleEmptyRow);
+      rowProjects.add(
+          LogicalProject.create(
```
nielspardon, line 113: Uniquify the field names via `SqlValidatorUtil.F_SUGGESTER`, or reject duplicates in the `VirtualTable` constructor. `NamedStruct.names` is spec'd only as a dfs-ordered list with no uniqueness requirement and `VirtualTableScan.check` doesn't reject repeats, while `Project.isValid` requires them distinct — so such a table converts and round-trips but fails in the escape hatch, where it used to fail at conversion.

isthmus/src/main/java/io/substrait/isthmus/OuterReferenceResolver.java
```diff
@@ -115,6 +116,12 @@ public RelNode visit(Correlate correlate) throws RuntimeException {
 
   @Override
   public RelNode visitOther(RelNode other) throws RuntimeException {
+    if (other instanceof VirtualTable) {
```
nielspardon: Drop the type test and scan unconditionally. `AbstractRelNode.accept(RexShuttle)` returns `this` and the result is discarded here, and it would also cover a `Join`/`Calc`/`Sort` condition holding a subquery, which currently dies with the misleading "only partially decorrelated" message.

```suggestion
    other.accept(rexVisitor);
```

isthmus/src/main/java/io/substrait/isthmus/calcite/rel/VirtualTable.java
```diff
@@ -0,0 +1,140 @@
...
+   * Returns the table's rows.
+   *
+   * @return one list of values per row
+   */
+  public List<? extends List<RexNode>> getRows() {
+    return rows;
+  }
+
+  @Override
+  protected RelDataType deriveRowType() {
```
nielspardon: `deriveRowType` is unreachable — `getRowType()` is final and consults it only when the field is null.

```suggestion
```
Also in this file: `accept` can use `shuttle.apply(row)` instead of the inner loop (and its "or itself where nothing changed" branch is untested); `explainTerms` is missing Values's `.itemIf("type", rowType.getFieldList(), pw.nest())`, which is why the rewritten `LiteralNullabilityRoundtripTest` assertions no longer check the column name; declaring the field `ImmutableList<List<RexNode>>` drops the wildcard from `getRows()`; and a `create(cluster, rowType, rows)` factory removes the duplicated `traitSetOf(Convention.NONE)` and the `Convention` import from both call sites.

isthmus/src/main/java/io/substrait/isthmus/calcite/rel/rules/VirtualTableExpansionRule.java
```diff
@@ -0,0 +1,137 @@
...
+ *
+ * <p>This is for a consumer whose planner only knows Calcite's own relations. It is one-way: the
+ * expansion is a plan like any other, and converting it back to Substrait gives the relation it is,
+ * not the virtual table it came from. Isthmus never runs it -- a plan it converts keeps the {@link
+ * VirtualTable}, which is what makes the round trip exact.
+ */
+public class VirtualTableExpansionRule extends RelRule<VirtualTableExpansionRule.Config> {
+
+  /** The rule instance to add to a planner. */
+  public static final VirtualTableExpansionRule INSTANCE = Config.DEFAULT.toRule();
```
nielspardon: Make `Config` a `@Value.Immutable interface` with `@Value.Enclosing` on the rule and move `INSTANCE` to a holder — `RelRule`'s guidelines 2 and 5, and the processor is already wired. Also: `relBuilderFactory` (108) is inert since `expand` is static and never calls `call.builder()`, and the per-row `LogicalValues` (66) can be hoisted.

isthmus/src/test/java/io/substrait/isthmus/VirtualTableTest.java
```diff
@@ -0,0 +1,265 @@
...
+  }
+
+  /**
+   * The unexpanded relation is opaque to the rules that rewrite the expansion -- UNION_REMOVE
+   * strips the one-input union a single-row table would expand to, and PROJECT_MERGE takes a row's
+   * projection into whatever sits above it -- so the table is still a table after a planner has run
+   * over it.
+   */
+  @Test
+  void theRelationSurvivesPlanning() {
```
nielspardon, line 107: Plan the `Project`-over-`VirtualTable` tree that `theTableIsStillATableUnderAProjection` builds. As written the assertions pass with an empty rule set — the planner gets a bare leaf, so none of the three rules can match, and neither the Javadoc's claim nor the body's "both measured" is tested.

isthmus/src/test/java/io/substrait/isthmus/VirtualTableTest.java
```diff
@@ -0,0 +1,265 @@
...
+   * expansion is, not the table it came from. That is the reason isthmus does not run the rule
+   * itself.
+   */
+  @Test
+  void theExpansionDoesNotConvertBackToAVirtualTable() {
+    RelNode expanded =
+        plan(substraitToCalcite.convert(computedRows()), VirtualTableExpansionRule.INSTANCE);
+
+    assertInstanceOf(
+        io.substrait.relation.Set.class, SubstraitRelVisitor.convert(expanded, extensions));
```
nielspardon: Use `convert(rel, converterProvider)` — the `ExtensionCollection` overload is `@Deprecated` (97, 116, 156, 179, 200). Move `virtualTable(...)` (248) and `plan(...)` (256) to `PlanTestBase`, drop `aHandWrittenUnionOfSingleRowProjectionsStaysAUnion` (167) as redundant with line 92, and import `Set`.

```suggestion
        Set.class, SubstraitRelVisitor.convert(expanded, converterProvider));
```

isthmus/src/test/java/io/substrait/isthmus/OuterReferenceResolverTest.java
```diff
@@ -129,6 +133,40 @@ void nestedApplyJoinQuery() throws SqlParseException {
     Assertions.assertNotEquals(anchor0, anchor1);
   }
 
+  /**
+   * A virtual table's rows are expressions, and a subquery among them is a tree the resolver only
+   * reaches through the row: a subquery's relation is not an input, so walking inputs never gets
+   * there and a correlation declared inside it is left unbound.
+   */
+  @Test
+  void correlationInsideASubqueryInAVirtualTableRow() {
```
nielspardon, line 149: Build a plan that converts — this one throws `UnsupportedOperationException: Unable to handle node: LogicalCorrelate`. As it stands it shows an anchor being allocated rather than a correlated subquery in a row reaching Substrait, which is what the `accept(RexShuttle)` work is for.

isthmus/src/test/java/io/substrait/isthmus/VirtualTableScanTest.java
```diff
@@ -256,10 +248,7 @@ void structColumnWithAComputedField() {
 
     RelNode relNode = substraitToCalcite.convert(virtualTableScan);
     assertEquals(
-        "LogicalProject(inputs=[0])\n"
-            + "  LogicalUnion(all=[true])\n"
-            + "    LogicalProject(exprs=[[ROW(*(6, 2), 2.0E0:DOUBLE)]])\n"
-            + "      LogicalValues(type=[RecordType()], tuples=[[{  }]])\n",
+        "VirtualTable(type=[RecordType(RecordType(INTEGER a, DOUBLE b) outer)], rows=[[{ ROW(*(6, 2), 2.0E0:DOUBLE) }]])\n",
         explain(relNode));
```
nielspardon, line 250: Add `assertFullRoundTrip` here and at 261, 370 and 385 — all four round-trip today. And give `nullableStructInsideStructColumnConverts` (406) the note its sibling at 280 has, so the deliberate omissions are distinguishable from the accidental ones.

nielspardon (changes requested):
Run `VirtualTableExpansionRule` inside `SubstraitToSql` and `SubstraitSqlDialect.toSql` before `RelToSqlConverter` — both now throw on any plan with a computed virtual table, and that's the call I'd like your read on, since isthmus's own SQL renderer turns out to be the consumer the rule exists for.

Two more blockers inline: a computed field inside a nullable struct crashes converting back, and `accept(RexShuttle)` keeps a stale `rowType`.

Please also rebase — main has moved five commits since your base — and drop the "first two commits here are #1153" line from the description, since it's merged and there are two commits of your own now.

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelVisitor.java
alexandrefimov: Applied for the literal values — `literalConverter.convert(value, rowFields.get(column).getType())`, the call `visit(Values)` makes — but your example needs a decision I would rather not take alone.

A computed field inside a nullable struct arrives as a `ROW(...)` call rather than a literal, so there is no literal to convert at the declared type, and it still fails on this branch:

```
Row field type    (Struct{nullable=true, fields=[I32{nullable=false}, FP64{nullable=false}]})
does not match
schema field type (Struct{nullable=true, fields=[I32{nullable=true}, FP64{nullable=true}]})
```

Calcite pushes the struct's nullability down into its fields, so the schema taken from the row type has nullable fields while the expression keeps its own. Two ways out that I can see: cast the value to the declared type here, which puts a cast in the plan that the plan we read did not have; or let `VirtualTableScan.check` accept a row value narrower than the column it stands in, which is the reading `valueAsDeclared` already takes on the way in through `fitsDeclared` — so the two directions disagree today. The second reads right to me, but it is a core change and a question about what the spec means by a virtual table's rows matching its schema, so I would rather have your call than pick one.

Leaving this open until then; a computed variant of the nullable-struct test comes with whichever way it goes.

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
alexandrefimov: Both entry points expand now, and the note says SQL generation runs the rule. Measured before and after: a computed table went into `AssertionError: Need to implement io.substrait.isthmus.calcite.rel.VirtualTable` from `RelToSqlConverter.visit(RelNode)`, which is an explicit throw rather than an assertion, so `-da` would not have helped. It renders now, and `sqlGenerationExpandsTheTable` pins both methods.

alexandrefimov: Carried on the relation and passed through in `expand`. It also turns out to override `RelNode.getVariablesSet` — PMD caught the missing `@Override` — so a consumer's own decorrelation sees the variables too, not just our expansion.

Nothing populates it on the conversion path yet: a virtual table has no input to anchor a scope to, so there is no `exitScope` to take the ids from. That is the same question as the `OuterReferenceResolverTest` thread below.

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelVisitor.java
alexandrefimov: Added, with `SubstraitRelVisitor` overriding it instead of matching in `visitOther`. A consumer copying the sanctioned pattern now gets the compile-time signal.

isthmus/src/main/java/io/substrait/isthmus/calcite/rel/VirtualTable.java
alexandrefimov: `shuttle.apply(row)` now, and the row type is rebuilt only where the rewritten values no longer fit the declared one: a substitution that keeps the types keeps the schema's names, nested ones included. Re-deriving unconditionally would need a third depth-first name walker, which is what #1178 is about — happy to take that on there instead.

isthmus/src/main/java/io/substrait/isthmus/calcite/rel/VirtualTable.java
alexandrefimov: Costed as what the expansion is: three relations per row, each with its own row and its values. Pinned by comparing the cumulative cost with the expansion's — before this the table came out at {2 rows, 6 cpu} against the expansion's {6 rows, 8 cpu}, so a cost-based planner did exactly what you describe.

isthmus/src/main/java/io/substrait/isthmus/calcite/rel/VirtualTable.java
alexandrefimov: Arity and `RexUtil.compatibleTypes`, both reported as `IllegalArgumentException` (the `Litmus.THROW` form raises an `AssertionError`, which reads as a bug in the caller rather than a rejected input). Distinct names are the one part I left out — see the next thread.

isthmus/src/main/java/io/substrait/isthmus/calcite/rel/rules/VirtualTableExpansionRule.java
alexandrefimov: Took `F_SUGGESTER` over rejecting duplicates in the constructor. A schema may repeat a name, and a literal-only table with two columns called `c` converts and round-trips today — #1189 made its emit mapping select by index for exactly that reason — so refusing one here would leave the two encodings of the same table disagreeing. `theExpansionUniquifiesRepeatedFieldNames` pins the names on both sides.

isthmus/src/main/java/io/substrait/isthmus/OuterReferenceResolver.java
alexandrefimov: Unconditional now, and the comment says why: a join, calc or sort condition holding a subquery is reached the same way.

isthmus/src/main/java/io/substrait/isthmus/calcite/rel/VirtualTable.java
alexandrefimov: All of it: `deriveRowType` gone, `accept` on `shuttle.apply`, `explainTerms` with Values's nested `type` item, the field declared `ImmutableList<List<RexNode>>` so `getRows()` drops the wildcard, and a `create(...)` factory that owns the convention — both call sites are through it now.

isthmus/src/test/java/io/substrait/isthmus/VirtualTableTest.java
alexandrefimov: Rebuilt around the `Project`-over-`VirtualTable` tree, and it earns its keep now: the assertion is that the table comes back as the input of the projection after `UNION_REMOVE`, `UNION_MERGE` and `PROJECT_MERGE` have run.

isthmus/src/test/java/io/substrait/isthmus/VirtualTableTest.java
alexandrefimov: All four: the deprecated overload is gone from every call, `virtualTable(...)` and `plan(...)` are on `PlanTestBase`, and the hand-written-union test is dropped as redundant.

isthmus/src/test/java/io/substrait/isthmus/VirtualTableScanTest.java
alexandrefimov: Added at all four — they do round-trip — and `nullableStructInsideStructColumnConverts` now carries its sibling's note, so the two deliberate omissions read as deliberate.

isthmus/src/main/java/io/substrait/isthmus/calcite/rel/rules/VirtualTableExpansionRule.java
alexandrefimov: The instance is in a holder and the per-row `LogicalValues` is hoisted; `relBuilderFactory` says in its Javadoc that it is inert and why.

The generated config I could not land: it copies `RelRule.Config#description()`, which Calcite declares nullable, so `ImmutableVirtualTableExpansionRule` imports `javax.annotation.Nullable` — and isthmus has no `javax.annotation` on its compile classpath. The build stops at `package javax.annotation does not exist`. The processor is wired, the annotations are not. Say the word and I will add jsr305 to `isthmus/build.gradle.kts` and switch it over; a dependency for one config class did not look like my call to make.

isthmus/src/test/java/io/substrait/isthmus/OuterReferenceResolverTest.java
alexandrefimov: Not rebuilt yet, and it is the same question as the row-types thread. The plan you are describing is one where the correlation is declared by the virtual table itself, and nothing populates `variablesSet` on the conversion path — a virtual table has no input to anchor a scope to, so there is no `exitScope` to take the ids from. Should `visit(VirtualTableScan)` open a scope of its own for the rows, or does an outer reference in a row belong to the enclosing relation the way it does elsewhere? I will build the test around whichever answer you give.

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelVisitor.java
nielspardon: Refuse the shape, with a message naming isthmus's limit the way the new projection guard does. The validator asserts a virtual table's derived type equals its `base_schema` using the non-promoting comparison, so relaxing `check` would have us emit plans it rejects — and casting at the declared type puts a cast in the output the input did not have. `fitsDeclared`'s inbound leniency is the side that disagrees with everyone else; I will file that separately.

isthmus/src/test/java/io/substrait/isthmus/OuterReferenceResolverTest.java
nielspardon: Do not open a scope — build the test with the virtual table inside a correlated join, so the enclosing relation binds the id and a row's subquery references it. `bind` associates an id with a target whose fields the reference names, and a zero-input leaf has none, so a subquery in a row sits exactly where one in a `Project` expression does. That leaves `variablesSet` as something only a consumer populates, which is worth a line in its Javadoc.

isthmus/src/main/java/io/substrait/isthmus/calcite/rel/rules/VirtualTableExpansionRule.java
nielspardon: Add `@Value.Style(allowedClasspathAnnotations = {Override.class})` and the generated config should compile — that is the setting `io.substrait`'s `package-info.java` already uses in core, and it drops the `javax.annotation.Nullable` that `RelRule.Config#description()` carries rather than needing it on the classpath. No new dependency, so nothing to weigh separately. If it still does not build, keep the hand-written form and say so here.

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelVisitor.java
alexandrefimov: Refused there, with the value and both types in the message. The computed-field case is pinned as the shape that reaches it: a `ROW` call inside a nullable struct, where Calcite's pushed-down nullability and the expression's own type have no meeting point that is not a cast. Literals keep going through `literalConverter.convert(value, declaredType)`, so the only values that reach the guard are the ones assembled from expressions.

`fitsDeclared` left as it is for your issue.

isthmus/src/main/java/io/substrait/isthmus/calcite/rel/rules/VirtualTableExpansionRule.java
alexandrefimov: It still does not build, so the hand-written form stays. `allowedClasspathAnnotations = {Override.class}` on the config, and again on the enclosing class, leaves the annotation in the builder method that copies from the supertype — `@Nullable String descriptionValue = instance.description();`, twice, with the import at the top and nothing else using it. Overriding `description()` without Calcite's annotation, both as a mandatory attribute and as a `@Value.Default`, takes it off the field and the accessor but not out of that copy; `nullableAnnotation` and `fallbackNullableAnnotation` do not reach it either. The javadoc says which knobs were tried, so the next reader does not have to repeat it.

Leaving this open in case you would rather have jsr305 on the compile classpath and the generated config after all — the guidelines are on your side, it is the dependency I did not want to add on my own.

isthmus/src/test/java/io/substrait/isthmus/OuterReferenceResolverTest.java
alexandrefimov: Rebuilt that way. The assertion needed a second correlation to be worth anything: with only the correlate's own id, `visit(Correlate)` binds it whether the row is walked or not. So the row's subquery is a filter declaring an id of its own, over a scan the filter's own subquery references, and the test asserts both anchors and that they differ. Taking `other.accept(rexVisitor)` back out fails it, and nothing else in the suite.

The plan still does not convert, and not because of the row: `SubstraitRelVisitor.visit(Correlate)` ends in `super.visit`, so every `LogicalCorrelate` lands in `visitOther` and "Unable to handle node". That is why the three tests above it in this file stop at the resolver as well.

`getVariablesSet` now says only a consumer populates it, and why a conversion has nothing to put there.

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelVisitor.java
nielspardon: Scratch the separate issue — this is #1154, and `fitsDeclared`'s leniency is the accommodation for it rather than a disagreement: Calcite pushes a struct's nullability into its fields, so a required value does have to stand in a column we were forced to declare nullable. The refusal you added stays right for the outbound direction, where we would otherwise emit rows and a schema that disagree. Nothing to change here.

isthmus/src/main/java/io/substrait/isthmus/calcite/rel/rules/VirtualTableExpansionRule.java
```diff
@@ -0,0 +1,200 @@
...
+   * has no case for a relation it does not know and throws an {@link AssertionError} naming it.
+   *
+   * @param relNode the tree to expand
+   * @return the tree with every virtual table expanded
+   */
+  public static RelNode expandAll(RelNode relNode) {
+    HepPlanner planner =
+        new HepPlanner(new HepProgramBuilder().addRuleInstance(instance()).build());
+    planner.setRoot(relNode);
+    return planner.findBestExp();
```
nielspardon: Give isthmus its own `RelToSqlConverter` subclass with a `visit(VirtualTable)` and drop `expandAll` from the call sites: `HepPlanner` walks `getInputs()` and never enters `RexSubQuery.rel`, so a table inside a subquery arrives unexpanded and both `SubstraitToSql.convert` and `SubstraitSqlDialect.toSql` throw `AssertionError: Need to implement io.substrait.isthmus.calcite.rel.VirtualTable` where main emitted SQL. `SqlImplementor` converts a subquery's relation through `visitRoot`, so one reflective `visit` covers every position a table can appear in.

That also keeps this class's "Isthmus never runs it" true, and reaches `ToOptimizedSql` in `examples/isthmus-api`, which builds a bare `RelToSqlConverter` and is the one call site `expandAll` was never added to.

isthmus/src/main/java/io/substrait/isthmus/calcite/rel/VirtualTable.java
```diff
@@ -0,0 +1,245 @@
...
+   * <p>Only a consumer populates it. A conversion never does: an id is bound to a relation whose
+   * fields the reference names, and a leaf with no inputs has none, so an outer reference in a row
+   * belongs to the relation around the table the way one in a projection's expression does.
+   *
+   * @return the correlation variables
+   */
+  @Override
+  public Set<CorrelationId> getVariablesSet() {
+    return variablesSet;
+  }
```
nielspardon: Drop this override and the field it reads. `RelNode.getVariablesSet()` is documented as "variables that are set in this relational expression but also used and therefore not available to parents", and `RelOptUtil.CorrelationCollector` does `vuv.variables.removeAll(other.getVariablesSet())` — so reporting the ids the rows *resolve against* makes `getVariablesUsed` come back empty and `notContainsCorrelation` answer **true** for a correlation that is right there in the rows.

The `accept(RexShuttle)` override already carries them without this: `RelOptUtil.VariableUsedVisitor` is a `RexShuttle`, and `CorrelationCollector` hands it to every relation, collecting each `RexCorrelVariable` out of the rows. The second paragraph of this Javadoc has the reason a leaf cannot populate it correctly — a leaf has no input whose fields a reference could name.

Same set, same fix at `VirtualTableExpansionRule.java:125`: copying it onto a `LogicalProject` over a zero-field `LogicalValues` trips `AssertionError: field ordinal 0 out of range {t=RecordType()}`. Pass `ImmutableSet.of()`.

isthmus/src/main/java/io/substrait/isthmus/calcite/rel/VirtualTable.java
```diff
@@ -0,0 +1,245 @@
...
+            String.format(
+                "A virtual table's row has %d values where its type declares %d columns: %s",
+                row.size(), rowType.getFieldCount(), row));
+      }
+      if (!RexUtil.compatibleTypes(row, rowType, Litmus.IGNORE)) {
+        throw new IllegalArgumentException(
+            String.format(
+                "A virtual table's row %s does not fit the type %s its columns are declared at",
+                row, rowType.getFullTypeString()));
+      }
```
nielspardon: Drop the `compatibleTypes` check and keep the field-count one; the leniency it removes is deliberate. `SubstraitRelNodeConverter.fitsDeclared` accepts a non-nullable value in a nullable column on the grounds that "Calcite makes a struct's fields nullable along with the struct, so the fields of a nullable struct are declared that way whatever the values in them are" — so a nullable struct column with a computed row now throws out of the constructor where main converted it.

```suggestion
      // Nothing else checks the count: AbstractRelNode.isValid succeeds unconditionally. The value
      // types are deliberately not checked -- the conversion accepts a non-nullable value in a
      // nullable column, which is how a row of a nullable struct column reaches here.
      if (row.size() != rowType.getFieldCount()) {
        throw new IllegalArgumentException(
            String.format(
                "A virtual table's row has %d values where its type declares %d columns: %s",
                row.size(), rowType.getFieldCount(), row));
      }
```

isthmus/src/main/java/io/substrait/isthmus/calcite/rel/VirtualTable.java
```diff
@@ -0,0 +1,245 @@
...
+      rewritten.add(rewrittenRow);
+    }
+    if (!changed) {
+      return this;
+    }
+    RelDataType rewrittenType =
+        rewritten.stream().allMatch(row -> RexUtil.compatibleTypes(row, rowType, Litmus.IGNORE))
+            ? rowType
+            : RexUtil.createStructType(
+                getCluster().getTypeFactory(), rewritten.get(0), rowType.getFieldNames(), null);
```
nielspardon: Derive the rebuilt type from every row rather than `rewritten.get(0)`: a shuttle that retypes a value in the first row need not retype the same column in the next, and the type built from row 0 alone then disagrees with the rows the constructor is about to check — an `IllegalArgumentException` thrown out of a rewrite. Take the least-restrictive type per column across the rewritten rows instead, so the result fits all of them.

isthmus/src/main/java/io/substrait/isthmus/OuterReferenceResolver.java
```diff
@@ -115,6 +115,12 @@ public RelNode visit(Correlate correlate) throws RuntimeException {
 
   @Override
   public RelNode visitOther(RelNode other) throws RuntimeException {
+    // A relation's own expressions can hold a subquery binding outer references, and a subquery's
+    // relation is not an input, so walking inputs never reaches it. Filter and Project scan theirs
+    // before they get here; a virtual table's rows, and a join, calc or sort condition, arrive
+    // here. AbstractRelNode.accept(RexShuttle) returns the relation itself, so the result is the
+    // one already in the tree.
+    other.accept(rexVisitor);
```
nielspardon, line 125: The last sentence is wrong, and what it gets wrong is the gap. `Join`, `Calc`, `Sort`, `Filter`, `Project`, `Snapshot`, `TableFunctionScan` and the new `VirtualTable` all return a **copy** from `accept(RexShuttle)`; discarding it is safe because `RexVisitor.visitSubQuery` returns the same instance, so no copy is ever made. The relations that do inherit `AbstractRelNode.accept` are exactly the ones this misses — `TableModify`, `Window`, `Match` and `Aggregate` never hand their expressions over, so a correlated subquery in an `UPDATE ... SET` value stays unscanned, as it is on main.

```suggestion
    // A relation's own expressions can hold a subquery binding outer references, and a subquery's
    // relation is not an input, so walking inputs never reaches it. Filter and Project scan theirs
    // before they get here; a virtual table's rows, and a join, calc or sort condition, arrive
    // here. The rewrite is an identity -- RexVisitor.visitSubQuery returns the subquery it was
    // given -- so the relation the tree holds is unchanged and the result is discarded. Relations
    // that do not override accept(RexShuttle) at all -- TableModify, Window, Match, Aggregate --
    // keep their expressions to themselves; a correlation among those is still unresolved.
    other.accept(rexVisitor);
```

nielspardon (commented):
Five things, three of them blocking: a virtual table inside a subquery still reaches `RelToSqlConverter` unexpanded, `getVariablesSet` reports the rows' outer references as variables the table *sets*, and the constructor's type check rejects rows the converter deliberately allows. I'll file the rest as issues — the shared empty row inflating cumulative cost, `Config`'s order-dependent static init, and the two tests that assert less than the ones they replaced.

isthmus/src/main/java/io/substrait/isthmus/calcite/rel/VirtualTable.java
alexandrefimov: Taking both -- the override with its field, and `ImmutableSet.of()` at 125.

isthmus/src/main/java/io/substrait/isthmus/calcite/rel/rules/VirtualTableExpansionRule.java
alexandrefimov: Taking the subclass.

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
alexandrefimov: Wrong: populating the set is what empties `getVariablesUsed`, and I had not run it before writing that. Dropping the override, https://github.com/substrait-io/substrait-java/pull/1151#discussion_r3921898901.

isthmus/src/main/java/io/substrait/isthmus/calcite/rel/VirtualTable.java
alexandrefimov: Taking it. It does not bring the nullable-struct table back, though: with the check gone, converting it still fails at `SubstraitRelVisitor:971`, a guard this PR adds. So the choice is refusing that value or casting it to its column's type; I would keep the refusal and name it in the breaking-change note -- a cast puts a `RexNode` in the output the plan did not carry -- unless you would rather it converted.

```
A virtual table's value ROW(*(6, 2), 2.0E0:DOUBLE) converts to
Struct{nullable=true, fields=[I32{nullable=false}, FP64{nullable=false}]} where its column is
declared Struct{nullable=true, fields=[I32{nullable=true}, FP64{nullable=true}]}
```

alexandrefimov: One clause of the code comment is wrong, though: a row of a nullable struct column does not reach the constructor as a non-nullable value in a nullable column. `renamedAsDeclared` rebuilds the `ROW` with `makeCall(declaredType, ROW, ...)`, so by the time the constructor sees that table the value's type and the column's are `equals`. The clause goes out, the rest stays.

nielspardon: Keep the refusal, and name it in the note — though not on the grounds I gave earlier in this PR, which leaned on the validator.

The spec does not answer this directly. `ReadRel.VirtualTable` carries one line of comment ("A table composed of expressions"), `base_schema` carries none, and the read-relation page says nothing about the two agreeing, so whether a row's expression may be narrower than the column it stands in is undefined for a virtual table.

Where the spec does address this exact shape, though, it requires a cast. `DISCRETE` nullability in `expressions/scalar_functions.md`: "if a type input is declared as `i64?` and one has an `i64` literal, the `i64` literal must be cast to `i64?` to allow the operation to bind." Substrait defaults to explicit casting rather than implicit throughout — the design principles put it as "excluding concepts like overloading, type coercion, implicit casting" — so a required value standing in a nullable column without a cast is the shape the spec declines to make implicit everywhere it has had occasion to say.

Which makes your reason the right one, and it needs no appeal to the spec at all: a cast puts an expression in the output the input never carried. Silently widening instead would have isthmus encode a promotion rule the spec never stated; an `UnsupportedOperationException` naming isthmus's limit says only what is true. It is also the reversible direction — if the spec later allows a narrower row, the guard comes out.

Two asks. Name the refusal in the breaking-change note, and say in the description that the spec does not define whether a virtual table's row may be narrower than its column — that gap is worth sharpening in the spec rather than settling here, and I will open a spec issue for it unless you would rather.

And your correction is right, drop the clause. Precisely, the leniency is one level down from where I put it: `valueAsDeclared` keeps `*(6, 2)` at `INTEGER NOT NULL` against a nullable declared field, and `nameRowFieldsAsDeclared` then builds `makeCall(declaredType, ROW, renamed)`, so the call is typed nullable-fields over required operands. That is exactly why the outbound conversion derives required fields — it reads the operands, not the call's type.

alexandrefimov: Filed as substrait-io/substrait#1211 -- the three plans, what the four implementations do with them, and the scope question, since a rule stated for nullability alone would leave width and decimal precision open.

isthmus/src/main/java/io/substrait/isthmus/calcite/rel/rules/VirtualTableExpansionRule.java
alexandrefimov: Done. `sqlGenerationReachesATableInsideASubquery` pins both entry points at that position, and `sqlGenerationRendersTheTable` now asserts the SQL on the `SubstraitToSql` half instead of a row count.

isthmus/src/main/java/io/substrait/isthmus/calcite/rel/VirtualTable.java
alexandrefimov: Done, both. `aCorrelationInARowIsVisibleToTheCollector` asserts `getVariablesUsed` returns the id and `notContainsCorrelation` answers false.

isthmus/src/main/java/io/substrait/isthmus/calcite/rel/VirtualTable.java
alexandrefimov: Least restrictive per column across every rewritten row now, with the declared type kept where the rows still fit it. `aShuttleRetypingOneRowRebuildsATypeThatFitsThemAll` retypes the first row only.

isthmus/src/main/java/io/substrait/isthmus/OuterReferenceResolver.java
alexandrefimov: Taken as written.

isthmus/src/main/java/io/substrait/isthmus/calcite/rel/rules/VirtualTableExpansionRule.java
```diff
@@ -0,0 +1,217 @@
...
+  }
```
nielspardon: Nothing calls this now and no test reaches it, and the summary line promises more than it delivers -- the paragraph below it is where a reader finds out that a table inside a `RexSubQuery` is left alone. It is new public API in a breaking release, so it is cheaper to drop than to keep as a partial guarantee: a consumer wanting a pass adds `instance()` to their own planner, which is what that is for, and one table is `expand`.

```suggestion
```

isthmus/src/main/java/io/substrait/isthmus/calcite/rel/VirtualTable.java
```diff
@@ -0,0 +1,222 @@
...
+    RelDataTypeFactory typeFactory = getCluster().getTypeFactory();
+    List<RelDataType> columnTypes = new ArrayList<>(rowType.getFieldCount());
+    for (int column = 0; column < rowType.getFieldCount(); column++) {
+      List<RelDataType> valueTypes = new ArrayList<>(rewritten.size());
+      for (List<RexNode> row : rewritten) {
+        valueTypes.add(row.get(column).getType());
+      }
+      RelDataType columnType = typeFactory.leastRestrictive(valueTypes);
+      // Nothing in Calcite's type system unifies every pair -- a rewrite that leaves two rows with
+      // no common type keeps the declared one, and the constructor is left to report the row.
```
nielspardon: The constructor checks the count and not the types now, so nothing reports this row. With assertions off it reaches `LogicalProject.create` in the expansion as a projection whose declared type does not fit its expressions -- worth saying here, rather than naming the guard that went away in this commit.

```suggestion
      // Nothing in Calcite's type system unifies every pair -- a rewrite that leaves two rows with
      // no common type keeps the declared one, which then fits neither. The constructor checks the
      // count and not the types, so the expansion is where that surfaces: Project.isValid with
      // assertions on, and a projection whose declared type does not fit its expressions without.
```

nielspardon (commented):
The case on the converter is a better fix than the pass I asked for -- it reaches the subquery position a pass structurally cannot -- and a row type built across all of the rewritten rows is better than the first-row type I suggested. Two small things left, both leftovers rather than new problems: `expandAll` has no caller or test now, and one comment names a check this same commit removed.

#1227 and #1228 are fixed here, so they are worth adding to the closes list. I will close #1225 myself: the digest argument now in the code is right, a planner merges the empty rows whether the rule builds one or many.

isthmus/src/main/java/io/substrait/isthmus/calcite/rel/rules/VirtualTableExpansionRule.java
alexandrefimov: Dropped. The breaking note names the rule rather than the pass, so it still reads right.

isthmus/src/main/java/io/substrait/isthmus/calcite/rel/VirtualTable.java
alexandrefimov: Taken as written.

isthmus/src/test/java/io/substrait/isthmus/VirtualTableTest.java
```diff
@@ -0,0 +1,415 @@
...
+                  firstRow = false;
+                  return rexBuilder.makeCast(bigint, literal);
+                }
+                return literal;
+              }
+            });
+
+    assertEquals(
+        SqlTypeName.BIGINT,
+        rewritten.getRowType().getFieldList().get(0).getType().getSqlTypeName());
```
nielspardon, line 177: Record that the widened table no longer converts back, since that is what the rebuilt type costs and nothing else states it. `rebuiltRowType` widens the column to fit every row, but only the row the shuttle retyped carries the wider type, so the way back refuses on the other one: `A virtual table's value *(6, 2) converts to I32{nullable=false} where its column is declared I64{nullable=false}`. That is the refusal this PR settles on, so it belongs in the test rather than nowhere.

```suggestion
    assertEquals(
        SqlTypeName.BIGINT,
        rewritten.getRowType().getFieldList().get(0).getType().getSqlTypeName());
    assertThrows(
        UnsupportedOperationException.class,
        () -> SubstraitRelVisitor.convert(rewritten, converterProvider),
        "the widened column fits every row, but only the row the shuttle retyped carries it");
```

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
```diff
@@ -924,45 +924,13 @@ public RelNode visit(VirtualTableScan virtualTableScan, Context context) {
...
-          LogicalProject.create(
-              union, Collections.emptyList(), topProjectExprs, rowType, Collections.emptySet());
-      return applyRelCommon(topProject, virtualTableScan, topProject);
+      // A row that does not fit a LogicalValues tuple keeps its expressions, in a relation of our
+      // own: Calcite has none that holds them, and expanding the table into a projection per row
+      // does not come back -- the projection is what converts back, and the table is gone. A
+      // consumer whose planner only knows Calcite's own relations can expand it with
+      // VirtualTableExpansionRule.
+      return applyRelCommon(
+          VirtualTable.create(relBuilder.getCluster(), rowType, convertedRows), virtualTableScan);
```
nielspardon, line 933: Cover a remap over a computed table: this is the branch whose `applyRelCommon` inputs went from the top projection to none, and both emit tests use literal rows, so they take the `LogicalValues` branch above instead. It works as it stands -- the projection lands over the relation and the hint names it -- so this is a test for the branch, not a fix.

```java
  /** The same for a computed table, where the projection has no table to merge into. */
  @Test
  void outputNamesReachTheProjectionTheMappingAddsOnAComputedTable() {
    NamedStruct schema = NamedStruct.of(List.of("col1", "col2"), R.struct(R.I32, R.FP64));
    VirtualTableScan table =
        VirtualTableScan.builder()
            .from(
                createVirtualTableScan(
                    schema, List.of(sb.i32(2), sb.add(sb.fp64(4.4), sb.fp64(4.5)))))
            .remap(Rel.Remap.of(List.of(1)))
            .hint(Hint.builder().addOutputNames("label").build())
            .build();

    RelNode relNode = substraitToCalcite.convert(table);

    assertEquals(List.of("label"), relNode.getRowType().getFieldNames());
    assertEquals(
        List.of(R.FP64),
        SubstraitRelVisitor.convert(relNode, extensions).getRecordType().fields());
  }
```

nielspardon (commented):
This round I read the whole diff rather than what had just changed, and it holds together: the visitor wiring, both proto directions, the digest contract a hand-written `AbstractRelNode` has to meet, the SQL entry points (no `new RelToSqlConverter(` is left anywhere), and `applyRelCommon` taking no inputs for a leaf. Three notes inline, none blocking. One more that GitHub will not let me anchor, since it sits outside every hunk: the javadoc of `outputNamesWithoutAMappingAreLeftAloneOnAComputedTable` (`VirtualTableScanTest:551-553`) still says a computed table "comes back under a projection of its own" -- it comes back as a bare `VirtualTable` now, so the test passes for the same reason as the one it contrasts itself with.

isthmus/src/test/java/io/substrait/isthmus/PlanTestBase.java
```diff
@@ -369,6 +374,44 @@ protected String toSql(Plan plan) {
...
+    List<io.substrait.expression.Expression.NestedStruct> structs =
+        java.util.Arrays.stream(rows)
+            .map(
+                row ->
+                    io.substrait.expression.Expression.NestedStruct.builder()
+                        .addAllFields(row)
+                        .build())
+            .collect(java.util.stream.Collectors.toList());
+    return VirtualTableScan.builder().initialSchema(schema).addAllRows(structs).build();
+  }
```
nielspardon: Import these rather than writing them out: nothing in `PlanTestBase` is called `Expression` or `Collectors`, and `java.util.Arrays` is imported already. `VirtualTableScanTest`'s private `createVirtualTableScan` is then this method under another name -- dropping it, and the `Arrays` and `Collectors` imports it was the last user of, leaves one helper for both files.

```suggestion
  protected final VirtualTableScan virtualTable(NamedStruct schema, List<Expression>... rows) {
    List<Expression.NestedStruct> structs =
        Arrays.stream(rows)
            .map(row -> Expression.NestedStruct.builder().addAllFields(row).build())
            .collect(Collectors.toList());
    return VirtualTableScan.builder().initialSchema(schema).addAllRows(structs).build();
  }
```

with `import io.substrait.expression.Expression;` and `import java.util.stream.Collectors;` added.

isthmus/src/test/java/io/substrait/isthmus/VirtualTableTest.java
alexandrefimov: Recorded, and the assertion pins the refusal itself: `A virtual table's value *(6, 2) converts to I32{nullable=false} where its column is declared I64{nullable=false}`, with the tail about a value not carrying its column's type. On the class alone an unrelated throw would pass it, which is your own point from #1249. Leaving this open in case you would rather it stayed at the class.

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
alexandrefimov: Added as written, with `virtualTable` in place of `createVirtualTableScan` since the next thread removes that one. It passes as it stands.

isthmus/src/test/java/io/substrait/isthmus/PlanTestBase.java
alexandrefimov: Taken as written, and the private copy is gone with the `Arrays` and `Collectors` imports it was the last user of. The stale javadoc you could not anchor is in the same commit: `outputNamesWithoutAMappingAreLeftAloneOnAComputedTable` now says the table comes back bare, so it passes for the reason its sibling does.

nielspardon (approved):

PR #1153, opened 2026-08-25
fix(isthmus): convert a virtual table with a struct column

### LINKED ISSUE #1152: isthmus: a virtual table with a struct column throws a bare AssertionError, because its row type pairs top-level types with flattened names
`SubstraitRelNodeConverter.visit(VirtualTableScan)` builds the Calcite row type from two sources that count fields differently, so a virtual table whose schema contains a struct fails with a bare `AssertionError` before any row is looked at:

```java
final RelDataType rowType =
    typeConverter.toCalcite(typeFactory, virtualTableScan.getInitialSchema().struct());
final List<String> correctFieldNames = virtualTableScan.getInitialSchema().names();
final List<RelDataType> fieldTypes =
    rowType.getFieldList().stream().map(RelDataTypeField::getType).collect(Collectors.toList());
final RelDataType rowTypeWithNames = typeFactory.createStructType(fieldTypes, correctFieldNames);
```

`fieldTypes` has one entry per top-level field, while `NamedStruct.names()` is the flattened depth-first list and carries the nested names too. `createStructType` asserts the two lists are the same length, so it throws as soon as the schema is anything but flat.

Reproduces on main at `d331f47`:

```java
VirtualTableScan.builder()
    .initialSchema(NamedStruct.of(List.of("outer", "a", "b"), R.struct(R.struct(R.I32, R.FP64))))
    .addRows(ExpressionCreator.nestedStruct(false,
        List.of(ExpressionCreator.nestedStruct(false, List.of(sb.i32(1), sb.fp64(2.0))))))
    .build();

// substraitToCalcite.convert(...) -> java.lang.AssertionError: null
```

The message is worth noting on its own: it is Calcite's bare `assert`, with no indication of which relation or which schema was at fault, and it disappears entirely if assertions are off — in which case `createStructType` is being handed mismatched lists with no complaint at all.

`TypeConverter.toNamedStruct` builds that flattened list by collecting names as it descends, and it is what the reverse conversion produces, so the two halves disagree only here. What the row type needs is one name per top-level field, which means walking the struct alongside the flattened list rather than slicing a prefix off it — with two struct fields at the top level, the second one's name is not at index 1.

Found while working on #631, which does not touch this path.

alexandrefimov:
The Calcite row type for a `VirtualTableScan` was assembled from two lists that count fields differently: one type per top-level field, paired with the schema's flattened depth-first names. `createStructType` asserts the two are the same length, so a schema with a struct anywhere in it died on a bare `AssertionError` before a row was looked at -- a plain assert, so with assertions off the mismatched lists go straight on to be canonized. The row type is now built by the conversion that takes those names, the one the named-scan path already uses.

Past that, the choice of encoding was made on the Substrait expressions rather than on what they convert to. A struct or list literal converts to a Calcite call, and a `LogicalValues` tuple holds nothing but literals, so those rows reached the tuple builder and died on a cast. The choice now looks at the converted values, and a row that does not fit a tuple takes the projection encoding that already existed for computed rows.

Then the names themselves. A struct value carries the field names Calcite derives for it while the row type carries the schema's, and Calcite compares those names when it checks a value against its column, so the values are renamed to the schema's -- inside a list or a map as well, where the schema names a struct just the same.

A struct column takes the projection encoding rather than a tuple. Calcite has a row literal, but it orders tuples by casting each value to `Comparable` and a row literal's value is a list of `RexLiteral`s, which are not: a two-row struct table throws out of `RelMdCollation.values`, and so does whatever a consumer's planner does with the relation afterwards. Round-tripping such a table is #631's business.

A value that cannot take the type its column is declared at is now reported instead of stamped with it. `LogicalProject` and `LogicalValues` check the row type they are handed with an `assert`, which says nothing unless assertions are on, so a struct-typed call whose names are not the declared ones used to build a plan Calcite considers invalid and say nothing about it.

Two smaller ones: the relation's row type stays NOT NULL, which is what `createStructType` produced, so a nullable schema struct contributes its fields and not its nullability; and the tuple encoding no longer takes the nullability cast off a cast the plan carries itself, which looks the same but has a failure behavior and is part of what round-trips.

Closes #1152

### DIFF
--- a/isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
@@ -84,6 +84,7 @@ import org.apache.calcite.rex.RexInputRef;
 import org.apache.calcite.rex.RexLiteral;
 import org.apache.calcite.rex.RexNode;
 import org.apache.calcite.rex.RexSlot;
+import org.apache.calcite.rex.RexUtil;
 import org.apache.calcite.sql.SqlAggFunction;
 import org.apache.calcite.sql.SqlKind;
 import org.apache.calcite.sql.SqlOperator;
@@ -796,34 +797,48 @@ public class SubstraitRelNodeConverter
     // they have to be handed to the conversion rather than paired with the row type afterwards:
     // with a nested struct anywhere in the schema the two lists do not even have the same length.
     final NamedStruct schema = virtualTableScan.getInitialSchema();
+    // A relation's row type says what its columns are, not whether a value is there, and Calcite
+    // builds one NOT NULL wherever it derives one -- so a nullable schema struct contributes its
+    // fields and not its nullability.
     final RelDataType rowType =
-        typeConverter.toCalcite(typeFactory, schema.struct(), schema.names());
+        typeFactory.createTypeWithNullability(
+            typeConverter.toCalcite(typeFactory, schema.struct(), schema.names()), false);
 
     List<List<RexNode>> convertedRows = new ArrayList<>();
+    // The same values with the cast a nullable literal converts as taken off, which is the form a
+    // LogicalValues tuple would hold them in -- where they are literals at all.
+    List<List<RexNode>> tupleValues = new ArrayList<>();
     for (final Expression.NestedStruct rowExpr : virtualTableScan.getRows()) {
       List<RexNode> convertedRow = new ArrayList<>();
+      List<RexNode> tupleRow = new ArrayList<>();
       for (int column = 0; column < rowExpr.fields().size(); column++) {
-        RexNode value = rowExpr.fields().get(column).accept(expressionRexConverter, context);
-        convertedRow.add(
-            nameStructFieldsAsDeclared(value, rowType.getFieldList().get(column).getType()));
+        Expression field = rowExpr.fields().get(column);
+        RelDataType declaredType = rowType.getFieldList().get(column).getType();
+        RexNode value =
+            valueAsDeclared(field.accept(expressionRexConverter, context), declaredType);
+        convertedRow.add(value);
+        // Only a converted literal has its nullability cast taken off. A cast the plan carries
+        // itself is doing work -- it has a failure behavior, and it is part of what round-trips --
+        // so a row holding one is computed rather than tabulated.
+        tupleRow.add(field instanceof Expression.Literal ? unwrapNullabilityCast(value) : value);
       }
       convertedRows.add(convertedRow);
+      tupleValues.add(tupleRow);
     }
 
-    // A LogicalValues tuple holds nothing but literals, and whether the values are literals is a
-    // property of what they convert to rather than of the Substrait expressions they came from: a
-    // struct or a list converts to a call, both when it is built from expressions and when Calcite
-    // has no literal of its own for it.
+    // A LogicalValues tuple holds nothing but literals, and whether a value is one is a property of
+    // what it converts to rather than of the Substrait expression it came from: a struct converts
+    // to a ROW call and a list to an array constructor, literal or not. Neither belongs in a tuple
+    // anyway -- Calcite orders tuples by casting each value to Comparable, and the value of a row
+    // literal is a list of RexLiterals, which are not.
     boolean encodableAsTuples =
-        convertedRows.stream()
-            .flatMap(List::stream)
-            .allMatch(value -> unwrapNullabilityCast(value) instanceof RexLiteral);
+        tupleValues.stream().flatMap(List::stream).allMatch(value -> value instanceof RexLiteral);
     if (encodableAsTuples) {
       ImmutableList.Builder<ImmutableList<RexLiteral>> tuplesBuilder = ImmutableList.builder();
-      for (final List<RexNode> convertedRow : convertedRows) {
+      for (final List<RexNode> tupleRow : tupleValues) {
         ImmutableList.Builder<RexLiteral> tupleBuilder = ImmutableList.builder();
-        for (RexNode rexNode : convertedRow) {
-          tupleBuilder.add((RexLiteral) unwrapNullabilityCast(rexNode));
+        for (RexNode value : tupleRow) {
+          tupleBuilder.add((RexLiteral) value);
         }
         tuplesBuilder.add(tupleBuilder.build());
       }
@@ -870,7 +885,7 @@ public class SubstraitRelNodeConverter
   }
 
   /**
-   * Renames the struct fields of a converted row value to the names its column is declared with.
+   * Gives a converted row value the type its column is declared at.
    *
    * <p>A Substrait struct type has no field names of its own -- a schema names every field once, at
    * the relation level -- so a struct-valued expression converts carrying the placeholder names
@@ -878,18 +893,53 @@ public class SubstraitRelNodeConverter
    * Calcite compares field names when it checks a value against the type it is declared at, so a
    * struct is rejected on that difference alone, wherever in the column's type it sits.
    *
-   * <p>Only names change. A row's field types are checked against the schema when the {@link
-   * VirtualTableScan} is built, so a value that is the row, list or map its column declares differs
-   * from it in nothing else. A value that is something else -- a call returning a struct, say -- is
-   * returned untouched, for Calcite to report as the mismatch it is.
+   * <p>Names are what a row, list or map value is rebuilt for: a row's field types are checked
+   * against the schema when the {@link VirtualTableScan} is built, so a value that is one of those
+   * agrees with its declared type on everything a Substrait type says. It can still be stamped
+   * nullable where it was not -- Calcite makes a struct's fields nullable along with the struct,
+   * and a value that cannot be null stands in a column that can. A value that cannot take the
+   * declared type at all -- a call returning a struct, say, whose names are not the declared ones
+   * -- is reported here: {@link LogicalProject} and {@link LogicalValues} check the row type they
+   * are handed with an {@code assert}, which says nothing at all unless assertions are on.
    *
    * @param value the converted row value
    * @param declaredType the type its column is declared at
-   * @return the value, renamed where it has to be
+   * @return the value, at {@code declaredType}
+   * @throws IllegalArgumentException if the value cannot be given that type
    */
-  private RexNode nameStructFieldsAsDeclared(RexNode value, RelDataType declaredType) {
+  private RexNode valueAsDeclared(RexNode value, RelDataType declaredType) {
+    RexNode declared = renamedAsDeclared(value, declaredType);
+    if (!fitsDeclared(declared.getType(), declaredType)) {
+      throw new IllegalArgumentException(
+          String.format(
+              "A virtual table's value %s converts to the type %s, which is not the %s its column is declared at",
+              value, declared.getType().getFullTypeString(), declaredType.getFullTypeString()));
+    }
+    return declared;
+  }
+
+  /**
+   * Whether a converted value can stand at the type it is declared at.
+   *
+   * <p>A value that cannot be null stands in a column that can: Calcite makes a struct's fields
+   * nullable along with the struct, so the fields of a nullable struct are declared that way
+   * whatever the values in them are. The other direction is a mismatch, and so is any difference
+   * beyond nullability -- a field name included, which is the one this rename is about.
+   */
+  private boolean fitsDeclared(RelDataType valueType, RelDataType declaredType) {
+    return SqlTypeUtil.equalSansNullability(typeFactory, valueType, declaredType)
+        && (declaredType.isNullable() || !valueType.isNullable());
+  }
+
+  private RexNode renamedAsDeclared(RexNode value, RelDataType declaredType) {
     if (value.getType().equals(declaredType)) {
       return value;
+    }
+    if (value instanceof RexLiteral && ((RexLiteral) value).isNull()) {
+      // A null value has no fields to rename, only a type to take -- and a list or a map converts
+      // to a literal when it is null and to a constructor call otherwise, so this is where those
+      // two are named as much as a struct is.
+      return rexBuilder.makeNullLiteral(declaredType);
     }
     switch (declaredType.getSqlTypeName()) {
       case ROW:
@@ -904,44 +954,17 @@ public class SubstraitRelNodeConverter
   }
 
   private RexNode nameRowFieldsAsDeclared(RexNode value, RelDataType declaredType) {
-    List<? extends RexNode> fields;
-    if (value instanceof RexLiteral) {
-      @SuppressWarnings("unchecked")
-      List<RexLiteral> literals = (List<RexLiteral>) ((RexLiteral) value).getValueAs(List.class);
-      if (literals == null) {
-        // A null struct has no fields to rename, only a type to take
-        return rexBuilder.makeNullLiteral(declaredType);
-      }
-      fields = literals;
-    } else if (value.isA(SqlKind.ROW)) {
-      fields = ((RexCall) value).getOperands();
-    } else {
+    if (!value.isA(SqlKind.ROW)) {
       return value;
     }
+    List<RexNode> fields = ((RexCall) value).getOperands();
     if (fields.size() != declaredType.getFieldCount()) {
       return value;
     }
-
     List<RexNode> renamed = new ArrayList<>();
     for (int field = 0; field < fields.size(); field++) {
       renamed.add(
-          nameStructFieldsAsDeclared(
-              fields.get(field), declaredType.getFieldList().get(field).getType()));
-    }
-
-    // Rebuilt as a literal only while the parts are literals: renaming a struct can give back a
-    // ROW call, and a literal cannot hold one
-    if (value instanceof RexLiteral
-        && renamed.stream().allMatch(field -> field instanceof RexLiteral)) {
-      List<RexLiteral> literals =
-          renamed.stream().map(RexLiteral.class::cast).collect(Collectors.toList());
-      RexNode asLiteral = rexBuilder.makeLiteral(literals, declaredType);
-      // Calcite has no nullable literal of a row: a value in a nullable column comes back typed NOT
-      // NULL, which neither a tuple nor a projection accepts. A ROW call is built at the type it is
-      // given, so a column like that keeps the projection encoding.
-      if (asLiteral.getType().equals(declaredType)) {
-        return asLiteral;
-      }
+          valueAsDeclared(fields.get(field), declaredType.getFieldList().get(field).getType()));
     }
     return rexBuilder.makeCall(declaredType, SqlStdOperatorTable.ROW, renamed);
   }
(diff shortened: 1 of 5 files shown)


isthmus/src/main/java/io/substrait/isthmus/expression/ExpressionRexConverter.java
```diff
@@ -446,6 +446,21 @@ public RexNode visit(Expression.StructLiteral expr, Context context) throws Runt
...
+    // of a row, and building one at that type gives a literal typed NOT NULL, which would drop
+    // the nullability this expression carries.
+    if (fieldNodes.stream().allMatch(field -> field instanceof RexLiteral)) {
+      List<RexLiteral> fieldLiterals =
+          fieldNodes.stream().map(RexLiteral.class::cast).collect(Collectors.toList());
+      RexNode asLiteral = rexBuilder.makeLiteral(fieldLiterals, structType);
+      if (asLiteral.getType().equals(structType)) {
+        return asLiteral;
+      }
+    }
```
nielspardon: Two rows are enough to break this: `LogicalValues.create` computes the collation trait, which compares tuple[0] to tuple[1] through `getValueAs(Comparable.class)`, and a ROW literal's value is a list of `RexLiteral`s. `RexSimplify` hits the same cast on `equal(struct(...), struct(...))`, so this reaches any plan with two struct literals, not only virtual tables. Letting `nameRowFieldsAsDeclared` build the literal instead keeps the struct column a `LogicalValues` and stops the change being breaking.

```suggestion
```

Worth pinning either way: a struct virtual table with two rows, and a filter comparing two struct literals. The comment can move with it, cut to the two conditions that decide the branch — see my comment on `nameRowFieldsAsDeclared`.

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
```diff
@@ -873,12 +865,137 @@ public RelNode visit(VirtualTableScan virtualTableScan, Context context) {
...
+      List<RexLiteral> literals =
+          renamed.stream().map(RexLiteral.class::cast).collect(Collectors.toList());
+      RexNode asLiteral = rexBuilder.makeLiteral(literals, declaredType);
+      // Calcite has no nullable literal of a row: a value in a nullable column comes back typed NOT
+      // NULL, which neither a tuple nor a projection accepts. A ROW call is built at the type it is
+      // given, so a column like that keeps the projection encoding.
+      if (asLiteral.getType().equals(declaredType)) {
+        return asLiteral;
+      }
+    }
```
nielspardon: Dropping `value instanceof RexLiteral` promotes a renamed ROW *call* of literals here, which is all the tuple encoding needs. Testing `declaredType.isNullable()` up front replaces `asLiteral.getType().equals(declaredType)`, which means the same thing — `makeLiteral` strips nullability for every type — but only after building a literal that a nullable column then discards. It also lets the comment stop explaining the probe.

```suggestion
    // A LogicalValues tuple holds literals, not calls, so build Calcite's row literal where we
    // can. Calcite has no nullable row literal, so a nullable column keeps the ROW call and with
    // it the projection encoding.
    if (!declaredType.isNullable()
        && renamed.stream().allMatch(field -> field instanceof RexLiteral)) {
      List<RexLiteral> literals =
          renamed.stream().map(RexLiteral.class::cast).collect(Collectors.toList());
      return rexBuilder.makeLiteral(literals, declaredType);
    }
```

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
```diff
@@ -873,12 +865,137 @@ public RelNode visit(VirtualTableScan virtualTableScan, Context context) {
...
+   * returned untouched, for Calcite to report as the mismatch it is.
+   *
+   * @param value the converted row value
+   * @param declaredType the type its column is declared at
+   * @return the value, renamed where it has to be
+   */
+  private RexNode nameStructFieldsAsDeclared(RexNode value, RelDataType declaredType) {
+    if (value.getType().equals(declaredType)) {
+      return value;
+    }
```
nielspardon: A typed-null list or map converts to a `RexLiteral`, not a constructor call, so it is handed back carrying `f0` and the projection then fails on the nested name — `nameRowFieldsAsDeclared` handles this for ROW but the array and map twins do not. Hoisting it fixes both and lets the ROW copy go.

```suggestion
  private RexNode nameStructFieldsAsDeclared(RexNode value, RelDataType declaredType) {
    if (value.getType().equals(declaredType)) {
      return value;
    }
    if (value instanceof RexLiteral && ((RexLiteral) value).isNull()) {
      // A null value has no fields to rename, only a type to take
      return rexBuilder.makeNullLiteral(declaredType);
    }
```

`makeNullLiteral` widens a NOT NULL `declaredType` to nullable, so the literal can still come back typed differently from its column.

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
```diff
@@ -791,51 +792,46 @@ public RelNode visit(NamedDdl namedDdl, Context context) {
...
-            .allMatch(row -> row.fields().stream().allMatch(e -> e instanceof Expression.Literal));
-    if (allLiterals) {
+    // A LogicalValues tuple holds nothing but literals, and whether the values are literals is a
+    // property of what they convert to rather than of the Substrait expressions they came from: a
+    // struct or a list converts to a call, both when it is built from expressions and when Calcite
+    // has no literal of its own for it.
+    boolean encodableAsTuples =
+        convertedRows.stream()
+            .flatMap(List::stream)
+            .allMatch(value -> unwrapNullabilityCast(value) instanceof RexLiteral);
```
nielspardon: An `Expression.Cast` whose input is a literal and whose type differs only in nullability is unwrapped straight into a tuple, so the cast and its `failureBehavior` disappear from the plan with no error — previously it forced the projection encoding and survived. Should a `Cast` keep the row off the tuple path?

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
```diff
@@ -873,12 +865,137 @@ public RelNode visit(VirtualTableScan virtualTableScan, Context context) {
...
+    }
+    switch (declaredType.getSqlTypeName()) {
+      case ROW:
+        return nameRowFieldsAsDeclared(value, declaredType);
+      case ARRAY:
+        return nameArrayItemsAsDeclared(value, declaredType);
+      case MAP:
+        return nameMapEntriesAsDeclared(value, declaredType);
+      default:
+        return value;
```
nielspardon, line 952: The description's point about `createStructType`'s plain assert applies to this branch too: `Project`'s check is `assert isValid(...)` and `Values`' is `assert assertRowType()`, so the Javadoc's "for Calcite to report as the mismatch it is" holds only with `-ea`. A struct-typed `Cast`, `IfThen`, or scalar function reaches this `default`, so it is a live path — better to recurse or throw here than to leave it to Calcite.

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
```diff
@@ -791,51 +792,46 @@ public RelNode visit(NamedDdl namedDdl, Context context) {
...
-        typeConverter.toCalcite(typeFactory, virtualTableScan.getInitialSchema().struct());
-
-    final List<String> correctFieldNames = virtualTableScan.getInitialSchema().names();
-
-    final List<RelDataType> fieldTypes =
-        rowType.getFieldList().stream().map(RelDataTypeField::getType).collect(Collectors.toList());
-
-    final RelDataType rowTypeWithNames =
-        typeFactory.createStructType(fieldTypes, correctFieldNames);
+        typeConverter.toCalcite(typeFactory, schema.struct(), schema.names());
```
nielspardon: This carries the schema struct's nullability through, where `createStructType` always produced a NOT NULL record type — a nullable base schema now yields a nullable relation row type, and `SetOp.deriveRowType` renormalizes nullability independently of it. Intended? It wants a test either way.

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
```diff
@@ -873,12 +865,137 @@ public RelNode visit(VirtualTableScan virtualTableScan, Context context) {
...
+  private RexNode nameArrayItemsAsDeclared(RexNode value, RelDataType declaredType) {
+    if (!value.isA(SqlKind.ARRAY_VALUE_CONSTRUCTOR)) {
+      return value;
+    }
+    RelDataType itemType = Objects.requireNonNull(declaredType.getComponentType());
+    List<RexNode> renamed = new ArrayList<>();
+    for (RexNode item : ((RexCall) value).getOperands()) {
+      renamed.add(nameStructFieldsAsDeclared(item, itemType));
+    }
+    return rexBuilder.makeCall(declaredType, SqlStdOperatorTable.ARRAY_VALUE_CONSTRUCTOR, renamed);
```
nielspardon, line 981: `makeCall(declaredType, ...)` stamps the declared type on even when a nested rename bailed, so a mismatch passes `Project.isValid` by construction instead of being reported. "Only names change" in the Javadoc is not quite true either: a NOT NULL element in a nullable-component list comes back nullable.

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
```diff
@@ -873,12 +865,137 @@ public RelNode visit(VirtualTableScan virtualTableScan, Context context) {
...
+  private static RexNode unwrapNullabilityCast(RexNode rexNode) {
+    if (rexNode.isA(SqlKind.CAST)) {
+      RexNode castOperand = ((RexCall) rexNode).getOperands().get(0);
+      if (castOperand instanceof RexLiteral
+          && SqlTypeUtil.equalSansNullability(rexNode.getType(), castOperand.getType())) {
+        return castOperand;
+      }
     }
+    return rexNode;
   }
```
nielspardon: This is `RexUtil.removeNullabilityCast`, which also loops and so strips a stacked nullability cast. Needs a `RexUtil` import, and the method stops being `static` since it reads `typeFactory`.

```suggestion
  private RexNode unwrapNullabilityCast(RexNode rexNode) {
    return RexUtil.removeNullabilityCast(typeFactory, rexNode);
  }
```

isthmus/src/test/java/io/substrait/isthmus/LiteralNullabilityRoundtripTest.java
```diff
@@ -128,7 +130,13 @@ void tupleUnwrapLeavesATruncatingCastAlone() {
...
                     false, ExpressionCreator.varChar(false, "abcdef", 3)))
             .build();
-    assertThrows(ClassCastException.class, () -> substraitToCalcite.convert(overlong));
+
+    assertEquals(
+        "LogicalProject(A=[$0])\n"
+            + "  LogicalUnion(all=[true])\n"
+            + "    LogicalProject(A=[CAST('abcdef'):VARCHAR(3) NOT NULL])\n"
+            + "      LogicalValues(tuples=[[{  }]])\n",
+        RelOptUtil.toString(substraitToCalcite.convert(overlong)));
```
nielspardon, line 139: The `assertThrows` was the only thing rejecting an over-length varchar literal; the row now converts to a plan that truncates the value at execution. Pinning the truncating plan makes that the contract — an explicit check of each value against its declared column type would be better.

isthmus/src/test/java/io/substrait/isthmus/NestedExpressionsTest.java
```diff
@@ -283,6 +283,23 @@ void nullableStructLiteralTest() {
     assertFullRoundTrip(project);
   }
 
+  @Test
+  void nullableStructLiteralWithNonNullFieldsTest() {
+    // The fields being non-null is what makes this one different: every field converts to a
+    // Calcite literal, and the struct's own nullability has to survive that.
+    Expression.StructLiteral structLiteral =
+        ExpressionCreator.struct(true, ExpressionCreator.i32(false, 7));
```
nielspardon: `struct(true, ...)` is nullable, so `makeLiteral` returns a NOT NULL type, the guard fails and this takes the unchanged ROW-call path — the new literal path has no direct test. A non-nullable all-literal struct is the case to pin.

nielspardon (changes requested):
The struct literal in `ExpressionRexConverter` applies to every plan, not just virtual tables, and a ROW `RexLiteral` holds a list of `RexLiteral`s while `RexLiteral` is not `Comparable` — a two-row struct virtual table and a filter comparing two struct literals both throw `ClassCastException` on this branch. Scoping it to `nameRowFieldsAsDeclared` fixes both and drops the need for the breaking change. Two other things want a decision rather than a fix: whether an unrenamable value should throw instead of relying on Calcite's asserts, and whether the row type should now inherit the schema struct's nullability.

isthmus/src/main/java/io/substrait/isthmus/expression/ExpressionRexConverter.java
alexandrefimov: Measured all three. The filter and the projection both throw out of `RexSimplify.simplifyComparison`; the two-row table throws out of `RelMdCollation.values` -> `Ordering.isOrdered` -> `FlatLists$ComparableListImpl.compare`.

That third one is not fixed by moving the literal, though: the tuple is where it breaks, not the place the literal is built. A `LogicalValues` holding a row literal cannot take a second row at all, and any consumer rule that compares tuples hits the same cast afterwards.

So I dropped the row literal instead of moving it. `visit(StructLiteral)` is back to the ROW call, a struct column takes the projection encoding, and the breaking change goes with it. Both cases are pinned -- `twoStructRowsConvert` here, and the two comparison tests in `NestedExpressionsTest` for the plan-wide one. Round-tripping a struct table then rides on #1151, which is where it lands.

Left open since it is not the shape you suggested -- say if you would rather keep the literal for the single-row case.

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
alexandrefimov: This goes with the row literal: there is no literal path left to guard, so both the probe and the nullability check are gone and `nameRowFieldsAsDeclared` only renames a ROW call. The reasoning is in the `ExpressionRexConverter` thread.

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
alexandrefimov: Hoisted as suggested, into the entry point, so the null check sits before the ROW/ARRAY/MAP switch and a typed-null list or map takes its column's type too -- `nullListColumnConverts` and `nullMapColumnConverts` cover those.

On `makeNullLiteral` widening a NOT NULL declared type: the literal then fails the type check below and is reported. It should not be reachable, since a row's field types are checked against the schema when the scan is built and a null value only stands in a nullable column, but it is no longer silent if it is.

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
alexandrefimov: Yes -- it keeps the row off the tuple path now. The unwrap reads the Substrait expression rather than only the Calcite node, so only a converted literal has its nullability cast taken off and a row carrying an `Expression.Cast` is computed instead, failure behavior intact. `aDeclaredNullabilityCastKeepsTheRowOffTheTuplePath` pins it.

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
alexandrefimov: Throwing. `valueAsDeclared` reports a value that cannot take its column's declared type instead of stamping it with it.

A real case turned up while testing that, and it is a better example than the struct-typed call: Calcite types a scalar subquery nullable whatever the Substrait expression's own type says, so a row declared `i32` NOT NULL cannot hold one. On main that is `AssertionError: type mismatch: INTEGER / INTEGER NOT NULL` with assertions on, and an invalid plan without. `aValueThatCannotTakeItsColumnsTypeIsReported`.

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
alexandrefimov: Not intended -- the row type is forced back to NOT NULL, which is what `createStructType` gave. What the schema struct's nullability does still reach is the fields, since Calcite makes a struct's fields nullable along with the struct, and that part is the same as before. `nullableSchemaStructGivesANotNullRowType` pins both halves.

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
alexandrefimov: Each child is now checked against its declared field type before the parent is stamped, so a nested rename that bails is reported rather than masked by the stamp.

The Javadoc dropped the "only names change" claim as well. A value that cannot be null does get stamped nullable -- Calcite makes a struct's fields nullable along with the struct -- and that is the one difference the check lets through, spelled out where the check is.

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
alexandrefimov: Applied as suggested.

isthmus/src/test/java/io/substrait/isthmus/LiteralNullabilityRoundtripTest.java
alexandrefimov: A check of the converted value against its declared column type does not catch this one: the literal converts as `CAST('abcdef'):VARCHAR(3)`, so its type is the declared type and only the value is too long. Nothing rejects a varchar literal longer than the length it declares -- not this conversion, and not the POJO holding it.

There is a check now for the case it does catch, and I reworded the test so it pins the cast, which is the part this conversion decides, and says the over-length value goes unchecked. Would a `@Value.Check` on `VarCharLiteral` in core be worth a follow-up, or do you read the declared length as advisory there?

isthmus/src/test/java/io/substrait/isthmus/NestedExpressionsTest.java
alexandrefimov: Gone with the literal path. The ROW-call path is pinned instead by `comparingTwoStructLiteralsInAProjectTest` and `comparingTwoStructLiteralsInAFilterTest` -- the two shapes the row literal broke.

isthmus/src/main/java/io/substrait/isthmus/expression/ExpressionRexConverter.java
nielspardon: Dropping it is right, and my suggestion was wrong: moving the literal into `nameRowFieldsAsDeclared` would have left the two-row table broken, since the tuple is where it breaks. No reason to keep it for the single-row case either — that would make the encoding depend on the row count, and a consumer rule that compares tuples would still hit the cast.

isthmus/src/test/java/io/substrait/isthmus/LiteralNullabilityRoundtripTest.java
nielspardon: Agreed a type check cannot catch it, and the reworded test says what it pins. Worth a follow-up on `VarCharLiteral`: its `length()` is documented as the maximum, so a longer `value()` contradicts the POJO's own contract, and a `@Value.Check` is the place. `FixedCharLiteral` needs nothing — it derives its type from `value().length()`, so it cannot disagree with itself.

nielspardon (approved):
Everything I raised is addressed, and dropping the row literal is a better answer than the one I suggested.

PR #1155, opened 2026-08-25
build(core): declare shadowJar as core's outgoing artifact

### LINKED ISSUE #1142: build: :core:jar and :core:shadowJar write the same file, so :isthmus:build runs its tests without the ANTLR runtime
`:core:shadowJar` sets `archiveClassifier` to `""` so it replaces the plain jar instead of producing `-all.jar`. That gives `:core:jar` and `:core:shadowJar` the same output file, `core/build/libs/core-<version>.jar`, with nothing ordering them against each other — so whichever runs last is what consumers see. Only the shadowed one carries the `io.substrait.antlr` parser and its relocated runtime; the plain jar has neither, and `shadowImplementation` keeps the un-relocated `org.antlr.v4.runtime` off the compile and runtime classpaths by design.

`./gradlew :isthmus:build` on a clean tree runs them in the order that loses:

```
> Task :core:shadowJar
> Task :isthmus:testClasses
> Task :core:jar
> Task :isthmus:test
```

Every isthmus test then fails at class initialization — 906 of them on `origin/main` at 3b46130:

```
java.lang.NoClassDefFoundError: org/antlr/v4/runtime/tree/ParseTreeVisitor
	at io.substrait.type.Deserializers.<clinit>(Deserializers.java:25)
	at io.substrait.extension.SimpleExtension.objectMapper(SimpleExtension.java:76)
	at io.substrait.extension.SimpleExtension.load(SimpleExtension.java:1472)
	at io.substrait.extension.DefaultExtensionCatalog.loadDefaultCollection(DefaultExtensionCatalog.java:97)
	at io.substrait.isthmus.ConverterProvider$Builder.<init>(ConverterProvider.java:576)
```

The root `./gradlew build --rerun-tasks` that CI runs passes, because there the two land the other way round (`:core:jar` first, `:core:shadowJar` after, then `:isthmus:test`) and the shadowed jar is the one left on disk. `./gradlew :isthmus:test` on its own passes for the same reason. So the green build depends on an ordering nothing declares, and the per-module build a contributor is likely to reach for first does not work at all.

Giving the shadowed jar its own classifier and wiring `:isthmus` (and the other consumers) to depend on that artifact would remove the shared output file. Keeping the current single-jar layout instead would need `:core:jar` ordered before `:core:shadowJar`, which is a constraint on a shared output rather than a fix for it.

Reproduced on macOS, Gradle 9.7.0, daemon on Java 17.

alexandrefimov:
`shadowJar` sets its classifier to the empty string on purpose, so that it replaces the plain jar instead of sitting beside it as `-all.jar`. Both tasks then write `core/build/libs/core-<version>.jar`, nothing orders them, and which one survives depends on where each ends up in the task graph. Only shadowJar's output carries the generated `io.substrait.antlr` parser and its relocated ANTLR runtime.

Building a downstream module on its own gets the plain one. On a clean tree at `d331f47`, `./gradlew :isthmus:build` fails 930 of 989 tests -- 918 of them on a `NoClassDefFoundError`, one naming `org/antlr/v4/runtime/tree/ParseTreeVisitor` outright and 917 reporting the 60 isthmus classes whose static initialization it broke, plus 12 assertions that catch the same exception -- and `./gradlew :spark:build --no-build-cache` fails all three variants with `ClassSelector ... resolution failed`. The root build CI runs happens to order the two the other way round, which is why none of this shows up on a pull request. What has kept the downstream modules working is 14 hand-wired `dependsOn(":core:shadowJar")` across six build files.

So give the two tasks separate paths instead of ordering them: the plain jar takes the classifier `plain`, and shadowJar becomes the outgoing artifact of `apiElements` and `runtimeElements`. Gradle derives the dependency for every consumer from that, and all 14 hooks come out, along with the two `mustRunAfter(":core:compileJava")` lines in spark-3.4. The one hook that was not redundant is `examples/substrait-spark`'s fat jar, where `zipTree` drops the task dependencies the resolved classpath carried; it becomes `dependsOn(configurations.runtimeClasspath)`.

Three things follow that the ordering could not fix.

`./gradlew :core:jar` on its own no longer leaves an ANTLR-less jar where consumers read it. It writes `core-<version>-plain.jar` (1,820,901 bytes, no `io/substrait/antlr` entries) and leaves `core-<version>.jar` (2,229,802 bytes, 62 of them) alone. Nothing in a normal build asks for it any more: `:core:jar` does not run during `./gradlew build` at all.

Both jar tasks used to re-execute on every run, because Gradle will not cache a task whose output path it cannot attribute — `--info` spells it out: *"Gradle does not know how file 'build/libs/core-0.101.0.jar' was created (output property 'archiveFile'). Task output caching requires exclusive access to output paths"*. With separate paths both go `UP-TO-DATE` on a repeat run and `shadowJar` is cacheable.

Shadow injects a `shadowRuntimeElements` variant into the java component, and `from(components["java"])` publishes it. Its dependency list in the module metadata is empty, because the bundled deps are declared on the custom `shadowImplementation` configuration — so a Gradle consumer asking for `Bundling.SHADOWED` on 0.101.0 gets the jar with no Jackson, protobuf, extensions or slf4j behind it. The regular runtime variant now serves that same jar with the real dependency list, so the injected variant is skipped.

The release path is the clearest case for doing it this way. Today the publication declares the plain jar as its artifact (`builtBy [:core:jar]`) and `signMaven-publishPublication` depends on `:core:jar`, while `shadowJar` is reached only transitively through `generateMetadataFileForMaven-publishPublication`: a signature over a file identified by the wrong task, correct only because both tasks write one path and the ordering decides who wrote last. The artifact is now `builtBy [:core:shadowJar]`, and `:core:jar` leaves the publish graph entirely, so there is no longer a file a wrong-jar publish could select. The release path did order them the right way -- the jar published for 0.101.0 is 2,229,802 bytes, the shadowed one -- so this was a latent risk there rather than a defect in what has been released.

One thing left as it is: `runtimeElements` still carries `org.gradle.dependency.bundling=external` while the jar it serves bundles the relocated ANTLR runtime. That is the attribute and the bytes consumers already resolve today, and the relocated classes are not visible to them as a dependency, so nothing in resolution turns on it.

Closes #1142

### DIFF
--- a/core/build.gradle.kts
+++ b/core/build.gradle.kts
@@ -162,7 +162,12 @@ tasks {
     relocate("org.antlr.v4.runtime", "io.substrait.org.antlr.v4.runtime")
   }
 
-  jar { manifest { from("build/generated/sources/manifest/META-INF/MANIFEST.MF") } }
+  jar {
+    // shadowJar takes the unclassified name, so the plain jar needs one of its own: two tasks
+    // writing one path leaves it to the task graph which of them consumers end up with.
+    archiveClassifier.set("plain")
+    manifest { from("build/generated/sources/manifest/META-INF/MANIFEST.MF") }
+  }
 
   // Set the release instead of using a Java 8 toolchain since ANTLR requires Java 11+ to run.
   // Only set the compile release since JUnit 6 requires Java 17 to run tests.
@@ -170,6 +175,25 @@ tasks {
     options.release = 8
     dependsOn("writeManifest")
   }
+}
+
+// Only shadowJar's output carries the io.substrait.antlr parser and its relocated ANTLR runtime,
+// so it is what consumers and the publication have to resolve. Declaring it as the outgoing
+// artifact is what lets Gradle derive the task dependency, for every consumer and for signing.
+listOf("apiElements", "runtimeElements").forEach { variant ->
+  configurations.named(variant) {
+    outgoing.artifacts.clear()
+    outgoing.artifact(tasks.shadowJar)
+  }
+}
+
+// The bundled deps are declared on shadowImplementation, which is kept out of the POM on purpose,
+// so shadow's injected variant would be published advertising no dependencies at all. Its file is
+// the jar the runtime variant now serves anyway.
+(components["java"] as AdhocComponentWithVariants).withVariantsFromConfiguration(
+  configurations["shadowRuntimeElements"]
+) {
+  skip()
 }
 
 java {
--- a/examples/isthmus-api/build.gradle.kts
+++ b/examples/isthmus-api/build.gradle.kts
@@ -18,5 +18,3 @@ application { mainClass = "io.substrait.examples.IsthmusAppExamples" }
 tasks.named<Test>("test") { useJUnitPlatform() }
 
 java { toolchain { languageVersion.set(JavaLanguageVersion.of(17)) } }
-
-tasks.pmdMain { dependsOn(":core:shadowJar") }
--- a/examples/substrait-spark/build.gradle.kts
+++ b/examples/substrait-spark/build.gradle.kts
@@ -35,7 +35,8 @@ dependencies {
 }
 
 tasks.jar {
-  dependsOn("$sparkProject:jar", ":core:jar", ":core:shadowJar")
+  // zipTree drops the task dependencies the resolved classpath carried, so declare it here.
+  dependsOn(configurations.runtimeClasspath)
 
   isZip64 = true
   exclude("META-INF/*.RSA")
@@ -53,5 +54,3 @@ tasks.named<Test>("test") {
 }
 
 java { toolchain { languageVersion.set(JavaLanguageVersion.of(17)) } }
-
-tasks.pmdMain { dependsOn(":core:shadowJar") }
--- a/isthmus/build.gradle.kts
+++ b/isthmus/build.gradle.kts
@@ -131,8 +131,6 @@ dependencies {
 }
 
 tasks {
-  classes { dependsOn(":core:shadowJar") }
-
   jar {
     manifest {
       from("../core/build/generated/sources/manifest/META-INF/MANIFEST.MF")
--- a/spark/spark-3.4_2.12/build.gradle.kts
+++ b/spark/spark-3.4_2.12/build.gradle.kts
@@ -146,16 +146,6 @@ dependencies {
 spotless { isEnforceCheck = false }
 
 tasks {
-  // Ensure shadowJar runs before compilation, but skip core tests
-  named("compileJava") {
-    dependsOn(":core:shadowJar")
-    mustRunAfter(":core:compileJava")
-  }
-  named("compileScala") {
-    dependsOn(":core:shadowJar")
-    mustRunAfter(":core:compileJava")
-  }
-
   jar {
     manifest {
       from("../../core/build/generated/sources/manifest/META-INF/MANIFEST.MF")
@@ -164,7 +154,6 @@ tasks {
   }
 
   test {
-    dependsOn(":core:shadowJar")
     useJUnitPlatform { includeEngines("scalatest") }
 
     // Set system properties for variant identification
--- a/spark/spark-3.5_2.12/build.gradle.kts
+++ b/spark/spark-3.5_2.12/build.gradle.kts
@@ -147,10 +147,6 @@ dependencies {
 spotless { isEnforceCheck = false }
 
 tasks {
-  // Ensure shadowJar runs before compilation
-  named("compileJava") { dependsOn(":core:shadowJar") }
-  named("compileScala") { dependsOn(":core:shadowJar") }
-
   jar {
     manifest {
       from("../../core/build/generated/sources/manifest/META-INF/MANIFEST.MF")
@@ -159,7 +155,6 @@ tasks {
   }
 
   test {
-    dependsOn(":core:shadowJar")
     useJUnitPlatform { includeEngines("scalatest") }
 
     // Set system properties for variant identification
--- a/spark/spark-4.0_2.13/build.gradle.kts
+++ b/spark/spark-4.0_2.13/build.gradle.kts
@@ -147,17 +147,12 @@ dependencies {
 spotless { isEnforceCheck = false }
 
 tasks.register<JavaExec>("dialect") {
-  dependsOn(":core:shadowJar")
   classpath = java.sourceSets["main"].runtimeClasspath
   mainClass = "io.substrait.spark.utils.DialectGenerator"
   args = listOf("../spark_dialect.yaml")
 }
 
 tasks {
-  // Ensure shadowJar runs before compilation
-  named("compileJava") { dependsOn(":core:shadowJar") }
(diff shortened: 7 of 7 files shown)


nielspardon (changes requested):
Diagnosis is exactly right — thanks for chasing it down. Before you polish it, I'd like to propose fixing the cause rather than the ordering, because it ends up being a *smaller* change than this one.

`mustRunAfter` is inert whenever `shadowJar` isn't already in the graph, so `./gradlew :core:jar` still leaves the ANTLR-less jar at that path. And it can't help what actually keeps consumers correct today: 14 hand-wired `dependsOn(":core:shadowJar")` across six build files. Strip just `isthmus/build.gradle.kts:134` and `:core:shadowJar` drops out of `:isthmus:test` — and out of `nativeCompile`, which release.yml, pr.yml and the weekly macOS run all use.

The cause is two tasks declaring one output path. Give them separate paths and make shadowJar the artifact, and Gradle derives the dependency for every consumer:

```kotlin
// core/build.gradle.kts -- replaces the mustRunAfter
tasks.jar {
  // shadowJar is what consumers get; keep the plain jar out of its way.
  archiveClassifier.set("plain")
}

// Only shadowJar's output carries the io.substrait.antlr parser and its relocated ANTLR runtime,
// so it is what consumers and the publication must resolve. Declaring it as the outgoing artifact
// lets Gradle derive the task dependency.
listOf("apiElements", "runtimeElements").forEach { variant ->
  configurations.named(variant) {
    outgoing.artifacts.clear()
    outgoing.artifact(tasks.shadowJar)
  }
}

// The bundled deps are declared on the custom shadowImplementation configuration, so shadow's
// injected variant would be published advertising zero dependencies.
(components["java"] as AdhocComponentWithVariants).withVariantsFromConfiguration(
  configurations["shadowRuntimeElements"]
) {
  skip()
}
```

All 14 hooks then come out (`isthmus:134`, `spark-3.4:151,155,167`, `spark-3.5:151,152,162`, `spark-4.0:150,158,159,169`, `examples/isthmus-api:22`, `examples/substrait-spark:38,57`), plus the two `mustRunAfter(":core:compileJava")` lines only spark-3.4 has. Net −5 lines, and it fixes three things ordering can't: the standalone `./gradlew :core:jar` hole; the permanent thrash (both jar tasks currently re-execute forever and `shadowJar` is uncacheable, since Gradle requires exclusive output paths for caching — with separate paths both go `UP-TO-DATE`); and the published `shadowRuntimeElements` variant, which today advertises **zero** dependencies, so a Gradle consumer asking for `Bundling.SHADOWED` on 0.101.0 gets the jar with no Jackson, protobuf, extensions or slf4j.

One non-obvious step: `examples/substrait-spark`'s fat jar hook is the one that is *not* redundant, because `zipTree` drops the task dependencies from the resolved classpath. Replace it with `dependsOn(configurations.runtimeClasspath)` rather than deleting it.

The release path is the clearest case for doing it this way. Today, and still with this PR applied, the publication declares **the plain jar** as its artifact (`builtBy [:core:jar]`) and `signMaven-publishPublication` depends on `:core:jar`; `shadowJar` is reached only transitively via `generateMetadataFileForMaven-publishPublication`. The right bytes get published only because both tasks write one path and the ordering decides who wrote last — a signature over a file identified by `:core:jar`. With the restructure the artifact becomes `builtBy [:core:shadowJar]`, `sign` gains a direct edge to `shadowJar`, and `:core:jar` leaves the publish graph entirely, so there is no longer a file a wrong-jar publish could select.

alexandrefimov:
Restructured as you proposed. All 14 hooks are gone, along with the two `mustRunAfter(":core:compileJava")` lines, and `examples/substrait-spark` keeps its one as `dependsOn(configurations.runtimeClasspath)`.

Numbers from checking the three claims the change hangs on:

- The publication artifact went from `builtBy [task ':core:jar']` to `builtBy [task ':core:shadowJar']`, and `:core:jar` no longer shows up in `publishMaven-publishPublicationToLocalRepository --dry-run`, in `:isthmus-cli:nativeCompile --dry-run` (which still pulls `:core:shadowJar`, now through the artifact), or in `assemble`. The generated POM is byte-identical to main's.
- `shadowRuntimeElements` had no `dependencies` key at all while pointing at the same 2,229,802-byte file. Its block is the only thing the module metadata loses -- everything else, hashes included, diffs clean against main.
- `--info` on main names the reason for the thrash: *"Gradle does not know how file 'build/libs/core-0.101.0.jar' was created (output property 'archiveFile'). Task output caching requires exclusive access to output paths"*. Both tasks now go `UP-TO-DATE` on a repeat run and `shadowJar` comes back `FROM-CACHE`.

`./gradlew :core:jar` on its own now writes `core-0.101.0-plain.jar` (1,820,901 bytes, zero `io/substrait/antlr` entries) and leaves `core-0.101.0.jar` (2,229,802, 62 of them) alone; from a clean tree `:isthmus:build --no-build-cache` and `:spark:build --no-build-cache` both pass, with `:core:jar` never entering either graph. The examples fat jar carries the 62 parser entries too, so the `runtimeClasspath` hook does what the old one did. nativeCompile I could only check as a task graph -- no GraalVM here -- so the Linux image on CI is the real test of that one.

While re-measuring the baseline I found the failure breakdown in the old description was overstated, and the new one says it properly: of the 930 failing isthmus tests, exactly one names `org/antlr/v4/runtime/tree/ParseTreeVisitor`; 917 are `Could not initialize class io.substrait.isthmus.…` across 60 classes, and 12 are assertions catching the same exception.

One thing I flagged rather than changed: `runtimeElements` still says `org.gradle.dependency.bundling=external` while the jar it serves bundles the relocated ANTLR. That is the attribute consumers match today, and flipping it to `shadowed` would leave nothing for a consumer asking for `external`, so keeping it looked right -- say if you read it differently.

I also put a short paragraph in `AGENTS.md` about where the shadowed jar comes from now, on the theory that the next person to hit a classpath error is otherwise going to reach for `dependsOn(":core:shadowJar")` again. Happy to drop it if you'd rather keep that file to the model and not the build.

AGENTS.md
```diff
@@ -139,6 +139,12 @@ Run `./gradlew build` and make sure it passes before pushing — narrower tasks
 CI runs, and a `:core` change can break the visitor implementors in the other modules. See
 [`CONTRIBUTING.md`](CONTRIBUTING.md#building-and-testing).
 
+What `:core` hands its consumers is `shadowJar`'s output — it is the outgoing artifact of
+`apiElements` and `runtimeElements`, while the plain `jar` is classified `plain` and consumed by
+nothing. Downstream modules resolve the shadowed jar, `io.substrait.antlr` parser included,
+through the ordinary project dependency, so don't wire `dependsOn(":core:shadowJar")` into a
+module to make a build work.
+
```
nielspardon: I would drop this. We already have a statement above on the shadowing of the ANTLR runtime and the rest is discoverable from `core/build.gradle.kts`:

alexandrefimov: Dropped -- `AGENTS.md` is back to what it is on main.

nielspardon (approved):
LGTM

PR #1156, opened 2026-08-25
feat(core)!: support offset_expr on window bounds (spec v0.102.0)

### LINKED ISSUE #1143: Update to Substrait v0.102.0
# Release Notes

### ⚠ BREAKING CHANGES

* **protos:** deprecates the `offset` field of the `Preceding` and
`Following` window bounds in favor of the new `offset_expr` field, which
accepts an arbitrary expression instead of an integer literal. Consumers
should add `offset_expr` support before producers switch to it. The PR
also requires explicit window frame bounds type. Producers should set
the bounds type and Consumers should reject `BOUNDS_TYPE_UNSPECIFIED` in
such cases.
 

### Features

* **extensions:** add a decimal overload for negate ([#1177](https://github.com/substrait-io/substrait/issues/1177)) ([6092523](https://github.com/substrait-io/substrait/commit/60925234a0a2c02a41400df349b0bad1e674581b))
* **protos:** support expression in window aggregate bounds ([#1105](https://github.com/substrait-io/substrait/issues/1105)) ([6706015](https://github.com/substrait-io/substrait/commit/6706015ffdb19345fef399ef8238e3e6269cb380)), closes [#748](https://github.com/substrait-io/substrait/issues/748) [#617](https://github.com/substrait-io/substrait/issues/617) [#930](https://github.com/substrait-io/substrait/issues/930)

See https://github.com/substrait-io/substrait/releases/tag/v0.102.0

anasik:
- `Preceding/Following` now prefer `offset_expr` over the deprecated `offset` when set, per the spec's precedence rule 
- A literal offset still populates the legacy offset field too, for producers that need backward compatibility 
- Reject `BOUNDS_TYPE_UNSPECIFIED` when a bound is `CurrentRow`, `Preceding,` or `Following`; still allowed when both bounds are `Unbounded` (needed for Spark's `UnspecifiedFrame)` 
- Update Spark's window frame conversion, which still only supports a literal integer offset in its window frame API; isthmus's producer side (`WindowBoundConverter`) now builds a type-matched `offset_expr` from the ordering column's type instead of assuming `i64`
- Bumping to spec v0.102.0 also exposes `negate(decimal<P,S>)` via `FunctionMappings`'s `UNARY_MINUS`→`negate` mapping, so `SELECT -decimal_col` may now resolve a different function; untested by any isthmus test

Closes #1143

BREAKING CHANGE: `WindowBound.Preceding`/`Following.offset()` now returns `Expression` instead of `long`. `ExpressionProtoConverter.BoundConverter`'s public `convert(WindowBound)` is now an instance method (reachable via `ExpressionProtoConverter.toProto(WindowBound)`) instead of a static utility. A plan with `bounds_type` unset (`BOUNDS_TYPE_UNSPECIFIED`) alongside a bound that isn't Unbounded no longer parses; it must set `bounds_type` to `ROWS` or `RANGE`, or use `Unbounded` on both sides.


nielspardon:
Can you check `ExpressionCopyOnWriteVisitor`, `RelCopyOnWriteVisitor` and `OuterReferenceConverter`? I think they need to be changed to handle copying / handling references in the expression trees in `offset_expr`.

anasik:
> Can you check `ExpressionCopyOnWriteVisitor`, `RelCopyOnWriteVisitor` and `OuterReferenceConverter`? I think they need to be changed to handle copying / handling references in the expression trees in `offset_expr`.

check now, @nielspardon

spark/src/main/scala/io/substrait/spark/expression/ToWindowFunction.scala
```diff
@@ -141,6 +141,16 @@ object ToWindowFunction {
...
 
+  private def toLiteralOffset(offset: SExpression): Int = {
+    WindowBound
+      .toLiteralOffset(offset)
+      .orElseThrow(
+        () =>
+          new UnsupportedOperationException(
+            s"Spark window bounds only support a literal positive offset, got: $offset"))
+      .intValue()
+  }
```
nielspardon, line 157: Check the range before narrowing to `Int`.

`Following.of(2147483648L)` currently becomes `Literal(-2147483648)`, and `fromSpark` (:106) reads a negative bound as PRECEDING, so a forward frame silently turns into a backward one.

```suggestion
  private def toLiteralOffset(offset: SExpression): Int = {
    val value = WindowBound
      .toLiteralOffset(offset)
      .orElseThrow(
        () =>
          new UnsupportedOperationException(
            s"Spark window bounds only support a literal positive offset, got: $offset"))
      .longValue()
    if (value > Int.MaxValue) {
      throw new UnsupportedOperationException(
        s"Spark window bounds only support offsets up to ${Int.MaxValue}, got: $value")
    }
    value.toInt
  }
```

core/src/main/java/io/substrait/expression/WindowBound.java
```diff
@@ -15,6 +16,46 @@ public interface WindowBound {
...
+   */
+  static Optional<Long> toLiteralOffset(Expression offset) {
+    if (offset instanceof Expression.I64Literal) {
+      long value = ((Expression.I64Literal) offset).value();
+      if (value > 0) {
+        return Optional.of(value);
+      }
+    }
+    return Optional.empty();
+  }
```
nielspardon, line 42: Accept any integral literal, not just i64.

An i32 offset (the natural width for a Spark `IntegerType` frame) currently leaves the deprecated `offset` field at 0, which the spec reads as unset, so an older consumer sees no offset at all. It also makes isthmus and Spark throw on a plain constant.

```suggestion
  static Optional<Long> toLiteralOffset(Expression offset) {
    long value;
    if (offset instanceof Expression.I64Literal) {
      value = ((Expression.I64Literal) offset).value();
    } else if (offset instanceof Expression.I32Literal) {
      value = ((Expression.I32Literal) offset).value();
    } else if (offset instanceof Expression.I16Literal) {
      value = ((Expression.I16Literal) offset).value();
    } else if (offset instanceof Expression.I8Literal) {
      value = ((Expression.I8Literal) offset).value();
    } else {
      return Optional.empty();
    }
    return value > 0 ? Optional.of(value) : Optional.empty();
  }
```

core/src/main/java/io/substrait/expression/proto/ProtoExpressionConverter.java
```diff
@@ -457,9 +463,14 @@ public ConsistentPartitionWindow.WindowRelFunctionInvocation fromWindowRelFuncti
...
-        return WindowBound.Preceding.of(bound.getPreceding().getOffset());
+        // Per the spec, consumers must use offset_expr when it is set and ignore offset.
+        return bound.getPreceding().hasOffsetExpr()
+            ? WindowBound.Preceding.of(from(bound.getPreceding().getOffsetExpr()))
+            : WindowBound.Preceding.of(bound.getPreceding().getOffset());
       case FOLLOWING:
-        return WindowBound.Following.of(bound.getFollowing().getOffset());
+        return bound.getFollowing().hasOffsetExpr()
+            ? WindowBound.Following.of(from(bound.getFollowing().getOffsetExpr()))
+            : WindowBound.Following.of(bound.getFollowing().getOffset());
```
nielspardon: Treat a zero `offset` as `CurrentRow`.

`offset` has no presence in proto3, so `preceding {}` and `offset: 0` are the same bytes, and the spec says zero was never a valid offset and is equivalent to `CurrentRow`. Today both become `Preceding.of(0)`, which we write back as a zero `offset_expr` — something the spec tells producers not to emit — and which makes isthmus and Spark throw later on. Rejecting it outright would also be defensible; this keeps older plans readable.

```suggestion
      case PRECEDING:
        {
          io.substrait.proto.Expression.WindowFunction.Bound.Preceding preceding =
              bound.getPreceding();
          // Per the spec, consumers must use offset_expr when it is set and ignore offset.
          if (preceding.hasOffsetExpr()) {
            return WindowBound.Preceding.of(from(preceding.getOffsetExpr()));
          }
          // offset has no presence, so zero means unset, and a zero offset means CurrentRow.
          return preceding.getOffset() == 0
              ? WindowBound.CURRENT_ROW
              : WindowBound.Preceding.of(preceding.getOffset());
        }
      case FOLLOWING:
        {
          io.substrait.proto.Expression.WindowFunction.Bound.Following following =
              bound.getFollowing();
          if (following.hasOffsetExpr()) {
            return WindowBound.Following.of(from(following.getOffsetExpr()));
          }
          return following.getOffset() == 0
              ? WindowBound.CURRENT_ROW
              : WindowBound.Following.of(following.getOffset());
        }
```

core/src/main/java/io/substrait/expression/WindowBound.java
```diff
@@ -15,6 +16,46 @@ public interface WindowBound {
...
+   */
+  static void checkBoundsType(
+      Expression.WindowBoundsType boundsType, WindowBound lowerBound, WindowBound upperBound) {
+    if (boundsType == Expression.WindowBoundsType.UNSPECIFIED
+        && (!(lowerBound instanceof Unbounded) || !(upperBound instanceof Unbounded))) {
+      throw new IllegalArgumentException(
+          "bounds_type is required when either window bound is CurrentRow, Preceding, or"
+              + " Following, but was BOUNDS_TYPE_UNSPECIFIED");
+    }
+  }
```
nielspardon, line 63: Move this into a `@Value.Check` on `Expression.WindowFunctionInvocation` and `ConsistentPartitionWindow.WindowRelFunctionInvocation`, and drop the two calls in `ProtoExpressionConverter` (and the locals added to feed them).

It only runs when reading a plan today, so `…boundsType(UNSPECIFIED).lowerBound(Preceding.of(5))` builds and serializes fine, then fails in every consumer including us. `Plan.java:70` rejects an `UNSPECIFIED` enum the same way.

isthmus/src/main/java/io/substrait/isthmus/expression/ExpressionRexConverter.java
```diff
@@ -811,16 +811,25 @@ private ToRexWindowBound(RexBuilder rexBuilder, RexWindowBound unboundedVariant)
...
     }
 
+    private static long toLiteralOffset(io.substrait.expression.Expression offset) {
+      return WindowBound.toLiteralOffset(offset)
+          .orElseThrow(
+              () ->
+                  new UnsupportedOperationException(
+                      "Calcite window bounds only support a literal positive offset, got: "
+                          + offset));
+    }
```
nielspardon: Pass the enclosing `ExpressionRexConverter` and `Context` into `ToRexWindowBound` and use `RexWindowBounds.preceding(offset.accept(converter, context))` instead of unwrapping to a literal.

`RexWindowBounds.preceding` takes any `RexNode`, so the literal-only limit is ours, not Calcite's, and it blocks the interval and decimal RANGE offsets this spec version adds. The repo did it this way for `FetchRel`'s `offset_expr` in `SubstraitRelNodeConverter:658`. While you're in there: the helper sits between the `visit` overloads (Google style asks that overloads stay together) and fully qualifies `Expression`, which the file already imports.

core/src/main/java/io/substrait/expression/WindowBound.java
```diff
@@ -65,23 +106,33 @@ interface WindowBoundVisitor<R, E extends Throwable> {
...
+    public abstract Expression offset();
 
     /**
-     * Creates a {@link Preceding} bound with the given offset.
+     * Creates a {@link Preceding} bound with the given literal row offset.
      *
      * @param offset the number of rows preceding the current row
      * @return the preceding bound
      */
     public static Preceding of(long offset) {
```
nielspardon, line 133: The `long` overloads are really the ROWS-mode convenience, so worth saying so — RANGE bounds want `of(Expression)` with the ordering column's type. i64 is right for a row count, but for RANGE `add(decimal, i64)` isn't defined, so `rangePrecedingToCurrent` (`ORDER BY O_TOTALPRICE`, DECIMAL) emits a plan a strict consumer rejects.

```suggestion
    /**
     * Creates a {@link Preceding} bound from a literal row offset. For {@code BOUNDS_TYPE_ROWS}
     * only: a RANGE bound's offset must be type-compatible with the ordering expression, so use
     * {@link #of(Expression)} there.
     *
     * @param offset the number of rows preceding the current row
     * @return the preceding bound
     */
    public static Preceding of(long offset) {
```

core/src/main/java/io/substrait/expression/WindowBound.java
```diff
@@ -91,15 +142,15 @@ public <R, E extends Throwable> R accept(WindowBoundVisitor<R, E> visitor) {
...
+     * Returns the expression evaluating to the distance following the current row.
      *
-     * @return the offset
+     * @return the offset expression
      */
-    public abstract long offset();
+    public abstract Expression offset();
 
     /**
      * Creates a {@link Following} bound with the given offset.
```
nielspardon: This summary was left as "with the given offset" while `Preceding.of(long)`'s was updated. The `@param` on the next line should lose "number of rows" too, for the same reason as the RANGE comment above.

```suggestion
     * Creates a {@link Following} bound from a literal row offset. For {@code BOUNDS_TYPE_ROWS}
     * only: a RANGE bound's offset must be type-compatible with the ordering expression, so use
     * {@link #of(Expression)} there.
```

core/src/main/java/io/substrait/relation/RelProtoConverter.java
```diff
@@ -780,8 +780,8 @@ private List<ConsistentPartitionWindowRel.WindowRelFunction> toProtoWindowRelFun
                   .addAllOptions(options)
                   .setFunctionReference(extensionCollector.getFunctionReference(f.declaration()))
                   .setBoundsType(f.boundsType().toProto())
-                  .setLowerBound(BoundConverter.convert(f.lowerBound()))
-                  .setUpperBound(BoundConverter.convert(f.upperBound()))
+                  .setLowerBound(BoundConverter.convert(f.lowerBound(), exprProtoConverter))
+                  .setUpperBound(BoundConverter.convert(f.upperBound(), exprProtoConverter))
```
nielspardon: Go through the converter rather than the static, so this uses the same `toProto` path as every other expression here. A subclass that overrides `toProto(Expression)` (what `StringHolderHandlingRelProtoConverter` does) otherwise never sees bound offsets. Needs the instance method from the `BoundConverter` comment.

```suggestion
                  .setLowerBound(exprProtoConverter.toProto(f.lowerBound()))
                  .setUpperBound(exprProtoConverter.toProto(f.upperBound()))
```

core/src/main/java/io/substrait/expression/proto/ExpressionProtoConverter.java
```diff
@@ -844,34 +844,42 @@ public Expression visit(
...
     @Override
     public Expression.WindowFunction.Bound visit(WindowBound.Preceding preceding) {
-      return Expression.WindowFunction.Bound.newBuilder()
-          .setPreceding(
-              Expression.WindowFunction.Bound.Preceding.newBuilder().setOffset(preceding.offset()))
-          .build();
+      Expression.WindowFunction.Bound.Preceding.Builder builder =
+          Expression.WindowFunction.Bound.Preceding.newBuilder()
+              .setOffsetExpr(exprProtoConverter.toProto(preceding.offset()));
+      WindowBound.toLiteralOffset(preceding.offset()).ifPresent(builder::setOffset);
```
nielspardon, line 883: Hold `BoundConverter` in a `private final` field and expose an instance `toProto(WindowBound)`, next to the existing `toProto(Expression)` and `toProto(Type)`.

That removes the `convert(bound, this)` self-pass and the two allocations per window function, and keeps the public `convert(WindowBound)` signature you had before.

core/src/test/java/io/substrait/type/proto/ConsistentPartitionWindowRelRoundtripTest.java
```diff
@@ -141,4 +147,183 @@ void consistentPartitionWindowRoundtripMulti() {
...
+                Arrays.asList(
+                    ConsistentPartitionWindow.WindowRelFunctionInvocation.builder()
+                        .declaration(windowFunctionDeclaration)
+                        .arguments(Arrays.asList(argAdd))
+                        .outputType(R.I64)
+                        .aggregationPhase(Expression.AggregationPhase.INITIAL_TO_RESULT)
+                        .invocation(Expression.AggregationInvocation.ALL)
+                        .lowerBound(WindowBound.Preceding.of(offsetAdd))
+                        .upperBound(WindowBound.CURRENT_ROW)
+                        .boundsType(Expression.WindowBoundsType.RANGE)
```
nielspardon, line 319: Give this RANGE fixture a single sort whose type matches the offset (the other RANGE test sorts i32 against an i64 offset).

The spec text `checkBoundsType` paraphrases also wants exactly one ordering expression for a RANGE Preceding/Following bound, and `add(T, D) -> T` to be defined. As written `verifyRoundTrip` asserts fidelity for a plan that isn't valid. Worth noting in the PR body which of those rules you're leaving for later.

core/src/test/java/io/substrait/type/proto/ConsistentPartitionWindowRelRoundtripTest.java
```diff
@@ -141,4 +147,183 @@ void consistentPartitionWindowRoundtripMulti() {
...
+        protoRel.getWindow().getWindowFunctions(0);
+    assertTrue(protoWindowFunction.getLowerBound().getPreceding().hasOffsetExpr());
+    assertEquals(0, protoWindowFunction.getLowerBound().getPreceding().getOffset());
+
+    io.substrait.relation.Rel rel2 = protoRelConverter.from(protoRel);
+    assertEquals(rel1, rel2);
+  }
+
+  @Test
+  void precedingOffsetExprRoundtripsThroughLiteralOffsetForBackwardCompat() {
```
nielspardon, line 223: Add an `Expression.WindowFunctionInvocation` case, and either finish this test's proto→POJO leg or rename it — nothing in core builds a `WindowFunctionInvocation`, so the expression-level half of the change is untested. `:193-194` can also just call `verifyRoundTrip`.

core/src/test/java/io/substrait/type/proto/ConsistentPartitionWindowRelRoundtripTest.java
```diff
@@ -141,4 +147,183 @@ void consistentPartitionWindowRoundtripMulti() {
...
+    SimpleExtension.ScalarFunctionVariant addDeclaration =
+        extensions.scalarFunctions().stream()
+            .filter(s -> s.name().equalsIgnoreCase("add"))
+            .findFirst()
+            .orElseThrow(AssertionError::new);
+    SimpleExtension.ScalarFunctionVariant subtractDeclaration =
+        extensions.scalarFunctions().stream()
+            .filter(s -> s.name().equalsIgnoreCase("subtract"))
+            .findFirst()
+            .orElseThrow(AssertionError::new);
```
nielspardon: Use `sb.add(...)` / `sb.subtract(...)`, or an exact anchor like `getScalarFunction(FunctionAnchor.of(FUNCTIONS_ARITHMETIC, "add:i64_i64"))`.

The name scan picks `add:dec_dec`, since `functions_arithmetic_decimal.yaml` loads first, while the test passes i64 arguments and declares `R.I64`. Also: `:203` can use `TestBase`'s `expressionProtoConverter` field, and `:175`'s `WindowBound.Following.CURRENT_ROW` is just `CURRENT_ROW`.

core/src/test/java/io/substrait/type/proto/ConsistentPartitionWindowRelRoundtripTest.java
```diff
@@ -141,4 +147,183 @@ void consistentPartitionWindowRoundtripMulti() {
...
+    assertTrue(protoWindowFunction.getLowerBound().getPreceding().hasOffsetExpr());
+    assertEquals(0, protoWindowFunction.getLowerBound().getPreceding().getOffset());
+
+    io.substrait.relation.Rel rel2 = protoRelConverter.from(protoRel);
+    assertEquals(rel1, rel2);
+  }
+
+  @Test
+  void precedingOffsetExprRoundtripsThroughLiteralOffsetForBackwardCompat() {
+    // A literal offset_expr must also populate the legacy field, per the spec's migration note.
```
nielspardon: Reword: we populate the legacy field for compatibility. The spec says producers *may* set both, and its "must" only constrains the value when both are set.

gradle/libs.versions.toml
```diff
@@ -31,7 +31,7 @@ spark-3-5 = "3.5.4"
 spark-4-0 = "4.0.2"
 spotless = "8.10.0"
 # substrait-packaging artifacts, versioned by the Substrait spec release they are generated from.
-substrait-packaging = "0.101.0"
+substrait-packaging = "0.102.0"
```
nielspardon, line 34: Please mention in the PR body that this also brings `negate(decimal<P,S>)`. `FunctionMappings:93` maps `UNARY_MINUS` to `negate`, so `SELECT -decimal_col` can now resolve a different function, and no isthmus test covers it.

nielspardon (changes requested):
Thanks for the copy-on-write fix — that part looks good.

The main thing left is on the producer side, in isthmus `WindowBoundConverter`. It still turns every Calcite offset into a `long`, so `RANGE BETWEEN 1.5 PRECEDING` silently becomes `1 PRECEDING` (`DECIMAL` is in `SqlTypeName.EXACT_TYPES`), and it can never emit an `offset_expr` for a decimal or interval offset. Since the deprecated `offset` field is an int64 and has to stay one, I'd build the bound from the offset's own type: integral literals become an i64 `offset_expr`, which the writer still mirrors into `offset`; everything else sets only `offset_expr`. The same place should map a zero offset to `WindowBound.CURRENT_ROW`, since the spec says zero isn't an offset. One caveat for RANGE: an integral offset needs the ordering column's type, not a blanket i64, or `add(T, D) -> T` isn't defined.

Three body-level ones: please add an offset expression to the `RelSamples` `ConsistentPartitionWindow` sample so the shared round-trip tests cover one; the `BREAKING CHANGE:` footer should also mention the removed public `BoundConverter.convert(WindowBound)` and that plans with `bounds_type` unset no longer parse; and `ExpressionRexConverter.isRowMode` (:760) still rejects `UNSPECIFIED` even when both bounds are unbounded, which `checkBoundsType` now allows.

The rest is inline. Suggestions are written against `62e2aa9e` but not compiled — worth a local build before committing them.

core/src/main/java/io/substrait/expression/WindowBound.java
anasik: Yeah, I actually wanted to do this. Idk why I let Claude advise me against it.

spark/src/main/scala/io/substrait/spark/expression/ToWindowFunction.scala
anasik: I noticed this and wondered if this could cause problems but didn't think of throwing.

nielspardon: we probably need to think about whether such limits are something we should capture in the Substrait dialect

isthmus/src/main/java/io/substrait/isthmus/expression/WindowBoundConverter.java
```diff
@@ -1,69 +1,115 @@
...
+  private static Optional<Expression> integralLiteralOfType(Type type, long value) {
+    if (type instanceof Type.I64) {
+      return Optional.of(ExpressionCreator.i64(false, value));
+    } else if (type instanceof Type.I32) {
+      return Optional.of(ExpressionCreator.i32(false, (int) value));
+    } else if (type instanceof Type.I16) {
+      return Optional.of(ExpressionCreator.i16(false, (int) value));
+    } else if (type instanceof Type.I8) {
+      return Optional.of(ExpressionCreator.i8(false, (int) value));
     }
```
nielspardon, line 138: Guard the narrowing before casting, or leave the offset alone when the value doesn't fit.

The `instanceof` chain tests `orderingType` — the ordering column's type — while `value` comes from the offset literal, and Calcite never reconciles the two. `SqlWindow.validateFrameBoundary` only checks the bound's type *family* against `orderTypeFam.allowableDifferenceTypes()` for RANGE; there's no range check and no coercion to the ordering type. So:

- `ORDER BY smallint_col RANGE BETWEEN 100000 PRECEDING` — literal is INTEGER, NUMERIC family, accepted. Cast is to `int`, not `short`, and `Expression.I16Literal` doesn't validate, so the POJO ends up holding 100000 in an i16 literal.
- `ORDER BY int_col RANGE BETWEEN 3000000000 PRECEDING` — literal is BIGINT, also accepted. `(int) 3000000000L` is -1294967296, so a preceding offset goes negative.

Both frames are degenerate (the distance exceeds the column's domain), so this is low-stakes — it's the silence I'd rather avoid, same as the `.intValue()` case last round.

isthmus/src/main/java/io/substrait/isthmus/expression/WindowBoundConverter.java
```diff
@@ -1,69 +1,115 @@
...
+    return Optional.empty();
```
nielspardon, line 139: Consider a decimal branch here, or a cast to the ordering type, so a non-integral ordering column doesn't fall through to `.orElse(offset)`.

An integral literal offset over a decimal, fp, or temporal ordering column keeps whatever Calcite produced. If Calcite types the `10` in `ORDER BY O_TOTALPRICE RANGE BETWEEN 10 PRECEDING` as INTEGER — it doesn't coerce the bound to the ordering type, so I believe it does — that's `add(decimal, i32)`, which isn't defined, i.e. the same problem this method solves for i32 columns. Worth checking what `rangePrecedingToCurrent` emits now.

isthmus/src/main/java/io/substrait/isthmus/expression/WindowFunctionConverter.java
```diff
@@ -108,10 +108,19 @@ protected Expression.WindowFunctionInvocation generateBinding(
...
-    WindowBound lowerBound = toWindowBound(window.getLowerBound());
-    WindowBound upperBound = toWindowBound(window.getUpperBound());
+        isRows ? Expression.WindowBoundsType.ROWS : Expression.WindowBoundsType.RANGE;
+    // The spec requires a RANGE bound's offset type to be compatible with the single ordering
+    // expression's type (add(T, D) -> T).
+    Optional<RelDataType> orderingType =
+        window.orderKeys == null || window.orderKeys.isEmpty()
+            ? Optional.empty()
+            : Optional.of(window.orderKeys.get(0).left.getType());
+    WindowBound lowerBound =
```
nielspardon: `orderKeys.get(0)` picks the first key even when there are several, but the spec allows a RANGE `Preceding`/`Following` bound only with exactly one ordering expression. Worth rejecting that case, or a note on why it's safe to take the first.

nielspardon (commented):
This looks good — the producer side is better than what I suggested; keying the offset type off the ordering expression is the right call, and `WindowBoundConverterTest` covers the cases individually.

Three follow-ups, none of which I'd block on — see inline.

Two notes outside the diff: the bullet "Update isthmus and Spark conversions, since both only support literal offsets in their own window frame APIs" isn't true for isthmus any more, and the body becomes the commit message. And `WindowBoundConverter.toWindowBound` is public and gained three parameters — probably fine to leave out of the footer given it's effectively isthmus-internal, but worth a thought.

nielspardon (approved):
LGTM, thanks @anasik

Issue #1157, opened 2026-08-25
isthmus: hint.output_names is never written on export


alexandrefimov:
Isthmus does not look at `RelCommon.hint` at all — `hint` does not appear anywhere in `isthmus/src/main` — so the alternative output field names a producer puts on a relation are lost, and none are written back.

A `Project` carrying `hint.output_names = [total_price]`:

```
Project(remap=[2], expressions=[add(a, 1)], hint=[output_names=[total_price]])
  NamedScan(t, [a, b])
```

converts to `LogicalProject($f2=[+($0, 1)])` and renders as

```sql
SELECT a + 1 AS $f2 FROM t
```

The name the producer supplied is replaced by one Calcite derives positionally. `Plan.Root.names` covers the top of the plan, so this shows up on every relation below the root.

This matters inside this repository as much as outside it: the Spark module both writes these names (`ToSubstraitRel` puts them on `Project` and `Expand`) and reads them back (`ToLogicalPlan.fieldNames`), so a plan that travels Spark → Substrait → Isthmus loses the column names Spark took care to record.

Per the spec, `output_names` "[a]ssigns alternative output field names for any relation. Equivalent to the names field in RelRoot but applies to the output of the relation this RelCommon is attached to." Being equivalent to `RelRoot.names` makes it a depth-first flattened list that includes the names of nested struct fields, which is the same walk `SubstraitToCalcite` already does when it applies `Plan.Root.names`.

### Import

The names fit where the conversion already produces a projection: `visit(Project)` and the projection `applyRemap` adds for a relation with an emit mapping. There they cost nothing — rebuilding that `Project` with a renamed row type keeps the plan shape:

```
LogicalProject(total_price=[+($0, 1)])
  LogicalTableScan(table=[[t]])
```

Renaming any other relation would mean adding a projection the plan never asked for, which is a worse trade than dropping a hint, so on a `Filter` or a `Join` the names would be dropped.

Only the top-level names can be applied. `output_names` is written depth-first and names the fields nested inside a struct column as well, but those belong to the type of the expression producing the column: Calcite compares a projection's row type against the types of its expressions, nested names included, and rejects the projection when they disagree.

A name list that does not fit the relation would be dropped rather than fail the conversion — a hint carries no meaning of its own and should not be able to break an otherwise valid plan (#1071 is the same concern for `ComputationType`). That covers a list of the wrong length, and also one that repeats a name, which Calcite does not allow a projection to have.

### Export

[...]

Issue #1154, opened 2026-08-25
isthmus: a nullable struct loses its fields' non-nullability converting to Calcite


alexandrefimov:
`struct?<i32, fp64>` does not survive a trip through Calcite. The struct's own nullability is kept, its fields' is not: what comes back is `struct?<i32?, fp64?>`.

Verified on `main` at `d331f47`.

## Reproduction

A named scan is enough — nothing about this is specific to a relation type, the schema goes through `TypeConverter` either way:

```java
NamedStruct schema = NamedStruct.of(List.of("s", "a", "b"), R.struct(N.struct(R.I32, R.FP64)));
Rel scan = NamedScan.builder().addNames("t").initialSchema(schema).build();

SubstraitRelVisitor.convert(substraitToCalcite.convert(scan), extensions);
```

| | schema |
| --- | --- |
| in | `Struct{nullable=false, fields=[Struct{nullable=true, fields=[I32{nullable=false}, FP64{nullable=false}]}]}` |
| out | `Struct{nullable=false, fields=[Struct{nullable=true, fields=[I32{nullable=true}, FP64{nullable=true}]}]}` |

## Where it comes from

`TypeConverter.ToRelDataType` applies nullability with `typeFactory.createTypeWithNullability`, and for a record type Calcite pushes it down into the fields:

```
createStructType([INTEGER, DOUBLE], [a, b])  ->  RecordType(INTEGER NOT NULL a, DOUBLE NOT NULL b) NOT NULL
createTypeWithNullability(that, true)        ->  RecordType(INTEGER a, DOUBLE b)
```

That is Calcite being consistent with SQL, where a null row has null fields. Substrait says something narrower — the struct may be absent, and where it is present its fields are not — and the conversion widens it silently instead of reporting that the target cannot hold it.

It is specific to structs. A nullable list and a nullable map of non-nullable parts both come back as they went in:

```
N.list(R.I32)           ->  INTEGER NOT NULL ARRAY                 ->  ListType{nullable=true, elementType=I32{nullable=false}}
N.map(R.STRING, R.I32)  ->  (VARCHAR NOT NULL, INTEGER NOT NULL) MAP  ->  Map{nullable=true, key=Str{nullable=false}, value=I32{nullable=false}}
```

## One note for whoever takes this

[...]

PR #1158, opened 2026-08-25
feat(isthmus): apply hint.output_names when converting a plan to Calcite

alexandrefimov:
Isthmus did not look at `RelCommon.hint` at all, so the alternative output field names a producer put on a relation were dropped: a `Project` carrying `output_names = [total_price]` came out as `SELECT a + 1 AS $f2`. The Spark module writes those names and reads them back, so a plan crossing from Spark into Isthmus lost the column names it carried.

The names are applied to the projection this relation's own conversion produced — the one a `Project` becomes, or the one added for a relation with an emit mapping — so the plan keeps the shape it had. Which projection that is is decided by the node, not by the relation: where Calcite builds no operator at all — a filter that cannot filter, a sort with no sort fields, an identity emit mapping — it hands back the node it was given, and that one belongs to the input.

They are dropped where the relation's record type does not describe that projection's columns, type by type. An aggregate over several grouping sets orders its grouping columns by first appearance where Calcite orders them by group key, so the two disagree on which column is which and the names would land on columns the plan does not name.

Only the top-level names are applied. `output_names` is written depth first and names the fields inside a struct column as well, but those belong to the type of the expression producing the column, and Calcite rejects a projection whose row type restates them differently.

A name list that does not fit the relation is dropped as well, rather than failing the conversion: the wrong length, a repeated name, or an empty one. Calcite requires a projection's field names to be distinct, and reads an empty name as the star identifier — `output_names = ["", "x"]` renders as `SELECT "a" + 1 AS *, "a" + 2 AS "x"`, which does not parse. The spec says neither, so both are this repo's reading of it; substrait-io/substrait#1188 asks for it to be settled.

The names live in the row type of the projection, which Calcite treats as non-semantic on purpose, so they describe the tree as it is returned here and do not survive a planner: two projections differing only in their names share one digest — `LogicalProject.NONE.[](input=LogicalTableScan#3,exprs=[+($0, 1)])` for both — and `deepEquals` holds between them, so a planner is free to collapse the two and let one inherit the other's names. Neither forcing a projection nor attaching a `RelHint` changes that, since the digest omits hints too.

`visit(VirtualTableScan)`, `NamedDdl`, `NamedWrite` and `NamedUpdate` route neither their emit mapping nor their hint through this; that is #1160.

The export half, writing the names back from the Calcite row type, is left out until the question in #1157 is settled: always, or behind a switch on `ConverterProvider`.

Sits on top of #1181: without it the names applied here would reach the schema of a created table, which is declared and must not depend on a hint.

Part of #1157

### DIFF
--- a/isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
@@ -198,7 +198,7 @@ public class SubstraitRelNodeConverter
     context.enterScope(AnchoredInput.of(filter.getInput().getRelAnchor(), input.getRowType()));
     RexNode filterCondition = filter.getCondition().accept(expressionRexConverter, context);
     RelNode node = relBuilder.push(input).filter(context.exitScope(), filterCondition).build();
-    return applyRelCommon(node, filter);
+    return applyRelCommon(node, filter, input);
   }
 
   @Override
@@ -229,7 +229,7 @@ public class SubstraitRelNodeConverter
 
     RelNode node =
         relBuilder.push(child).project(rexExprs, List.of(), false, context.exitScope()).build();
-    return applyRelCommon(node, project);
+    return applyRelCommon(node, project, child);
   }
 
   @Override
@@ -239,7 +239,7 @@ public class SubstraitRelNodeConverter
     // Calcite represents CROSS JOIN as the equivalent INNER JOIN with true condition
     RelNode node =
         relBuilder.push(left).push(right).join(JoinRelType.INNER, relBuilder.literal(true)).build();
-    return applyRelCommon(node, cross);
+    return applyRelCommon(node, cross, left, right);
   }
 
   @Override
@@ -256,7 +256,7 @@ public class SubstraitRelNodeConverter
     JoinRelType joinType = asJoinRelType(join);
     RelNode node =
         relBuilder.push(left).push(right).join(joinType, condition, context.exitScope()).build();
-    return applyRelCommon(node, join);
+    return applyRelCommon(node, join, left, right);
   }
 
   private JoinRelType asJoinRelType(Join join) {
@@ -289,14 +289,15 @@ public class SubstraitRelNodeConverter
 
   @Override
   public RelNode visit(Set set, Context context) throws RuntimeException {
-    set.getInputs()
-        .forEach(
-            input -> {
-              relBuilder.push(input.accept(this, context));
-            });
+    List<RelNode> inputs = new ArrayList<>(set.getInputs().size());
+    for (Rel input : set.getInputs()) {
+      RelNode converted = input.accept(this, context);
+      inputs.add(converted);
+      relBuilder.push(converted);
+    }
     RelBuilder builder = getRelBuilder(set);
     RelNode node = builder.build();
-    return applyRelCommon(node, set);
+    return applyRelCommon(node, set, inputs.toArray(new RelNode[0]));
   }
 
   private RelBuilder getRelBuilder(Set set) {
@@ -410,7 +411,7 @@ public class SubstraitRelNodeConverter
     RelNode node = aggregateBuilder.push(child).aggregate(groupKey, aggregateCalls).build();
     // Not applyRelCommon: the mapping applied here is the one rewritten above, not the one the
     // relation carries.
-    return applyOutputNames(applyRemap(node, remap), aggregate);
+    return applyOutputNames(applyRemap(node, remap), aggregate, child);
   }
 
   /**
@@ -628,7 +629,7 @@ public class SubstraitRelNodeConverter
             .collect(Collectors.toList());
     exitUncorrelatedScope(context, Sort.class);
     RelNode node = relBuilder.push(child).sort(sortExpressions).build();
-    return applyRelCommon(node, sort);
+    return applyRelCommon(node, sort, child);
   }
 
   private RexNode directedRexNode(Expression.SortField sortField, Context context) {
@@ -669,7 +670,7 @@ public class SubstraitRelNodeConverter
         fetch.getCount().map(e -> e.accept(expressionRexConverter, context)).orElse(null);
     exitUncorrelatedScope(context, Fetch.class);
     RelNode node = relBuilder.push(child).sortLimit(offset, count, ImmutableList.of()).build();
-    return applyRelCommon(node, fetch);
+    return applyRelCommon(node, fetch, child);
   }
 
   private RelFieldCollation toRelFieldCollation(Expression.SortField sortField, Context context) {
@@ -1143,15 +1144,17 @@ public class SubstraitRelNodeConverter
    * first, and then the alternative output field names of its hint.
    *
    * <p>The rest is dropped. A Calcite {@code RelNode} has no place for a relation's alias, its
-   * statistics, the computations it saves or loads, or its anchor, and inventing one would mean
-   * defining a convention on every consumer's behalf.
+   * statistics, the computations it saves or loads, or its common advanced extension, and inventing
+   * one would mean defining a convention on every consumer's behalf. Its anchor is consumed by the
+   * correlation machinery instead.
    *
    * @param relNode the node the relation was converted into
    * @param rel the relation being converted
+   * @param inputs the nodes this relation's inputs were converted into
    * @return the node, remapped and renamed
    */
-  protected RelNode applyRelCommon(RelNode relNode, Rel rel) {
-    return applyOutputNames(applyRemap(relNode, rel.getRemap()), rel);
+  protected RelNode applyRelCommon(RelNode relNode, Rel rel, RelNode... inputs) {
+    return applyOutputNames(applyRemap(relNode, rel.getRemap()), rel, inputs);
   }
 
   /**
@@ -1161,9 +1164,17 @@ public class SubstraitRelNodeConverter
    * <p>The names are applied to the projection this relation's own conversion produced: the one a
    * {@link Project} becomes, or the one {@link #applyRemap(RelNode, Optional)} adds for a relation
    * with an emit mapping. Anywhere else they are dropped, rather than renaming a node that stands
-   * for another relation or adding a projection the plan never asked for. That leaves a relation
-   * converted into a bare Calcite operator, such as an aggregate that emits its output directly,
-   * without its names.
+   * for another relation or adding a projection the plan never asked for. A relation converted into
+   * a bare Calcite operator therefore keeps the names Calcite derives, and so does one Calcite
+   * builds no operator for at all -- a filter that cannot filter, a sort with no sort fields, an
+   * identity emit mapping -- where the node handed back is the input's, which is what the given
+   * inputs are compared against.
+   *
+   * <p>They are dropped as well where the columns of that projection are not the columns of the
+   * relation's record type, type by type. An aggregate over several grouping sets orders its
+   * grouping columns by first appearance where Calcite orders them by group key, so the two
+   * disagree on what the third column is, and binding the names by position would name columns the
+   * plan does not name.
    *
    * <p>Only the names of the top-level fields are applied. The names of the fields nested inside
    * them belong to the type of the expression that produces the field, which a projection cannot
@@ -1174,23 +1185,31 @@ public class SubstraitRelNodeConverter
    * conversion: a hint holds no meaning of its own and should not be able to break an otherwise
    * valid plan.
    *
+   * <p>The names live in the row type of the projection, which Calcite treats as non-semantic: a
+   * planner may discard them, and two projections that differ only in their names share one digest.
+   * They describe the tree as it is returned here, and do not survive planning.
+   *
    * @param relNode the node to rename, with any emit mapping already applied
    * @param rel the relation being converted
+   * @param inputs the nodes this relation's inputs were converted into
    * @return the renamed node, or the node unchanged where the names do not apply
    */
-  protected RelNode applyOutputNames(RelNode relNode, Rel rel) {
+  protected RelNode applyOutputNames(RelNode relNode, Rel rel, RelNode... inputs) {
     List<String> names = rel.getHint().map(Hint::getOutputNames).orElse(List.of());
-    boolean ownsProjection = rel instanceof Project || rel.getRemap().isPresent();
-    if (names.isEmpty()
-        || !ownsProjection
-        || !(relNode instanceof org.apache.calcite.rel.core.Project)) {
+    if (names.isEmpty() || !(relNode instanceof org.apache.calcite.rel.core.Project)) {
       return relNode;
+    }
+    for (RelNode input : inputs) {
+      if (relNode == input) {
+        return relNode;
+      }
     }
     org.apache.calcite.rel.core.Project project = (org.apache.calcite.rel.core.Project) relNode;
     RelDataType rowType = project.getRowType();
     List<Type> fields = rel.getRecordType().fields();
     if (names.size() != NamedFieldCountingTypeVisitor.countNames(rel.getRecordType())
-        || fields.size() != rowType.getFieldCount()) {
+        || fields.size() != rowType.getFieldCount()
+        || !describesColumnsOf(fields, rowType)) {
       return relNode;
     }
     List<String> fieldNames = new ArrayList<>(fields.size());
@@ -1198,9 +1217,11 @@ public class SubstraitRelNodeConverter
       fieldNames.add(names.get(nameIndex));
       nameIndex += 1 + NamedFieldCountingTypeVisitor.countNames(fields.get(field));
     }
-    if (new HashSet<>(fieldNames).size() != fieldNames.size()) {
-      // Calcite requires the field names of a projection to be distinct, and uniquifying the names
-      // here would hand back names the producer did not ask for.
+    if (fieldNames.stream().anyMatch(String::isEmpty)
+        || new HashSet<>(fieldNames).size() != fieldNames.size()) {
+      // Calcite requires the field names of a projection to be distinct and non-empty -- it reads
+      // an empty one as the star identifier -- and uniquifying the names here would hand back names
+      // the producer did not ask for.
       return relNode;
     }
     return project.copy(
@@ -1209,6 +1230,26 @@ public class SubstraitRelNodeConverter
         project.getProjects(),
         typeFactory.createStructType(
             rowType.getStructKind(), RelOptUtil.getFieldTypeList(rowType), fieldNames));
+  }
+
+  /**
+   * Returns whether the given row type holds the given fields, one for one and in the same order,
+   * ignoring nullability: Calcite derives that from the expression producing the column, and a
+   * relation's record type says what its columns are, not whether a value is there.
+   *
+   * @param fields the field types of a relation's record type
+   * @param rowType the row type of the node it was converted into
+   * @return whether the two describe the same columns in the same order
+   */
+  private boolean describesColumnsOf(List<Type> fields, RelDataType rowType) {
+    for (int field = 0; field < fields.size(); field++) {
+      RelDataType declared = typeConverter.toCalcite(typeFactory, fields.get(field));
+      if (!SqlTypeUtil.equalSansNullability(
+          declared, rowType.getFieldList().get(field).getType())) {
+        return false;
+      }
+    }
+    return true;
   }
 
   /**
--- a/isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
@@ -64,6 +64,7 @@ import java.util.stream.IntStream;
 import java.util.stream.Stream;
 import org.apache.calcite.plan.RelOptSchema;
 import org.apache.calcite.plan.RelOptTable;
+import org.apache.calcite.plan.RelOptUtil;
 import org.apache.calcite.plan.RelTraitDef;
 import org.apache.calcite.prepare.Prepare;
 import org.apache.calcite.rel.RelCollation;
@@ -407,6 +408,8 @@ public class SubstraitRelNodeConverter
             : relBuilder;
 
     RelNode node = aggregateBuilder.push(child).aggregate(groupKey, aggregateCalls).build();
+    // Not applyRelCommon: the mapping applied here is the one rewritten above, not the one the
+    // relation carries.
     return applyOutputNames(applyRemap(node, remap), aggregate);
   }
 
@@ -1136,8 +1139,12 @@ public class SubstraitRelNodeConverter
   }
 
   /**
-   * Applies the {@code RelCommon} data of a relation to the node it was converted into: its emit
-   * mapping first, and then the alternative output field names of its hint.
+   * Applies the parts of a relation's {@code RelCommon} that Calcite can hold: its emit mapping
+   * first, and then the alternative output field names of its hint.
+   *
+   * <p>The rest is dropped. A Calcite {@code RelNode} has no place for a relation's alias, its
+   * statistics, the computations it saves or loads, or its anchor, and inventing one would mean
+   * defining a convention on every consumer's behalf.
    *
    * @param relNode the node the relation was converted into
    * @param rel the relation being converted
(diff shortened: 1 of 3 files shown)


isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
```diff
@@ -987,6 +993,79 @@ public RelNode visitFallback(Rel rel, Context context) throws RuntimeException {
...
+   * conversion: a hint holds no meaning of its own and should not be able to break an otherwise
+   * valid plan.
+   *
+   * @param relNode the node to rename, with any emit mapping already applied
+   * @param rel the relation being converted
+   * @return the renamed node, or the node unchanged where the names do not apply
+   */
+  protected RelNode applyOutputNames(RelNode relNode, Rel rel) {
+    List<String> names = rel.getHint().map(Hint::getOutputNames).orElse(List.of());
+    boolean ownsProjection = rel instanceof Project || rel.getRemap().isPresent();
```
nielspardon: Thread the converted input in — `applyRelCommon(node, rel, child)` — and test `relNode != child` rather than inferring ownership from the POJO. When Calcite elides this relation's own operator (`filter` on an always-true condition, an identity remap, an empty sort field list), `relNode` is the *child's* projection and the parent's names overwrite it. `Filter(condition=bool(true), remap=Remap.of(List.of(0)))` over the hinted Project in `leavesTheNamesOfAnotherRelationAlone` returns `[outer]`, not `[inner]`.

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
```diff
@@ -402,7 +406,9 @@ public RelNode visit(Aggregate aggregate, Context context) throws RuntimeExcepti
             : relBuilder;
 
     RelNode node = aggregateBuilder.push(child).aggregate(groupKey, aggregateCalls).build();
-    return applyRemap(node, remap);
+    // Not applyRelCommon: the mapping applied here is the one rewritten above, not the one the
+    // relation carries.
+    return applyOutputNames(applyRemap(node, remap), aggregate);
```
nielspardon: Compare `fields.get(i)` against `rowType.getFieldList().get(i).getType()` rather than only the counts. `Aggregate.deriveRecordType` orders grouping columns by first appearance while Calcite orders them by ascending `ImmutableBitSet`, so with two grouping sets and a remap the names bind across type boundaries: `groupings=[[b],[a]]` over `(a i64, b string, c fp64)` with `output_names=[k_b,k_a,n,gs]` yields `k_b BIGINT, k_a VARCHAR`. Both count guards pass at 4 == 4.

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
```diff
@@ -987,6 +993,79 @@ public RelNode visitFallback(Rel rel, Context context) throws RuntimeException {
...
+    List<String> fieldNames = new ArrayList<>(fields.size());
+    for (int field = 0, nameIndex = 0; field < fields.size(); field++) {
+      fieldNames.add(names.get(nameIndex));
+      nameIndex += 1 + NamedFieldCountingTypeVisitor.countNames(fields.get(field));
+    }
+    if (new HashSet<>(fieldNames).size() != fieldNames.size()) {
+      // Calcite requires the field names of a projection to be distinct, and uniquifying the names
+      // here would hand back names the producer did not ask for.
+      return relNode;
+    }
```
nielspardon: Reject empty names alongside duplicates — Calcite reads `""` as the star identifier, so `output_names=["", "x"]` converts to `SELECT "a" + 1 AS *, "a" + 2 AS "x"`, which does not parse. The spec never says a name must be non-empty, so I've raised substrait-io/substrait#1188 to settle it; dropping the list is the safe reading meanwhile.

```suggestion
    if (fieldNames.stream().anyMatch(String::isEmpty)
        || new HashSet<>(fieldNames).size() != fieldNames.size()) {
      // Calcite requires the field names of a projection to be distinct and non-empty, and
      // uniquifying the names here would hand back names the producer did not ask for.
      return relNode;
    }
```

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
```diff
@@ -987,6 +993,79 @@ public RelNode visitFallback(Rel rel, Context context) throws RuntimeException {
             rel, rel.getClass().getCanonicalName(), this.getClass().getCanonicalName()));
   }
 
+  /**
+   * Applies the parts of a relation's {@code RelCommon} that Calcite can hold: its emit mapping
+   * first, and then the alternative output field names of its hint.
+   *
+   * <p>The rest is dropped. A Calcite {@code RelNode} has no place for a relation's alias, its
+   * statistics, the computations it saves or loads, or its anchor, and inventing one would mean
+   * defining a convention on every consumer's behalf.
```
nielspardon: Drop `or its anchor` — it is read at seven sites in this class and turned into a `CorrelationId` on `variablesSet`; only its storage on the `RelNode` is absent. The common advanced extension is the part that belongs in that list.

```suggestion
   * <p>The rest is dropped. A Calcite {@code RelNode} has no place for a relation's alias, its
   * statistics, the computations it saves or loads, or its common advanced extension, and inventing
   * one would mean defining a convention on every consumer's behalf. Its anchor is consumed by the
   * correlation machinery instead.
```

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
```diff
@@ -987,6 +993,79 @@ public RelNode visitFallback(Rel rel, Context context) throws RuntimeException {
...
+   * without its names.
+   *
+   * <p>Only the names of the top-level fields are applied. The names of the fields nested inside
+   * them belong to the type of the expression that produces the field, which a projection cannot
+   * restate: Calcite compares the two, nested names included, and rejects a projection whose row
+   * type says something the expressions do not.
+   *
+   * <p>A name list that does not fit the relation is dropped as well, rather than failing the
+   * conversion: a hint holds no meaning of its own and should not be able to break an otherwise
+   * valid plan.
```
nielspardon, line 1186: Say here that the rename does not survive planning, rather than leaving a consumer to find out. Calcite treats a projection's field names as non-semantic on purpose — `Project.explainTerms` skips them at `DIGEST_ATTRIBUTES` "so the optimizer can reuse the projects that differ in field names only", `deepEquals0` compares with `equalsSansFieldNames`, and `RelOptUtil.verifyTypeEquivalence` passes `compareNames=false` — so two subtrees differing only in `output_names` share one digest, and the `HepPlanner` that `examples/isthmus-api` runs collapses them with one side inheriting the other's names. Neither forcing a projection nor attaching a `RelHint` changes that, since the digest omits hints too; `RelRoot.fields` is the only carrier outside the graph, and there is one per plan, so it covers the outermost relation only.

```suggestion
   * <p>A name list that does not fit the relation is dropped as well, rather than failing the
   * conversion: a hint holds no meaning of its own and should not be able to break an otherwise
   * valid plan.
   *
   * <p>The names live in the row type of the projection, which Calcite treats as non-semantic: a
   * planner may discard them, and two projections that differ only in their names share one digest.
   * They describe the tree as it is returned here, and do not survive planning.
```

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
```diff
@@ -987,6 +993,79 @@ public RelNode visitFallback(Rel rel, Context context) throws RuntimeException {
...
+   * @param relNode the node the relation was converted into
+   * @param rel the relation being converted
+   * @return the node, remapped and renamed
+   */
+  protected RelNode applyRelCommon(RelNode relNode, Rel rel) {
+    return applyOutputNames(applyRemap(relNode, rel.getRemap()), rel);
+  }
+
+  /**
+   * Applies the alternative output field names a relation carries in its hint to the node it was
```
nielspardon, line 1161: Keep the rename off the write and DDL paths: a CTAS declaring `tableSchema` names `["b","a"]` over an input hinted `["x","y"]` now round-trips as `["x","y"]`, and since the spec makes hints "optionally provided, optionally consumed", a created object's schema must not depend on one. The pair that causes it predates this PR — inbound `handleCreateTableAs` (890) drops the declared `tableSchema`, keeping only `getNames()`, so outbound `handleCreateTable` has to rebuild it from the input's row type via `getSchema` (`SubstraitRelVisitor.java:892`) — but the rename is what makes it observable, so either carry `tableSchema` through the Calcite node or stop deriving those names from the input. While here, the comment at `SubstraitRelVisitor.java:218` ("Substrait carries output names on Plan.Root, not on the Project") now states the opposite of this PR.

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
```diff
@@ -987,6 +993,79 @@ public RelNode visitFallback(Rel rel, Context context) throws RuntimeException {
...
+   *
+   * <p>The rest is dropped. A Calcite {@code RelNode} has no place for a relation's alias, its
+   * statistics, the computations it saves or loads, or its anchor, and inventing one would mean
+   * defining a convention on every consumer's behalf.
+   *
+   * @param relNode the node the relation was converted into
+   * @param rel the relation being converted
+   * @return the node, remapped and renamed
+   */
+  protected RelNode applyRelCommon(RelNode relNode, Rel rel) {
```
nielspardon: Route `visit(VirtualTableScan)` (799) through this as well — it drops the emit mapping, not just the names, so a scan with `remap=Remap.of([1])` converts carrying all its columns while `getRecordType()` reports one. `NamedDdl` (777), `NamedWrite` (908) and `NamedUpdate` (707) are in the same position.

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
```diff
@@ -987,6 +993,79 @@ public RelNode visitFallback(Rel rel, Context context) throws RuntimeException {
...
+    boolean ownsProjection = rel instanceof Project || rel.getRemap().isPresent();
+    if (names.isEmpty()
+        || !ownsProjection
+        || !(relNode instanceof org.apache.calcite.rel.core.Project)) {
+      return relNode;
+    }
+    org.apache.calcite.rel.core.Project project = (org.apache.calcite.rel.core.Project) relNode;
+    RelDataType rowType = project.getRowType();
+    List<Type> fields = rel.getRecordType().fields();
+    if (names.size() != NamedFieldCountingTypeVisitor.countNames(rel.getRecordType())
```
nielspardon, line 1210: Reuse `SubstraitToCalcite.renameFields` (`SubstraitToCalcite.java:139`) — it already walks a `RelDataType` against a flat depth-first name list and recurses through ROW/ARRAY/MAP, so it covers the nested names this version documents giving up on. It replaces both the guard and the loop, which today walk the record type twice over.

isthmus/src/test/java/io/substrait/isthmus/OutputNamesTest.java
```diff
@@ -0,0 +1,226 @@
...
+            .build();
+
+    RelNode node = substraitToCalcite.convert(filter);
+
+    assertEquals(List.of("a", "b"), node.getRowType().getFieldNames());
+    assertEquals(0, countProjections(node));
+  }
+
+  @Test
+  void leavesTheNamesOfAnotherRelationAlone() {
```
nielspardon, line 224: Make the inner relation a `Project` here. With a `NamedScan` input `relNode instanceof Project` is already false, so this test and `leavesARelationWithAnIdentityEmitMappingAlone` both pass with `|| !ownsProjection` deleted from line 1040.

isthmus/src/test/java/io/substrait/isthmus/OutputNamesTest.java
```diff
@@ -0,0 +1,226 @@
...
+ */
+class OutputNamesTest extends PlanTestBase {
+
+  private final Rel scan = sb.namedScan(List.of("t"), List.of("a", "b"), List.of(R.I64, N.STRING));
+
+  private Project projectWithHint(Optional<Hint> hint) {
+    ImmutableProject.Builder builder =
+        Project.builder()
+            .input(scan)
+            .remap(Rel.Remap.offset(2, 1))
```
nielspardon, line 30: Add a `Project` case carrying `output_names` with no emit mapping. Every fixture here sets `.remap(...)` (29, 89, 133, 205), so `rel instanceof Project ||` never decides and can be deleted with the suite still green.

isthmus/src/test/java/io/substrait/isthmus/OutputNamesTest.java
```diff
@@ -0,0 +1,226 @@
...
+            .addExpressions(sb.fieldReference(structScan, 0));
+    hint.ifPresent(builder::hint);
+    return builder.build();
+  }
+
+  @Test
+  void keepsCalciteNamesWithoutAHint() {
+    RelNode node = substraitToCalcite.convert(projectWithHint(Optional.empty()));
+
+    assertEquals(List.of("$f2"), node.getRowType().getFieldNames());
```
nielspardon, line 102: Assert against the un-hinted conversion rather than `$f2`/`$f3` (99, 108, 142), the way `namesTopLevelFieldsOfANestedRow` already does.

---

nielspardon (changes requested):
Three reproducible defects below, all in `applyOutputNames`: names land on a child relation's projection when Calcite elides this relation's operator, a multi-grouping-set aggregate binds them to the wrong columns, and an empty name renders as `AS *`. Neither disjunct of `ownsProjection` is constrained by a test — deleting either one leaves the suite green. Dropping names rather than forcing them is the right call — the spec makes hints "optionally consumed" — but two things belong in the description: that the rename does not survive a planner, since Calcite keeps projection field names out of the rel digest by design, and that the empty-name and duplicate-name cases are this repo's reading of a spec that does not cover them (substrait-io/substrait#1188).

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
alexandrefimov: Applied as suggested: `applyRelCommon(node, rel, inputs...)` threads the converted inputs in, and the node is renamed only where it is none of them. `ownsProjection` is gone.

Your case reproduces as written -- `Filter(condition=bool(true), remap=Remap.of(List.of(0)))` over the hinted Project returns `[outer]`. `leavesTheNamesOfAnotherRelationAloneWhenItsOwnOperatorIsElided` pins it: put the old rule back and it is the one test that fails.

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
alexandrefimov: Applied, comparing the other way round: each record-type field is converted to Calcite and compared against the row type's field with `SqlTypeUtil.equalSansNullability`. Converting the row type back to Substrait instead drops the names for a nullable struct column, since Calcite makes a struct's fields nullable along with the struct (#1154) where the record type does not.

Your example reproduces as written: `k_b BIGINT | k_a VARCHAR | n BIGINT | gs VARCHAR`, both counts at 4. `dropsNamesWhereTheColumnsAreNotTheRelationsColumns` pins it.

One thing that turned up while measuring it, on `main` and not from this PR: that fourth column is `b0=[$1]`, a copy of the second grouping column. Where an emit mapping keeps the grouping-set index, `visit(Aggregate)` rewrites its index to `aggregateCalls.size() - 1`, which counts aggregate calls rather than output columns, so it lands on a grouping column; where the aggregate emits directly the same column comes out as the constant `0:BIGINT`. Raising it separately.

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
alexandrefimov: Applied as suggested. `dropsNamesThatAreEmpty` pins it and also asserts the plan still renders as SQL -- without the check it renders as `SELECT "a" + 1 AS *, "a" + 2 AS "x"`, as you said.

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
alexandrefimov: Applied as suggested.

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
alexandrefimov: Applied as suggested, and the PR description says it too. Measured the digest while writing it: the hinted and un-hinted conversions of the same projection both print `LogicalProject.NONE.[](input=LogicalTableScan#3,exprs=[+($0, 1)])` and `deepEquals` holds between them.

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
alexandrefimov: Reproduced, and the hint is not needed to make it go wrong: a CTAS declaring `[b, a]` over a plain projection comes back as `[$f2, $f3]`, because the schema is rebuilt from the input's row type whatever names are on it.

Carried it instead, in #1181: `CreateTable` and `CreateView` hold the declared schema, the conversion fills it from `tableSchema` and reads it back. This PR sits on top of that one, so the rename cannot reach a created object's schema. The comment at `SubstraitRelVisitor.java:218` is updated here -- output names live on `Plan.Root` and in a relation's hint, and this direction writes neither.

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
alexandrefimov: Left for #1160, which is the emit-mapping half of this and predates the review -- I would rather fix it there, with a test per relation.

Two of the four also want a decision rather than a routing change: a `NamedWrite` becomes either a `TableModify`, whose row type is a single ROWCOUNT column, or a `CreateTable`, and a `NamedDdl` becomes a `CreateView`, so an emit mapping over any of them has no columns to select, and rejecting it the way `visit(NamedUpdate)` already does looks more honest than applying it. Leaving this open for your read on that.

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
alexandrefimov: Measured it, and it does not hold here: renaming the nested names makes Calcite reject the projection.

```
AssertionError: type mismatch:
type1:
RecordType(BIGINT NOT NULL f0, VARCHAR f1) NOT NULL
type2:
RecordType(BIGINT NOT NULL first, VARCHAR second) NOT NULL
```

That is the difference between the two callers: `RelRoot.fields` is a row type nothing has to agree with, and a `Project`'s row type is compared against its expressions. So the walk stays top-level here.

The second half of your point stands though -- the count and the loop do walk the record type twice. Leaving open in case you want them merged into one pass.

isthmus/src/test/java/io/substrait/isthmus/OutputNamesTest.java
alexandrefimov: Measured this one before touching it: it does fail with `|| !ownsProjection` deleted -- it and `leavesAnAggregateThatEmitsDirectlyAlone` are the two that do. What no test decides is the other disjunct, `rel instanceof Project ||`, which is your point at line 29; deleting that one leaves the suite green.

Both are moot now that the node decides, and the elided-operator case you describe has its own test.

isthmus/src/test/java/io/substrait/isthmus/OutputNamesTest.java
alexandrefimov: Added `namesAProjectionWithoutAnEmitMapping`. It is also what the `rel instanceof Project` disjunct was for, and I measured that nothing else covered it.

isthmus/src/test/java/io/substrait/isthmus/OutputNamesTest.java
alexandrefimov: Applied at 108 and 142.

At 99 the test is the un-hinted conversion, so comparing it against itself would say nothing: what it pins is that a relation without a hint keeps the name Calcite derives, and `$f2` is that name. If the hardcoded name is what bothers you there, I would rather drop the test than assert it against itself -- the two below cover the same ground now. Leaving it open.

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
nielspardon: Filed #1183 for the dropped emit mapping — it predates this PR and is not about naming, so it should not hold this up.

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
nielspardon: Filed #1184 rather than holding this PR for it.

nielspardon (approved):
Fixes verified. Deciding ownership by node identity also covers the multi-input relations my suggestion would have missed. The two remaining threads are tracked in #1160 and #1178.

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
nielspardon: Correction: #1183 was a duplicate — #1160 already tracks this. Closed mine.

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
nielspardon: Correction: #1184 was a duplicate of #1178, which already covers the walker duplication. Closed mine and added the extra detail there.

alexandrefimov:
This branch carries #1181's commit, so squashing this one first would fold the CTAS schema fix into this entry and leave #1180 open. Merging #1181 first works better -- I rebase this onto it and it drops back to its own three commits.

nielspardon:
> This branch carries #1181's commit, so squashing this one first would fold the CTAS schema fix into this entry and leave #1180 open. Merging #1181 first works better -- I rebase this onto it and it drops back to its own three commits.

I noticed that. Reviewing #1181 in parallel so we can merge it first.

nielspardon:
`8121d198` ("read the column list a create statement declares") looks like it wants to be its own PR, the way the CTAS schema fix became #1181: it fixes a distinct defect in a different file set (`DdlSqlToRelConverter`, `DdlSchemas`, the two DDL nodes), and it corrects an assumption #1180 stated explicitly rather than following from this PR's subject. Filed #1187 for it.

Concretely, this PR's title and body become the squash commit `semantic-release` reads, so a `fix` folded under `feat(isthmus): apply hint.output_names ...` ships with no changelog entry of its own.

One finding on that commit, carried over to #1187 so it is not lost: `declaredSchema` ends in `createStructType(types, names)` with no distinctness check, and `create table d (a, a) as select intcol, intcol from src1` yields `RecordType(INTEGER A, INTEGER A) NOT NULL` — the case `applyOutputNames` rejects a few hundred lines away, for the same reason.

The three output-names commits are approved as they stand; dropping `8121d198` leaves this ready to merge.

alexandrefimov:
Splitting it was the right call, and the changelog reason is the one that decides it: a `fix` folded under a `feat` title ships with no entry of its own. The duplicate-name finding is real too -- `create table d (a, a) as select intcol, intcol from src1` gives `RecordType(INTEGER A, INTEGER A)` and records the schema as `[A, A]` on main today.

It went in with #1181, though: `8121d198` was the head of that branch when you merged it, and the squash `40bd9da` carries `DdlSchemas` and the column-list handling in `DdlSqlToRelConverter`. So there is nothing to drop here -- rebasing onto main was enough, and this PR is now its three output-names commits.

That leaves #1187 with the duplicate check; happy to take it there.

PR #1161, opened 2026-08-25
fix(isthmus)!: keep the declared column order of an aggregate over grouping sets

### LINKED ISSUE #1159: isthmus: an aggregate over several grouping sets converts to Calcite in a different column order, so references over it read the wrong column
An aggregate over more than one grouping set declares its output as the unique grouping expressions in the order they first appear, then the measures, then the grouping-set index — that is what `Aggregate.deriveRecordType` builds, from a `LinkedHashSet` of the expressions. The conversion to Calcite hands those expressions to a group key, which is an `ImmutableBitSet`, so Calcite orders the grouping columns by field index. When the two orders disagree, nothing complains and the columns are misaligned.

A scan `t(a i64, b string)` grouped by the sets `[[b], [a]]` with `count(a)` as its measure declares `[string?, i64?, i64, i32]`, and converts to:

```
RecordType(BIGINT a, VARCHAR b, BIGINT NOT NULL $f2, BIGINT NOT NULL $f3)
```

`b` is declared first and lands second. A parent that reads field 0 of that aggregate is asking for `b`:

```
Project(remap=[4], expressions=[field(0)])   declares [string?]
```

and after conversion reads `a` instead, with the type quietly changing with it:

```
LogicalProject(a0=[$0])
  LogicalAggregate(group=[{0, 1}], groups=[[{0}, {1}]], agg#0=[COUNT($0)])
    LogicalTableScan(table=[[t]])

RecordType(BIGINT a0)
```

A single grouping set is fine: grouping by `(b, a)` converts to `RecordType(VARCHAR b0, BIGINT a0, BIGINT $f2)`, and a reference to field 0 stays on `b`.

Reproduced on `main` at 4242051. This is the Substrait → Calcite direction, so it shows up on plans that come from another producer rather than on ones Isthmus emitted itself. Found while reviewing #1158, which wanted to name those columns and could only pin the names onto whatever order came out.

alexandrefimov:
substrait-java takes the grouping columns of an aggregate to be the distinct grouping expressions in the order they first appear across its grouping sets, while Calcite takes them from a bit set and emits them ordered by field index. (The spec orders them by the relation's shared grouping-expression list, which each set's expression references index into; the POJO models a per-set expression list and cannot hold that order, so first-appearance is this library's reconstruction — spec v0.101.0.) Neither direction accounted for that, so a plan whose sets first mention field 1 and then field 0 changed meaning on the way through: a reference to the aggregate's first column reached the column Calcite had put there instead, and its type changed with it.

Both directions now carry the difference in the emit mapping rather than in the shape of the plan. On the way in, the mapping a relation carries is translated into the order the converted aggregate emits, and a relation that emits directly gets the mapping that puts its columns back in the declared order. On the way out, the mapping presents the aggregate's output in Calcite's order, so a parent converted from the same Calcite plan finds its columns where it left them. Neither direction adds a relation the plan did not have, so a plan that already agrees with Calcite round-trips unchanged — which is what `Substrait2SqlTest.simpleTestGroupingSets` checks, and it is also how I found that fixing only one direction is not enough.

The pre-aggregate projection reuses one column for a field grouped on by several sets — two copies of it would each be missing from a grouping set, and Calcite would make both nullable — and only there, because that is where `Aggregate.deriveRecordType` dedups: a lone grouping set gives every mention a column of its own.

That leaves the order itself no reason to send an aggregate through the projection, since the emit mapping now carries it, and it no longer does. What a rewrite is still needed for is a grouping set naming an expression twice, which Calcite's bit set cannot hold. Measured against `origin/main` (934a60e):

| aggregate over `foo(a i64, b i64, c string)` | declares | `main` | now |
| --- | --- | --- | --- |
| one set `(2, 0, 2)` | 3 columns | 3 | 3 |
| one set `(2, 0, 2)`, emit `(0, 1, 2)` | 3 columns | 3 | 3 |
| one set `(0, 2, 2)` | 3 columns | **2** | 3 |
| one set, `a + 42` twice | 2 columns | 2 | 2 |
| two sets `(2, 0)` and `(2)` | 3 columns | **4** | 3 |

The `(0, 2, 2)` row is wrong on `main` as well: its fields are ascending, so the aggregate went to Calcite directly and the bit set ate a column. The visible change from the order no longer triggering a rewrite is where the reordering sits — `LogicalProject(c=[$1], a=[$0])` over `LogicalAggregate(group=[{0, 2}])` over the scan, where the projection used to sit under the aggregate.

The per-mention count that keeps a repeated grouping expression a column of its own is a convention of this library rather than something the plan carries. `RelProtoConverter.visit(Aggregate)` dedups `grouping_expressions` unconditionally, where `Aggregate.deriveRecordType` dedups only across several sets, so the one-set `(2, 0, 2)` above writes two grouping expressions with references `[0, 1, 0]`: three columns here, two to a consumer reading the proto as the spec's list of grouping expressions in declaration order. A round trip through proto still compares equal, because both directions flatten per-`Grouping`. That split is #1210.

Worth a second look, since the change touches a branch that had no coverage: the GROUP_ID branch of `SubstraitRelVisitor.visit(Aggregate)` used a grouping-field count that included a field shared by several sets more than once, which disagrees with `Aggregate.deriveRecordType`. It now uses the distinct count. No query reaches that branch — Calcite folds `GROUP_ID()` into a literal wherever it can work out the answer, including over duplicated grouping sets — so the test that covers it builds the Calcite aggregate directly.

Two more defects in the grouping-set index column came out of the same math and are fixed here. The mapping that keeps that column replaced its index with `aggregateCalls.size() - 1`, which counts aggregate calls rather than output columns, so what came back was a copy of a grouping column. And the index the relation declares for it counted every mention of a grouping expression, so a field grouped on by several sets shifted it past the end: such an aggregate with a mapping that keeps the index threw `ArrayIndexOutOfBoundsException`, on `main` as well. Both counts are over the distinct grouping columns now. What the column holds is still Calcite's folded `GROUP_ID()` literal, which is #1182.

One difference is left alone: the grouping-set index comes back as an i64. The conversion builds Calcite's `GROUP_ID` call as a BIGINT while the spec gives the aggregate an i32 column, and Calcite folds the call to a literal of its own type. That is #1162.

Closes #1159

BREAKING CHANGE: an aggregate whose grouping sets do not first mention its fields in ascending order now converts with its grouping columns in the order the relation declares them, in both directions. For the sets `{0, 3}` and `{1, 2}`, converting the relation to Calcite emits fields 0, 3, 1, 2 where it emitted 0, 1, 2, 3, and converting the Calcite aggregate back gives the emit mapping `[0, 2, 3, 1]` where it gave `[0, 1, 2, 3]`. An aggregate whose grouping sets already agree with Calcite's field order converts as before.

### DIFF
--- a/isthmus/src/main/java/io/substrait/isthmus/PreCalciteAggregateValidator.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/PreCalciteAggregateValidator.java
@@ -8,6 +8,7 @@ import io.substrait.relation.Aggregate;
 import io.substrait.relation.Project;
 import java.util.ArrayList;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
@@ -36,7 +37,28 @@ public class PreCalciteAggregateValidator {
     return aggregate.getMeasures().stream()
             .allMatch(PreCalciteAggregateValidator::isValidCalciteMeasure)
         && aggregate.getGroupings().stream()
-            .allMatch(PreCalciteAggregateValidator::isValidCalciteGrouping);
+            .allMatch(PreCalciteAggregateValidator::isValidCalciteGrouping)
+        && aLoneGroupingSetNamesEachExpressionOnce(aggregate);
+  }
+
+  /**
+   * Checks that an aggregate holding one grouping set does not name an expression in it twice.
+   *
+   * <p>Calcite holds a grouping set in an {@link org.apache.calcite.util.ImmutableBitSet}, which
+   * cannot hold a field twice, while a lone grouping set gives each mention a column of its own --
+   * {@code Aggregate.deriveRecordType} dedups the grouping expressions only across several sets. So
+   * a repeat has to reach Calcite as two columns of the relation underneath, which is what the
+   * transformer makes of it.
+   *
+   * @param aggregate the aggregate relation
+   * @return {@code true} if valid, {@code false} otherwise
+   */
+  private static boolean aLoneGroupingSetNamesEachExpressionOnce(Aggregate aggregate) {
+    if (aggregate.getGroupings().size() != 1) {
+      return true;
+    }
+    List<Expression> expressions = aggregate.getGroupings().get(0).getExpressions();
+    return new HashSet<>(expressions).size() == expressions.size();
   }
 
   /**
@@ -62,32 +84,19 @@ public class PreCalciteAggregateValidator {
   }
 
   /**
-   * Checks if an {@link Aggregate.Grouping} uses only {@link FieldReference}s and ensures grouping
-   * fields are in ascending order.
+   * Checks if an {@link Aggregate.Grouping} uses only {@link FieldReference}s.
+   *
+   * <p>The order the fields are grouped in is not a reason to rewrite the aggregate. Calcite holds
+   * a grouping set in an {@link org.apache.calcite.util.ImmutableBitSet} and emits its grouping
+   * columns in ascending field order whatever order they were declared in, so a plan grouping on
+   * (0, 2, 1) reaches Calcite as (0, 1, 2); the conversion carries the declared order in the emit
+   * mapping instead.
    *
    * @param grouping the aggregate grouping to validate
    * @return {@code true} if valid, {@code false} otherwise
    */
   private static boolean isValidCalciteGrouping(Aggregate.Grouping grouping) {
-    if (!grouping.getExpressions().stream().allMatch(e -> isSimpleFieldReference(e))) {
-      return false;
-    }
-
-    // Calcite stores grouping fields in an ImmutableBitSet and does not track the order of the
-    // grouping fields. The output record shape that Calcite generates ALWAYS has the groupings in
-    // ascending field order. This causes issues with Substrait in cases where the grouping fields
-    // in Substrait are not defined in ascending order.
-
-    // For example, if a grouping is defined as (0, 2, 1) in Substrait, Calcite will output it as
-    // (0, 1, 2), which means that the Calcite output will no longer line up with the expectations
-    // of the Substrait plan.
-
-    List<Integer> groupingFields =
-        grouping.getExpressions().stream()
-            .map(expr -> getFieldRefOffset((FieldReference) expr))
-            .collect(Collectors.toList());
-
-    return isOrdered(groupingFields);
+    return grouping.getExpressions().stream().allMatch(e -> isSimpleFieldReference(e));
   }
 
   private static boolean isSimpleFieldReference(FunctionArg e) {
@@ -99,33 +108,24 @@ public class PreCalciteAggregateValidator {
     return segments.size() == 1 && segments.get(0) instanceof FieldReference.StructField;
   }
 
-  private static int getFieldRefOffset(FieldReference fr) {
-    return ((FieldReference.StructField) fr.segments().get(0)).offset();
-  }
-
-  private static boolean isOrdered(List<Integer> list) {
-    for (int i = 1; i < list.size(); i++) {
-      if (list.get(i - 1) > list.get(i)) {
-        return false;
-      }
-    }
-    return true;
-  }
-
   /**
-   * Transforms invalid aggregates into Calcite-compatible form by projecting non-field expressions
-   * and reordering groupings.
+   * Transforms invalid aggregates into Calcite-compatible form by projecting out the grouping
+   * expressions Calcite cannot hold as they are.
    */
   public static class PreCalciteAggregateTransformer {
 
     // New expressions to include in the project before the aggregate
     private final List<Expression> newExpressions;
 
-    // The field reference each grouping expression was projected out to. A field grouped on by
-    // several grouping sets is one column of the aggregate's output, so it has to stay one column
-    // of the project underneath it: two copies of it would each be missing from a grouping set,
-    // and Calcite would make both of them nullable.
+    // The field reference each grouping expression was projected out to, kept only where the
+    // aggregate's own record type shares a column between mentions of one expression: with several
+    // grouping sets a field grouped on by two of them is one column of the output, so it has to
+    // stay one column of the project underneath -- two copies of it would each be missing from a
+    // set, and Calcite would make both nullable. A lone grouping set gives every mention a column
+    // of its own, so there the map is not consulted.
     private final Map<Expression, Expression> projectedGroupingExpressions;
+
+    private final boolean groupingColumnsAreShared;
 
     // Tracks the offset of the next expression added
     private int expressionOffset;
@@ -133,16 +133,14 @@ public class PreCalciteAggregateValidator {
     private PreCalciteAggregateTransformer(Aggregate aggregate) {
       this.newExpressions = new ArrayList<>();
       this.projectedGroupingExpressions = new HashMap<>();
+      this.groupingColumnsAreShared = aggregate.getGroupings().size() > 1;
       this.expressionOffset = aggregate.getInput().getRecordType().fields().size();
     }
 
     /**
-     * Rewrites an {@link Aggregate} so that it can be converted to Calcite by:
-     *
-     * <ul>
-     *   <li>Projecting non-field references before aggregation
-     *   <li>Ensuring groupings are in ascending order
-     * </ul>
+     * Rewrites an {@link Aggregate} so that it can be converted to Calcite by projecting the
+     * grouping expressions and the measures' non-field arguments out before the aggregation, so
+     * that each is a field reference of its own.
      *
      * @param aggregate the original Substrait aggregate
      * @return a transformed Calcite-compatible aggregate
@@ -203,7 +201,11 @@ public class PreCalciteAggregateValidator {
     private Aggregate.Grouping updateGrouping(Aggregate.Grouping grouping) {
       List<Expression> newGroupingExpressions =
           grouping.getExpressions().stream()
-              .map(expr -> projectedGroupingExpressions.computeIfAbsent(expr, this::projectOut))
+              .map(
+                  expr ->
+                      groupingColumnsAreShared
+                          ? projectedGroupingExpressions.computeIfAbsent(expr, this::projectOut)
+                          : projectOut(expr))
               .collect(Collectors.toList());
       return Aggregate.Grouping.builder().expressions(newGroupingExpressions).build();
     }
--- a/isthmus/src/main/java/io/substrait/isthmus/PreCalciteAggregateValidator.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/PreCalciteAggregateValidator.java
@@ -7,7 +7,9 @@ import io.substrait.expression.FunctionArg;
 import io.substrait.relation.Aggregate;
 import io.substrait.relation.Project;
 import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 import java.util.Optional;
 import java.util.stream.Collectors;
 
@@ -119,11 +121,18 @@ public class PreCalciteAggregateValidator {
     // New expressions to include in the project before the aggregate
     private final List<Expression> newExpressions;
 
+    // The field reference each grouping expression was projected out to. A field grouped on by
+    // several grouping sets is one column of the aggregate's output, so it has to stay one column
+    // of the project underneath it: two copies of it would each be missing from a grouping set,
+    // and Calcite would make both of them nullable.
+    private final Map<Expression, Expression> projectedGroupingExpressions;
+
     // Tracks the offset of the next expression added
     private int expressionOffset;
 
     private PreCalciteAggregateTransformer(Aggregate aggregate) {
       this.newExpressions = new ArrayList<>();
+      this.projectedGroupingExpressions = new HashMap<>();
       this.expressionOffset = aggregate.getInput().getRecordType().fields().size();
     }
 
@@ -193,7 +202,9 @@ public class PreCalciteAggregateValidator {
 
     private Aggregate.Grouping updateGrouping(Aggregate.Grouping grouping) {
       List<Expression> newGroupingExpressions =
-          grouping.getExpressions().stream().map(this::projectOut).collect(Collectors.toList());
+          grouping.getExpressions().stream()
+              .map(expr -> projectedGroupingExpressions.computeIfAbsent(expr, this::projectOut))
+              .collect(Collectors.toList());
       return Aggregate.Grouping.builder().expressions(newGroupingExpressions).build();
     }
 
--- a/isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
@@ -418,10 +418,10 @@ public class SubstraitRelNodeConverter
    * expressions in the order they first appear across its grouping sets. The spec orders them by
    * the relation's shared grouping-expression list, which each set's expression references index
    * into; the POJO models a per-set expression list and cannot hold that list's order, so
-   * first-appearance is the reconstruction this library reads and writes (spec v0.101.0). Calcite
-   * takes them from a bit set, so it emits them ordered by field index. A relation whose grouping
-   * sets first mention field 1 and then field 0 declares them in that order, and its emit mapping
-   * indexes that order, while the aggregate underneath emits field 0 first.
+   * first-appearance is the reconstruction this library reads and writes. Calcite takes them from a
+   * bit set, so it emits them ordered by field index. A relation whose grouping sets first mention
+   * field 1 and then field 0 declares them in that order, and its emit mapping indexes that order,
+   * while the aggregate underneath emits field 0 first.
    *
    * <p>An aggregate that emits directly and declares an order Calcite does not produce gets a
    * mapping it did not carry, which is what puts the columns back in the declared order.
--- a/isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
@@ -55,7 +55,6 @@ import java.util.Deque;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.LinkedHashSet;
-import java.util.LinkedList;
 import java.util.List;
 import java.util.Locale;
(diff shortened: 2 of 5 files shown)


isthmus/src/main/java/io/substrait/isthmus/PreCalciteAggregateValidator.java
```diff
@@ -193,7 +202,9 @@ private Aggregate.Measure updateMeasure(Aggregate.Measure measure) {
 
     private Aggregate.Grouping updateGrouping(Aggregate.Grouping grouping) {
       List<Expression> newGroupingExpressions =
-          grouping.getExpressions().stream().map(this::projectOut).collect(Collectors.toList());
+          grouping.getExpressions().stream()
+              .map(expr -> projectedGroupingExpressions.computeIfAbsent(expr, this::projectOut))
```
nielspardon: Drop the `isOrdered` trigger in `isValidCalciteGrouping` rather than adding this dedup map — the emit mapping added in this PR already handles a reorder within one grouping set (with `isOrdered` stubbed to return true, `outOfOrderGroupingKeysHaveCorrectCalciteType` still passes), so the map has no job left. As written it dedups unconditionally while `Aggregate.deriveRecordType` dedups only when `getGroupings().size() > 1`: an aggregate with one grouping set naming an expression twice declares two columns and converts to one, and with an emit mapping over it the conversion throws `ArrayIndexOutOfBoundsException`. `main` converted the same plan correctly.

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
```diff
@@ -411,7 +418,65 @@ public RelNode visit(Aggregate aggregate, Context context) throws RuntimeExcepti
...
+        applyRemap(node, inConvertedGroupingOrder(remap, groupExprs, aggregateCalls.size())),
+        aggregate,
+        child);
+  }
+
+  /**
+   * Returns the emit mapping of a converted aggregate with its indices translated from the order
+   * the relation declares its output in to the order the converted aggregate emits it.
+   *
+   * <p>Substrait takes the grouping columns of an aggregate to be the distinct grouping expressions
```
nielspardon: Attribute this to substrait-java rather than to Substrait, or drop the claim: the spec's aggregate Direct Output Order is "the list of grouping expressions in declaration order", meaning the order of `AggregateRel.grouping_expressions`, which `Grouping.expression_references` indexes into. First-appearance order is this library's reconstruction, forced because the POJO models per-`Grouping` expression lists and cannot hold the shared list's order — so a foreign aggregate whose shared list is not in first-appearance order has its columns reordered on the way in. Same wording at `SubstraitRelVisitor` 418 and 506 and in the PR description (spec v0.101.0).

nielspardon (commented):
The unconditional dedup in `updateGrouping` needs a decision before the rest is worth reading: a single grouping set that names a grouping expression twice now converts to fewer Calcite columns than the relation declares, where `main` converted the same plan correctly. Dropping the `isOrdered` trigger in `isValidCalciteGrouping` looks like the cleanest way out, since the emit mapping added here already handles a within-set reorder and the dedup map then has no job — but that reaches past the bug, so it is your call.

isthmus/src/main/java/io/substrait/isthmus/PreCalciteAggregateValidator.java
alexandrefimov: Dropped. It needed one more piece: with the order no longer a trigger, a single set `(2, 0, 2)` reaches Calcite directly and the bit set collapses the repeat, so a relation declaring three columns converts to two. What the transformer keeps as a trigger is that repeat — a lone grouping set naming an expression twice — and the map is consulted only where `deriveRecordType` dedups, which is across several sets.

Measured against `origin/main` (934a60e), with the branch as you read it in the middle column:

| aggregate over `foo(a i64, b i64, c string)` | declares | `main` | branch | now |
| --- | --- | --- | --- | --- |
| one set `(2, 0, 2)` | 3 | 3 | 2 | 3 |
| one set `(2, 0, 2)`, emit `(0, 1, 2)` | 3 | 3 | AIOOBE | 3 |
| one set `(0, 2, 2)` | 3 | 2 | 2 | 3 |
| one set, `a + 42` twice | 2 | 2 | 1 | 2 |
| two sets `(2, 0)` and `(2)` | 3 | 4 | 3 | 3 |

The `(0, 2, 2)` row was wrong on `main` as well: ascending, so `isOrdered` let it through and the bit set ate a column. Five tests in `ComplexAggregateTest` hold the rows, and each fails on the branch as it stood.

The other thing the trigger took with it is where the reordering sits — `LogicalProject(c=[$1], a=[$0])` over `LogicalAggregate(group=[{0, 2}])` over the scan, where the projection used to sit under the aggregate.

Leaving this open in case you would rather a repeated grouping expression were refused rather than projected out: it is the one shape that still rewrites the input, and refusing it would take the transformer out of the grouping path entirely.

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
alexandrefimov: Attributed to substrait-java in the three comments and in the description. `RelProtoConverter.visit(Aggregate)` made it concrete while I was there: it builds the shared list as `.distinct()` over the sets, so first-appearance is what this library writes as well as what it reads back.

isthmus/src/main/java/io/substrait/isthmus/PreCalciteAggregateValidator.java
nielspardon: Keep projecting it out — refusing would reject a plan `Aggregate.deriveRecordType` considers valid and `main` converted. Worth recording in the description that the per-mention count is an in-process convention that does not survive serialization: `RelProtoConverter.visit(Aggregate)` dedups `grouping_expressions` unconditionally, so your one-set `(2, 0, 2)` writes `grouping_expressions` of 2 with `expression_references [0, 1, 0]` — three columns here, two to a consumer reading the proto per the spec's "list of grouping expressions in declaration order", and `pojo.equals(back)` still true because both directions flatten per-`Grouping`. That split is #1210, not yours to fix here.

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
```diff
@@ -382,7 +387,9 @@ public RelNode visit(Aggregate aggregate, Context context) throws RuntimeExcepti
               RelCollations.EMPTY,
               typeConverter.toCalcite(typeFactory, TypeCreator.REQUIRED.I64),
               null));
-      final int groupingCallIndex = aggregateCalls.size() - 1;
+      // The call was appended, so it is the last column of the converted aggregate: the grouping
+      // columns come first, then the calls.
+      final int groupingCallIndex = groupColumnCount + aggregateCalls.size() - 1;
```
nielspardon: Delete lines 393-401 and `groupingCallIndex`, keep `groupingSetIndexGetsRemapped` only as the append condition, and rename `lastFieldIndex` to `groupingSetIndex`. `lastFieldIndex` is `groupColumnCount + aggregateCalls.size()` taken before the call is appended and `groupingCallIndex` is `groupColumnCount + aggregateCalls.size() - 1` taken after, so they are the same integer and the loop replaces every match with itself. That also removes the file's last `LinkedList` use.

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelVisitor.java
```diff
@@ -414,32 +415,35 @@ public Rel visit(org.apache.calcite.rel.core.Aggregate aggregate) {
         Aggregate.builder().input(input).addAllGroupings(groupings).addAllMeasures(aggCalls);
 
     if (groupings.size() > 1) {
+      // substrait-java declares the grouping columns of an aggregate as the distinct grouping
```
nielspardon, line 416: Three comments still describe the behaviour this PR removes, all outside the diff so I cannot put suggestions on them: `SubstraitRelVisitor:396-398` still says the GROUP_ID remap branch deliberately uses a non-distinct count and points at a comment this PR deleted; `SubstraitRelNodeConverter:1263` gives grouping-column order as the reason `applyOutputNames` drops names, which is what you just made agree; and `OutputNamesTest:272-274` states the same premise, so that test now passes only on the i32/i64 gate and no longer covers the case it is named for.

isthmus/src/test/java/io/substrait/isthmus/ComplexAggregateTest.java
```diff
@@ -214,6 +224,273 @@ void handleOutOfOrderGroupingArguments() {
...
+            .getInput();
+
+    List<Type> declared = aggregate.getRecordType().fields();
+    List<Type> roundTripped = converted.getRecordType().fields();
+    assertEquals(declared.size(), roundTripped.size());
+    assertEquals(declared.subList(0, 2), roundTripped.subList(0, 2));
+  }
+
+  @Test
+  void anExplicitGroupIdCallKeepsTheDeclaredColumnOrder() {
```
nielspardon, line 355: Add one fixture per direction with three or more grouping sets in a non-swap order — every fixture here is a 2-element transposition, which is its own inverse, so replacing either permutation helper with its inverse leaves the whole isthmus suite green. Calcite sets `[{0,3},{1,2}]` must give `Remap.of([0,2,3,1])`, where the inverse gives `[0,3,1,2]`. It has to be built the way this test does rather than from SQL: `SqlToRelConverter` projects the aggregate's input into first-appearance order, so every SQL-derived plan comes out with the identity mapping.

nielspardon (commented):
Add the `BREAKING CHANGE:` footer — the `!` cuts the release, but the footer text is what fills the ⚠ section of the notes, and without it they ship empty for a change that rewrites the emit mapping in both directions. The rest is three comments the fix outran and one test gap; the type-blind grouping-column count is #1210's, not this PR's.

isthmus/src/main/java/io/substrait/isthmus/PreCalciteAggregateValidator.java
alexandrefimov: Recorded in the description, with the numbers measured rather than restated: the one-set `(2, 0, 2)` writes two `grouping_expressions` and references `[0, 1, 0]`, the POJO keeps three columns, and `pojo.equals(back)` holds after the round trip. #1210 is named there as where the split belongs.

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
alexandrefimov: Gone, along with `groupingCallIndex`; `groupingSetIndexGetsRemapped` is only the append condition now and `lastFieldIndex` is `groupingSetIndex`. The arithmetic checks out: the first is taken before the call is appended and the second after, so `groupColumnCount + (n + 1) - 1` is `groupColumnCount + n`. The `LinkedList` import went with it, and `remap` is final now that nothing reassigns it.

One more line the removal outran: the comment above the guard said "map grouping set index if it is not removed via remap", which described the loop rather than what is left, so it now says what the condition decides.

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelVisitor.java
alexandrefimov: All three updated. `SubstraitRelVisitor` now says both remap branches read the distinct count, which is what they do. The `applyOutputNames` javadoc keeps the rule and changes the example to the one that still holds: the relation types the grouping-set index i32 where the GROUP_ID call standing for it is i64, so the disagreement is over the last column rather than the third.

`OutputNamesTest` I split rather than repointed, and the naming is yours to settle. The case the old test is named for now goes the other way -- with the index mapped away, every column is one the relation declares, and the names do land on (b, a, count) -- so that became a test of its own, `namesAnAggregateWhoseGroupingColumnsCalciteOrdersDifferently`, which fails if `inConvertedGroupingOrder` is stubbed to return its input. The old test keeps the shape that still drops the names and says i32/i64 is why, so it no longer claims to cover the order. Happy to rename either if you would rather the old one kept the name.

isthmus/src/test/java/io/substrait/isthmus/ComplexAggregateTest.java
alexandrefimov: Two fixtures, one per direction, on your sets. Calcite `[{0,3},{1,2}]` gives `Remap.of([0,2,3,1])`, which passed as written. The Substrait side declares (a, d, b, c) where the aggregate underneath emits (a, b, c, d), asserted on the plan text; the scan columns have four different types so the inverse cannot pass by coincidence.

Both are measured against the inverse rather than argued: swapping `converted.indexOf` for `declared.indexOf` in `inConvertedGroupingOrder` fails the first and nothing else in the isthmus suite, and sorting-then-`indexOf` in `calciteGroupingOrder` fails the second and nothing else.

nielspardon (commented):
All three carried findings are fixed and the `BREAKING CHANGE:` footer is in place; one import nit below. The type-blind grouping-column count at `SubstraitRelNodeConverter:368` stays with #1210 rather than growing this PR.

isthmus/src/test/java/io/substrait/isthmus/ComplexAggregateTest.java
```diff
@@ -214,6 +224,325 @@ void handleOutOfOrderGroupingArguments() {
...
+        "LogicalProject(a=[$0], d=[$3], b=[$1], c=[$2], $f4=[0:BIGINT])\n"
+            + "  LogicalAggregate(group=[{0, 1, 2, 3}], groups=[[{0, 3}, {1, 2}]])\n"
+            + "    LogicalTableScan(table=[[foo]])\n",
+        RelOptUtil.toString(relNode));
+  }
+
+  /** The same shape in the other direction, asserted on the mapping the conversion produces. */
+  @Test
+  void groupingSetsInANonSwapOrderGiveAMappingThatIsNotItsOwnInverse() {
+    org.apache.calcite.tools.RelBuilder relBuilder =
```
nielspardon: Import `org.apache.calcite.tools.RelBuilder` and `org.apache.calcite.sql.SqlKind` rather than naming them inline, and drop the `io.substrait.relation.Aggregate` qualifier on line 455 — that one is already imported at line 9. Neither of the other two collides with anything this file imports, and the same pattern is at 343, 358, 384, 394 and 449; base `main` has none of it in this file.

alexandrefimov: All of them, plus `io.substrait.relation.Project` at 316, which is the same thing and also ours -- `RelBuilder`, `SqlKind` and `Project` are imported now and `Aggregate` uses the import it already had. Base `main` had none of these, so the file is back to naming its types the way it did.

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
```diff
@@ -410,7 +404,68 @@ public RelNode visit(Aggregate aggregate, Context context) throws RuntimeExcepti
...
+   * the relation declares its output in to the order the converted aggregate emits it.
+   *
+   * <p>substrait-java takes the grouping columns of an aggregate to be the distinct grouping
+   * expressions in the order they first appear across its grouping sets. The spec orders them by
+   * the relation's shared grouping-expression list, which each set's expression references index
+   * into; the POJO models a per-set expression list and cannot hold that list's order, so
+   * first-appearance is the reconstruction this library reads and writes (spec v0.101.0). Calcite
+   * takes them from a bit set, so it emits them ordered by field index. A relation whose grouping
+   * sets first mention field 1 and then field 0 declares them in that order, and its emit mapping
+   * indexes that order, while the aggregate underneath emits field 0 first.
```
nielspardon: Drop `(spec v0.101.0)` here -- the marker reads as "changed in v0.101.0", and nothing changed. The shared grouping-expression list and the declaration-order rule both arrived in spec #706, released in v0.57.0, and the Direct Output Order sentence is byte-identical from there through v0.101.0; naming the pinned version means this line goes stale on every bump while the rule it describes never moves. The prose already says what the spec orders them by, which is the part a reader needs.

```suggestion
   * first-appearance is the reconstruction this library reads and writes. Calcite takes them from a
   * bit set, so it emits them ordered by field index. A relation whose grouping sets first mention
   * field 1 and then field 0 declares them in that order, and its emit mapping indexes that order,
   * while the aggregate underneath emits field 0 first.
```

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelVisitor.java
```diff
@@ -497,6 +499,43 @@ public Rel visit(org.apache.calcite.rel.core.Aggregate aggregate) {
...
 
+  /**
+   * Returns, for each grouping column of the converted Calcite aggregate, the position that column
+   * holds in the output the Substrait aggregate declares.
+   *
+   * <p>substrait-java takes the grouping columns to be the distinct grouping expressions in the
+   * order they first appear across the grouping sets, reconstructing the shared list the spec
+   * orders them by (spec v0.101.0); Calcite takes them from a bit set and so emits them ordered by
+   * field index. Reading the result as an emit mapping presents the aggregate's output in Calcite's
+   * order.
```
nielspardon: Same marker here.

```suggestion
   * orders them by; Calcite takes them from a bit set and so emits them ordered by field index.
   * Reading the result as an emit mapping presents the aggregate's output in Calcite's order.
```

nielspardon (commented):
Drop the two `(spec v0.101.0)` markers: the rule they annotate has read identically since v0.57.0, so the version names what substrait-java pins rather than when the ordering changed.

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
alexandrefimov: Both markers dropped. Checked the provenance first: 65a7d38 (spec #706) is an ancestor of v0.57.0, released 2024-10-02, and the Direct Output Order row in `logical_relations.md` is byte-identical between v0.57.0 and v0.101.0 -- so the version named nothing that had changed.

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelVisitor.java
alexandrefimov: Dropped with the other one.

nielspardon (approved):
LGTM

Issue #1048, opened 2026-07-31
refactor(isthmus): unify ConverterProvider customization on the builder


nielspardon:
## Context

Review discussion on #1036: now that `ConverterProvider` has a builder, there are two
customization mechanisms — the builder and subclassing — and they can be mixed. As
[@vbarua noted](https://github.com/substrait-io/substrait-java/pull/1036#discussion_r3686978392),
it isn't obvious which one a user should reach for, and the built-in subclasses are
themselves an example of the ambiguity:

```java
// today
new DynamicConverterProvider(ConverterProvider.builder().extensions(extensions))

// could plausibly be
ConverterProvider.builder()
    .extensions(extensions)
    .scalarFunctionConverter(...)
    .sqlOperatorTable(...)
    .build()
```

## What's missing

`DynamicConverterProvider` and `AutomaticDynamicFunctionMappingConverterProvider` subclass
rather than configure because the builder can't express what they override:

- `getSqlOperatorTable()` — both chain dynamically generated operators onto the base table.
- `getCallConverters()` — `DynamicConverterProvider` appends an extra `ScalarFunctionConverter`.
  Addressed by #1166; see the proposal below. That append turns out to be a duplicate that never
  fires (#1208), so this half may be a deletion rather than a migration.

`getSqlOperatorTable()` needs the *derived* extensions/type factory to compute its value, so
exposing it as a plain builder setting also means giving callers a way to build those pieces
(e.g. a public helper that turns an extension collection into the dynamic operators).

## Proposal

1. ~~Add `callConverters` to `ConverterProvider.Builder`.~~ **Done** in #1166, as
   `callConverters(UnaryOperator<List<CallConverter>>)` — a *transform* over the assembled list
   rather than a replacement for it. A plain list cannot work here: `scalarFunctionConverter` goes
   last in `getCallConverters()`, is derived during construction, and `DynamicConverterProvider`
   reassigns it after `super(builder)` returns, so anything frozen at `build()` captures the
   pre-override converter; a caller supplying the whole list themselves would also trip the #1036
   guards. The transform is handed the built-ins with late-bound values in place and runs per
   conversion, so it expresses insert-at-position, append, prepend, replace and filter.
2. Add `sqlOperatorTable` to `ConverterProvider.Builder`. Whether this wants the same transform
   shape is open — unlike the converter list it has no derived-late component, so a plain setting
   may be enough.
3. Expose the derivation helpers the dynamic providers use so a caller can assemble the same
[...]

PR #1166, opened 2026-08-26
feat(isthmus): let the builder transform the call converter list

alexandrefimov:
`ConverterProvider.getCallConverters()` assembles its list internally, so a consumer whose Calcite dialect emits calls the built-in converters do not handle has to subclass the provider to add one. The builder already takes the scalar, aggregate and window function converters; call converters are the remaining piece. This is item 1 of #1048 — `sqlOperatorTable` is untouched here.

The builder takes a transform over the assembled list rather than converters to prepend. Where a caller's converter sits is a decision only the caller can make: ahead of a built-in one it claims a call that one would otherwise take, which is what a dialect giving a call different semantics needs, and also what puts it ahead of `CAST`, `REINTERPRET` and `ROW` — this repository's encoding of nullable and user-defined literals rather than dialect behaviour. Behind them it catches only what none of them claims. A transform says both, and it says replace and remove as well, which is what retiring the dynamic subclasses in item 3 of that issue needs.

The transform is applied where the list is assembled rather than at build time, so it sees the type converter and the scalar function converter as they are then. That matters here: the scalar function converter is derived during construction, and `DynamicConverterProvider` reassigns it after `super(builder)` returns, so anything frozen at `build()` would capture the wrong one. Both dynamic providers keep working — `AutomaticDynamicFunctionMappingConverterProvider` does not override `getCallConverters()`, and `DynamicConverterProvider` appends to what `super` returns, which is now the caller's list.

The case this comes from is an external consumer exporting Impala's Calcite plans through Isthmus: Impala rewrites `IF` and `cast` into operators carrying `SqlKind.OTHER`, which `CallConverters.CASE` and `CallConverters.CAST` decline, so today it subclasses `ConverterProvider` for both.

What does not reach these converters is narrower than an earlier revision of this description claimed: the `RexOver` node of a windowed call, an aggregate call, a subquery, and the reference expression of a field access outside the item, input-reference and field-access kinds it walks through. A windowed call's operands, partition keys and sort keys are converted through the list as usual, which `areConsultedForTheExpressionsInsideAWindowedCall` pins. The extended-expression path is honoured too now that #1168 takes its converter from the provider.

Part of #1048.

### DIFF
--- a/isthmus/src/main/java/io/substrait/isthmus/ConverterProvider.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/ConverterProvider.java
@@ -18,6 +18,7 @@ import java.util.List;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.function.Function;
+import java.util.function.UnaryOperator;
 import org.apache.calcite.avatica.util.Casing;
 import org.apache.calcite.config.CalciteConnectionConfig;
 import org.apache.calcite.config.CalciteConnectionProperty;
@@ -106,10 +107,10 @@ public class ConverterProvider {
   protected TypeConverter typeConverter;
 
   /**
-   * Caller-supplied {@link CallConverter}s, consulted ahead of the built-in ones. See {@link
+   * The transform applied to the assembled {@link CallConverter} list. See {@link
    * #getCallConverters()}.
    */
-  protected final List<CallConverter> additionalCallConverters;
+  protected final UnaryOperator<List<CallConverter>> callConverters;
 
   /** The execution behavior configuration for plans created by this converter. */
   protected final Plan.ExecutionBehavior executionBehavior;
@@ -234,7 +235,7 @@ public class ConverterProvider {
             .orElse(builder.sqlParserConfig);
     this.typeObserver = builder.typeObserver;
     this.aggregateConversion = builder.aggregateConversion;
-    this.additionalCallConverters = List.copyOf(builder.additionalCallConverters);
+    this.callConverters = builder.callConverters;
 
     this.scalarFunctionConverter =
         builder.scalarFunctionConverter.orElseGet(
@@ -378,24 +379,24 @@ public class ConverterProvider {
    * {@link CallConverter}s are used to convert Calcite {@link org.apache.calcite.rex.RexCall}s to
    * Substrait equivalents.
    *
-   * <p>The converters are consulted in order and the first one to return a result wins. Any {@link
-   * Builder#additionalCallConverters(List) additional converters} come first, so a caller whose
-   * dialect gives a call different semantics can claim it before the built-in converter for it
-   * does.
+   * <p>The converters are consulted in order and the first one to return a result wins. The
+   * built-in ones are assembled here and handed to the {@link Builder#callConverters(UnaryOperator)
+   * transform} a caller configured, which decides where its own converters sit: ahead of a built-in
+   * one to claim a call it would otherwise take, or behind them to catch what none of them claims.
    *
-   * <p>This list covers plain calls only. A windowed call is a {@link
-   * org.apache.calcite.rex.RexOver}, which {@link
-   * io.substrait.isthmus.expression.RexExpressionConverter} routes to the window function converter
-   * without consulting these.
+   * <p>Some calls never reach this list: the {@link org.apache.calcite.rex.RexOver} node of a
+   * windowed call, an aggregate call, a {@link org.apache.calcite.rex.RexSubQuery}, and the
+   * reference expression of a {@link org.apache.calcite.rex.RexFieldAccess} outside the item,
+   * input-reference and field-access kinds it walks through. A windowed call's operands, partition
+   * keys and sort keys are converted through this list as usual.
    *
    * @return a list of CallConverter instances
    */
   public List<CallConverter> getCallConverters() {
-    ArrayList<CallConverter> callConverters = new ArrayList<>(additionalCallConverters);
-    callConverters.addAll(CallConverters.defaults(typeConverter));
-    callConverters.add(CallConverters.CREATE_SEARCH_CONV.apply(new RexBuilder(typeFactory)));
-    callConverters.add(scalarFunctionConverter);
-    return callConverters;
+    ArrayList<CallConverter> builtIns = new ArrayList<>(CallConverters.defaults(typeConverter));
+    builtIns.add(CallConverters.CREATE_SEARCH_CONV.apply(new RexBuilder(typeFactory)));
+    builtIns.add(scalarFunctionConverter);
+    return callConverters.apply(builtIns);
   }
 
   // Substrait To Calcite Processing
@@ -602,7 +603,7 @@ public class ConverterProvider {
     private TypeObserver typeObserver = TypeObserver.NOOP;
     private AggregateConversion aggregateConversion = AggregateConversion.DEFAULT;
     private Optional<Casing> unquotedCasing = Optional.empty();
-    private List<CallConverter> additionalCallConverters = List.of();
+    private UnaryOperator<List<CallConverter>> callConverters = UnaryOperator.identity();
 
     // Derived from the extensions and type factory at build time when left unset.
     private Optional<ScalarFunctionConverter> scalarFunctionConverter = Optional.empty();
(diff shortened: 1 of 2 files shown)


isthmus/src/main/java/io/substrait/isthmus/ConverterProvider.java
```diff
@@ -685,6 +703,24 @@ public Builder aggregateConversion(AggregateConversion aggregateConversion) {
...
+     * {@link ConverterProvider#getCallConverters()}.
+     *
+     * @param additionalCallConverters the call converters to consult first
+     * @return this builder
+     */
+    public Builder additionalCallConverters(List<CallConverter> additionalCallConverters) {
+      this.additionalCallConverters =
+          List.copyOf(Objects.requireNonNull(additionalCallConverters, "additionalCallConverters"));
+      return this;
+    }
```
nielspardon, line 737: Take a transform over the assembled list rather than a list to prepend. A plain full-list setter can't work here: `scalarFunctionConverter` goes last, is derived during construction, and `DynamicConverterProvider` reassigns it after `super(builder)` returns, so anything frozen at `build()` captures the wrong one — and a caller supplying their own would trip the #1036 guards. Rename the field, the builder field (`UnaryOperator.identity()` default) and the constructor assignment to match, drop the `List.copyOf` in both places, and import `java.util.function.UnaryOperator`.

```suggestion
    /**
     * Sets a transform over the assembled {@link CallConverter} list, letting a caller insert,
     * replace, reorder or remove converters. The list passed to the operator is mutable and holds
     * the built-in converters in the order {@link ConverterProvider#getCallConverters()} assembles
     * them; the returned list must also be mutable, since a subclass may append to it.
     *
     * @param callConverters a transform over the built-in call converter list
     * @return this builder
     */
    public Builder callConverters(UnaryOperator<List<CallConverter>> callConverters) {
      this.callConverters = Objects.requireNonNull(callConverters, "callConverters");
      return this;
    }
```

nielspardon (commented):
Replace `additionalCallConverters(List)` with `callConverters(UnaryOperator<List<CallConverter>>)` applied inside `getCallConverters()`, so a caller orders converters explicitly instead of merging against a list they can't name. Prepending is what I'd hold this on — it puts caller converters ahead of `CAST`, `REINTERPRET` and `ROW`, which are Isthmus' encoding of nullable and user-defined literals rather than dialect behaviour. A transform also makes append expressible, which is what #1048 item 3 needs to retire the dynamic subclasses.

isthmus/src/main/java/io/substrait/isthmus/ConverterProvider.java
```diff
@@ -368,11 +375,21 @@ public RexExpressionConverter getRexExpressionConverter(SubstraitRelVisitor srv)
...
    * @return a list of CallConverter instances
    */
   public List<CallConverter> getCallConverters() {
-    ArrayList<CallConverter> callConverters =
-        new ArrayList<>(CallConverters.defaults(typeConverter));
+    ArrayList<CallConverter> callConverters = new ArrayList<>(additionalCallConverters);
+    callConverters.addAll(CallConverters.defaults(typeConverter));
     callConverters.add(CallConverters.CREATE_SEARCH_CONV.apply(new RexBuilder(typeFactory)));
     callConverters.add(scalarFunctionConverter);
     return callConverters;
```
nielspardon: Apply the operator to the assembled built-ins here, so it sees the late-bound `typeConverter` and `scalarFunctionConverter` and runs on every call rather than once at `build()`. `DynamicConverterProvider` keeps working unchanged — it appends to what `super` returns, which is now the caller's list.

```suggestion
  public List<CallConverter> getCallConverters() {
    ArrayList<CallConverter> builtIns = new ArrayList<>(CallConverters.defaults(typeConverter));
    builtIns.add(CallConverters.CREATE_SEARCH_CONV.apply(new RexBuilder(typeFactory)));
    builtIns.add(scalarFunctionConverter);
    return callConverters.apply(builtIns);
```

isthmus/src/main/java/io/substrait/isthmus/ConverterProvider.java
```diff
@@ -368,11 +375,21 @@ public RexExpressionConverter getRexExpressionConverter(SubstraitRelVisitor srv)
...
    *
+   * <p>The converters are consulted in order and the first one to return a result wins. Any {@link
+   * Builder#additionalCallConverters(List) additional converters} come first, so a caller whose
+   * dialect gives a call different semantics can claim it before the built-in converter for it
+   * does.
+   *
+   * <p>This list covers plain calls only. A windowed call is a {@link
+   * org.apache.calcite.rex.RexOver}, which {@link
+   * io.substrait.isthmus.expression.RexExpressionConverter} routes to the window function converter
+   * without consulting these.
```
nielspardon: Rewrite this paragraph — it's inaccurate in both directions. A `RexOver`'s operands, partition keys and ORDER BY keys do go through the list; what never reaches it is the `RexOver` node itself, aggregate calls, `RexSubQuery` (so `SOME`/`ALL` and the query constructors stay unclaimable), and a call under a `RexFieldAccess` whose reference expr isn't whitelisted. The paragraph above also needs its `Builder#additionalCallConverters(List)` link updated.

```suggestion
   * <p>Some calls never reach this list: the {@link org.apache.calcite.rex.RexOver} node of a
   * windowed call, an aggregate call, a {@link org.apache.calcite.rex.RexSubQuery}, and a call
   * under a {@link org.apache.calcite.rex.RexFieldAccess}. A windowed call's operands, partition
   * keys and sort keys are converted through this list as usual.
   *
   * <p>A converter is reused for the lifetime of the provider and across threads, so it must be
   * stateless, and it must not pass the call it is claiming back to the top-level converter, which
   * would re-enter this list from the start and recurse without end.
```

isthmus/src/main/java/io/substrait/isthmus/ConverterProvider.java
alexandrefimov: Applied as suggested: `callConverters(UnaryOperator<List<CallConverter>>)`, `UnaryOperator.identity()` as the builder default, no copy in either place, and the field renamed to match. Your reason for it holds beyond ordering -- freezing anything at `build()` captures a scalar function converter `DynamicConverterProvider` has not reassigned yet.

isthmus/src/main/java/io/substrait/isthmus/ConverterProvider.java
alexandrefimov: Applied as suggested. The tests now go through `provider.getRexExpressionConverter(null)` rather than assembling a converter out of the provider's parts, so they also pin that the transform reaches a conversion: routing that method around `getCallConverters()` used to leave them green.

isthmus/src/main/java/io/substrait/isthmus/ConverterProvider.java
alexandrefimov: Rewritten, and measured rather than reasoned this time. A recording converter prepended through the transform sees `[+, -]` for `ROW_NUMBER() OVER (PARTITION BY a + 1 ORDER BY a - 2)` while the call itself converts to a `WindowFunctionInvocation`, so the operands and both key lists do go through the list and the `RexOver` node does not. `areConsultedForTheExpressionsInsideAWindowedCall` pins it.

One narrowing to your wording: `visitFieldAccess` walks through `ITEM`, `INPUT_REF` and `FIELD_ACCESS` reference expressions with `accept(this)`, so a call of those kinds does reach the list; what never reaches it is a reference expression outside those. The paragraph says that, and the link to the old setter is gone with it.

isthmus/src/main/java/io/substrait/isthmus/ConverterProvider.java
```diff
@@ -685,6 +704,30 @@ public Builder aggregateConversion(AggregateConversion aggregateConversion) {
       return this;
     }
 
+    /**
+     * Sets a transform over the assembled {@link CallConverter} list, letting a caller insert,
+     * replace, reorder or remove converters. The list handed to the operator is mutable and holds
+     * the built-in converters in the order {@link ConverterProvider#getCallConverters()} assembles
+     * them; the list the operator returns is the one used.
```
nielspardon: Say the returned list must be mutable too. `DynamicConverterProvider.getCallConverters()` appends to what `super` returns, so `callConverters(List::copyOf)` — or any transform ending in `toUnmodifiableList()` — throws `UnsupportedOperationException` at conversion time, for that provider only and with nothing pinning it.

```suggestion
     * them; the list the operator returns is the one used, and must be mutable too, since a
     * subclass may append to it.
```

nielspardon (commented):
Transform shape looks right, and the window test earns its place. One contract left implicit.

isthmus/src/main/java/io/substrait/isthmus/ConverterProvider.java
alexandrefimov: Applied as suggested, and pinned: `leaveTheListMutableForASubclassToAppendTo` adds to what `getCallConverters()` returns, so a transform that hands back an unmodifiable list fails here rather than at conversion time in one subclass.

isthmus/src/main/java/io/substrait/isthmus/ConverterProvider.java
```diff
@@ -685,6 +704,31 @@ public Builder aggregateConversion(AggregateConversion aggregateConversion) {
...
+     * replace, reorder or remove converters. The list handed to the operator is mutable and holds
+     * the built-in converters in the order {@link ConverterProvider#getCallConverters()} assembles
+     * them; the list the operator returns is the one used, and must be mutable too, since a
+     * subclass may append to it.
+     *
+     * <p>Where a caller's converter sits is its own decision. Ahead of the built-in ones it claims
+     * a call one of them would otherwise take -- which is what a dialect that gives a call
+     * different semantics needs, and also what puts it ahead of {@code CAST}, {@code REINTERPRET}
+     * and {@code ROW}, this repository's encoding of nullable and user-defined literals rather than
+     * dialect behaviour. Behind them it catches only what none of them claims.
```
nielspardon, line 721: This paragraph invites building on a built-in's result, and doing it the obvious way is a `StackOverflowError` rather than a diagnosable error. Nothing documents that today — `CallConverter`'s own `@param` calls `topLevelConverter` "a function for converting nested `RexNode` operands", which is true but doesn't warn against passing the claimed call itself.

```suggestion
     * dialect behaviour. Behind them it catches only what none of them claims.
     *
     * <p>A converter must not pass the call it is claiming to the top-level converter, which
     * re-enters this list from the start and recurses without end. To build on a built-in
     * converter's result, call that converter directly rather than routing back through the list.
```

nielspardon (changes requested):
Add one sentence saying a converter must not hand the call it is claiming back to `topLevelConverter`. Prepending to adjust a built-in's result is exactly the idiom the paragraph above invites, and it recurses without end — `topLevelConverter` is `rexNode -> rexNode.accept(this)`, which re-enters the list from the start. Otherwise this is ready: the transform shape and the mutability test both read right.

isthmus/src/main/java/io/substrait/isthmus/ConverterProvider.java
alexandrefimov: Applied as suggested. Measured it too: a converter prepended through the transform that claims `CAST` and passes it to `top.apply(call)` ends in `StackOverflowError` — `visitCall` hands out `rexNode -> rexNode.accept(this)`, so the call re-enters at the top of the list.

I did not keep that as a test. Pinning a `StackOverflowError` felt like the wrong thing to have in the suite, but say the word if you would rather have it pinned.

nielspardon (approved):
LGTM

Issue #1093, opened 2026-08-13
Aggregate measure sorts and options are dropped on POJO to proto for extended expressions


nielspardon:
## Summary

Two independent gaps in core's sort-field and aggregate-measure proto conversion: one POJO → proto writer drops fields its paired reader consumes, and the `sort_kind` oneof is read without checking which arm is set. Found while reviewing #1090.

## 1. `AggregateFunctionProtoConverter.toProto` never writes `sorts` or `options`

`AggregateFunctionProtoConverter.toProto(Aggregate.Measure)` (`core/src/main/java/io/substrait/relation/AggregateFunctionProtoConverter.java:42-62`) builds the proto `AggregateFunction` from phase, invocation, output type, arguments and function reference — and nothing else. Its paired reader `ProtoAggregateFunctionConverter` reads `measure.getOptionsList()` (`:72`) and `measure.getSortsList()` (`:76`) straight back, so both fields are write-never / read-always.

The other POJO → proto path for the same proto message does write them: `RelProtoConverter.java:281` calls `.addAllSorts(toProtoS(measure.getFunction().sort()))` and `:284` `.addAllOptions(…)`. So the two writers for one message disagree, and which one you get depends on how you got there.

`ExtendedExpressionProtoConverter.java:56` is the only caller of the broken one, which bounds the blast radius to extended expressions — but that is exactly where an aggregate measure carrying sort fields is now reachable, since #1090 makes isthmus produce them. An extended expression holding `sum(x ORDER BY y)` serializes with `sorts` empty and deserializes as an unordered aggregate: wrong results for any order-sensitive aggregate, no exception, nothing in the proto to show what was lost. The same applies to `options`.

**Fix:** add `.addAllSorts(…)` and `.addAllOptions(…)` to `AggregateFunctionProtoConverter.toProto`, mirroring `RelProtoConverter.java:281-284`. Better still, have the two writers share one measure-to-proto helper so the next field added cannot land in only one of them.

## 2. The `comparison_function_reference` arm of `SortField.sort_kind` produces a nonsense error

`SortField` is a oneof over `direction` and `comparison_function_reference`, but every reader takes `getDirection()` unconditionally and never consults `getSortKindCase()`:

| Reader | Sort field construction |
| --- | --- |
| `ProtoRelConverter.newSort` | `ProtoRelConverter.java:1039-1047` — hand-inlined, `SortDirection.fromProto(field.getDirection())` at `:1044` |
[...]

PR #1168, opened 2026-08-26
fix(isthmus): convert extended expressions with the provider's converter

### LINKED ISSUE #1165: isthmus: SqlExpressionToSubstrait skips the default call converters, so CAST and CASE do not convert as extended expressions
`SqlExpressionToSubstrait` builds its own `RexExpressionConverter` from the scalar function converter alone:

```java
public SqlExpressionToSubstrait(ConverterProvider converterProvider) {
  super(converterProvider);
  this.rexConverter = new RexExpressionConverter(converterProvider.getScalarFunctionConverter());
}
```

That varargs constructor takes the whole call-converter list, so the list here *is* the scalar function converter — `CallConverters.defaults(...)` never reaches this path. Anything the default converters handle is unconvertible as an extended expression while converting fine inside a plan:

```
CREATE TABLE orders (o_orderkey BIGINT, o_comment VARCHAR(79))

                                       SqlExpressionToSubstrait   SqlToSubstrait
o_orderkey + 1                         converts                   converts
CAST(o_orderkey AS VARCHAR)            IllegalArgumentException:  converts
                                       Unable to convert call CAST(i64?).
CASE WHEN o_orderkey > 1               IllegalArgumentException:  converts
     THEN 1 ELSE 2 END                 Unable to convert call CASE(boolean?, i32, i32).
```

The same gap covers field selection, `REINTERPRET`, the execution-context variables, the array and map constructors and the SEARCH converter — everything `ConverterProvider.getCallConverters()` assembles and this constructor skips. `isthmus-cli` reaches it directly: `IsthmusEntryPoint:112` passes a provider into this constructor for the expression mode.

Routing the constructor through `converterProvider.getCallConverters()` is the obvious fix, and it also makes the path honour whatever the provider is configured with rather than only its scalar function converter.

The line dates to #649, which introduced `ConverterProvider` and replaced the previous `(FeatureBoard, ExtensionCollection)` constructor.

alexandrefimov:
`SqlExpressionToSubstrait` built its own `RexExpressionConverter` from the scalar function converter alone. That varargs constructor takes the whole call-converter list, so the list on this path was just that one converter and `CallConverters.defaults(...)` never reached it — `CAST`, `CASE`, the execution-context variables and the array and map constructors were unconvertible as extended expressions while converting fine inside a plan.

Taking the converter from `ConverterProvider.getRexExpressionConverter` rather than assembling a second one here also means a provider that overrides that method is honoured on this path, as it already is when converting a plan, and that the window function converter and type converter are the provider's rather than `null` and `TypeConverter.DEFAULT`. Only the call-converter half and the provider hook are pinned by tests; a marker `TypeConverter` cannot isolate the third, because `SqlConverterBase` already uses the provider's type converter for the schema, so such a test passes against the old constructor too.

`getRexExpressionConverter` is passed `null` for the relation visitor: an expression that stands alone has no relation to visit. Nothing in this path produces a subquery to convert -- validating an extended expression that holds one throws inside Calcite first, a scalar subquery and `IN`/`EXISTS` alike -- so the converter names what is missing instead of dereferencing it. `visitFieldAccess` did the same dereference for an outer reference, and `visitOver` for the window function converter that three of the four constructors leave null; both say so now.

Naming what is missing has to stay out of the way of the message that names the call. `callConversionFailureMessage` converted each operand again to render its type, so an unconvertible call with a subquery operand reported the operand's failure rather than the call's; the types come off the nodes now.

The expressions the scalar function converter does not claim are pinned by what they convert into -- `CAST`, `IF_THEN`, `EXECUTION_CONTEXT_VARIABLE`, two `LITERAL`s and, since the same wiring admits it, the `WINDOW_FUNCTION` of `ROW_NUMBER() OVER (...)`. The round trip they used to share compares a proto against the one converting it back produces, which holds whatever the conversion made of them, and they no longer feed the duplicate-column test, where they threw before the expression was parsed.

Closes #1165.

### DIFF
--- a/isthmus/src/main/java/io/substrait/isthmus/ConverterProvider.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/ConverterProvider.java
@@ -33,6 +33,7 @@ import org.apache.calcite.sql.validate.SqlConformanceEnum;
 import org.apache.calcite.sql2rel.SqlToRelConverter;
 import org.apache.calcite.tools.Frameworks;
 import org.apache.calcite.tools.RelBuilder;
+import org.jspecify.annotations.Nullable;
 
 /**
  * ConverterProvider provides a single-point of configuration for a number of conversions: {@code
@@ -361,7 +362,7 @@ public class ConverterProvider {
    *     rather than converting one
    * @return a new RexExpressionConverter instance
    */
-  public RexExpressionConverter getRexExpressionConverter(SubstraitRelVisitor srv) {
+  public RexExpressionConverter getRexExpressionConverter(@Nullable SubstraitRelVisitor srv) {
     return new RexExpressionConverter(
         srv, getCallConverters(), getWindowFunctionConverter(), getTypeConverter());
   }
--- a/isthmus/src/main/java/io/substrait/isthmus/ConverterProvider.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/ConverterProvider.java
@@ -356,8 +356,9 @@ public class ConverterProvider {
    * A {@link RexExpressionConverter} converts Calcite {@link org.apache.calcite.rex.RexNode}s to
    * Substrait equivalents.
    *
-   * @param srv the SubstraitRelVisitor to use for nested relation conversions, or {@code null} when
-   *     the expressions being converted stand alone and cannot contain a subquery
+   * @param srv the SubstraitRelVisitor to use for nested relation conversions, or {@code null}
+   *     where the expressions being converted stand alone; the converter then rejects a subquery
+   *     rather than converting one
    * @return a new RexExpressionConverter instance
    */
   public RexExpressionConverter getRexExpressionConverter(SubstraitRelVisitor srv) {
--- a/isthmus/src/main/java/io/substrait/isthmus/ConverterProvider.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/ConverterProvider.java
@@ -356,7 +356,8 @@ public class ConverterProvider {
    * A {@link RexExpressionConverter} converts Calcite {@link org.apache.calcite.rex.RexNode}s to
    * Substrait equivalents.
    *
-   * @param srv the SubstraitRelVisitor to use for nested relation conversions
+   * @param srv the SubstraitRelVisitor to use for nested relation conversions, or {@code null} when
+   *     the expressions being converted stand alone and cannot contain a subquery
    * @return a new RexExpressionConverter instance
    */
   public RexExpressionConverter getRexExpressionConverter(SubstraitRelVisitor srv) {
--- a/isthmus/src/main/java/io/substrait/isthmus/expression/RexExpressionConverter.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/expression/RexExpressionConverter.java
@@ -265,10 +265,19 @@ public class RexExpressionConverter implements RexVisitor<Expression> {
    *
    * @param subQuery the subquery node
    * @return the converted Substrait expression
-   * @throws UnsupportedOperationException for unsupported subquery operators
+   * @throws UnsupportedOperationException for an unsupported subquery operator, or when this
+   *     converter was built without a relation visitor
    */
   @Override
   public Expression visitSubQuery(RexSubQuery subQuery) {
+    if (relVisitor == null) {
+      // A converter built without one converts expressions that stand alone, an extended
+      // expression among them. Say so rather than dereferencing it.
+      throw new UnsupportedOperationException(
+          String.format(
+              "This converter has no relation visitor, so it cannot convert the subquery %s",
+              subQuery));
+    }
     Rel rel = relVisitor.apply(subQuery.rel);
 
     if (subQuery.getOperator() == SqlStdOperatorTable.EXISTS) {
--- a/isthmus/src/main/java/io/substrait/isthmus/expression/RexExpressionConverter.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/expression/RexExpressionConverter.java
@@ -217,7 +217,8 @@ public class RexExpressionConverter implements RexVisitor<Expression> {
    *
    * @param fieldAccess the field access
    * @return the converted Substrait expression
-   * @throws UnsupportedOperationException for unsupported reference kinds
+   * @throws UnsupportedOperationException for an unsupported reference kind, or when this converter
+   *     was built without a relation visitor and the reference is an outer one
    */
   @Override
   public Expression visitFieldAccess(RexFieldAccess fieldAccess) {
@@ -225,6 +226,15 @@ public class RexExpressionConverter implements RexVisitor<Expression> {
     switch (kind) {
       case CORREL_VARIABLE:
         {
+          if (relVisitor == null) {
+            // As for a subquery: a converter built without one converts expressions that stand
+            // alone, and an outer reference is bound by a relation there is none of here.
+            throw new UnsupportedOperationException(
+                String.format(
+                    "This converter has no relation visitor, so it cannot convert the outer"
+                        + " reference %s",
+                    fieldAccess));
+          }
           CorrelationId correlationId = ((RexCorrelVariable) fieldAccess.getReferenceExpr()).id;
           Integer anchor = relVisitor.getOuterReferenceAnchor(correlationId);
           if (anchor == null) {
--- a/isthmus/src/main/java/io/substrait/isthmus/expression/RexExpressionConverter.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/expression/RexExpressionConverter.java
@@ -139,9 +139,24 @@ public class RexExpressionConverter implements RexVisitor<Expression> {
     return String.format(
         "Unable to convert call %s(%s).",
         call.getOperator().getName(),
-        call.getOperands().stream()
-            .map(t -> t.accept(this).getType().accept(new StringTypeVisitor()))
-            .collect(Collectors.joining(", ")));
+        call.getOperands().stream().map(this::operandTypeName).collect(Collectors.joining(", ")));
+  }
+
+  /**
+   * Names an operand's type for a failure message, reading it off the node rather than converting
+   * the operand again. The call being reported is one this converter could not convert, and an
+   * operand it cannot convert either -- a subquery where there is no relation visitor, say -- would
+   * report its own failure in place of the one being described.
+   *
+   * @param operand the operand to name the type of
+   * @return the Substrait name of its type, or the Calcite one where the type does not convert
+   */
+  private String operandTypeName(RexNode operand) {
+    try {
+      return typeConverter.toSubstrait(operand.getType()).accept(new StringTypeVisitor());
+    } catch (RuntimeException e) {
+      return operand.getType().getFullTypeString();
+    }
   }
 
   /**
@@ -161,9 +176,21 @@ public class RexExpressionConverter implements RexVisitor<Expression> {
    * @param over the windowed call
    * @return the converted Substrait expression
    * @throws IllegalArgumentException if {@code IGNORE NULLS} is used or conversion fails
+   * @throws UnsupportedOperationException if this converter was built without a window function
+   *     converter
    */
   @Override
   public Expression visitOver(RexOver over) {
+    if (windowFunctionConverter == null) {
+      // As for a subquery and an outer reference: three of this class's four constructors leave
+      // the collaborator null, and a converter built by one of them cannot convert a windowed call.
+      throw new UnsupportedOperationException(
+          String.format(
+              "This converter has no window function converter, so it cannot convert the windowed"
+                  + " call %s",
+              over));
+    }
+
     if (over.ignoreNulls()) {
       throw new IllegalArgumentException("IGNORE NULLS cannot be expressed in Substrait");
     }
--- a/isthmus/src/test/java/io/substrait/isthmus/SimpleExtendedExpressionsTest.java
+++ b/isthmus/src/test/java/io/substrait/isthmus/SimpleExtendedExpressionsTest.java
@@ -4,10 +4,19 @@ import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertThrows;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 
+import com.google.common.collect.ImmutableList;
 import io.substrait.isthmus.expression.RexExpressionConverter;
 import java.io.IOException;
 import java.util.stream.Stream;
+import org.apache.calcite.plan.RelOptCluster;
+import org.apache.calcite.plan.hep.HepPlanner;
+import org.apache.calcite.plan.hep.HepProgram;
+import org.apache.calcite.rel.RelNode;
+import org.apache.calcite.rel.logical.LogicalValues;
+import org.apache.calcite.rex.RexBuilder;
+import org.apache.calcite.rex.RexSubQuery;
 import org.apache.calcite.sql.parser.SqlParseException;
+import org.apache.calcite.sql.type.SqlTypeName;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.params.ParameterizedTest;
 import org.junit.jupiter.params.provider.Arguments;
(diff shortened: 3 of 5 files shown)


isthmus/src/test/java/io/substrait/isthmus/SimpleExtendedExpressionsTest.java
```diff
@@ -21,7 +37,14 @@ private static Stream<Arguments> expressionTypeProvider() {
...
         Arguments.of("L_ORDERKEY is not null"), // ScalarFunctionExpressionIsNotNull
-        Arguments.of("L_ORDERKEY is null")); // ScalarFunctionExpressionIsNull
+        Arguments.of("L_ORDERKEY is null"), // ScalarFunctionExpressionIsNull
+        // The rest are handled by CallConverters.defaults rather than by the scalar function
+        // converter, and so reach this path only when it is assembled from the provider.
+        Arguments.of("CAST(L_ORDERKEY AS VARCHAR)"), // CallConverters.CAST
+        Arguments.of("CASE WHEN L_ORDERKEY > 10 THEN 1 ELSE 2 END"), // CallConverters.CASE
+        Arguments.of("CURRENT_DATE"), // CallConverters.EXECUTION_CONTEXT_VARIABLE
+        Arguments.of("ARRAY[1, 2]"), // SqlArrayValueConstructorCallConverter
+        Arguments.of("MAP['a', 1]")); // SqlMapValueConstructorCallConverter
```
nielspardon: Give these their own `@MethodSource`: they also feed `testExtendedExpressionsDuplicateColumnIdentifierRoundTrip`, which throws from schema registration before `sqlExpression` is parsed, so all twelve runs there take the same path and these five assert nothing.

isthmus/src/test/java/io/substrait/isthmus/SimpleExtendedExpressionsTest.java
```diff
@@ -21,7 +37,14 @@ private static Stream<Arguments> expressionTypeProvider() {
...
   }
 
   @ParameterizedTest
```
nielspardon, line 61: Assert on the produced expression, not only on the round trip — the helper compares `ExtendedExpressionProtoConverter(ProtoExtendedExpressionConverter(proto))` against `proto`, which holds for any conversion, right or wrong. A wrong index, type or operator round-trips identically, so these cases can only catch an outright throw. Same blindness as #1093.

isthmus/src/test/java/io/substrait/isthmus/SimpleExtendedExpressionsTest.java
```diff
@@ -44,6 +67,27 @@ void testExtendedExpressionsDuplicateColumnIdentifierRoundTrip(String sqlExpress
...
+          public RexExpressionConverter getRexExpressionConverter(SubstraitRelVisitor srv) {
+            throw new UnsupportedOperationException(MARKER);
+          }
+        };
+
+    UnsupportedOperationException e =
+        assertThrows(
+            UnsupportedOperationException.class, () -> new SqlExpressionToSubstrait(provider));
+    assertEquals(MARKER, e.getMessage());
+  }
```
nielspardon: Drop `throws IOException` (nothing here throws it) and pin the contract the javadoc documents — otherwise a regression passing a non-null visitor still passes. Needs `assertNull` added to the static imports.

```suggestion
  void usesTheProvidersRexExpressionConverter() {
    SubstraitRelVisitor[] seen = new SubstraitRelVisitor[1];
    ConverterProvider provider =
        new ConverterProvider(ConverterProvider.builder()) {
          @Override
          public RexExpressionConverter getRexExpressionConverter(SubstraitRelVisitor srv) {
            seen[0] = srv;
            throw new UnsupportedOperationException(MARKER);
          }
        };

    UnsupportedOperationException e =
        assertThrows(
            UnsupportedOperationException.class, () -> new SqlExpressionToSubstrait(provider));
    assertEquals(MARKER, e.getMessage());
    assertNull(seen[0]);
  }
```

nielspardon (commented):
Does an execution-context variable belong in an `ExtendedExpression` at all? `algebra.proto` ties it to `ExecutionBehavior`, which only `Plan` carries, so `CURRENT_DATE` now converts into a message with nowhere to record whether it is evaluated once or per record — and the same widening makes `ROW_NUMBER() OVER (...)` convert, untested.

The description needs a rewrite before merge: its subquery paragraph describes the javadoc that commit 2 replaced, and `i + (SELECT 1)` is not rejected by the validator — it converts, to a field reference.

The half-honoured provider getters behind `getCallConverters()` are #1048; this also unblocks dropping #1166's javadoc caveat that nothing configured on the provider reaches this path.

isthmus/src/test/java/io/substrait/isthmus/SimpleExtendedExpressionsTest.java
```diff
@@ -58,4 +102,54 @@ void testExtendedExpressionsListExpressionRoundTrip() throws SqlParseException,
...
+                    .visitFieldAccess(fieldAccess));
+    assertTrue(rejected.getMessage().contains("no relation visitor"), rejected.getMessage());
+  }
+
+  /**
+   * The converter this path builds has no relation to visit, so a subquery cannot be converted on
+   * it. What the path accepts today never produces one -- a constant scalar subquery folds before
+   * the conversion sees it, and IN and EXISTS are rejected by Calcite's own validation -- so this
+   * pins the converter rather than a SQL expression that reaches it.
+   */
```
nielspardon: `(SELECT 1)` neither folds nor is rejected — it becomes `RexInputRef(0)` and converts to a field reference — and IN/EXISTS fail with an internal Calcite NPE/CCE rather than validation.

```suggestion
  /**
   * The converter this path builds has no relation to visit, so a subquery cannot be converted on
   * it. No SQL expression reaches this guard today, so this pins the converter directly.
   */
```

isthmus/src/test/java/io/substrait/isthmus/SimpleExtendedExpressionsTest.java
```diff
@@ -58,4 +102,54 @@ void testExtendedExpressionsListExpressionRoundTrip() throws SqlParseException,
...
+    RelNode oneRow =
+        LogicalValues.create(
+            cluster,
+            SubstraitTypeSystem.TYPE_FACTORY.builder().add("a", SqlTypeName.INTEGER).build(),
+            ImmutableList.of(
+                ImmutableList.of(
+                    rexBuilder.makeExactLiteral(
+                        java.math.BigDecimal.ONE,
+                        SubstraitTypeSystem.TYPE_FACTORY.createSqlType(SqlTypeName.INTEGER)))));
+    RexSubQuery subQuery = RexSubQuery.scalar(oneRow);
```
nielspardon: `RelCreator` already builds this cluster, so the planner, the row type and the literal can all go. Better still, move both new tests to `SubqueryConversionTest` — it extends `PlanTestBase`, which hands you `creator`/`builder`, and it is where the other `visitSubQuery` cases live. Either way the `RelOptCluster`, `HepPlanner`, `HepProgram`, `LogicalValues`, `ImmutableList` and `BigDecimal` imports go with it.

```suggestion
    RexSubQuery subQuery =
        RexSubQuery.scalar(
            new RelCreator().createRelBuilder().values(new String[] {"a"}, 1).build());
```

isthmus/src/main/java/io/substrait/isthmus/ConverterProvider.java
```diff
@@ -356,7 +356,9 @@ public SubstraitRelVisitor getSubstraitRelVisitor() {
    * A {@link RexExpressionConverter} converts Calcite {@link org.apache.calcite.rex.RexNode}s to
    * Substrait equivalents.
    *
-   * @param srv the SubstraitRelVisitor to use for nested relation conversions
+   * @param srv the SubstraitRelVisitor to use for nested relation conversions, or {@code null}
+   *     where the expressions being converted stand alone; the converter then rejects a subquery
+   *     rather than converting one
    * @return a new RexExpressionConverter instance
    */
   public RexExpressionConverter getRexExpressionConverter(SubstraitRelVisitor srv) {
```
nielspardon: Mark `srv` `@Nullable` (jspecify is already an `api` dependency of `:isthmus`, though the import is new here) — a downstream NullAway run otherwise rejects the `null` this javadoc now blesses.

```suggestion
  public RexExpressionConverter getRexExpressionConverter(@Nullable SubstraitRelVisitor srv) {
```

isthmus/src/main/java/io/substrait/isthmus/expression/RexExpressionConverter.java
```diff
@@ -217,14 +217,24 @@ public Expression visitRangeRef(RexRangeRef rangeRef) {
...
+   * @throws UnsupportedOperationException for an unsupported reference kind, or when this converter
+   *     was built without a relation visitor and the reference is an outer one
    */
   @Override
   public Expression visitFieldAccess(RexFieldAccess fieldAccess) {
     SqlKind kind = fieldAccess.getReferenceExpr().getKind();
     switch (kind) {
       case CORREL_VARIABLE:
         {
+          if (relVisitor == null) {
```
nielspardon, line 256: Worth the same treatment for `windowFunctionConverter` at :171 — it is the remaining nullable collaborator this class dereferences unguarded, and three of the four constructors leave it null, so `visitOver` still throws a bare NPE where these two commits now name the missing one.

isthmus/src/main/java/io/substrait/isthmus/expression/RexExpressionConverter.java
```diff
@@ -265,10 +275,19 @@ public Expression visitFieldAccess(RexFieldAccess fieldAccess) {
    *
    * @param subQuery the subquery node
    * @return the converted Substrait expression
-   * @throws UnsupportedOperationException for unsupported subquery operators
+   * @throws UnsupportedOperationException for an unsupported subquery operator, or when this
+   *     converter was built without a relation visitor
    */
   @Override
   public Expression visitSubQuery(RexSubQuery subQuery) {
+    if (relVisitor == null) {
```
nielspardon, line 310: `callConversionFailureMessage` (:138) re-converts each operand to render its type, so with no relation visitor a `RexSubQuery` operand now hits this guard and an unconvertible call reports "no relation visitor" instead of "Unable to convert call". Reading the type off the `RexNode` avoids both the masking and the duplicate walk.

isthmus/src/test/java/io/substrait/isthmus/SimpleExtendedExpressionsTest.java
alexandrefimov: Split off, and the new source pins what each expression converts into rather than that it converts at all: `CAST`, `IF_THEN`, `EXECUTION_CONTEXT_VARIABLE`, two `LITERAL`s, and `WINDOW_FUNCTION` for the `ROW_NUMBER() OVER (...)` you point at below, which the same wiring admits.

isthmus/src/test/java/io/substrait/isthmus/SimpleExtendedExpressionsTest.java
alexandrefimov: Done, in the same split: the five convert through `SqlExpressionToSubstrait` and the test asserts the `RexTypeCase` of the referred expression. The round trip stays for the cases that were there before it.

isthmus/src/test/java/io/substrait/isthmus/SimpleExtendedExpressionsTest.java
alexandrefimov: Applied as suggested.

isthmus/src/test/java/io/substrait/isthmus/SimpleExtendedExpressionsTest.java
alexandrefimov: You are right that the comment was wrong, and I have replaced it -- but the `RexInputRef(0)` half does not reproduce for me through this path. Converting `L_ORDERKEY + (SELECT 1)`, `L_ORDERKEY IN (SELECT ...)` and `EXISTS (SELECT ...)` with `SqlExpressionToSubstrait.convert(sql, tpchSchemaCreateStatements())` all three throw the same way:

```
java.lang.NullPointerException: namespace for SELECT 1, scope org.apache.calcite.sql.validate.ParameterScope@...
```

from validation, before any conversion. Which entry point gives you the field reference -- the plan path, where the subquery is expanded? The comment now says what I measured: nothing reaches the guard because validating an extended expression that holds a subquery throws inside Calcite first. Leaving this open until we agree on which of the two the comment should describe.

isthmus/src/test/java/io/substrait/isthmus/SimpleExtendedExpressionsTest.java
alexandrefimov: Moved, with one correction: there is no `SubqueryConversionTest` in the repo -- the subquery cases live in `SubqueryPlanTest`, which extends `PlanTestBase`, so that is where all three went. The cluster, planner, row type and literal are gone with them; the nodes come from `builder` now.

isthmus/src/main/java/io/substrait/isthmus/ConverterProvider.java
alexandrefimov: Applied as suggested.

isthmus/src/main/java/io/substrait/isthmus/expression/RexExpressionConverter.java
alexandrefimov: Added, in the same shape: `visitOver` names the missing converter instead of dereferencing it. `aWindowedCallIsRejectedWhereThereIsNoWindowFunctionConverter` builds the `RexOver` through `builder` and converts it on a `new RexExpressionConverter()`, which is one of the three constructors that leave the field null.

isthmus/src/main/java/io/substrait/isthmus/expression/RexExpressionConverter.java
alexandrefimov: Fixed: the message reads each operand's type off the node and converts the type, rather than converting the operand again. `anUnconvertibleCallWithASubqueryOperandReportsTheCall` pins it -- with the old rendering it reports "no relation visitor" for a call the message is supposed to name. A type the conversion cannot express falls back to the Calcite type string, so the message cannot fail in place of the failure it describes.

alexandrefimov:
On the execution-context variable: the gap you point at is there. `ExtendedExpression` declares `version`, `extension_urns`, `extensions`, `referred_expr`, `base_schema`, `advanced_extensions` and `expected_type_urls` -- `execution_behavior` is on `Plan` alone -- so a `CURRENT_DATE` in an extended expression carries no statement of whether it is evaluated once or per record.

I would rather not refuse it here, though. The message the spec defines admits an `Expression`, and this conversion has no rule the plan path does not; refusing would make the extended-expression path narrower than the message it produces, on a question the spec has not answered. What the wiring can do is stop being silent about it, which is what the split test now does: `CURRENT_DATE` is pinned as `EXECUTION_CONTEXT_VARIABLE` and `ROW_NUMBER() OVER (...)` as `WINDOW_FUNCTION`, so neither arrives untested and both are visible to whoever settles the semantics.

Happy to raise the behaviour question upstream, or to refuse both here if you would rather this path stay to what an extended expression can fully describe -- your call, and I have no strong second preference.

isthmus/src/test/java/io/substrait/isthmus/SimpleExtendedExpressionsTest.java
nielspardon: Still not right for a bare `(SELECT 1)` — it converts, to `struct_field {}`, i.e. base-schema field 0. `i + (SELECT 1)`, IN and EXISTS do throw, but as a raw `NullPointerException: namespace for SELECT …` rather than validation. The silent conversion looks like a bug of its own; happy to file it so this javadoc can just say the guard is pinned directly.

isthmus/src/test/java/io/substrait/isthmus/SubqueryPlanTest.java
```diff
@@ -350,4 +358,91 @@ void correlatedScalarSubQueryInSelect() throws Exception {
...
+   */
+  @Test
+  void anUnconvertibleCallWithASubqueryOperandReportsTheCall() {
+    RexBuilder rexBuilder = builder.getRexBuilder();
+    RexNode subQuery = RexSubQuery.scalar(builder.values(new String[] {"a"}, 1).build());
+    RexNode call =
+        rexBuilder.makeCall(
+            SqlStdOperatorTable.SESSION_USER.getReturnTypeInference() == null
+                ? subQuery.getType()
+                : subQuery.getType(),
```
nielspardon: Both arms of this conditional are `subQuery.getType()`, so it reads as an unfinished thought — drop it.

```suggestion
            subQuery.getType(),
```

nielspardon (commented):
Thanks — the rest all look good.

isthmus/src/test/java/io/substrait/isthmus/SimpleExtendedExpressionsTest.java
alexandrefimov: You are right, and the difference was in the expression I measured: bare `(SELECT 1)` converts to `selection { direct_reference { struct_field { } } }` -- the base schema's first field -- while `L_ORDERKEY + (SELECT 1)`, `IN` and `EXISTS` throw the raw `NullPointerException`. I was converting the embedded form both times.

The javadoc now just says the guard is pinned directly, with no account of what SQL reaches it. Please do file the silent conversion; you have the sharper description of it, and I will link the test to it once it exists if that reads better than saying nothing.

isthmus/src/test/java/io/substrait/isthmus/SubqueryPlanTest.java
alexandrefimov: Dropped -- it was a leftover from working out which operator to build the call with, and both arms were the subquery's own type.

nielspardon:
Withdrawing the execution-behavior question from my review body: `ExtendedExpression` describes a collection of expressions rather than a plan, so per-plan versus per-record evaluation has no referent there, and the absent `ExecutionBehavior` field is by design rather than a gap. `CURRENT_DATE` converting with no mode declared is correct — Arrow Java takes these as `ScanOptions.substraitFilter` / `substraitProjection` bytes, where the engine doing the pushdown owns the evaluation model. Worth noting that Arrow's cookbook builds them with `new SqlExpressionToSubstrait().convert(...)`, so this fix lands squarely on that path.

isthmus/src/test/java/io/substrait/isthmus/SimpleExtendedExpressionsTest.java
nielspardon: Filed the silent conversion as #1201, which also covers why no SQL reaches the guard.

nielspardon (approved):
LGTM, thanks @alexandrefimov

Issue #1124, opened 2026-08-18
isthmus: derive the Calcite type system's limits from a Substrait dialect instead of hardcoding them


nielspardon:
## Thesis

Isthmus should be initializable from a Substrait dialect, and that dialect's declared type limits should drive the Calcite type system it builds — rather than the limits being hardcoded in `SubstraitTypeSystem` as they are today.

The dialect model already carries exactly the information needed: `SupportedType` exposes `maxPrecision()`, `maxScale()` and `maxLength()` as `Optional<Integer>` (added in #808). Nothing consumes them — `grep -rl 'io.substrait.dialect' isthmus/src/main/java` returns no files, and `SubstraitTypeSystem` is a hardcoded `switch`: 6 for the `TIME`/`TIMESTAMP`/`TIMESTAMP_WITH_LOCAL_TIME_ZONE`/`INTERVAL_*` family, 38 for `DECIMAL`, with `getMaxScale` overriding only `DECIMAL`.

The result is that "what precision can this target actually represent" is currently a property of Isthmus rather than of the engine being targeted, even though we publish per-engine dialects that are entitled to say so.

## Why this is worth doing

Every hardcoded limit is a place where Isthmus either rejects a plan the target could execute, or accepts one it cannot. Three existing issues are all instances of the same missing seam:

- **#995** — Substrait→Calcite rejects `precision_time`/`precision_timestamp`/`precision_timestamp_tz` above 6, though the spec allows 0–12. That issue asks whether the hardcoded ceiling can be raised; dialect-driven limits reframe it as "whose ceiling is it", which is the more answerable question.
- **#1113** — the `precision_timestamp_tz` literal guard reads `SqlTypeName.TIMESTAMP_TZ`, which `SubstraitTypeSystem` does not override, so it caps at Calcite's default 3 while the type conversion caps at 6. A single declared source of limits removes the class of bug where two lookups disagree.
- **Spark** hardcodes the same number independently: `Util.assertMicroseconds` throws for any `interval_day` precision but 6, and `ToSubstraitType` only ever emits 6. Spark reaches Substrait without going through Isthmus, so it would need to consume the dialect on its own, but the hardcode is the same shape.

## Known blocker: the type system is a reflectively-constructed singleton

[...]

PR #1169, opened 2026-08-26
fix(isthmus)!: keep the declared length of a wide character or binary type

### LINKED ISSUE #1167: isthmus: a varchar or fixedchar wider than 65536 is silently narrowed converting to Calcite
A `varchar` or `fixedchar` longer than 65536 comes back from `SubstraitRelNodeConverter` as a 65536-wide one. No error is raised — the column simply has a different type than the plan declared, and a consumer that built a Calcite tree from the plan has no way to notice.

```
varchar<10>         -> VARCHAR(10)     -> varchar<10>         round-trips
varchar<65536>      -> VARCHAR(65536)  -> varchar<65536>      round-trips
varchar<65537>      -> VARCHAR(65536)  -> varchar<65536>      silently narrowed
varchar<2147483647> -> VARCHAR(65536)  -> varchar<65536>      silently narrowed
fixedchar<100000>   -> CHAR(65536)     -> fixedchar<65536>    silently narrowed
```

`TypeConverter` builds the type through the type factory, which caps the precision at `RelDataTypeSystem.getMaxPrecision(VARCHAR)`. `SubstraitTypeSystem` overrides `getMaxPrecision`, but only for the temporal types and `DECIMAL`; `VARCHAR` and `CHAR` fall through to `RelDataTypeSystemImpl`, whose default is 65536. Substrait's own limit is wider — `length` is an `int` in both the POJO and the proto, so the spec admits up to `Integer.MAX_VALUE`.

Adding `VARCHAR` and `CHAR` to that switch works: with them returning `Integer.MAX_VALUE` every width above round-trips and `./gradlew build` stays green. Whether that is the right ceiling is the open part — it is the hardcoded-limits question of #1124, and a Calcite type carrying `VARCHAR(2147483647)` deserves a second look before it is settled. Failing the conversion instead of narrowing would also be defensible, though a plan carrying a width Substrait allows does not look like an error.

Raised by an external consumer exporting Impala's Calcite plans through Isthmus: Impala's `string` is a maximum-width varchar and is written to the plan as `varchar<2147483647>`, so every `string` column comes back narrowed. The repro above needs nothing beyond Isthmus.

alexandrefimov:
`SubstraitTypeSystem` overrides `getMaxPrecision` for the temporal types and `DECIMAL`; `CHAR`, `VARCHAR` and `BINARY` fell through to `RelDataTypeSystemImpl`, whose default caps them at 65536. The type factory then narrowed anything wider without raising an error, so a plan declaring `varchar<2147483647>` came back as `VARCHAR(65536)` and a consumer building a Calcite tree from it had no way to notice. The boundary is exact: 65536 round-trips, 65537 does not.

The cap reaches the conversion from SQL too, which the issue does not mention: Calcite narrows a declared width there the same silent way, so `SELECT CAST(a AS VARCHAR(100000))` came out of the conversion as a `varchar<65536>`. It now comes out as `varchar<100000>`. What SQL is accepted does not change -- Calcite never rejected the wider declaration, it just did not keep it.

Those three are the types whose length crosses the boundary — holding `fixedchar`, `varchar` and `fixedbinary` — and Substrait declares each length as a 32-bit integer, so they now report `Integer.MAX_VALUE`. `VARBINARY` is raised with them for a different reason: Substrait's `binary` carries no length of its own, but Calcite unifies a union of fixed-width binaries of different widths into a `VARBINARY` of the widest, so leaving it at the default reimposed the cap one step later — `leastRestrictive(BINARY(100000), BINARY(5))` returned `VARBINARY(65536)`, narrower than one of its own inputs. That shape was unreachable before this change, since no `BINARY` wider than 65536 could exist. Raising it opens no new failure at the extreme: a `VARBINARY` is not padded, so `cast(x'01' AS VARBINARY(2147483647))` projects fine, where the fixed-width `BINARY(2147483647)` runs out of memory exactly as `CHAR` does. The issue was filed for the character types; `fixedbinary` turned out to have the same defect at the same boundary.

One consequence, and where the ceiling belongs. `CHAR` and `BINARY` are fixed-width, so Calcite pads a literal cast to them out to the declared length, and the padding is now allowed to reach the declared width. At the extreme, `RelBuilder.project(cast('a' AS CHAR(2147483647)))` throws `OutOfMemoryError: Required length exceeds implementation limit`, where the old cap silently made it `CHAR(65536)` and it converted; `CHAR(100000)` is fine, and the varying types are unaffected at any width. No literal need appear in the plan for this either: `RexBuilder.zeroValue` allocates `new byte[precision]` for a `BINARY`, so `makeZeroLiteral(BINARY(2147483647))` throws the same way and a planner rule reaches it from a column type alone, where the old cap made it a 65536-byte array. A `VARBINARY`'s zero value is `ByteString.EMPTY`, so raising that type costs nothing here. `Integer.MAX_VALUE` stays here because it is the spec's own bound on a length -- a narrower one is a dialect's answer, which `SupportedType.maxLength()` already carries for `FIXED_CHAR`, `VARCHAR` and `FIXED_BINARY`, and reading it is #1124. A cap on this type system would not bound all of it in any case: `UserTypeMapper.toSubstrait` returns a Substrait type without going through the type factory at all. The padding isthmus does itself, on the `LogicalValues` path where Calcite does none, is bounded in #1171.

One more thing the raise changes, and it reaches plans carrying no wide type at all: `getMaxPrecision(VARCHAR)` is Calcite's overflow threshold for concatenation as well, in `ReturnTypes.DYADIC_STRING_SUM_PRECISION`. A sum of widths past it falls back to an unparameterised type, so `a || b` over two `varchar<40000>` columns converts to `varchar<80000>` where it converted to `string`; `CONCAT` does the same. Neither operand is anywhere near the old cap.

A `RelDataTypeFactory` narrows a width past its own maximum without saying so, and this conversion takes whatever factory it is handed, so raising the maximum fixes the narrowing only for the factory isthmus builds. A declared length the factory cannot hold is now reported instead of returned quietly, and a negative one is refused before the factory is asked at all. Asking cannot tell the two apart: with assertions off — the normal production JVM — the factory stores the negative and reports it back, so `varChar(-5)` certified itself as `VARCHAR(-5)`, while under `-ea` Calcite raised a bare `AssertionError` in place of the message. A length of -1 is Calcite's unspecified precision besides, so `varChar(-1)` passed and became a Substrait `string` where `fixedChar(-1)` was reported against a bound its length is below. A zero length is left to the factory, for the reason #1171 leaves it alone: ordinary SQL produces a `CHAR(0)` today, and substrait-io/substrait#1199 has the floor open. A decimal went the same silent way and now reports too: `decimal<38,10>` through a factory on Calcite's default type system came back `DECIMAL(19,10)` and read back as a precision the plan never declared. Its two ends go with it, both found checking the guard's boundaries rather than in review: `decimal<-1,0>` converted to a bare `DECIMAL` and read back as the maximum through any factory, `-1` being Calcite's unspecified precision here as well, and a scale above the precision was built as asked -- `decimal<19,25>` came out as `DECIMAL(19,25)`. That one is bounded by the spec rather than by the factory: scale sits in `0 <= S <= P`, and no maximum catches it, since both type systems here set `maxScale` equal to `maxPrecision`, so a scale within the precision is within the maximum too. Calcite reports a zero precision and a negative scale itself, so those are left to it. The precision's own ceiling is the spec's too, not the factory's: a type system allowing more than 38 -- a `getMaxPrecision(DECIMAL)` of 76 -- let `decimal<45,2>` convert to `DECIMAL(45,2)`, which `toSubstrait` then refused with "unsupported decimal precision 45", so the type converted in with no way back. Both ends of that clause read one constant now.

The conversion test asserts the resulting precision directly rather than through `testType`, whose expected type is built with the same type factory and would be narrowed alongside the value under test.

Closes #1167.

BREAKING CHANGE: `TypeConverter.toCalcite` throws `IllegalArgumentException` where it previously returned a narrowed type — for a declared length the given type factory cannot hold, and for a negative length, which a factory running without assertions accepted as a type of that width. It also throws for a decimal precision the given factory cannot hold and for a negative one, where it returned a narrowed type, for a scale above the precision, which it built as asked, and for a precision above the 38 the spec allows, which only a type factory more permissive than the spec could reach. Converted output types change with the raised bound: `SELECT CAST(a AS VARCHAR(100000))` converts to `varchar<100000>` where it converted to `varchar<65536>`, `a || b` over two `varchar<40000>` columns converts to `varchar<80000>` where it converted to `string`, and the least restrictive type of `BINARY(100000)` and `BINARY(5)` is `VARBINARY(100000)` where it was `VARBINARY(65536)`.

### DIFF
--- a/isthmus/src/main/java/io/substrait/isthmus/SubstraitTypeSystem.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/SubstraitTypeSystem.java
@@ -81,12 +81,24 @@ public class SubstraitTypeSystem extends RelDataTypeSystemImpl {
   /**
    * Returns the maximum precision for the given SQL type.
    *
+   * <p>For the three types that carry a length across the Substrait boundary — {@link
+   * SqlTypeName#CHAR}, {@link SqlTypeName#VARCHAR} and {@link SqlTypeName#BINARY}, holding {@code
+   * fixedchar}, {@code varchar} and {@code fixed_binary} — this is Substrait's own limit: those
+   * lengths are 32-bit integers. Calcite's default of 65536 is narrower, and the type factory caps
+   * a converted type at it rather than reporting that it cannot represent the declared width.
+   * {@link SqlTypeName#VARBINARY} is left alone because Substrait's {@code binary} carries no
+   * length for it to lose.
+   *
    * @param typeName The {@link SqlTypeName} for which precision is requested.
    * @return Maximum precision for the type.
    */
   @Override
   public int getMaxPrecision(final SqlTypeName typeName) {
     switch (typeName) {
+      case CHAR:
+      case VARCHAR:
+      case BINARY:
+        return Integer.MAX_VALUE;
       case INTERVAL_DAY:
       case INTERVAL_YEAR:
       case INTERVAL_YEAR_MONTH:
--- a/isthmus/src/main/java/io/substrait/isthmus/SubstraitTypeSystem.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/SubstraitTypeSystem.java
@@ -86,8 +86,14 @@ public class SubstraitTypeSystem extends RelDataTypeSystemImpl {
    * fixedchar}, {@code varchar} and {@code fixed_binary} — this is Substrait's own limit: those
    * lengths are 32-bit integers. Calcite's default of 65536 is narrower, and the type factory caps
    * a converted type at it rather than reporting that it cannot represent the declared width.
-   * {@link SqlTypeName#VARBINARY} is left alone because Substrait's {@code binary} carries no
-   * length for it to lose.
+   *
+   * <p>{@link SqlTypeName#VARBINARY} is raised with them even though Substrait's {@code binary}
+   * carries no length of its own, because the cap bites inside Calcite's own type unification:
+   * {@link #shouldConvertRaggedUnionTypesToVarying()} is true here, so a union of fixed-width
+   * binaries of different widths is unified as a {@code VARBINARY} of the widest. Left at 65536,
+   * the least restrictive type of {@code BINARY(100000)} and {@code BINARY(5)} is {@code
+   * VARBINARY(65536)} -- narrower than one of its own inputs, and the cap this method removes
+   * reimposed.
    *
    * @param typeName The {@link SqlTypeName} for which precision is requested.
    * @return Maximum precision for the type.
@@ -98,6 +104,7 @@ public class SubstraitTypeSystem extends RelDataTypeSystemImpl {
       case CHAR:
       case VARCHAR:
       case BINARY:
+      case VARBINARY:
         return Integer.MAX_VALUE;
       case INTERVAL_DAY:
       case INTERVAL_YEAR:
--- a/isthmus/src/main/java/io/substrait/isthmus/SubstraitTypeSystem.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/SubstraitTypeSystem.java
@@ -79,25 +79,6 @@ public class SubstraitTypeSystem extends RelDataTypeSystemImpl {
   }
 
   /**
-   * Checks that a Substrait decimal scale is one the Calcite type system in effect allows, and
-   * reports the bound it exceeds if it is not.
-   *
-   * @param typeSystem the type system the converted type will live under, which need not be this
-   *     one
-   * @param scale the scale carried by the Substrait type
-   * @throws IllegalArgumentException if the scale exceeds what the type system allows
-   */
-  public static void requireSupportedScale(final RelDataTypeSystem typeSystem, final int scale) {
-    int maxScale = typeSystem.getMaxScale(SqlTypeName.DECIMAL);
-    if (scale > maxScale) {
-      throw new IllegalArgumentException(
-          String.format(
-              "unsupported decimal scale %s, max scale in Calcite type system is set to %s",
-              scale, maxScale));
-    }
-  }
-
-  /**
    * Returns the maximum precision for the given SQL type.
    *
    * <p>For the three types that carry a length across the Substrait boundary — {@link
--- a/isthmus/src/main/java/io/substrait/isthmus/SubstraitTypeSystem.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/SubstraitTypeSystem.java
@@ -83,7 +83,7 @@ public class SubstraitTypeSystem extends RelDataTypeSystemImpl {
    *
    * <p>For the three types that carry a length across the Substrait boundary — {@link
    * SqlTypeName#CHAR}, {@link SqlTypeName#VARCHAR} and {@link SqlTypeName#BINARY}, holding {@code
-   * fixedchar}, {@code varchar} and {@code fixed_binary} — this is Substrait's own limit: those
+   * fixedchar}, {@code varchar} and {@code fixedbinary} — this is Substrait's own limit: those
    * lengths are 32-bit integers. Calcite's default of 65536 is narrower, and the type factory caps
    * a converted type at it rather than reporting that it cannot represent the declared width.
    *
--- a/isthmus/src/main/java/io/substrait/isthmus/SubstraitTypeSystem.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/SubstraitTypeSystem.java
@@ -79,6 +79,25 @@ public class SubstraitTypeSystem extends RelDataTypeSystemImpl {
   }
 
   /**
+   * Checks that a Substrait decimal scale is one the Calcite type system in effect allows, and
+   * reports the bound it exceeds if it is not.
+   *
+   * @param typeSystem the type system the converted type will live under, which need not be this
+   *     one
+   * @param scale the scale carried by the Substrait type
+   * @throws IllegalArgumentException if the scale exceeds what the type system allows
+   */
+  public static void requireSupportedScale(final RelDataTypeSystem typeSystem, final int scale) {
+    int maxScale = typeSystem.getMaxScale(SqlTypeName.DECIMAL);
+    if (scale > maxScale) {
+      throw new IllegalArgumentException(
+          String.format(
+              "unsupported decimal scale %s, max scale in Calcite type system is set to %s",
+              scale, maxScale));
+    }
+  }
+
+  /**
    * Returns the maximum precision for the given SQL type.
    *
    * <p>For the three types that carry a length across the Substrait boundary — {@link
--- a/isthmus/src/main/java/io/substrait/isthmus/TypeConverter.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/TypeConverter.java
@@ -31,6 +31,9 @@ import org.jspecify.annotations.Nullable;
  */
 public class TypeConverter {
 
+  /** The widest precision the spec gives a decimal: {@code DECIMAL<P, S>} puts P at 38 or less. */
+  private static final int MAX_DECIMAL_PRECISION = 38;
+
   private final UserTypeMapper userTypeMapper;
 
   /**
@@ -143,7 +146,7 @@ public class TypeConverter {
         return creator.FP64;
       case DECIMAL:
         {
-          if (type.getPrecision() > 38) {
+          if (type.getPrecision() > MAX_DECIMAL_PRECISION) {
             throw new UnsupportedOperationException(
                 "unsupported decimal precision " + type.getPrecision());
           }
@@ -441,6 +444,15 @@ public class TypeConverter {
             String.format(
                 "A decimal cannot declare a negative precision, and this one is %d",
                 expr.precision()));
+      }
+      // The spec's own ceiling, not just the factory's: handed a type system whose DECIMAL maximum
+      // is above it, the factory builds the type and the outbound conversion above then refuses it,
+      // so the type would convert in and have no way back.
+      if (expr.precision() > MAX_DECIMAL_PRECISION) {
+        throw new IllegalArgumentException(
+            String.format(
+                "A decimal cannot declare a precision of %d, above the %d the spec allows",
+                expr.precision(), MAX_DECIMAL_PRECISION));
       }
       SubstraitTypeSystem.requireSupportedPrecision(
           typeFactory.getTypeSystem(), SqlTypeName.DECIMAL, "decimal", expr.precision());
--- a/isthmus/src/main/java/io/substrait/isthmus/TypeConverter.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/TypeConverter.java
@@ -376,17 +376,41 @@ public class TypeConverter {
 
     @Override
     public RelDataType visit(Type.FixedChar expr) {
-      return t(n(expr), SqlTypeName.CHAR, expr.length());
+      return withLength(n(expr), SqlTypeName.CHAR, expr.length());
     }
 
     @Override
     public RelDataType visit(Type.VarChar expr) {
-      return t(n(expr), SqlTypeName.VARCHAR, expr.length());
+      return withLength(n(expr), SqlTypeName.VARCHAR, expr.length());
     }
 
     @Override
     public RelDataType visit(Type.FixedBinary expr) {
-      return t(n(expr), SqlTypeName.BINARY, expr.length());
+      return withLength(n(expr), SqlTypeName.BINARY, expr.length());
+    }
+
+    /**
+     * Returns the type the given factory builds for a declared length, having checked that it holds
+     * it. A factory caps a width at its type system's maximum without saying so, and this
+     * conversion takes whatever factory it is handed, so a factory whose limits are not Substrait's
+     * would otherwise return a type narrower than the plan declares.
+     *
+     * @param nullable whether the type is nullable
+     * @param typeName the Calcite type name to build
+     * @param length the declared length
+     * @return the built type
+     * @throws IllegalArgumentException if the factory built a type of another length
+     */
+    private RelDataType withLength(boolean nullable, SqlTypeName typeName, int length) {
+      RelDataType type = t(nullable, typeName, length);
+      if (type.getPrecision() != length) {
+        throw new IllegalArgumentException(
+            String.format(
+                "The type factory cannot hold %s(%d), which it narrowed to %s; its type system"
+                    + " allows up to %d",
+                typeName, length, type, typeFactory.getTypeSystem().getMaxPrecision(typeName)));
+      }
+      return type;
     }
 
     @Override
--- a/isthmus/src/main/java/io/substrait/isthmus/TypeConverter.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/TypeConverter.java
@@ -433,8 +433,18 @@ public class TypeConverter {
 
     @Override
     public RelDataType visit(Type.Decimal expr) {
+      // Before the factory, for the reason the lengths are: -1 is Calcite's unspecified precision,
+      // so a negative one is answered with an unparameterised DECIMAL whose precision reads back as
+      // the type system's maximum. A zero or negative scale Calcite reports itself.
+      if (expr.precision() < 0) {
+        throw new IllegalArgumentException(
+            String.format(
+                "A decimal cannot declare a negative precision, and this one is %d",
+                expr.precision()));
+      }
       SubstraitTypeSystem.requireSupportedPrecision(
           typeFactory.getTypeSystem(), SqlTypeName.DECIMAL, "decimal", expr.precision());
+      SubstraitTypeSystem.requireSupportedScale(typeFactory.getTypeSystem(), expr.scale());
       return t(n(expr), SqlTypeName.DECIMAL, expr.precision(), expr.scale());
     }
 
--- a/isthmus/src/main/java/io/substrait/isthmus/TypeConverter.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/TypeConverter.java
@@ -395,13 +395,27 @@ public class TypeConverter {
      * conversion takes whatever factory it is handed, so a factory whose limits are not Substrait's
      * would otherwise return a type narrower than the plan declares.
      *
+     * <p>A negative length is refused before the factory is asked, because asking tells us nothing:
+     * with assertions off the factory stores the negative and reports it back, so the width below
+     * certifies itself, and with them on Calcite raises a bare {@code AssertionError} in place of
+     * this message. A length of -1 is Calcite's unspecified precision besides, so the factory
+     * answers with an unparameterised type whose precision equals what was asked for. A zero length
+     * is left to the factory: the spec puts a fixedchar's width at 1 or more, but Calcite types the
+     * empty character literal as a {@code CHAR(0)}, so plans carrying one exist.
+     *
      * @param nullable whether the type is nullable
      * @param typeName the Calcite type name to build
      * @param length the declared length
      * @return the built type
-     * @throws IllegalArgumentException if the factory built a type of another length
+     * @throws IllegalArgumentException if the length is negative, or the factory built a type of
+     *     another length
      */
     private RelDataType withLength(boolean nullable, SqlTypeName typeName, int length) {
+      if (length < 0) {
+        throw new IllegalArgumentException(
+            String.format(
+                "A %s cannot declare a negative length, and this one is %d", typeName, length));
+      }
       RelDataType type = t(nullable, typeName, length);
       if (type.getPrecision() != length) {
         throw new IllegalArgumentException(
--- a/isthmus/src/main/java/io/substrait/isthmus/TypeConverter.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/TypeConverter.java
@@ -226,6 +226,8 @@ public class TypeConverter {
    * @return Calcite relational type.
    * @throws UnsupportedOperationException if the expression contains unsupported precision or
    *     user-defined types cannot be mapped.
+   * @throws IllegalArgumentException if a declared length or precision is negative, or the given
+   *     factory cannot hold it.
    */
   public RelDataType toCalcite(
       RelDataTypeFactory relDataTypeFactory, TypeExpression typeExpression) {
@@ -242,6 +244,8 @@ public class TypeConverter {
    * @return Calcite relational type.
    * @throws UnsupportedOperationException if the expression contains unsupported precision or
    *     user-defined types cannot be mapped.
+   * @throws IllegalArgumentException if a declared length or precision is negative, or the given
+   *     factory cannot hold it.
    */
   public RelDataType toCalcite(
       RelDataTypeFactory relDataTypeFactory,
@@ -429,6 +433,8 @@ public class TypeConverter {
 
     @Override
     public RelDataType visit(Type.Decimal expr) {
+      SubstraitTypeSystem.requireSupportedPrecision(
+          typeFactory.getTypeSystem(), SqlTypeName.DECIMAL, "decimal", expr.precision());
       return t(n(expr), SqlTypeName.DECIMAL, expr.precision(), expr.scale());
     }
 
--- a/isthmus/src/main/java/io/substrait/isthmus/TypeConverter.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/TypeConverter.java
@@ -444,7 +444,15 @@ public class TypeConverter {
       }
       SubstraitTypeSystem.requireSupportedPrecision(
           typeFactory.getTypeSystem(), SqlTypeName.DECIMAL, "decimal", expr.precision());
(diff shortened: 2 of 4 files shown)


nielspardon:
The ceiling question you flag belongs to #1124 — `SupportedType.maxLength()` already carries a dialect's answer for `FIXED_CHAR`, `VARCHAR` and `FIXED_BINARY`, the exact three types this touches, and nothing in isthmus reads it. Deriving `Integer.MAX_VALUE` from the proto's int32 length makes the limit a property of the spec rather than of the engine being targeted, which is the same hardcoding as the 65536 it replaces, just at the permissive end. Whether that means holding this or landing it and letting #1124 narrow it later is a maintainer call.

alexandrefimov:
Measured on this branch, since the ceiling reaching values turns out to have sharper consequences than a wrong type.

`SELECT CAST(x'01' AS BINARY(2147483647))` and `SELECT CAST('a' AS CHAR(2147483647))` both end in `OutOfMemoryError`, and in Calcite rather than here: `RexSimplify.simplifyCast` reduces the cast through `RexExecutable`, and `SqlFunctions.truncateOrPad` materialises the padded value — `Spaces.padRight` for the char, `Arrays.copyOf` for the binary. Under Calcite's default type system the width is capped to 65536 before any of that, so the same literal is built without trouble; the cap is what bounds the padding today.

The other cases are unaffected: `CAST('a' AS VARCHAR(2147483647))` converts fine since varchar is not padded, and so do column casts to `BINARY(2147483647)` or `CHAR(2147483647)`, which have no constant to reduce. So it is the two padded types, and only for literals.

That reads to me as an argument for taking the limit from `SupportedType.maxLength()` in one go rather than landing `Integer.MAX_VALUE` and narrowing in #1124 later. Happy either way — say which and I will hold this or bound it here.

nielspardon:
The padding on that path is Calcite's, agreed.

One thing to weigh before you pick, though it doesn't change the direction: a cap on `SubstraitTypeSystem` won't bound all of it. `UserTypeMapper.toSubstrait` returns a Substrait `Type` directly, so a mapper answering `TypeCreator.REQUIRED.fixedChar(Integer.MAX_VALUE)` never goes through the type factory and no `getMaxPrecision` ceiling sees it. On the `LogicalValues` path Calcite never pads either — it builds a CHAR literal at its text length, and #1171 is what widens it to the row field — so that pad wants its own bound whichever way this lands.

alexandrefimov:
Keeping `Integer.MAX_VALUE`, then, and the description says why rather than asking.

The reasoning I landed on: the number here is the spec's bound on a length, and a narrower one is a dialect's, which is `SupportedType.maxLength()` and #1124 — putting a second hardcoded ceiling in `SubstraitTypeSystem` now would only be something to take out there. Your point that a cap here does not bound the mapper path settles the other half: `UserTypeMapper.toSubstrait` hands back a Substrait type without the type factory seeing it, so whatever number this class carries, a mapper can ask for a `fixedchar<2147483647>` anyway. The pad that isthmus does itself, on the `LogicalValues` path, is refused in #1171 at the largest array a JVM allocates.

Rebased on `main` while I was here.

isthmus/src/main/java/io/substrait/isthmus/SubstraitTypeSystem.java
```diff
@@ -81,12 +81,24 @@ public static void requireSupportedPrecision(
...
+   *
    * @param typeName The {@link SqlTypeName} for which precision is requested.
    * @return Maximum precision for the type.
    */
   @Override
   public int getMaxPrecision(final SqlTypeName typeName) {
     switch (typeName) {
+      case CHAR:
+      case VARCHAR:
+      case BINARY:
```
nielspardon, line 106: Thread the provider's type factory into `SubstraitSqlToCalcite.createDefaultRelOptCluster`, which hardcodes `new JavaTypeFactoryImpl(SubstraitTypeSystem.TYPE_SYSTEM)` and takes no argument, so the validator and the converter agree on a bound again. `a || b` over two `varchar<40000>` columns — no wide type anywhere — now throws `AssertionError: Conversion to relational algebra failed to preserve datatypes` for any `ConverterProvider` built with its own factory: the validator infers unbounded `VARCHAR` from the provider's factory while the cluster infers `VARCHAR(80000)`. It converted before the raise and no test covers a provider whose type system is not Substrait's; filed as #1220 with the Substrait→SQL half, so it can land there instead if you would rather keep this PR to the type system.

isthmus/src/main/java/io/substrait/isthmus/SubstraitTypeSystem.java
```diff
@@ -81,12 +81,24 @@ public static void requireSupportedPrecision(
...
+        return Integer.MAX_VALUE;
```
nielspardon, line 108: Raise `VARBINARY` alongside `BINARY`.

```suggestion
      case VARBINARY:
        return Integer.MAX_VALUE;
```

`shouldConvertRaggedUnionTypesToVarying()` is true here, so Calcite unifies a ragged fixed-width binary union through `createSqlType(VARBINARY, precision)` — with `VARBINARY` left at 65536, `leastRestrictive(BINARY(100000), BINARY(5))` returns `VARBINARY(65536)`, narrower than one of its own inputs and re-imposing the cap this removes. That was unreachable before, since no `BINARY` wider than 65536 could exist, and the `CHAR` side has no such gap because `VARCHAR` was raised with it. The description's reasoning holds at the POJO boundary — `binary` carries no length — but not inside Calcite's own type unification, which is where this bites.

isthmus/src/main/java/io/substrait/isthmus/TypeConverter.java
```diff
@@ -368,17 +368,41 @@ public RelDataType visit(Type.IntervalDay expr) {
...
+     * conversion takes whatever factory it is handed, so a factory whose limits are not Substrait's
+     * would otherwise return a type narrower than the plan declares.
+     *
+     * @param nullable whether the type is nullable
+     * @param typeName the Calcite type name to build
+     * @param length the declared length
+     * @return the built type
+     * @throws IllegalArgumentException if the factory built a type of another length
+     */
+    private RelDataType withLength(boolean nullable, SqlTypeName typeName, int length) {
```
nielspardon, line 420: Pre-check the length before calling the factory rather than comparing its echo afterwards. A post-check can only report what the factory survives: with assertions off — the normal production JVM — `varChar(-5)` is certified as `VARCHAR(-5)`, because `getPrecision()` returns the stored negative and so equals `length`, while under `-ea` it escapes as a bare `AssertionError: null` from Calcite rather than the guard's message; the same `-1` double duty is why `varChar(-1)` passes and silently becomes Substrait `string` while `fixedChar(-1)` reports a bound its length is below. Keep it to the negatives rather than the spec's stated floor of 1, though — substrait-io/substrait#1199 has that lower bound open, and ordinary SQL already emits `fixedchar<0>` from `SELECT ''`, `CAST(a AS CHAR(0))` and a `CHAR(0)` column, so a floor here would leave isthmus unable to read back plans it produces itself.

nielspardon (commented):
Mark this breaking and add the footer — public `toCalcite` now throws where it previously returned a narrowed type, and SQL→Substrait output types change, which is the shape #1120 and #1121 both carried as `fix(isthmus)!:`. Keeping `Integer.MAX_VALUE` with the reasoning now in the description reads right to me. The three comments below are asymmetries in the raise I would fix before it lands.

isthmus/src/main/java/io/substrait/isthmus/SubstraitTypeSystem.java
alexandrefimov: Leaving that to #1220, so this PR stays on the type system.

isthmus/src/main/java/io/substrait/isthmus/SubstraitTypeSystem.java
alexandrefimov: Raised, and your example reproduces exactly: `leastRestrictive(BINARY(100000), BINARY(5))` was `VARBINARY(65536)` here, where the char side already gave `VARCHAR(100000)`. `shouldConvertRaggedUnionTypesToVarying()` is true, as you say.

The javadoc now gives the union as the reason the type is raised -- `binary` carrying no length of its own is still true, and on its own reads as a reason to leave the type alone. One thing I checked before raising it, since it widens a type nothing else bounds: a `VARBINARY` is not padded, so `cast(x'01' AS VARBINARY(2147483647))` projects fine, where the fixed-width `BINARY(2147483647)` runs out of memory exactly as `CHAR` does. That is in the description now.

isthmus/src/main/java/io/substrait/isthmus/TypeConverter.java
alexandrefimov: Pre-checked now, and both halves reproduce. With `enableAssertions = false` the factory answers `VARCHAR(-5)` and `CHAR(-5)`; under `-ea` it is a bare `AssertionError: null` rather than the guard's message. `varChar(-1)` came back as an unparameterised `VARCHAR`, whose precision is -1 and which converts back to a Substrait `string`, and `fixedChar(-1)` reported the 2147483647 bound as you describe.

Kept to the negatives. `SELECT '' FROM t` gives a `fixedchar<0>` today and substrait-io/substrait#1199 has the floor open, which is where #1171 landed as well, so the two PRs treat a zero width the same way. Measured across the range afterwards: -5 and -1 are refused for all three types, and 0, 1, 65536, 65537 and 2147483647 convert.

isthmus/src/main/java/io/substrait/isthmus/SubstraitTypeSystem.java
```diff
@@ -81,12 +81,31 @@ public static void requireSupportedPrecision(
...
+   * reimposed.
+   *
    * @param typeName The {@link SqlTypeName} for which precision is requested.
    * @return Maximum precision for the type.
    */
   @Override
   public int getMaxPrecision(final SqlTypeName typeName) {
     switch (typeName) {
+      case CHAR:
+      case VARCHAR:
```
nielspardon, line 105: Worth a test pinning the concat result type, and a line in the description. `getMaxPrecision(VARCHAR)` is also Calcite's overflow threshold in `ReturnTypes.DYADIC_STRING_SUM_PRECISION`, so `a || b` over two `varchar<40000>` columns now converts to `varchar<80000>` where it converted to `string` — neither operand is anywhere near the old cap, so this changes the emitted type for plans carrying no wide type at all. The multivalent variants do the same for `CONCAT` and `CONCAT_WS`.

isthmus/src/main/java/io/substrait/isthmus/SubstraitTypeSystem.java
```diff
@@ -81,12 +81,31 @@ public static void requireSupportedPrecision(
...
+      case BINARY:
```
nielspardon, line 106: Worth adding the zero-literal path to the description's `OutOfMemoryError` note. `RexBuilder.zeroValue` allocates `new byte[precision]` for `BINARY`, so `makeZeroLiteral(BINARY(2147483647))` now throws `OutOfMemoryError: Requested array size exceeds VM limit` where the old cap made it a 65536-byte array — and no literal need appear in the plan, since planner rules an embedder runs reach it from a column type alone. `VARBINARY` is unaffected for the same reason it is not padded: its zero value is `ByteString.EMPTY`, so raising it costs nothing here.

isthmus/src/main/java/io/substrait/isthmus/SubstraitTypeSystem.java
```diff
@@ -81,12 +81,31 @@ public static void requireSupportedPrecision(
   /**
    * Returns the maximum precision for the given SQL type.
    *
+   * <p>For the three types that carry a length across the Substrait boundary — {@link
+   * SqlTypeName#CHAR}, {@link SqlTypeName#VARCHAR} and {@link SqlTypeName#BINARY}, holding {@code
+   * fixedchar}, {@code varchar} and {@code fixed_binary} — this is Substrait's own limit: those
```
nielspardon:
```suggestion
   * fixedchar}, {@code varchar} and {@code fixedbinary} — this is Substrait's own limit: those
```
The spec and this repo's own `StringTypeVisitor` both spell it `fixedbinary` — `fixed_binary` is the dialect enum's `FIXED_BINARY`, not the type name. Two more in the description.

isthmus/src/main/java/io/substrait/isthmus/TypeConverter.java
```diff
@@ -368,17 +368,55 @@ public RelDataType visit(Type.IntervalDay expr) {
...
+        throw new IllegalArgumentException(
+            String.format(
+                "The type factory cannot hold %s(%d), which it narrowed to %s; its type system"
+                    + " allows up to %d",
+                typeName, length, type, typeFactory.getTypeSystem().getMaxPrecision(typeName)));
+      }
+      return type;
     }
 
     @Override
```
nielspardon, line 437: `visit(Type.Decimal)` just below still hands its precision straight to the factory, so the narrowing this guard reports for a length stays silent for a decimal: `decimal<38,10>` through a factory on Calcite's default type system comes back `DECIMAL(19,10)` and round-trips as `precision=19`. Same foreign-factory path and the same silent loss the guard was added for, and `requireSupportedPrecision(typeFactory.getTypeSystem(), SqlTypeName.DECIMAL, "decimal", expr.precision())` is the one-line form the temporal visits already use.

isthmus/src/test/java/io/substrait/isthmus/CalciteTypeTest.java
```diff
@@ -202,6 +203,42 @@ void varchar(boolean nullable) {
...
+  void wideLengthCarryingTypesKeepTheirLength(SqlTypeName typeName, int length) {
+    TypeExpression substrait;
+    switch (typeName) {
+      case CHAR:
+        substrait = TypeCreator.REQUIRED.fixedChar(length);
+        break;
+      case VARCHAR:
+        substrait = TypeCreator.REQUIRED.varChar(length);
+        break;
+      default:
```
nielspardon: Make the unmatched case throw rather than fall through to `fixedBinary`.

```suggestion
      default:
        throw new IllegalArgumentException("no Substrait type mapped for " + typeName);
```
A row added later — `VARBINARY, 65537` being the obvious one now that `VARBINARY` is raised too — otherwise builds a `fixed_binary` silently and fails against a `VARBINARY` expectation, which reads as a bug in the type system rather than in the test's own mapping table.

nielspardon (commented):
Five smaller things below, plus two I will leave unless you want them: `assertTrue(e.getMessage().contains("65536"))` passes on the `VARCHAR(65536)` already in that message rather than on the bound it means to pin, and both public `toCalcite` overloads still document only `@throws UnsupportedOperationException` now that two conditions raise `IllegalArgumentException`.

isthmus/src/main/java/io/substrait/isthmus/SubstraitTypeSystem.java
alexandrefimov: Test and description line added, on the numbers rather than the mechanism: `SELECT a || b` over two `varchar<40000>` columns converts to `varchar<80000>` here and to `string` on `origin/main` (934a60e), and `CONCAT(a, b)` matches.

`CONCAT_WS` I left out of the test -- it does not reach a Substrait type at all, failing earlier with `Unable to convert call CONCAT_WS(char<1>, varchar<40000>?, varchar<40000>?)`, so the width never gets to matter.

One end I checked while adding it, since raising the ceiling makes the sum reachable: `varchar<2000000000> || varchar<2000000000>` does not overflow into a negative width, it falls back to `string` the way the old cap did.

isthmus/src/main/java/io/substrait/isthmus/SubstraitTypeSystem.java
alexandrefimov: In the description now, with both ends measured: `makeZeroLiteral(BINARY(2147483647))` is `OutOfMemoryError: Requested array size exceeds VM limit` here and a 65536-byte array on `origin/main`, and `makeZeroLiteral(VARBINARY(2147483647))` is `X''` at any width, as you say.

isthmus/src/main/java/io/substrait/isthmus/SubstraitTypeSystem.java
alexandrefimov: Applied, and the two in the description with it. The `FIXED_CHAR` / `VARCHAR` / `FIXED_BINARY` further down stay as they are -- those name `SupportedType`'s enum rather than the types.

isthmus/src/main/java/io/substrait/isthmus/TypeConverter.java
alexandrefimov: Added, in the one-line form. Measured first: `decimal<38,10>` through a factory on Calcite's default type system came back `DECIMAL(19,10)` and round-tripped as `precision=19`; it now reports the 19 it exceeds.

Checking the ends of that guard turned up two more, both silent and neither needing a foreign factory. `decimal<-1,0>` converted to a bare `DECIMAL` and read back as the type system's maximum -- `-1` is Calcite's unspecified precision here the same way it is for a length, so this went through `TYPE_FACTORY` as well. And a scale past the maximum was narrowed rather than reported: `decimal<19,25>` through the default type system came back with a scale of 19. Both are closed here, the scale through a `requireSupportedScale` next to the existing precision one. Calcite reports a zero precision and a negative scale itself, so those are left to it.

The footer covers the lot, since the same public method now throws where it narrowed.

isthmus/src/test/java/io/substrait/isthmus/CalciteTypeTest.java
alexandrefimov: Applied. I left the `VARBINARY, 65537` row itself out: Substrait's `binary` carries no length, so there is no type to build the case from -- which is what the throw now says to whoever adds the row anyway.

alexandrefimov:
Took both of the two you left rather than leaving them: the length test asserts `allows up to 65536`, which the narrowed `VARCHAR(65536)` in the same message no longer satisfies, and both public `toCalcite` overloads document the `IllegalArgumentException` alongside the existing `UnsupportedOperationException`.

nielspardon:
looks like this needs to be reconciled with latest main

alexandrefimov:
Rebased onto `main` at 85c03c7. One conflict, in `CalciteTypeTest`: #1171 added `aJavaCharColumnTakesCalcitesDefaultWidth` where this branch adds `wideLengthCarryingTypesKeepTheirLength` — both kept, each with its own javadoc.

`./gradlew build` and `./gradlew integrationTest` are both green on the rebased branch, which should also clear the red Integration tests job: it failed on `429 Too Many Requests` from Maven Central resolving `spotless-lib` while configuring `:build-logic`, not on anything here.

isthmus/src/test/java/io/substrait/isthmus/SubstraitTypeSystemTest.java
```diff
@@ -60,4 +178,103 @@ void decimalMaxPrecisionAndScaleDifferentFromDefaultTypeSystem() {
     assertEquals(38, typeSystem.getMaxPrecision(SqlTypeName.DECIMAL));
     assertEquals(38, typeSystem.getMaxScale(SqlTypeName.DECIMAL));
   }
+
+  /**
+   * The cap reaches the conversion from SQL as well. Calcite narrows a declared width to its
+   * maximum silently rather than reporting that it cannot hold it, so before this a cast wider than
+   * the default came out of the conversion as a {@code varchar<65536>}.
+   */
```
nielspardon: Delete this block — it documents `aWideVarcharDeclaredInSqlKeepsItsLength` at the bottom of the file, not this test. Of two consecutive doc comments only the second applies, so the first is dead text.

```suggestion
```

nielspardon (commented):
Everything from round 2 is in, and the description now carries the zero-literal and concatenation paths. Two small things left: a stray Javadoc block in the new test, and a scale check no valid decimal can reach.

isthmus/src/main/java/io/substrait/isthmus/TypeConverter.java
```diff
@@ -376,21 +380,71 @@ public RelDataType visit(Type.IntervalDay expr) {
...
+      // the type system's maximum. A zero or negative scale Calcite reports itself.
+      if (expr.precision() < 0) {
+        throw new IllegalArgumentException(
+            String.format(
+                "A decimal cannot declare a negative precision, and this one is %d",
+                expr.precision()));
+      }
+      SubstraitTypeSystem.requireSupportedPrecision(
+          typeFactory.getTypeSystem(), SqlTypeName.DECIMAL, "decimal", expr.precision());
+      SubstraitTypeSystem.requireSupportedScale(typeFactory.getTypeSystem(), expr.scale());
```
nielspardon: No spec-valid decimal can reach this check: the spec puts scale at `0 <= S <= P <= 38`, and both type systems here set `maxScale` equal to `maxPrecision` (19 and 38), so a scale that survives the precision check above is already within `maxScale`. The one input that does reach it — `decimal<19,25>`, in the test and the description — has scale above precision, which the spec disallows and which `TYPE_FACTORY` converts to `DECIMAL(19,25)` without complaint. `expr.scale() > expr.precision()` is the bound actually missing.

isthmus/src/test/java/io/substrait/isthmus/SubstraitTypeSystemTest.java
alexandrefimov: Moved rather than deleted -- it is `aWideVarcharDeclaredInSqlKeepsItsLength`'s own javadoc, added with that test in the SQL-cast commit, and the decimal test landed between the two. The test at the bottom of the file had none left, so the block goes back to it.

isthmus/src/main/java/io/substrait/isthmus/TypeConverter.java
alexandrefimov: The spec bound is in instead, and `requireSupportedScale` is gone with it -- nothing else called it.

Both halves check out: `maxScale` is 38 against a `maxPrecision` of 38 here and 19 against 19 on Calcite's default, so a scale that survives the precision check is inside `maxScale` by construction, and the only input that reached the old check was one the spec already disallows. `type_classes.md` gives `0 <= S <= P` for the same reading.

`decimal<19,25>` now reports its scale against its precision. Measured across the boundary afterwards: `decimal<38,38>` and `decimal<19,19>` convert, `decimal<19,20>` and `decimal<38,39>` are refused, and a zero precision or a negative scale still comes back from Calcite itself.

isthmus/src/main/java/io/substrait/isthmus/TypeConverter.java
```diff
@@ -376,21 +380,79 @@ public RelDataType visit(Type.IntervalDay expr) {
...
+      // so a negative one is answered with an unparameterised DECIMAL whose precision reads back as
+      // the type system's maximum. A zero or negative scale Calcite reports itself.
+      if (expr.precision() < 0) {
+        throw new IllegalArgumentException(
+            String.format(
+                "A decimal cannot declare a negative precision, and this one is %d",
+                expr.precision()));
+      }
+      SubstraitTypeSystem.requireSupportedPrecision(
+          typeFactory.getTypeSystem(), SqlTypeName.DECIMAL, "decimal", expr.precision());
```
nielspardon, line 458: Bound the precision by the spec's 38 too, not only by the factory's maximum — `P <= 38` is the same clause as the `0 <= S <= P` just added. Handed a type system whose `DECIMAL` maximum is above 38, this returns `DECIMAL(45,2)` for `decimal<45,2>`, and `toSubstrait` then refuses it against the hardcoded `> 38` at line 146 — so the type converts in and cannot come back. That literal makes a shared constant the tidier form than the two spellings below.

```suggestion
      if (expr.precision() > 38) {
        throw new IllegalArgumentException(
            String.format(
                "A decimal cannot declare a precision of %d, above the 38 the spec allows",
                expr.precision()));
      }
      SubstraitTypeSystem.requireSupportedPrecision(
          typeFactory.getTypeSystem(), SqlTypeName.DECIMAL, "decimal", expr.precision());
```

nielspardon (commented):
Both fixes read right, and moving the stray Javadoc onto its own test beats deleting it. One thing left — the other half of the spec clause the scale bound came from.

isthmus/src/main/java/io/substrait/isthmus/TypeConverter.java
alexandrefimov: Applied, with the constant. Reproduced it first: with a type system whose `getMaxPrecision(DECIMAL)` is 76, `decimal<45,2>` converted to `DECIMAL(45,2)` and came back as `UnsupportedOperationException: unsupported decimal precision 45` -- in with no way out, as you describe.

`MAX_DECIMAL_PRECISION` sits on `TypeConverter` and both ends of the clause read it. The test builds that wide type system rather than asserting on the message alone; taking the inbound bound out fails it.

nielspardon (approved):
LGTM

Issue #200, opened 2023-11-08
How to carry aliases in internal projects


xiedeyantu:
There is only names in root path. child node can not know aliases info.

nielspardon:
The Substrait specification now supports specifying intermediate field names in the `Hint` for a relation operation which was introduced in this PR: https://github.com/substrait-io/substrait/pull/696

Accordingly, substrait-java added support for specifying intermediate output names in this PR https://github.com/substrait-io/substrait-java/pull/295

With those changes I would consider this issue resolved. Let us know if you think there is still something missing.

PR #1171, opened 2026-08-26
fix(isthmus)!: build a character literal from the type its conversion produced

### LINKED ISSUE #1170: isthmus: a UserTypeMapper is applied to a relation's schema but dropped from its literals, so the two disagree
A `UserTypeMapper` configured on the `TypeConverter` is applied when converting a relation's schema but discarded when converting its literals, so a plan comes out disagreeing with itself. `VirtualTableScan` catches it:

```java
UserTypeMapper charsToString = ... // maps CHAR and VARCHAR to Substrait's lengthless string
ConverterProvider p =
    ConverterProvider.builder().typeConverter(new TypeConverter(charsToString)).build();
```

```
SELECT * FROM (VALUES ('a'), ('b')) AS v(c)
-- IllegalArgumentException: Row field type (FixedChar{nullable=false, length=1})
--   does not match schema field type (Str{nullable=false})

SELECT i FROM t WHERE s IN ('v0', ..., 'v24')          -- 25 values, above Calcite's
-- IllegalArgumentException: Row field type            -- inSubQueryThreshold, so it
--   (VarChar{nullable=false, length=20})              -- becomes a virtual table
--   does not match schema field type (Str{nullable=false})

SELECT s FROM t                                        -- converts: no literal involved
SELECT i FROM t WHERE s IN ('a', 'b', 'c')             -- converts: expands to OR/EQUALS
```

The mapper is consulted for the literal — `LiteralConverter.convert(literal, resultType)` opens with `typeConverter.toSubstrait(resultType)` — but its answer is used only for nullability and for the null-literal case. Every non-null literal is then rebuilt by `switch (resultType.getSqlTypeName())` from the Calcite type, and the mapped type is dropped.

That produces a sharp asymmetry, which is the clearest way to see it:

```
VALUES (CAST(NULL AS VARCHAR(20)))                           converts   -- null honours the mapper
VALUES (CAST(NULL AS VARCHAR(20))), (CAST('b' AS VARCHAR(20)))
-- IllegalArgumentException: Row field type (VarChar{nullable=true, length=20})
--   does not match schema field type (Str{nullable=true})              -- the non-null one does not
```

So the extension point reads as working — a mapped column converts, the schema carries the mapped type — until a literal of that type appears in the same relation.

This is the same class of schema/literal mismatch as #1064 — and as #683 and #684 before it for nullability — but that fix does not cover it. It made the literal convert against the row schema's Calcite type, which is why the `resultType` overload exists; the mapped type is lost one step later, when the literal is rebuilt from that type's name. A mapping to `string` is not expressible as a Calcite type name at all, so passing the schema type down cannot carry it. The fix belongs in that switch rather than in what is handed to it, and it is not a regression from #1064: before it the schema went through the mapper and the literal did not, the same way.

Raised by an external consumer exporting Impala's Calcite plans through Isthmus. Impala's `string` reaches Calcite as `VARCHAR(2147483647)`, and mapping it to Substrait's `string` is the natural use of this hook; it broke TPC-DS q08, whose `IN` list is large enough to become a virtual table. The repro above needs nothing beyond Isthmus.

alexandrefimov:
`LiteralConverter.convert` derives the Substrait type twice. Once through `typeConverter.toSubstrait(resultType)`, which consults the `UserTypeMapper`, and again in the `switch` over `resultType.getSqlTypeName()`, which does not — and the second derivation is the one the literal gets. So a mapped character column produced a schema of the mapped type and literals of the unmapped one, and `VirtualTableScan` rejected the relation:

```
SELECT * FROM (VALUES ('a'), ('b')) AS v(c)     -- with CHAR/VARCHAR mapped to string
-- Row field type (FixedChar{nullable=false, length=1})
--   does not match schema field type (Str{nullable=false})
```

A null literal takes its type straight from the first derivation, so it already agreed; only rows carrying a value diverged. That asymmetry is what the `nullAndNonNullLiteralsCarryTheSameMappedType` test pins.

The character branches now build from the type already derived, which covers the three forms a Calcite character type can map to — `fixedchar`, `varchar` and `string`. It also removes a copy of the `PRECISION_NOT_SPECIFIED` rule that `TypeConverter` already applies — the `VARCHAR` one, which is the only arm that had it. `CHAR` had none, so a reflective schema's `char` or `Character` column, which carries no precision at all, derived a `fixedchar<-1>`: outside the `[1..2147483647]` the spec allows, and a width the check below then measured every value against, the empty string included. That arm now reads the unspecified precision as Calcite's default of 1, and each rule lives in one place.

A `fixedchar` literal needed one thing more: it carries no length of its own, since `Expression.FixedCharLiteral` derives the type from its text. Building one from the declared type therefore means padding the text to the declared width, which is also what `CHAR(n)` means — `'a'` in a `CHAR(3)` is `'a  '` — and what `padRightIfNeeded` already does for `BINARY` a few lines down. A value longer than the width it is declared as is rejected rather than truncated. The width is counted in characters, as the spec gives it -- a `FIXEDCHAR<L>` is `L` characters where a `string` is a count of UTF-8 bytes -- so one astral character fills a `fixedchar<1>`.

That half fixes a case needing no mapper at all. A `LogicalValues` row field wider than its literal — the shape #1064 was reported with — was still rejected for character types: a `CHAR(3)` field holding `'a'` produced `Row field type (FixedChar{length=1}) does not match schema field type (FixedChar{length=3})`. Isthmus' own SQL path does not reach it, because Calcite pads the literal itself; a plan arriving from another planner does.

What this does not cover, deliberately. A mapper that maps some other family still diverges the same way — mapping `INTEGER` to `i64` gives `Row field type (I32) does not match schema field type (I64)`. Closing that generally means dispatching every branch on the Substrait type rather than the Calcite type name, which is a much larger change to this method, and it still would not reach a genuine `Type.UserDefined`: there is no way to build a literal of one from a `RexLiteral` without knowing its encoding, so that needs a literal-side hook the interface does not have. The character family is the part that can be closed without deciding either of those. A mapping outside it leaves the literal in the form Calcite declares, which is what it converted to before this change; under a virtual table that still puts the schema and its rows out of step, and closing it needs the literal-side hook above.

Worth saying, since it bears on whether this use is meant to be supported at all: `UserTypeMapper.toSubstrait` returns a `Type`, so mapping to a built-in is expressible, while its reverse takes a `Type.UserDefined`. The consumer this came from maps Impala's `string` — reaching Calcite as `VARCHAR(2147483647)` — onto Substrait's `string`, which is a natural reading of the forward signature. That reading is the intended one: `TypeConverter` consults the mapper before its own switch precisely because a mapper may re-use a `SqlTypeName`, so mapping a Calcite built-in onto a Substrait built-in is what the forward direction offers. The reverse lags behind it — `toCalcite` takes a `Type.UserDefined`, so a mapped built-in has no hook to come back through — which is #1219.

A negative width is refused where the padding happens: a mapper hands the literal a Substrait type directly, so nothing between it and this method holds its width to what a `fixedchar` can declare. It threw before as well, but reported the value as longer than the `fixedchar<-5>` it was declared as, rather than the width as the problem.

The floor stops at the negatives rather than at the `[1..2147483647]` `type_classes.md` gives `L`. Calcite types `''` as a `CHAR(0)` and its DDL parser takes a `CHAR(0)` column, so a `fixedchar<0>` reaches a plan from ordinary SQL — on `origin/main` (934a60e) `SELECT '' FROM t`, `WHERE '' = ''`, `CHAR_LENGTH('')` and `'' || 'x'` all convert to one — and the bound itself is under discussion in substrait-io/substrait#1199, where `dialect_schema.yaml` allows a `max_length` of 0 for the same three types. A zero width therefore converts as it does today, rather than being refused now and legalized again if that issue lands the other way.

Closes #1170.

BREAKING CHANGE: a Calcite `CHAR(n)` literal converts to a `fixedchar<n>` whose text is padded to the declared width, where it previously carried the unpadded text and so a `fixedchar` of the text's own length. A character value longer than the width it is declared as is rejected rather than converted.

### DIFF
--- a/isthmus/src/main/java/io/substrait/isthmus/expression/LiteralConverter.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/expression/LiteralConverter.java
@@ -17,6 +17,7 @@ import java.time.format.DateTimeFormatter;
 import java.time.format.DateTimeFormatterBuilder;
 import java.time.temporal.ChronoField;
 import java.util.List;
+import java.util.Locale;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.concurrent.TimeUnit;
@@ -88,14 +89,17 @@ public class LiteralConverter {
    * type has no character literal form to build, and reaching one would need the value's encoding
    * in that type, which {@link io.substrait.isthmus.UserTypeMapper} has no way to give.
    *
+   * <p>Whether the literal is nullable comes from the mapped type either way, which is what the
+   * containing conversion took it from before the type reached here.
+   *
    * @param mappedType the Substrait type the containing conversion produced
-   * @param nullable whether the literal is nullable
    * @param value the literal's text
    * @param calciteType the Calcite type the literal was declared as
    * @return the literal
    */
   private static Expression.Literal characterLiteral(
-      Type mappedType, boolean nullable, String value, RelDataType calciteType) {
+      Type mappedType, String value, RelDataType calciteType) {
+    boolean nullable = mappedType.nullable();
     Type type =
         mappedType instanceof Type.Str
                 || mappedType instanceof Type.VarChar
@@ -115,14 +119,19 @@ public class LiteralConverter {
       // A fixedchar literal carries no length of its own — Expression.FixedCharLiteral derives it
       // from the text — so the text has to be the declared width or the two disagree. Padding is
       // also what CHAR(n) means: 'a' in a CHAR(3) is 'a  '.
+      // In characters rather than UTF-16 code units: the spec gives a fixedchar its length in
+      // characters, where it spells a string's out in UTF-8 bytes.
       int length = ((Type.FixedChar) type).length();
-      if (value.length() > length) {
+      int characters = value.codePointCount(0, value.length());
+      if (characters > length) {
         throw new IllegalArgumentException(
             String.format(
+                Locale.ROOT,
                 "Character value '%s' is longer than the fixedchar<%d> it is declared as",
-                value, length));
+                value,
+                length));
       }
-      return ExpressionCreator.fixedChar(nullable, value + " ".repeat(length - value.length()));
+      return ExpressionCreator.fixedChar(nullable, value + " ".repeat(length - characters));
     }
     throw new IllegalStateException(
         String.format(
@@ -185,7 +194,7 @@ public class LiteralConverter {
         {
           Comparable<?> val = literal.getValue();
           if (val instanceof NlsString) {
-            return characterLiteral(type, nullable, ((NlsString) val).getValue(), resultType);
+            return characterLiteral(type, ((NlsString) val).getValue(), resultType);
           }
           throw new UnsupportedOperationException("Unable to handle char type: " + val);
         }
@@ -202,7 +211,7 @@ public class LiteralConverter {
               nullable, bd, resultType.getPrecision(), resultType.getScale());
         }
       case VARCHAR:
-        return characterLiteral(type, nullable, s(literal), resultType);
+        return characterLiteral(type, s(literal), resultType);
       case BINARY:
         return ExpressionCreator.fixedBinary(
             nullable,
--- a/isthmus/src/main/java/io/substrait/isthmus/expression/LiteralConverter.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/expression/LiteralConverter.java
@@ -60,6 +60,19 @@ public class LiteralConverter {
           .append(CALCITE_LOCAL_TIME_FORMATTER)
           .toFormatter();
 
+  /**
+   * The longest text a {@code fixedchar} literal can be padded to.
+   *
+   * <p>A Java {@link String} holds its characters in an array, and this is the largest one a JVM
+   * allocates, so a wider literal cannot be built whatever the heap. {@code String.repeat} reports
+   * that as an {@link OutOfMemoryError}, which no {@code catch (Exception)} sees and which names
+   * neither the column nor the value. A width the heap alone cannot hold still fails as one.
+   *
+   * <p>The width a fixedchar may declare is the spec's {@code [1..2147483647]}; what a plan's
+   * target engine actually supports is a dialect's answer, not this.
+   */
+  private static final int MAX_PADDED_LENGTH = Integer.MAX_VALUE - 8;
+
   private final TypeConverter typeConverter;
 
   /**
@@ -130,6 +143,17 @@ public class LiteralConverter {
                 "Character value '%s' is longer than the fixedchar<%d> it is declared as",
                 value,
                 length));
+      }
+      long padded = (long) value.length() + ((long) length - characters);
+      if (padded > MAX_PADDED_LENGTH) {
+        throw new IllegalArgumentException(
+            String.format(
+                Locale.ROOT,
+                "A fixedchar<%d> literal cannot be built from '%s': padding it to that width takes "
+                    + "%d characters, more than a Java String holds",
+                length,
+                value,
+                padded));
       }
       return ExpressionCreator.fixedChar(nullable, value + " ".repeat(length - characters));
     }
--- a/isthmus/src/main/java/io/substrait/isthmus/expression/LiteralConverter.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/expression/LiteralConverter.java
@@ -135,6 +135,16 @@ public class LiteralConverter {
       // In characters rather than UTF-16 code units: the spec gives a fixedchar its length in
       // characters, where it spells a string's out in UTF-8 bytes.
       int length = ((Type.FixedChar) type).length();
+      // Only the negative end. The spec puts a fixedchar's width in [1..2147483647] (spec
+      // v0.101.0), but Calcite types the empty character literal as a CHAR(0) and its DDL parser
+      // takes a CHAR(0) column, so refusing a zero width here would stop ordinary SQL converting.
+      if (length < 0) {
+        throw new IllegalArgumentException(
+            String.format(
+                Locale.ROOT,
+                "A fixedchar cannot declare a negative width, and this one is %d",
+                length));
+      }
       int characters = value.codePointCount(0, value.length());
       if (characters > length) {
         throw new IllegalArgumentException(
@@ -150,7 +160,7 @@ public class LiteralConverter {
             String.format(
                 Locale.ROOT,
                 "A fixedchar<%d> literal cannot be built from '%s': padding it to that width takes "
-                    + "%d characters, more than a Java String holds",
+                    + "%d UTF-16 code units, more than a Java String holds",
                 length,
                 value,
                 padded));
--- a/isthmus/src/main/java/io/substrait/isthmus/expression/LiteralConverter.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/expression/LiteralConverter.java
@@ -135,9 +135,9 @@ public class LiteralConverter {
       // In characters rather than UTF-16 code units: the spec gives a fixedchar its length in
       // characters, where it spells a string's out in UTF-8 bytes.
       int length = ((Type.FixedChar) type).length();
-      // Only the negative end. The spec puts a fixedchar's width in [1..2147483647] (spec
-      // v0.101.0), but Calcite types the empty character literal as a CHAR(0) and its DDL parser
-      // takes a CHAR(0) column, so refusing a zero width here would stop ordinary SQL converting.
+      // Only the negative end. The spec puts a fixedchar's width in [1..2147483647], but Calcite
+      // types the empty character literal as a CHAR(0) and its DDL parser takes a CHAR(0) column,
+      // so refusing a zero width here would stop ordinary SQL converting.
       if (length < 0) {
         throw new IllegalArgumentException(
             String.format(
--- a/isthmus/src/main/java/io/substrait/isthmus/expression/LiteralConverter.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/expression/LiteralConverter.java
@@ -94,6 +94,9 @@ public class LiteralConverter {
       return ExpressionCreator.string(nullable, value);
     }
     if (type instanceof Type.VarChar) {
+      // Unlike a fixedchar, a varchar literal carries a length of its own, so a value shorter than
+      // it is what the type means. A value longer than it is malformed, and nothing rejects it --
+      // here or in the POJO that holds it.
       return ExpressionCreator.varChar(nullable, value, ((Type.VarChar) type).length());
     }
     if (type instanceof Type.FixedChar) {
--- a/isthmus/src/main/java/io/substrait/isthmus/expression/LiteralConverter.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/expression/LiteralConverter.java
@@ -84,12 +84,24 @@ public class LiteralConverter {
    * io.substrait.isthmus.UserTypeMapper} can map a Calcite character type to any of the three, and
    * only the first derivation consults it.
    *
-   * @param type the Substrait type this literal must carry
+   * <p>A mapping to something outside the character family keeps the form Calcite declares. Its
+   * type has no character literal form to build, and reaching one would need the value's encoding
+   * in that type, which {@link io.substrait.isthmus.UserTypeMapper} has no way to give.
+   *
+   * @param mappedType the Substrait type the containing conversion produced
    * @param nullable whether the literal is nullable
    * @param value the literal's text
+   * @param calciteType the Calcite type the literal was declared as
    * @return the literal
    */
-  private static Expression.Literal characterLiteral(Type type, boolean nullable, String value) {
+  private static Expression.Literal characterLiteral(
+      Type mappedType, boolean nullable, String value, RelDataType calciteType) {
+    Type type =
+        mappedType instanceof Type.Str
+                || mappedType instanceof Type.VarChar
+                || mappedType instanceof Type.FixedChar
+            ? mappedType
+            : TypeConverter.DEFAULT.toSubstrait(calciteType);
     if (type instanceof Type.Str) {
       return ExpressionCreator.string(nullable, value);
     }
@@ -112,11 +124,9 @@ public class LiteralConverter {
       }
       return ExpressionCreator.fixedChar(nullable, value + " ".repeat(length - value.length()));
     }
-    throw new UnsupportedOperationException(
+    throw new IllegalStateException(
         String.format(
-            "A Calcite character type converted to %s, which has no character literal form; a "
-                + "UserTypeMapper returning it cannot be applied to the literal '%s'",
-            type, value));
+            "A Calcite character type converted to %s, which is not a character type", type));
   }
 
   private static BigDecimal bd(RexLiteral literal) {
@@ -175,7 +185,7 @@ public class LiteralConverter {
         {
           Comparable<?> val = literal.getValue();
           if (val instanceof NlsString) {
-            return characterLiteral(type, nullable, ((NlsString) val).getValue());
+            return characterLiteral(type, nullable, ((NlsString) val).getValue(), resultType);
           }
           throw new UnsupportedOperationException("Unable to handle char type: " + val);
         }
@@ -192,7 +202,7 @@ public class LiteralConverter {
               nullable, bd, resultType.getPrecision(), resultType.getScale());
         }
       case VARCHAR:
-        return characterLiteral(type, nullable, s(literal));
+        return characterLiteral(type, nullable, s(literal), resultType);
       case BINARY:
         return ExpressionCreator.fixedBinary(
             nullable,
--- a/isthmus/src/main/java/io/substrait/isthmus/expression/LiteralConverter.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/expression/LiteralConverter.java
@@ -78,6 +78,44 @@ public class LiteralConverter {
     return ((NlsString) literal.getValue()).getValue();
   }
 
+  /**
+   * Builds a character literal of the Substrait type the containing conversion already produced,
+   * rather than deriving the form a second time from the Calcite type. A {@link
+   * io.substrait.isthmus.UserTypeMapper} can map a Calcite character type to any of the three, and
+   * only the first derivation consults it.
+   *
+   * @param type the Substrait type this literal must carry
+   * @param nullable whether the literal is nullable
+   * @param value the literal's text
+   * @return the literal
+   */
+  private static Expression.Literal characterLiteral(Type type, boolean nullable, String value) {
+    if (type instanceof Type.Str) {
+      return ExpressionCreator.string(nullable, value);
+    }
+    if (type instanceof Type.VarChar) {
+      return ExpressionCreator.varChar(nullable, value, ((Type.VarChar) type).length());
+    }
+    if (type instanceof Type.FixedChar) {
+      // A fixedchar literal carries no length of its own — Expression.FixedCharLiteral derives it
+      // from the text — so the text has to be the declared width or the two disagree. Padding is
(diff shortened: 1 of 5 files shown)


isthmus/src/main/java/io/substrait/isthmus/expression/LiteralConverter.java
```diff
@@ -78,6 +78,47 @@ private static String s(RexLiteral literal) {
...
+      // A fixedchar literal carries no length of its own — Expression.FixedCharLiteral derives it
+      // from the text — so the text has to be the declared width or the two disagree. Padding is
+      // also what CHAR(n) means: 'a' in a CHAR(3) is 'a  '.
+      int length = ((Type.FixedChar) type).length();
+      if (value.length() > length) {
+        throw new IllegalArgumentException(
+            String.format(
+                "Character value '%s' is longer than the fixedchar<%d> it is declared as",
+                value, length));
+      }
```
nielspardon, line 156: Fall back to the Calcite-derived form for a non-character type instead of throwing. A mapper returning `Type.UserDefined` for a character column is what the interface is for, and it converted before this PR — `WHERE c = 'a'` gave `Cast{UserDefined, VarCharLiteral}`. The message is right that a type like `uuid` has no character literal form, but not that the mapping cannot be applied: `UUID.fromString(value)` does it.

nielspardon (commented):
Whether a mapper returning a non-character type should fall back to the Calcite-derived literal instead of throwing is worth settling first — it converted before this PR, and the same question decides #1190. Worth marking this `!` either way, since the padding changes the plans existing consumers get. I have more on the padding, the width check and the tests, but let's start here.

isthmus/src/main/java/io/substrait/isthmus/expression/LiteralConverter.java
alexandrefimov: Fell back as you suggest, and the message claiming the mapping cannot be applied is gone with it. `characterLiteral` now takes the Calcite type too and builds from `TypeConverter.DEFAULT.toSubstrait(calciteType)` when the mapped type is outside the character family, so a mapper returning `Type.UserDefined` for a character column converts again rather than being refused.

Two tests in place of the rejection one: a projected character literal comes out in the Calcite form, and a virtual table under the same mapper still fails `VirtualTableScan` with `Row field type (FixedChar{nullable=false, length=1}) does not match schema field type (...)` — where it failed before this PR, and not something the fallback can close.

Marked `!`, with the fixedchar padding as the note.

isthmus/src/main/java/io/substrait/isthmus/expression/LiteralConverter.java
```diff
@@ -78,6 +78,57 @@ private static String s(RexLiteral literal) {
...
+      // it is what the type means. A value longer than it is malformed, and nothing rejects it --
+      // here or in the POJO that holds it.
+      return ExpressionCreator.varChar(nullable, value, ((Type.VarChar) type).length());
+    }
+    if (type instanceof Type.FixedChar) {
+      // A fixedchar literal carries no length of its own — Expression.FixedCharLiteral derives it
+      // from the text — so the text has to be the declared width or the two disagree. Padding is
+      // also what CHAR(n) means: 'a' in a CHAR(3) is 'a  '.
+      int length = ((Type.FixedChar) type).length();
+      if (value.length() > length) {
```
nielspardon: Count characters rather than UTF-16 code units — the spec defines `FIXEDCHAR<L>` in characters where it spells `string` out as UTF-8 bytes, so a CHAR(1) holding one astral character is rejected today. `Locale.ROOT` goes with it (and a `java.util.Locale` import): under `ar-EG` the bare `%d` renders `fixedchar<١>` and the test's substring assertion fails.

```suggestion
      int length = ((Type.FixedChar) type).length();
      int characters = value.codePointCount(0, value.length());
      if (characters > length) {
        throw new IllegalArgumentException(
            String.format(
                Locale.ROOT,
                "Character value '%s' is longer than the fixedchar<%d> it is declared as",
                value, length));
      }
      return ExpressionCreator.fixedChar(nullable, value + " ".repeat(length - characters));
```

The description says dropping the varchar copy means "the two cannot drift apart again for these types", but that holds for `VARCHAR` only — `TypeConverter`'s `CHAR` arm has no sentinel guard, so a reflective-schema `char`/`Character` column derives `fixedchar<-1>` and this check then rejects every value including `""`. Worth fixing there in this PR, same class as your `precisionOf` commit:

```java
      case CHAR:
        {
          // A char/Character JavaType carries no precision; Calcite reads that as its default of 1.
          if (type.getPrecision() == RelDataType.PRECISION_NOT_SPECIFIED) {
            return creator.fixedChar(1);
          }
          return creator.fixedChar(type.getPrecision());
        }
```

nielspardon (commented):
The fallback reads well. Three left, and the first is also a correction to the description: `TypeConverter` does not apply the `PRECISION_NOT_SPECIFIED` rule for `CHAR`.

isthmus/src/main/java/io/substrait/isthmus/expression/LiteralConverter.java
```diff
@@ -78,6 +78,57 @@ private static String s(RexLiteral literal) {
...
+   * in that type, which {@link io.substrait.isthmus.UserTypeMapper} has no way to give.
+   *
+   * @param mappedType the Substrait type the containing conversion produced
+   * @param nullable whether the literal is nullable
+   * @param value the literal's text
+   * @param calciteType the Calcite type the literal was declared as
+   * @return the literal
+   */
+  private static Expression.Literal characterLiteral(
+      Type mappedType, boolean nullable, String value, RelDataType calciteType) {
```
nielspardon: Drop `nullable` and read `mappedType.nullable()` inside. The fallback makes this sharper than before: the literal is now built at the Calcite-derived type while its nullability still comes from the mapped one, so a mapper whose nullability disagrees with the Calcite type gets its form from one source and its nullability from the other.

isthmus/src/test/java/io/substrait/isthmus/UserTypeMapperLiteralTest.java
```diff
@@ -0,0 +1,227 @@
...
+
+    VirtualTableScan converted = convert(ConverterProvider.DEFAULT, values);
+
+    assertEquals(List.of(R.fixedChar(3)), converted.getInitialSchema().struct().fields());
+    assertEquals(
+        List.of(ExpressionCreator.fixedChar(false, "a  ")), converted.getRows().get(0).fields());
+  }
+
+  @Test
+  void aCharacterValueWiderThanItsDeclaredTypeIsRejected() {
```
nielspardon, line 184: Assert on `LiteralConverter.convert` directly — `assertThrows(IllegalArgumentException.class, …)` passes on the base commit too, because `VirtualTableScan.check` throws the same type, so only the `fixedchar<1>` substring pins the new guard. Every column here is still `SqlTypeName.CHAR`, so `case VARCHAR` and the `PRECISION_NOT_SPECIFIED` rule you removed from it are untested.

isthmus/src/main/java/io/substrait/isthmus/expression/LiteralConverter.java
alexandrefimov: Both applied. The astral case is pinned in `aFixedCharWidthCountsCharactersRatherThanCodeUnits`, which needed the literal built through an explicit UTF-16 `NlsString` — Calcite's default charset is ISO-8859-1 and refuses the character outright.

The `CHAR` arm is fixed too, and the description no longer claims the removed rule covered both: `createJavaType(char.class)` and `Character.class` both come out at `PRECISION_NOT_SPECIFIED`, so `CalciteTypeTest.aJavaCharColumnTakesCalcitesDefaultWidth` fails with `fixedchar<-1>` without the guard.

Left the pad itself unbounded here. The spec puts a fixedchar's `L` in `[1..2147483647]`, so any number I picked in the literal converter would be arbitrary where a dialect's `max_length` is not — is #1124 where you want that bound, or would you rather this PR refuse a width it cannot materialise?

isthmus/src/main/java/io/substrait/isthmus/expression/LiteralConverter.java
alexandrefimov: Dropped, read from `mappedType` inside. Nothing pins it — the value is the one the caller was passing — so it stands on the reading you gave it.

isthmus/src/test/java/io/substrait/isthmus/UserTypeMapperLiteralTest.java
alexandrefimov: Moved onto `LiteralConverter.convert`, and you are right that it was passing for the wrong reason: with the guard removed the old assertion stayed green, the new one fails with "Expected java.lang.IllegalArgumentException to be thrown, but nothing was thrown".

`case VARCHAR` now has both halves covered in `CalciteLiteralTest` — a declared width, and a reflective `String` column for the `PRECISION_NOT_SPECIFIED` rule, which fails with a `VarCharLiteral` if that rule goes.

isthmus/src/main/java/io/substrait/isthmus/expression/LiteralConverter.java
nielspardon: Refuse here, and key it to the one limit that isn't a policy choice: a width whose padded text a Java `String` cannot hold. That is not a number you picked — it's where `repeat` fails anyway — and it turns an `OutOfMemoryError` that escapes `catch (Exception)` into something a caller can diagnose. #1124 then replaces that ceiling with the dialect's `max_length`, refusing far earlier, with the check staying where you put it.

Worth keeping the check on the pad rather than only on the type: a mapper hands `characterLiteral` a `fixedChar(...)` directly, without passing through Calcite's type factory, so a type-system ceiling — dialect-driven or not — never sees that path.

alexandrefimov: Refused there, keyed to the largest array a JVM allocates: padding past that cannot work whatever the heap, which is the part that is not a policy choice. Measured to be sure the line is real — `" ".repeat(Integer.MAX_VALUE)` gives `Requested array size exceeds VM limit`, and everything below it fails as heap space instead, so the guard turns the first into an `IllegalArgumentException` naming the width and the value while leaving the second what it is.

On the pad rather than the type, as you say: `aWidthNoStringCanHoldIsRefusedRatherThanPadded` maps a `CHAR(1)` column to `fixedChar(Integer.MAX_VALUE)` and never reaches Calcite's type factory. Without the guard it dies with the `OutOfMemoryError`.

isthmus/src/main/java/io/substrait/isthmus/expression/LiteralConverter.java
```diff
@@ -78,6 +92,76 @@ private static String s(RexLiteral literal) {
...
+      return ExpressionCreator.varChar(nullable, value, ((Type.VarChar) type).length());
+    }
+    if (type instanceof Type.FixedChar) {
+      // A fixedchar literal carries no length of its own — Expression.FixedCharLiteral derives it
+      // from the text — so the text has to be the declared width or the two disagree. Padding is
+      // also what CHAR(n) means: 'a' in a CHAR(3) is 'a  '.
+      // In characters rather than UTF-16 code units: the spec gives a fixedchar its length in
+      // characters, where it spells a string's out in UTF-8 bytes.
+      int length = ((Type.FixedChar) type).length();
+      int characters = value.codePointCount(0, value.length());
```
nielspardon, line 148: Reject `length < 1` before the width check — the spec puts `L` in `[1..2147483647]`, and the two ends disagree today. `fixedchar<0>` is reachable with no mapper at all: Calcite types `''` as `CHAR(0)`, and 0 is not `PRECISION_NOT_SPECIFIED`, so the guard you added to `TypeConverter` passes it through and this pads by nothing and emits it. A negative width does throw, but reports "longer than the fixedchar<-5>", which is not the problem — one floor check closes both and the message; `RelDataTypeSystemImpl.getMinPrecision(CHAR)` is 1 as well, so Calcite's floor and the spec's agree.

```suggestion
      int length = ((Type.FixedChar) type).length();
      if (length < 1) {
        throw new IllegalArgumentException(
            String.format(
                Locale.ROOT,
                "A fixedchar must declare a width of 1 or more, not %d", length));
      }
      int characters = value.codePointCount(0, value.length());
```

isthmus/src/main/java/io/substrait/isthmus/expression/LiteralConverter.java
```diff
@@ -78,6 +92,76 @@ private static String s(RexLiteral literal) {
...
+                value,
+                length));
+      }
+      long padded = (long) value.length() + ((long) length - characters);
+      if (padded > MAX_PADDED_LENGTH) {
+        throw new IllegalArgumentException(
+            String.format(
+                Locale.ROOT,
+                "A fixedchar<%d> literal cannot be built from '%s': padding it to that width takes "
+                    + "%d characters, more than a Java String holds",
```
nielspardon: `padded` is code units, not characters — worth saying so, since a value with an astral character makes the two differ, which is the distinction the check above it just drew.

```suggestion
                "A fixedchar<%d> literal cannot be built from '%s': padding it to that width takes "
                    + "%d UTF-16 code units, more than a Java String holds",
```

nielspardon (commented):
Two left, both on the width guard.

nielspardon:
No need — leave the mapper unrestricted. `TypeConverter` consults it before its own switch, with the call site saying it does so because mappers "may re-use SqlTypeNames"; were the mapper restricted to user-defined types, that ordering would buy nothing, since such a type's `SqlTypeName` is not one the switch handles and would fall through to `default` either way. So mapping a Calcite built-in onto a Substrait built-in is the intended forward contract, and the fallback here is the right shape.

What lags is the reverse, which is now #1219: `toCalcite` takes only a `Type.UserDefined` and `ToRelDataType` consults it from that one `visit`, so a mapped built-in has no hook to come back through. #1190 covers the other literal families.

isthmus/src/main/java/io/substrait/isthmus/expression/LiteralConverter.java
alexandrefimov: Refused there. The zero end turned out to want the width's source fixed with it: Calcite types `''` as a `CHAR(0)` and its DDL parser takes a `CHAR(0)` column, so on `origin/main` (934a60e) `SELECT '' FROM t`, `WHERE '' = ''`, `CHAR_LENGTH('')` and `'' || 'x'` all convert to a `fixedchar<0>` today, and the floor check on its own stops all four converting.

So a `CHAR(0)` now converts to a `string`: the empty string is the only value that type holds, a string holds it exactly, and the narrowest legal fixedchar pads it to a space — which is also the answer `TypeConverter` already gives a varchar carrying no width of its own. The floor then fires where you aimed it, at a width a mapper supplies, and `fixedchar<-5>` reports the width rather than the value's length.

If you would rather keep the two apart, I can split the `CHAR(0)` half into its own PR.

isthmus/src/main/java/io/substrait/isthmus/expression/LiteralConverter.java
alexandrefimov: Taken as written.

alexandrefimov:
Left unrestricted, and the description carries that now rather than the question.

nielspardon:
Filed the floor as a spec issue — substrait-io/substrait#1199. `type_classes.md` puts `L` at 1 or more, but `dialect_schema.yaml` allows `max_length: 0` for the same three types, so the bound this PR keys to is stated inconsistently. Worth saying in the body that mapping a `CHAR(0)` to `string` is what the spec allows today rather than the settled answer, since #1199 may legalize a zero width instead.

One consequence the footer does not cover: an empty character type no longer round-trips. `CHAR(0)` becomes a `string`, and `ToRelDataType.visit(Type.Str)` gives back an unparameterised `VARCHAR`, where before it came back as a `CHAR(0)`. Worth a clause there, since the footer describes the forward conversion only.

nielspardon:
Drop the `CHAR(0)` → `string` commit and narrow the floor to `length < 0`, leaving `fixedchar<0>` as it was pending substrait-io/substrait#1199. That reverses my round-three ask: the floor was right about negatives and wrong to include zero.

`type_classes.md`'s `[1..2147483647]` is the only thing asserting the floor. `dialect_schema.yaml` allows `max_length: 0` for FIXED_CHAR independently of it, and Calcite produces `CHAR(0)` from the four SQL forms you found — so refusing it is a breaking change #1199 may reverse, and undoing it would take a second one. Dropping the mapping also drops two costs it carried: the lost round trip, and re-binding string functions onto their `str` impls, since 16 functions in `functions_string.yaml` have a `fixedchar<L1>` impl distinct from their `string` one.

The negatives half stands on its own — unambiguous under any reading of the spec, and it fixes a message that reported the value's length where the width was the problem. That also answers the split question: with the `CHAR(0)` half gone there is nothing left to split.

isthmus/src/main/java/io/substrait/isthmus/expression/LiteralConverter.java
alexandrefimov: Narrowed to `length < 0`, and the `CHAR(0)` commit is out. `SELECT '' FROM t` gives a `fixedchar<0>` again, round trip included, so the four SQL forms are back where they were. The message names the width rather than the value now -- "A fixedchar cannot declare a negative width, and this one is -5" -- and the test that covered both ends keeps the negative one.

Your 16 checks out: `functions_string.yaml` has 16 functions whose `fixedchar<L1>` impl is distinct from the `string` one, 17 if `substring` counts, which spells its width differently.

alexandrefimov:
The description no longer presents the zero width as settled -- it says the bound is #1199's and that a `fixedchar<0>` converts as it does today -- and the footer is down to the padding and the refusal, so the round-trip clause has nothing left to describe.

One thing to weigh for #1199: `dialect_schema.yaml`'s `max_length` is a dialect's ceiling rather than a width the spec permits, and its `minimum: 0` reads as "this dialect does not support the type" about as easily as "a zero width is legal". The four SQL forms look like the stronger half of the case to me -- Calcite emits `CHAR(0)` whatever `type_classes.md` says, so any consumer reading a plan isthmus produced has to hold one either way.

isthmus/src/main/java/io/substrait/isthmus/expression/LiteralConverter.java
```diff
@@ -78,6 +92,86 @@ private static String s(RexLiteral literal) {
...
+    }
+    if (type instanceof Type.FixedChar) {
+      // A fixedchar literal carries no length of its own — Expression.FixedCharLiteral derives it
+      // from the text — so the text has to be the declared width or the two disagree. Padding is
+      // also what CHAR(n) means: 'a' in a CHAR(3) is 'a  '.
+      // In characters rather than UTF-16 code units: the spec gives a fixedchar its length in
+      // characters, where it spells a string's out in UTF-8 bytes.
+      int length = ((Type.FixedChar) type).length();
+      // Only the negative end. The spec puts a fixedchar's width in [1..2147483647] (spec
+      // v0.101.0), but Calcite types the empty character literal as a CHAR(0) and its DDL parser
```
nielspardon: Drop the `(spec v0.101.0)` here — the bound is not new in that version, it goes back to spec #200 in 2022, so the marker reads as provenance it does not have and would need editing on every unrelated bump. Line 71 and `TypeConverter`'s `CHAR` comment already cite the same range without one, which is the form to match.

The marker is worth keeping where it earns it, for a rule that was introduced, changed or deprecated in a given version — `ParseToPojo`'s "added to the grammar in spec v0.92.1" and `FunctionMappings`' v0.101.0 signature both read that way.

alexandrefimov: Dropped, and it was the only one this PR added -- the javadoc above and `TypeConverter`'s `CHAR` comment already cite the range bare, so the file is consistent again. `FunctionMappings`' marker is the other kind and stays.

nielspardon (approved):
LGTM, thanks

PR #1181, opened 2026-08-27
fix(isthmus): keep the schema a CTAS or a created view declares

### LINKED ISSUE #1180: isthmus: a CTAS or a created view comes back named after its input's columns, not the schema it declares
`NamedWrite.tableSchema` and `NamedDdl.tableSchema` are dropped converting to Calcite: `handleCreateTableAs` builds `new CreateTable(namedWrite.getNames(), relNode)` and `visit(NamedDdl)` builds `new CreateView(namedDdl.getNames(), relNode)`, and neither node has anywhere to put a schema. Converting back, `handleCreateTable` and `handleCreateView` rebuild it from the input's row type, so the object a statement creates is described by whatever Calcite derived for the columns filling it.

A CTAS declaring `[total, doubled]` over a projection of two computed columns comes back as `[$f2, $f3]` — the names Calcite gives expressions that have none. The schema is rebuilt whole, so declared types go the same way, but the names are what a CTAS's schema adds over its input in the first place: nothing else in the relation carries them.

Both nodes are this repo's own (`io.substrait.isthmus.calcite.rel`), so carrying the declared schema on them is available. A node built from a SQL statement (`DdlSqlToRelConverter`) can keep taking the input's row type, which is what the schema means there.

Checked on `main` @ e07d093.

alexandrefimov:
`CreateTable` and `CreateView` now carry the schema of the object they create: the conversion fills it from the relation's `tableSchema` on the way in and reads it back on the way out, where it used to rebuild it from the input's row type. A CTAS declaring `[total, doubled]` over a projection of two computed columns came back as `[$f2, $f3]`.

The SQL path declares one too. `create table dst1 (total, doubled) as select intcol + 1, intcol + 2 from src1` recorded `[EXPR$0, EXPR$1]`, and a column list may give types as well as names, so `DdlSqlToRelConverter` now reads it: a column named without a type takes the one the query produces, a column declared with one takes what it declares, and anything else in the list -- a constraint -- is left alone. A statement without a column list keeps the query's own row type, which is what it means there.

A DDL node's row type is now the schema of the object it creates rather than its input's. That is what `AbstractDdlRel.deriveRecordType()` reports on the Substrait side, so the two agree and the root over such a node is named by the declared schema instead of by the query filling it. What `RelRoot.names` should name over a DDL root is not settled by the spec -- substrait-io/substrait#1189 has that boundary open.

A schema the input cannot fill is rejected where the node is built: it has to be a struct, and its leaf fields have to be the ones the input produces. Names and types are the statement's own -- `create table t (a bigint) as select intcol from src1` is a legal CTAS -- so neither is compared. `WriteRel.table_schema` requires that alignment and `DdlRel` is silent about it; substrait-io/substrait#1190 asks for the same wording there.

Default values on a `NamedDdl` are refused rather than dropped, the way `visit(NamedUpdate)` refuses an emit mapping it cannot apply: a Calcite `CreateView` has nowhere to put them, and the spec has `table_defaults` report a full list of them.

The two-argument constructors keep taking the input's row type as the schema, which is what it means for a node the SQL path builds without a column list. The schema is carried through `copy()` and named in the explain terms, so rewriting what fills the object leaves its schema alone and two nodes that differ only in it do not share a digest.

Closes #1180

### DIFF
--- a/isthmus/src/main/java/io/substrait/isthmus/SubstraitRelVisitor.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/SubstraitRelVisitor.java
@@ -889,11 +889,6 @@ public class SubstraitRelVisitor extends RelNodeVisitor<Rel, RuntimeException> {
         .collect(Collectors.toList());
   }
 
-  private NamedStruct getSchema(final RelNode queryRelRoot) {
-    final RelDataType rowType = queryRelRoot.getRowType();
-    return typeConverter.toNamedStruct(rowType);
-  }
-
   /**
    * Handles Calcite {@link CreateTable} as Substrait CTAS. (Create Table As Select)
    *
@@ -903,7 +898,7 @@ public class SubstraitRelVisitor extends RelNodeVisitor<Rel, RuntimeException> {
   public Rel handleCreateTable(CreateTable createTable) {
     RelNode input = createTable.getInput();
     Rel inputRel = apply(input);
-    NamedStruct schema = getSchema(input);
+    NamedStruct schema = typeConverter.toNamedStruct(createTable.getTableSchema());
     return NamedWrite.builder()
         .input(inputRel)
         .tableSchema(schema)
@@ -928,7 +923,7 @@ public class SubstraitRelVisitor extends RelNodeVisitor<Rel, RuntimeException> {
 
     return NamedDdl.builder()
         .viewDefinition(inputRel)
-        .tableSchema(getSchema(input))
+        .tableSchema(typeConverter.toNamedStruct(createView.getViewSchema()))
         .tableDefaults(defaults)
         .operation(AbstractDdlRel.DdlOp.CREATE)
         .object(AbstractDdlRel.DdlObject.VIEW)
--- a/isthmus/src/main/java/io/substrait/isthmus/calcite/rel/CreateTable.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/calcite/rel/CreateTable.java
@@ -22,7 +22,7 @@ public class CreateTable extends SingleRel {
       RelNode input) {
     super(cluster, traitSet, input);
     this.tableName = tableName;
-    this.tableSchema = tableSchema;
+    this.tableSchema = DdlSchemas.requireFilledBy(tableSchema, input, "table");
   }
 
   /**
@@ -48,6 +48,17 @@ public class CreateTable extends SingleRel {
   }
 
   /**
+   * Returns the schema of the table this node creates, which is what it produces: the input fills
+   * it, and its own columns are the ones the statement declares.
+   *
+   * @return the schema of the table this node creates
+   */
+  @Override
+  protected RelDataType deriveRowType() {
+    return tableSchema;
+  }
+
+  /**
    * Explains the node terms for plan output.
    *
    * @param pw plan writer
@@ -57,7 +68,7 @@ public class CreateTable extends SingleRel {
   public RelWriter explainTerms(RelWriter pw) {
     return super.explainTerms(pw)
         .item("tableName", getTableName())
-        .item("tableSchema", getTableSchema());
+        .item("tableSchema", getTableSchema().getFullTypeString());
   }
 
   /**
@@ -87,8 +98,9 @@ public class CreateTable extends SingleRel {
   }
 
   /**
-   * Returns the schema of the table to create. A CTAS declares it, and it is not the row type of
-   * the input: the two hold the same columns, but the names are the ones the statement gives them.
+   * Returns the schema of the table to create: the one the statement declares where this node was
+   * built with it, and the row type of the input otherwise. A declared schema holds the same
+   * columns as the input, under the names -- and the types -- the statement gives them.
    *
    * @return the schema of the table this node creates
    */
--- a/isthmus/src/main/java/io/substrait/isthmus/calcite/rel/CreateTable.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/calcite/rel/CreateTable.java
@@ -6,26 +6,45 @@ import org.apache.calcite.plan.RelTraitSet;
 import org.apache.calcite.rel.RelNode;
 import org.apache.calcite.rel.RelWriter;
 import org.apache.calcite.rel.SingleRel;
+import org.apache.calcite.rel.type.RelDataType;
 
 /** Represents a CREATE TABLE DDL operation in Calcite's relational algebra. */
 public class CreateTable extends SingleRel {
 
   private final List<String> tableName;
+  private final RelDataType tableSchema;
 
   private CreateTable(
-      RelOptCluster cluster, RelTraitSet traitSet, List<String> tableName, RelNode input) {
+      RelOptCluster cluster,
+      RelTraitSet traitSet,
+      List<String> tableName,
+      RelDataType tableSchema,
+      RelNode input) {
     super(cluster, traitSet, input);
     this.tableName = tableName;
+    this.tableSchema = tableSchema;
+  }
+
+  /**
+   * CreateTable Constructor, taking the row type of the input as the schema of the table to create.
+   *
+   * @param tableName tablename components
+   * @param input RelNode input
+   */
+  public CreateTable(List<String> tableName, RelNode input) {
+    this(input.getCluster(), input.getTraitSet(), tableName, input.getRowType(), input);
   }
 
   /**
    * CreateTable Constructor.
    *
    * @param tableName tablename components
+   * @param tableSchema the schema of the table to create, which the input fills but need not name
+   *     the same way
    * @param input RelNode input
    */
-  public CreateTable(List<String> tableName, RelNode input) {
-    this(input.getCluster(), input.getTraitSet(), tableName, input);
+  public CreateTable(List<String> tableName, RelDataType tableSchema, RelNode input) {
+    this(input.getCluster(), input.getTraitSet(), tableName, tableSchema, input);
   }
 
   /**
@@ -36,7 +55,9 @@ public class CreateTable extends SingleRel {
    */
   @Override
   public RelWriter explainTerms(RelWriter pw) {
-    return super.explainTerms(pw).item("tableName", getTableName());
+    return super.explainTerms(pw)
+        .item("tableName", getTableName())
+        .item("tableSchema", getTableSchema());
   }
 
   /**
@@ -53,7 +74,7 @@ public class CreateTable extends SingleRel {
       throw new IllegalArgumentException(
           "CreateTable requires exactly one input, but got " + inputs.size());
     }
-    return new CreateTable(getCluster(), traitSet, tableName, inputs.get(0));
+    return new CreateTable(getCluster(), traitSet, tableName, tableSchema, inputs.get(0));
   }
 
   /**
@@ -63,5 +84,15 @@ public class CreateTable extends SingleRel {
    */
   public List<String> getTableName() {
     return tableName;
+  }
+
+  /**
+   * Returns the schema of the table to create. A CTAS declares it, and it is not the row type of
+   * the input: the two hold the same columns, but the names are the ones the statement gives them.
+   *
+   * @return the schema of the table this node creates
+   */
+  public RelDataType getTableSchema() {
+    return tableSchema;
   }
 }
--- a/isthmus/src/main/java/io/substrait/isthmus/calcite/rel/CreateView.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/calcite/rel/CreateView.java
@@ -21,7 +21,7 @@ public class CreateView extends SingleRel {
       RelNode input) {
     super(cluster, traitSet, input);
     this.viewName = viewName;
-    this.viewSchema = viewSchema;
+    this.viewSchema = DdlSchemas.requireFilledBy(viewSchema, input, "view");
   }
 
   /**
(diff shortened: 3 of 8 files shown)


isthmus/src/main/java/io/substrait/isthmus/calcite/rel/CreateView.java
```diff
@@ -6,25 +6,44 @@
...
+
+  /**
+   * CreateView Constructor.
+   *
+   * @param viewName view name components
+   * @param viewSchema the schema of the view to create, which the definition fills but need not
+   *     name the same way
+   * @param input RelNode input
+   */
+  public CreateView(List<String> viewName, RelDataType viewSchema, RelNode input) {
```
nielspardon, line 45: Worth settling before this lands: `getRowType()` stays the definition's while `AbstractDdlRel.deriveRecordType()` is now the declared schema, so `RelRoot.of(node, kind)` hands `Plan.Root` the definition's names, or throws its names-count check when the two arities differ. The spec does not decide it for us — `DdlRel` has no `input`, no `output_mode`, and nothing tying `table_schema` to `view_definition`, so what `RelRoot.names` names over a DDL root is undefined and `deriveRecordType()` returning the declared schema is our reading. Overriding `deriveRowType()` here is one way out; worth raising on substrait-io/substrait#1189, which already has the `DdlRel`/`WriteRel` boundary open.

isthmus/src/main/java/io/substrait/isthmus/calcite/rel/CreateTable.java
```diff
@@ -6,26 +6,45 @@
...
+
+  /**
+   * CreateTable Constructor.
+   *
+   * @param tableName tablename components
+   * @param tableSchema the schema of the table to create, which the input fills but need not name
+   *     the same way
+   * @param input RelNode input
+   */
+  public CreateTable(List<String> tableName, RelDataType tableSchema, RelNode input) {
```
nielspardon, line 46: Reject a schema that isn't a struct, or whose leaf-field count differs from the input's — `copy` already throws for the input count a few lines up. `WriteRel.table_schema` requires that alignment ("must align with Rel input (e.g., number of leaf fields must match)") and `DdlRel` is silent, but a view's columns are filled by its definition the same way, so check both; clarification filed as substrait-io/substrait#1190. Leaf count only, not names or types — declaring names the input does not carry is the point of this PR, and `create table t (a bigint) as select intcol from src1` is a legal CTAS.

isthmus/src/main/java/io/substrait/isthmus/calcite/rel/CreateTable.java
```diff
@@ -36,7 +55,9 @@ public CreateTable(List<String> tableName, RelNode input) {
    */
   @Override
   public RelWriter explainTerms(RelWriter pw) {
-    return super.explainTerms(pw).item("tableName", getTableName());
+    return super.explainTerms(pw)
+        .item("tableName", getTableName())
+        .item("tableSchema", getTableSchema());
```
nielspardon:
```suggestion
        .item("tableSchema", getTableSchema().getFullTypeString());
```

The plain type renders without nullability, so two nodes differing only in a column's `NOT NULL` explain identically. Same at `CreateView.java:59`.

isthmus/src/main/java/io/substrait/isthmus/calcite/rel/CreateTable.java
```diff
@@ -64,4 +85,14 @@ public RelNode copy(RelTraitSet traitSet, List<RelNode> inputs) {
   public List<String> getTableName() {
     return tableName;
   }
+
+  /**
+   * Returns the schema of the table to create. A CTAS declares it, and it is not the row type of
```
nielspardon: Say the schema is declared when the 3-arg constructor is used and defaults to the input's row type otherwise — via the 2-arg constructor above it, "it is not the row type of the input" is false for every node the SQL path builds. Same wording at `CreateView.java:89`.

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelVisitor.java
```diff
@@ -928,7 +923,7 @@ public Rel handleCreateView(CreateView createView) {
 
     return NamedDdl.builder()
         .viewDefinition(inputRel)
-        .tableSchema(getSchema(input))
+        .tableSchema(typeConverter.toNamedStruct(createView.getViewSchema()))
         .tableDefaults(defaults)
```
nielspardon, line 927: Give `CreateView` a slot for `tableDefaults`, or throw the way `visit(NamedUpdate)` does for what Calcite cannot carry — an empty struct is not a stand-in for a table with columns, since `table_defaults` is specified to "report a full list of default values".

isthmus/src/test/java/io/substrait/isthmus/DdlRoundtripTest.java
```diff
@@ -26,4 +39,70 @@ void testCreateView() throws Exception {
...
+            .tableSchema(declaredSchema())
+            .operation(AbstractWriteRel.WriteOp.CTAS)
+            .createMode(AbstractWriteRel.CreateMode.REPLACE_IF_EXISTS)
+            .outputMode(AbstractWriteRel.OutputMode.NO_OUTPUT)
+            .build();
+
+    RelNode calcite = new SubstraitToCalcite(converterProvider, catalogReader).convert(ctas);
+    Rel converted = SubstraitRelVisitor.convert(copied(calcite), converterProvider);
+
+    assertEquals(declaredSchema(), ((NamedWrite) converted).getTableSchema());
```
nielspardon: Use the inherited `assertFullRoundTrip(Rel)` (`PlanTestBase.java:286`) — hand-rolling skips the proto leg and narrows the assertion to one field, so a regression in `names`, `createMode`, `outputMode`, `tableDefaults` or the input rel passes here.

isthmus/src/test/java/io/substrait/isthmus/DdlRoundtripTest.java
```diff
@@ -26,4 +39,70 @@ void testCreateView() throws Exception {
...
+  /**
+   * Copies the node the way a planner rewriting its input does, so that the schema has to be
+   * carried rather than derived when the conversion reads it back.
+   */
+  private RelNode copied(RelNode node) {
+    return node.copy(node.getTraitSet(), node.getInputs());
+  }
+
+  private NamedStruct declaredSchema() {
+    return NamedStruct.of(List.of("total", "doubled"), TypeCreator.REQUIRED.struct(N.I32, N.I32));
```
nielspardon: Declare a type the input cannot produce, so the assertion covers types and not only names:

```suggestion
    return NamedStruct.of(List.of("total", "doubled"), TypeCreator.REQUIRED.struct(R.I64, N.I32));
```

`REQUIRED.struct(N.I32, N.I32)` is exactly what `computedColumns()` derives. It is also the one shape that hides #1154 on this path, which the declared schema now reaches.

isthmus/src/test/java/io/substrait/isthmus/DdlRoundtripTest.java
```diff
@@ -26,4 +39,70 @@ void testCreateView() throws Exception {
...
+    Rel converted = SubstraitRelVisitor.convert(copied(calcite), converterProvider);
+
+    assertEquals(declaredSchema(), ((NamedDdl) converted).getTableSchema());
+  }
+
+  /**
+   * Copies the node the way a planner rewriting its input does, so that the schema has to be
+   * carried rather than derived when the conversion reads it back.
+   */
+  private RelNode copied(RelNode node) {
```
nielspardon: Move this into `DdlRelCopyTest`, built with the 3-arg constructor and copied onto a *different* input — `copy` here gets the same inputs back, and both tests pass without it.

nielspardon (changes requested):
Read the column list on the SQL path too — `DdlSqlToRelConverter` (lines 94 and 106) still calls the 2-arg constructors and drops `sqlCreateTable.columnList` / `sqlCreateView.columnList`, so `create table dst1 (total, doubled) as select intcol + 1, intcol + 2 from src1` still records `[EXPR$0, EXPR$1]` and #1180 stays open for `SQL -> Substrait`. Declared types go the same way, since a `TableElement` can be a full `SqlColumnDeclaration`, and `convertQuery(...)`'s `RelRoot` is what carries a bare `select intcol as total`.

Keep the construction-time snapshot as it is: a rule may rename the input's fields but not change their count or types, so the frozen names are the declared ones.

isthmus/src/main/java/io/substrait/isthmus/calcite/rel/CreateView.java
alexandrefimov: Overrode `deriveRowType()` in both nodes, so a DDL relation produces the object it creates and the root over it is named by the declared schema. `theRootOverACreateStatementIsNamedByTheDeclaredSchema` pins it: without the override the root of `create table dst1 (total, doubled) as ...` comes back named `[EXPR$0, EXPR$1]`.

Agreed the spec does not settle what `RelRoot.names` names over a DDL root; nothing to add to substrait-io/substrait#1189 beyond what you wrote there, and the PR description now says this is our reading.

isthmus/src/main/java/io/substrait/isthmus/calcite/rel/CreateTable.java
alexandrefimov: Added, in a check the two nodes share: the schema has to be a struct, and its leaf fields have to be the ones the input produces. Names and types are not compared -- `create table t (a bigint) as select intcol from src1` is the case you name, and this PR is about declaring both. `aSchemaTheInputCannotFillIsRejected` covers a non-struct and a schema of the wrong width, in `DdlRelCopyTest` alongside the input-count checks.

isthmus/src/main/java/io/substrait/isthmus/calcite/rel/CreateTable.java
alexandrefimov: Applied as suggested, in both nodes.

isthmus/src/main/java/io/substrait/isthmus/calcite/rel/CreateTable.java
alexandrefimov: Reworded in both: the schema the statement declares where the node was built with one, and the row type of the input otherwise.

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelVisitor.java
alexandrefimov: Took the second option: `visit(NamedDdl)` now throws where `table_defaults` has any field, so a plan carrying defaults is refused instead of converted into one that has lost them. `aViewWithDefaultValuesIsRefused` pins it.

A slot would mean carrying Substrait literals on a Calcite node, and nothing on the SQL path fills them, so there would be no way to produce one except by reading a plan -- refusing says the same thing without the machinery. Happy to add the slot instead if you would rather have the round trip.

isthmus/src/test/java/io/substrait/isthmus/DdlRoundtripTest.java
alexandrefimov: Used it for the CTAS. The view cannot: `assertFullRoundTrip(Rel)` converts without a catalog, so the schema is collected by walking the plan with `RelCopyOnWriteVisitor`, whose `visit(NamedDdl)` throws `UnsupportedOperationException`. That test keeps the catalog-backed conversion and says why. Leaving this open -- the gap is in core, and it is the reason a `NamedDdl` cannot be round-tripped by the shared helper at all.

isthmus/src/test/java/io/substrait/isthmus/DdlRoundtripTest.java
alexandrefimov: Applied as suggested. The CTAS test now goes through `assertFullRoundTrip(Rel)` as well, so the declared types are compared along with everything else the relation carries.

isthmus/src/test/java/io/substrait/isthmus/DdlRoundtripTest.java
alexandrefimov: Moved: `aDeclaredSchemaSurvivesACopyOntoAnotherInput` in `DdlRelCopyTest`, built with the three-argument constructor and copied onto a different input, for both nodes.

One correction to "both tests pass without it": they did not. Mutating `copy()` to re-derive the schema from `inputs.get(0)` failed `createTableKeepsTheSchemaItDeclares`, since the same input's row type is `[$f2, $f3]` and not the declared names. Copying onto a different input is the stronger test either way, and it belongs where you put it.

isthmus/src/test/java/io/substrait/isthmus/DdlRoundtripTest.java
nielspardon: Filed as #1185 — it is four relations rather than just `NamedDdl` (`Expand`, `ExtensionWrite`, `ExtensionDdl` too), and `OuterReferenceConverter.RelRewriter` already overrides the same throw for `NamedDdl`, which is why the catalog-backed path converts at all while `SchemaCollector.TableGatherer` does not. Resolving here: the CTAS half is done and the view half is tracked there.

nielspardon (approved):
All addressed. The remaining half of the round-trip helper thread is #1185.

Issue #1178, opened 2026-08-27
isthmus: two depth-first schema-name walkers disagree on which type kinds can contain a struct


nielspardon:
There are two depth-first walkers that apply a `NamedStruct`'s flattened names, in the same package, and they cover different type kinds.

`SubstraitToCalcite.renameFields` walks *types*, and handles `ROW`/`STRUCTURED`, `ARRAY`/`MULTISET`, and `MAP`:

https://github.com/substrait-io/substrait-java/blob/main/isthmus/src/main/java/io/substrait/isthmus/SubstraitToCalcite.java#L139

`SubstraitRelNodeConverter.renamedAsDeclared` (added by #1153) walks *values* through the same shape, but its switch covers only `ROW`, `ARRAY` and `MAP`. A schema reaching a struct through a `STRUCTURED` or `MULTISET` type — which a `UserTypeMapper` can produce, since it may map a user-defined type to any `RelDataType` — is renamed by one walker and not the other, and any struct beneath it keeps Calcite's placeholder names. The two also disagree on nullability: `renameFields` rebuilds through `createArrayType`/`createMapType`/`createStructType` and drops it, while the value walker preserves it.

Adding a nested type kind currently means finding and editing both. `renameFields` also looks redundant: `TypeConverter.toCalcite(RelDataTypeFactory, TypeExpression, List<String>)` already threads the flattened name list through nested structs, which is what let #1153 replace the equivalent patch-up in `visit(VirtualTableScan)`. Replacing `renameFields` with that overload would leave one implementation.

Related but distinct: #1170 (a `UserTypeMapper` is dropped from literals) and #1154 (a nullable struct loses its fields' non-nullability) both touch this area.

nielspardon:
#1158 adds a third walker of this convention, so the count is now three rather than two: `SubstraitRelNodeConverter.applyOutputNames` slices a flat depth-first name list with a hand-rolled `nameIndex` over `NamedFieldCountingTypeVisitor.countNames`, for `hint.output_names`.

It differs from both of the walkers above in reach, deliberately: it applies only the top-level names, because a Calcite projection cannot restate the nested names of the expression producing a column. That is the right call there, but it means one name list now produces two different results depending on which field carried it — nested names applied via `Plan.Root.names`, top level only via a hint.

[...]

PR #1188, opened 2026-08-28
fix(isthmus): refuse a created object's schema that names two columns the same

### LINKED ISSUE #1187: isthmus: the SQL path drops the column list a CREATE TABLE ... AS SELECT or CREATE VIEW declares
`DdlSqlToRelConverter.handleCreateTable` and `handleCreateView` build `new CreateTable(sqlCreateTable.name.names, input)` and `new CreateView(sqlCreateView.name.names, input)`, ignoring `sqlCreateTable.columnList` / `sqlCreateView.columnList`. The object is therefore described by whatever the query produced, so

```sql
create table dst1 (total, doubled) as select intcol + 1, intcol + 2 from src1
```

records `[EXPR$0, EXPR$1]` — the names Calcite gives expressions that have none. A CTAS may also declare per-column *types*, parsed into `SqlColumnDeclaration.dataType`, and those go the same way.

**Why #1181 did not cover this.** #1180 scoped the SQL path out deliberately: "A node built from a SQL statement (`DdlSqlToRelConverter`) can keep taking the input's row type, which is what the schema means there", and #1181 kept the two-argument constructors on that basis. That assumption holds only where the statement declares no column list — SQL allows one, and Calcite parses it, so the input's row type is not the declared schema whenever it is present.

**A fix should also reject duplicate declared names.** Building the schema with `RelDataTypeFactory.createStructType(types, names)` does not uniquify, so

```sql
create table d (a, a) as select intcol, intcol from src1
```

yields `RecordType(INTEGER A, INTEGER A) NOT NULL`. Calcite requires a projection's field names to be distinct but only checks it inside `Project.isValid`, behind an `assert`, so a duplicate survives into the row type wherever assertions are off. #1158 rejects duplicates in `applyOutputNames` for this same reason and is a useful precedent for the shape of the check.

Found while reviewing #1158, which carries a fix for this in `8121d198`; splitting it out so that PR stays scoped to `hint.output_names`.

alexandrefimov:
`RelDataTypeFactory.createStructType` does not uniquify the names it is given, so `create table d (a, a) as select intcol, intcol from src1` built a row type holding two fields named `A`, and the schema recorded for the table read `[A, A]`. A plan whose `tableSchema` repeats a name went the same way. Calcite reads a row type as a scope and checks distinctness only inside `Project.isValid`, behind an `assert`, so the duplicate survives wherever assertions are off.

The check sits with the other checks on a declared schema rather than in the SQL path, so it covers a schema read from a plan as well as one a statement declares, and it recurses: every struct level is a scope of its own, so a struct column repeating a field name is refused too.

Closes #1187

### DIFF
--- a/isthmus/src/main/java/io/substrait/isthmus/calcite/rel/DdlSchemas.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/calcite/rel/DdlSchemas.java
@@ -2,6 +2,7 @@ package io.substrait.isthmus.calcite.rel;
 
 import java.util.HashSet;
 import java.util.List;
+import java.util.Objects;
 import org.apache.calcite.rel.RelNode;
 import org.apache.calcite.rel.type.RelDataType;
 import org.apache.calcite.rel.type.RelDataTypeField;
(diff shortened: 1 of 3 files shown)


isthmus/src/main/java/io/substrait/isthmus/calcite/rel/DdlSchemas.java
```diff
@@ -37,9 +40,32 @@ static RelDataType requireFilledBy(RelDataType schema, RelNode input, String wha
...
+  /**
+   * Checks that no two fields of a struct are named the same, at every level of it: Calcite reads a
+   * row type as a scope, and a repeated name there is one a reference cannot reach.
+   *
+   * @param type the type to check
+   * @param what the object being created, for the failure message
+   * @throws IllegalArgumentException if two fields of one struct share a name
+   */
+  private static void requireDistinctNames(RelDataType type, String what) {
+    if (!type.isStruct()) {
```
nielspardon: Recurse through collection types too — `requireDistinctNames` returns on any non-struct, so a struct reached through an ARRAY or MAP is never checked, though the Javadoc promises every level. Verified on this branch: a column of `array<row<a, a>>` and one of `map<int, row<a, a>>` are both accepted, while that same `row<a, a>` as a column is rejected. This is #1178's territory — a third walker with its own view of which kinds can hold a struct — so if you'd rather leave it there, "at every level of it" should say ROW instead.

```suggestion
    RelDataType component = type.getComponentType();
    if (component != null) {
      requireDistinctNames(component, what);
    }
    if (type.getKeyType() != null) {
      requireDistinctNames(type.getKeyType(), what);
      requireDistinctNames(type.getValueType(), what);
    }
    if (!type.isStruct()) {
```

nielspardon (changes requested):
One gap in the new check: it does not reach a struct nested under a collection, though the Javadoc says every level. Otherwise this reads well — moving the check into `DdlSchemas` so it covers a schema read from a plan, not just one a statement declares, is broader than what I raised on #1158.

isthmus/src/main/java/io/substrait/isthmus/calcite/rel/DdlSchemas.java
alexandrefimov: Recursed, through the component of an ARRAY or MULTISET and through both halves of a MAP. What decided it against narrowing the Javadoc is the name convention itself: `NamedFieldCountingTypeVisitor.countNames` counts two names for `list<struct<i32,i64>>`, the same as for the bare struct, so a schema does name those fields and a check that skips them disagrees with the list it is checking. `aSchemaThatRepeatsANameInsideACollectionIsRejected` covers your two cases.

It does add a third walker with its own view of which kinds hold a struct, which is #1178's point. This one is the smallest of the three -- it names nothing and rebuilds nothing -- so folding it into whatever #1178 settles on should be easier than leaving the gap open until then.

nielspardon (approved):
Verified: every collection kind is covered now, and the switch matches what `SubstraitToCalcite.renameFields` walks rather than adding a third view of which kinds hold a struct.

Issue #1186, opened 2026-08-28
isthmus: visit(NamedWrite) converts a CTAS input twice, so a CTAS over a correlated subquery fails with a duplicate rel_anchor


nielspardon:
`SubstraitRelNodeConverter.visit(NamedWrite)` converts the input before it looks at the write operation, and the CTAS branch then converts that same input a second time through the same `Context`. The first `RelNode` is dead on that path.

https://github.com/substrait-io/substrait-java/blob/main/isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java#L1031-L1045

That is not only wasted work, because `Context` is not idempotent. `enterScope` records each input's `rel_anchor` in a plan-wide `seenAnchors` set and rejects a repeat:

https://github.com/substrait-io/substrait-java/blob/main/isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java#L1101-L1119

`SubstraitToCalcite.convert(Rel)` runs `OuterReferenceConverter.toIdBased` first, which stamps anchors on the relations that bind correlated references, so any CTAS whose input contains a correlated subquery converts the anchored relation twice and dies on the second pass.

## Reproduction

Wrapping a correlated `SELECT` in a CTAS `NamedWrite` (`REPLACE_IF_EXISTS`, `NO_OUTPUT`) and converting it, against `TPCH_CATALOG`:

```
select l_partkey from lineitem where exists
  (select o_orderdate from orders where o_orderkey = l_orderkey)
```

```
=== converting CTAS over correlated subquery ===
RESULT: java.lang.UnsupportedOperationException: Duplicate rel_anchor=1; rel_anchors must be unique plan-wide for id-based outer references to resolve unambiguously
=== control: the same SELECT without the CTAS wrapper ===
CONTROL: converted with no error
```

The control is the point: the identical relation converts cleanly on its own, so the CTAS wrapper is what breaks it, and the message blames the plan for something the conversion did.

Reproduced on `main` at 9000232d.

## Fix

Dispatch on the write operation before converting the input, so `case CTAS` returns `handleCreateTableAs(write, context)` without a conversion having happened yet. That also drops the dead `RelNode` and the `requireRelOptSchema()` / `getTableForMember()` lookup, which currently runs for a CTAS to find a table the statement is about to create.

[...]

PR #1189, opened 2026-08-28
fix(isthmus)!: apply the emit mapping a virtual table carries

### LINKED ISSUE #1160: isthmus: visit(VirtualTableScan) ignores the emit mapping
`SubstraitRelNodeConverter.visit(VirtualTableScan)` returns its `LogicalValues` (or the projection built for non-literal rows) directly. It is the only `visit` that never calls `applyRemap`, so a virtual table with an emit mapping is converted as if it emitted every column.

A virtual table over `(a i64, b string)` with one row and `emit = [1]` declares `[string]`:

```
VirtualTableScan(schema=[a, b], rows=[[1, 'x']], remap=[1])
```

and converts to both columns:

```
LogicalValues(tuples=[[{ 1, 'x' }]])

RecordType(BIGINT NOT NULL a, VARCHAR NOT NULL b)
```

A parent relation resolves its field references positionally, so it reads `a` where the plan says `b`, and the count of columns it sees does not match the one it was built against.

Reproduced on `main` at 4242051. The method is being rewritten in #1151 and #1153; neither routes the mapping, so this fits on top of whichever lands first. Found while reviewing #1158.

alexandrefimov:
`visit(VirtualTableScan)` routed neither half of the relation's `RelCommon`: a scan whose emit mapping drops a column converted carrying all of them, under a row type that reported one, and the names of its hint were dropped with it. It goes through `applyRelCommon` now, like every relation whose conversion applies a mapping.

A `NamedWrite` and a `NamedDdl` cannot apply one, for reasons that are isthmus' rather than the relation's. A write's record type is its input's, but isthmus converts it to a `TableModify` whose row type is a single ROWCOUNT column, dropping the write's `output_mode` along the way, so the mapping has nothing left to select from; and there is no place for a projection between a `CreateView`'s definition and the view it creates. `visit(NamedUpdate)` already refuses one, and these two now do the same instead of dropping it.

A scan carrying a `projection` is refused as well. `AbstractReadRel.deriveRecordType()` masks the initial schema with it before anything else, so an emit mapping's indices count the columns the mask leaves -- while isthmus builds the row type from the unmasked schema and reads the projection nowhere. `ProtoRelConverter` fills the field from the proto, so such a relation reaches the conversion.

The mapping selects its columns by index rather than by name. A virtual table's row type carries the names of its schema, which Calcite never uniquifies, so resolving a field by name gave the projection the type of the first column sharing it: an `AssertionError` under `-ea`, a silently mistyped plan without one.

A mapping is materialised as a projection, so a scan carrying one comes back from Calcite as a projection over a virtual table rather than as the scan it went in as -- the shape every relation with a mapping returns as. Without a mapping there is no projection for the hint's names to land on, and the table keeps the names its schema gives it.

Closes #1160

BREAKING CHANGE: a NamedWrite or a NamedDdl carrying an emit mapping, and a VirtualTableScan carrying a projection, no longer convert to Calcite. They used to convert with the mapping or the projection dropped, which produced a plan that did not describe the relation it came from.

### DIFF
--- a/isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
@@ -789,6 +789,12 @@ public class SubstraitRelNodeConverter
               namedDdl.getObject()));
     }
 
+    if (namedDdl.getRemap().isPresent()) {
+      throw new UnsupportedOperationException(
+          "Emit mapping on a NamedDdl is not supported: a CreateView produces the view it creates, "
+              + "not columns to select from");
+    }
+
     if (namedDdl.getViewDefinition().isEmpty()) {
       throw new IllegalArgumentException("NamedDdl view definition must be set");
     }
@@ -855,7 +861,9 @@ public class SubstraitRelNodeConverter
         }
         tuplesBuilder.add(tupleBuilder.build());
       }
-      return LogicalValues.create(relBuilder.getCluster(), rowType, tuplesBuilder.build());
+      return applyRelCommon(
+          LogicalValues.create(relBuilder.getCluster(), rowType, tuplesBuilder.build()),
+          virtualTableScan);
     } else {
       // A row that does not fit a LogicalValues tuple is computed instead: we create a
       // LogicalProject for each row to compute its values, and combine them together using a
@@ -892,8 +900,10 @@ public class SubstraitRelNodeConverter
       for (int i = 0; i < rowType.getFieldCount(); i++) {
         topProjectExprs.add(rexBuilder.makeInputRef(union, i));
       }
-      return LogicalProject.create(
-          union, Collections.emptyList(), topProjectExprs, rowType, Collections.emptySet());
+      return applyRelCommon(
+          LogicalProject.create(
+              union, Collections.emptyList(), topProjectExprs, rowType, Collections.emptySet()),
+          virtualTableScan);
     }
   }
 
@@ -1052,6 +1062,12 @@ public class SubstraitRelNodeConverter
 
   @Override
   public RelNode visit(NamedWrite write, Context context) {
+    if (write.getRemap().isPresent()) {
+      throw new UnsupportedOperationException(
+          "Emit mapping on a NamedWrite is not supported: a TableModify's row type is a single "
+              + "ROWCOUNT column and a CreateTable produces the table it creates, neither of them "
+              + "columns to select from");
+    }
     RelNode input = write.getInput().accept(this, context);
     final RelOptSchema relOptSchema = requireRelOptSchema();
     final RelOptTable targetTable = relOptSchema.getTableForMember(write.getNames());
--- a/isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
@@ -81,7 +81,6 @@ import org.apache.calcite.rel.logical.LogicalUnion;
 import org.apache.calcite.rel.logical.LogicalValues;
 import org.apache.calcite.rel.type.RelDataType;
 import org.apache.calcite.rel.type.RelDataTypeFactory;
-import org.apache.calcite.rel.type.RelDataTypeField;
 import org.apache.calcite.rex.RexBuilder;
 import org.apache.calcite.rex.RexCall;
 import org.apache.calcite.rex.RexInputRef;
@@ -789,14 +788,14 @@ public class SubstraitRelNodeConverter
               namedDdl.getObject()));
     }
 
-    if (namedDdl.getRemap().isPresent()) {
-      throw new UnsupportedOperationException(
-          "Emit mapping on a NamedDdl is not supported: a CreateView produces the view it creates, "
-              + "not columns to select from");
-    }
-
     if (namedDdl.getViewDefinition().isEmpty()) {
       throw new IllegalArgumentException("NamedDdl view definition must be set");
+    }
+
+    if (namedDdl.getRemap().isPresent()) {
+      throw new UnsupportedOperationException(
+          "Emit mapping on a NamedDdl is not supported: isthmus has no place for a projection "
+              + "between a CreateView's definition and the view it creates");
     }
 
     if (!namedDdl.getTableDefaults().fields().isEmpty()) {
@@ -812,6 +811,11 @@ public class SubstraitRelNodeConverter
 
   @Override
   public RelNode visit(VirtualTableScan virtualTableScan, Context context) {
+    if (virtualTableScan.getProjection().isPresent()) {
+      throw new UnsupportedOperationException(
+          "Projection on a VirtualTableScan is not supported: its columns would have to be "
+              + "masked before an emit mapping selects from them");
+    }
     // A schema's names are one per field at every level of the struct, in depth-first order, so
     // they have to be handed to the conversion rather than paired with the row type afterwards:
     // with a nested struct anywhere in the schema the two lists do not even have the same length.
@@ -900,10 +904,10 @@ public class SubstraitRelNodeConverter
       for (int i = 0; i < rowType.getFieldCount(); i++) {
         topProjectExprs.add(rexBuilder.makeInputRef(union, i));
       }
-      return applyRelCommon(
+      RelNode topProject =
           LogicalProject.create(
-              union, Collections.emptyList(), topProjectExprs, rowType, Collections.emptySet()),
-          virtualTableScan);
+              union, Collections.emptyList(), topProjectExprs, rowType, Collections.emptySet());
+      return applyRelCommon(topProject, virtualTableScan, topProject);
     }
   }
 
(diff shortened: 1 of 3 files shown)


isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
```diff
@@ -855,7 +861,9 @@ public RelNode visit(VirtualTableScan virtualTableScan, Context context) {
         }
         tuplesBuilder.add(tupleBuilder.build());
       }
-      return LogicalValues.create(relBuilder.getCluster(), rowType, tuplesBuilder.build());
+      return applyRelCommon(
```
nielspardon, line 868: Refuse a scan carrying a `projection` at the top of this method, or mask its columns before the mapping selects from them. `AbstractReadRel.deriveRecordType()` masks first, so the emit indices are defined against the projected columns, while `rowType` here is the full `initialSchema` -- a scan declaring `[b string]` converts to `a i32`. `ProtoRelConverter.newVirtualTable` fills `projection` from the proto and isthmus reads it nowhere, so before this the mapping had nothing to be wrong about.

```java
if (virtualTableScan.getProjection().isPresent()) {
      "Projection on a VirtualTableScan is not supported: its columns would have to be "
          + "masked before an emit mapping selects from them");
}
```

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
```diff
@@ -892,8 +900,10 @@ public RelNode visit(VirtualTableScan virtualTableScan, Context context) {
       for (int i = 0; i < rowType.getFieldCount(); i++) {
         topProjectExprs.add(rexBuilder.makeInputRef(union, i));
       }
-      return LogicalProject.create(
-          union, Collections.emptyList(), topProjectExprs, rowType, Collections.emptySet());
+      return applyRelCommon(
+          LogicalProject.create(
+              union, Collections.emptyList(), topProjectExprs, rowType, Collections.emptySet()),
+          virtualTableScan);
```
nielspardon: Pass the pre-remap node as an input so the names land only on a projection the mapping added. Without it this branch's own top projection is a `core.Project` with nothing to veto against, so a hint takes effect here but not on the literal branch at 864 -- which is what `outputNamesWithoutAMappingAreLeftAlone` pins.

```suggestion
      RelNode topProject =
          LogicalProject.create(
              union, Collections.emptyList(), topProjectExprs, rowType, Collections.emptySet());
      return applyRelCommon(topProject, virtualTableScan, topProject);
```

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
```diff
@@ -855,7 +861,9 @@ public RelNode visit(VirtualTableScan virtualTableScan, Context context) {
         }
         tuplesBuilder.add(tupleBuilder.build());
       }
-      return LogicalValues.create(relBuilder.getCluster(), rowType, tuplesBuilder.build());
+      return applyRelCommon(
+          LogicalValues.create(relBuilder.getCluster(), rowType, tuplesBuilder.build()),
+          virtualTableScan);
```
nielspardon, line 870: `applyRemap` resolves each field's type by name at line 1313 while referencing it by index -- use `rowType.getFieldList().get(index).getType()`. A virtual table's row type is the one Calcite never uniquifies, since it comes straight from `NamedStruct`, so two columns sharing a name give a `RexInputRef` the wrong type: `AssertionError` under `-ea`, a silently mistyped plan without it.

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
```diff
@@ -1052,6 +1062,12 @@ private RelNode handleCreateTableAs(NamedWrite namedWrite, Context context) {
 
   @Override
   public RelNode visit(NamedWrite write, Context context) {
+    if (write.getRemap().isPresent()) {
+      throw new UnsupportedOperationException(
+          "Emit mapping on a NamedWrite is not supported: a TableModify's row type is a single "
+              + "ROWCOUNT column and a CreateTable produces the table it creates, neither of them "
+              + "columns to select from");
```
nielspardon: Name what actually blocks it -- isthmus drops `output_mode`, so nothing carries the columns a mapping would select. `OUTPUT_MODE_MODIFIED_RECORDS` has the write return the modified records and `AbstractWriteRel.deriveRecordType()` returns its input's, so the ROWCOUNT row type is Calcite's limit rather than a property of a `NamedWrite`.

```suggestion
          "Emit mapping on a NamedWrite is not supported: isthmus converts a write to a "
              + "TableModify whose row type is a single ROWCOUNT column, dropping the write's "
              + "output_mode, so the mapping has no columns to select from");
```

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
```diff
@@ -789,6 +789,12 @@ public RelNode visit(NamedDdl namedDdl, Context context) {
               namedDdl.getObject()));
     }
 
+    if (namedDdl.getRemap().isPresent()) {
+      throw new UnsupportedOperationException(
+          "Emit mapping on a NamedDdl is not supported: a CreateView produces the view it creates, "
+              + "not columns to select from");
```
nielspardon: Move this below the `viewDefinition` check so a relation missing its definition still reports that, and drop the claim about what a `CreateView` produces -- `AbstractDdlRel.deriveRecordType()` returns the declared schema and `CreateView.deriveRowType()` the view's. The spec says nothing either way here, which is worth saying rather than filling in.

```suggestion
          "Emit mapping on a NamedDdl is not supported: isthmus has no place for a projection "
              + "between a CreateView's definition and the view it creates");
```

isthmus/src/test/java/io/substrait/isthmus/DdlRoundtripTest.java
```diff
@@ -163,6 +163,40 @@ void theRootOverACreateStatementIsNamedByTheDeclaredSchema() throws SqlParseExce
...
+            .tableSchema(declaredSchema())
+            .tableDefaults(ExpressionCreator.struct(false))
+            .operation(AbstractDdlRel.DdlOp.CREATE)
+            .object(AbstractDdlRel.DdlObject.VIEW)
+            .remap(Rel.Remap.of(List.of(0)))
+            .build();
+    SubstraitToCalcite converter = new SubstraitToCalcite(converterProvider, catalogReader);
+
+    assertThrows(UnsupportedOperationException.class, () -> converter.convert(ctas));
+    assertThrows(UnsupportedOperationException.class, () -> converter.convert(createView));
```
nielspardon: Assert the message, as `CatalogReaderValidationTest.namedUpdateRejectsRemap` does, and report both rather than letting the first failure hide the second (needs static imports for `assertAll` and `assertTrue`).

```suggestion
    assertAll(
        () ->
            assertTrue(
                assertThrows(UnsupportedOperationException.class, () -> converter.convert(ctas))
                    .getMessage()
                    .contains("Emit mapping on a NamedWrite is not supported")),
        () ->
            assertTrue(
                assertThrows(
                        UnsupportedOperationException.class,
                        () -> converter.convert(createView))
                    .getMessage()
                    .contains("Emit mapping on a NamedDdl is not supported")));
```

nielspardon (commented):
One first: the mapping is applied against the wrong basis when the scan also carries a projection. The rest are smaller -- a type lookup by name, the hint diverging between the two encodings, and the two refusal messages. #1186 already tracks the CTAS input three lines below the new write guard being converted twice.

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
alexandrefimov: Refused, and pinned with a scan whose projection selects its second column. Took the refusal over masking here because isthmus reads `projection` on no read relation — the field is filled by `ProtoRelConverter` and consumed nowhere — so masking it for virtual tables alone would close a corner of what #1204 tracks while leaving the rest as it is.

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
alexandrefimov: Applied as suggested, with the computed-table twin of `outputNamesWithoutAMappingAreLeftAlone` to hold it: without the pre-remap input that plan comes back named `[x, y]` while the literal branch leaves `[col1, col2]`.

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
alexandrefimov: Fixed to `rowType.getFieldList().get(index).getType()`. `anEmitMappingSelectsByIndexWhereTwoColumnsShareAName` gives a table two columns called `c`; with the by-name lookup it fails with `AssertionError: type mismatch: ref: INTEGER NOT NULL`, which is your -ea case.

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
alexandrefimov: Taken as suggested. `AbstractWriteRel.deriveRecordType()` returning the input's is also what the description had wrong, so that paragraph is rewritten too.

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
alexandrefimov: Moved below the definition check and reworded as suggested.

isthmus/src/test/java/io/substrait/isthmus/DdlRoundtripTest.java
alexandrefimov: Applied as suggested, and the test's Javadoc no longer repeats the claim the message dropped.

alexandrefimov:
One thing to settle before either of these lands: #1151 replaces this method with an emitter for an isthmus-owned `VirtualTable`, so whichever goes first, the other's changes to it have to be redone. My preference is this one first — three files against nine, and the emit handling then moves onto the new shape with the projection guard and the by-index selection already settled. Happy to fold it into #1151 instead if you would rather have one review.

nielspardon:
This one first, agreed — #1151 can rebase onto it with the projection guard and the by-index selection already settled. Keeping them as two reviews.

nielspardon (approved):
LGTM

Issue #1198, opened 2026-08-31
core: RANGE window frames accept sorts and offset types the spec disallows


nielspardon:
## Description

Spec v0.102.0 states three rules for a RANGE window frame whose bounds are `Preceding`/`Following`. `core` enforces one of them (`bounds_type` must be set — added in #1156 as a `@Value.Check` on both invocation types) and none of the other two. Plans that violate them are constructible through the POJO builders and round-trip through proto unchallenged.

From `algebra.proto` at v0.102.0, on `bounds_type` (the same paragraph appears on `Expression.WindowFunction` and on `ConsistentPartitionWindowRel.WindowRelFunction`):

> For RANGE with a Preceding or Following bound, `sorts` on the enclosing ConsistentPartitionWindowRel must contain exactly one ordering expression and must not use SORT_DIRECTION_CLUSTERED or a custom comparison function.

`BOUNDS_TYPE_RANGE`'s own comment repeats the first half — "This bounds type requires exactly one ordering expression if Preceding or Following is used" — and the `offset_expr` comment adds the type rule:

> Its type D must be compatible with the type T of the single ordering expression, i.e. `add(T, D) -> T` and `subtract(T, D) -> T` must be defined (e.g. timestamp/interval_day, decimal/decimal, i64/i64).

## What core does today

Nothing checks any of the three. Concretely:

1. **Exactly one ordering expression.** A RANGE frame with zero sorts or with two is accepted. Both shapes were in #1156's own test fixtures before review caught them, which is a fair sign of how easily this is missed.
2. **Sort direction / comparison function.** `SortDirection` includes `CLUSTERED`, and nothing excludes it from a RANGE frame's sorts. The "custom comparison function" half is currently moot for a different reason: `Expression.SortField` models only `expr()` and `direction()`, so the proto's `comparison_function_reference` has no POJO representation at all — worth a separate issue if that's news.
3. **Offset/ordering type compatibility.** `add(T, D) -> T` is unchecked, so an i64 offset against an i32 or decimal ordering column is accepted. isthmus's producer side now derives the offset type from the ordering column (#1156), but a plan built directly, or read from proto, has no such guarantee.

## Suggested shape

The two window kinds need the check in different places, which is the interesting part:

- `Expression.WindowFunctionInvocation` carries `sort()`, the bounds, and `boundsType()` itself, so its existing `@Value.Check` can cover this.
[...]

PR #1205, opened 2026-08-31
fix(core)!: require exactly one ordering expression for RANGE window bounds

anasik:
- Spec v0.102.0 states this rule at site/docs/expressions/window_functions.md:34, and in algebra.proto at WindowRelFunction.bounds_type (:457) and Expression.WindowFunction.sorts (:1401)        
- Requires exactly one, non-CLUSTERED ordering expression for a RANGE bound with a Preceding/Following side
- Fixed 4 pre-existing test fixtures that were building this invalid shape      
- Adds a sorts-carrying overload to SubstraitBuilder.windowFn, since the existing one left every RANGE frame built through the DSL with no way to supply the ordering this check now requires     
  
BREAKING CHANGE: a ConsistentPartitionWindow or WindowFunctionInvocation with a RANGE bound's Preceding or Following side now requires exactly one, non-CLUSTERED ordering expression. A plan that
previously built or parsed with zero, multiple, or a CLUSTERED ordering expression in that position now throws IllegalArgumentException.

### DIFF
--- a/core/src/main/java/io/substrait/expression/WindowBound.java
+++ b/core/src/main/java/io/substrait/expression/WindowBound.java
@@ -1,5 +1,6 @@
 package io.substrait.expression;
 
+import java.util.List;
 import java.util.Optional;
 import org.immutables.value.Value;
 
@@ -59,6 +60,46 @@ public interface WindowBound {
       throw new IllegalArgumentException(
           "bounds_type is required when either window bound is CurrentRow, Preceding, or"
               + " Following, but was BOUNDS_TYPE_UNSPECIFIED");
+    }
+  }
+
+  /**
+   * Validates a RANGE window's ordering against its bounds, per the spec's rule that a RANGE frame
+   * with a {@link Preceding} or {@link Following} bound must have exactly one ordering expression,
+   * which must not use {@code SORT_DIRECTION_CLUSTERED}.
+   *
+   * @param boundsType the window's bounds type
+   * @param lowerBound the window's lower bound
+   * @param upperBound the window's upper bound
+   * @param sorts the window's ordering expressions
+   * @throws IllegalArgumentException if {@code boundsType} is {@code RANGE}, either bound is {@link
+   *     Preceding} or {@link Following}, and {@code sorts} does not contain exactly one ordering
+   *     expression, or that expression uses {@code SORT_DIRECTION_CLUSTERED}
+   */
+  static void checkRangeOrdering(
+      Expression.WindowBoundsType boundsType,
+      WindowBound lowerBound,
+      WindowBound upperBound,
+      List<Expression.SortField> sorts) {
+    boolean needsSingleOrdering =
+        boundsType == Expression.WindowBoundsType.RANGE
+            && (lowerBound instanceof Preceding
+                || lowerBound instanceof Following
+                || upperBound instanceof Preceding
+                || upperBound instanceof Following);
+    if (!needsSingleOrdering) {
+      return;
+    }
+    if (sorts.size() != 1) {
+      throw new IllegalArgumentException(
+          "a RANGE bound with a Preceding or Following side requires exactly one ordering"
+              + " expression, but found "
+              + sorts.size());
+    }
+    if (sorts.get(0).direction() == Expression.SortDirection.CLUSTERED) {
+      throw new IllegalArgumentException(
+          "a RANGE bound with a Preceding or Following side cannot use"
+              + " SORT_DIRECTION_CLUSTERED for its ordering expression");
     }
   }
 
--- a/core/src/main/java/io/substrait/expression/WindowBound.java
+++ b/core/src/main/java/io/substrait/expression/WindowBound.java
@@ -72,15 +72,17 @@ public interface WindowBound {
    * @param lowerBound the window's lower bound
    * @param upperBound the window's upper bound
    * @param sorts the window's ordering expressions
-   * @throws IllegalArgumentException if {@code boundsType} is {@code RANGE}, either bound is {@link
-   *     Preceding} or {@link Following}, and {@code sorts} does not contain exactly one ordering
-   *     expression, or that expression uses {@code SORT_DIRECTION_CLUSTERED}
+   * @param function identifies the window function being validated, for the exception message
+   * @throws IllegalArgumentException if {@code boundsType} is {@code RANGE} and either bound is
+   *     {@link Preceding} or {@link Following}, and {@code sorts} does not hold exactly one
+   *     ordering expression whose direction is not {@code SORT_DIRECTION_CLUSTERED}
    */
   static void checkRangeOrdering(
       Expression.WindowBoundsType boundsType,
       WindowBound lowerBound,
       WindowBound upperBound,
-      List<Expression.SortField> sorts) {
+      List<Expression.SortField> sorts,
+      String function) {
     boolean needsSingleOrdering =
         boundsType == Expression.WindowBoundsType.RANGE
             && (lowerBound instanceof Preceding
@@ -92,13 +94,15 @@ public interface WindowBound {
     }
     if (sorts.size() != 1) {
       throw new IllegalArgumentException(
-          "a RANGE bound with a Preceding or Following side requires exactly one ordering"
+          function
+              + ": a RANGE bound with a Preceding or Following side requires exactly one ordering"
               + " expression, but found "
               + sorts.size());
     }
     if (sorts.get(0).direction() == Expression.SortDirection.CLUSTERED) {
       throw new IllegalArgumentException(
-          "a RANGE bound with a Preceding or Following side cannot use"
+          function
+              + ": a RANGE bound with a Preceding or Following side cannot use"
               + " SORT_DIRECTION_CLUSTERED for its ordering expression");
     }
   }
--- a/core/src/main/java/io/substrait/expression/WindowBound.java
+++ b/core/src/main/java/io/substrait/expression/WindowBound.java
@@ -168,9 +168,8 @@ public interface WindowBound {
     public abstract Expression offset();
 
     /**
-     * Creates a {@link Preceding} bound from a literal row offset. For {@code BOUNDS_TYPE_ROWS}
-     * only: a RANGE bound's offset must be type-compatible with the ordering expression, so use
-     * {@link #of(Expression)} there.
+     * Creates a {@link Preceding} bound from a literal {@code i64} row offset. Valid for ROWS, or
+     * for RANGE over an {@code i64} ordering expression; use {@link #of(Expression)} otherwise.
      *
      * @param offset the row offset preceding the current row
      * @return the preceding bound
@@ -206,9 +205,8 @@ public interface WindowBound {
     public abstract Expression offset();
 
     /**
-     * Creates a {@link Following} bound from a literal row offset. For {@code BOUNDS_TYPE_ROWS}
-     * only: a RANGE bound's offset must be type-compatible with the ordering expression, so use
-     * {@link #of(Expression)} there.
+     * Creates a {@link Following} bound from a literal {@code i64} row offset. Valid for ROWS, or
+     * for RANGE over an {@code i64} ordering expression; use {@link #of(Expression)} otherwise.
      *
      * @param offset the row offset following the current row
      * @return the following bound
--- a/core/src/test/java/io/substrait/dsl/SubstraitBuilderTest.java
+++ b/core/src/test/java/io/substrait/dsl/SubstraitBuilderTest.java
@@ -8,6 +8,7 @@ import io.substrait.TestBase;
 import io.substrait.expression.AggregateFunctionInvocation;
 import io.substrait.expression.Expression;
 import io.substrait.expression.FieldReference;
+import io.substrait.expression.WindowBound;
 import io.substrait.extension.DefaultExtensionCatalog;
 import io.substrait.extension.SimpleExtension;
 import io.substrait.plan.Plan;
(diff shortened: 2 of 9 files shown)


alexandrefimov (commented):
Checked the three rules against `algebra.proto` at v0.102.0, the version the catalog pins: the `bounds_type` comment on `WindowRelFunction` carries both halves for the relation, and for the expression form the "exactly one ordering expression" half lives on the `BOUNDS_TYPE_RANGE` enum, which both kinds share. So both checks land where the spec puts them.

Three things I would add, none of them blocking.

**The expression-side check has no test behind it.** Neutering `checkRangeOrdering` and running `:core:test` fails exactly three of 743 tests — the three new ones, all on `ConsistentPartitionWindow`. `Expression.WindowFunctionInvocation.check()` gained the same call, and the fixture in `windowFunctionInvocationRoundtripWithNonLiteralOffsetExpr` needed its `sort` precisely because of it, but nothing pins the rejection there. The two checks sit in different classes and can regress independently, so one mirror of `rangePrecedingWithTwoOrderingExpressionsIsRejected` on the invocation would cover it.

**On the issue's open question about CLUSTERED reachability:** isthmus cannot emit it on a window — `WindowFunctionConverter` maps a window's collations through `SortFieldConverter.asSortDirection`, which throws for any direction but ASCENDING and DESCENDING. The CLUSTERED mapping isthmus does have, `SubstraitRelVisitor.asSortDirection`, is on the `SortRel` path. The one producer in the repo that builds a `ConsistentPartitionWindow` from user input is Spark's `visitWindow`, which takes its sorts straight from `window.orderSpec`. Worth a line in the description naming what this can newly reject, since the BREAKING CHANGE note reads as if any producer might be affected.

**The third rule in that paragraph cannot be checked at all:** `Expression.SortField` models only `expr()` and `direction()`, so a custom comparison function has no POJO representation to reject. The issue says so; the description does not, and the paragraph reads as fully enforced without it.

core/src/test/java/io/substrait/relation/RelCopyOnWriteVisitorTest.java
```diff
@@ -57,6 +57,12 @@ void consistentPartitionWindowBoundOffsetsAreRewritten() {
                         .upperBound(WindowBound.Following.of(sb.i32(7)))
                         .boundsType(Expression.WindowBoundsType.RANGE)
                         .build()))
+            .sorts(
+                Arrays.asList(
+                    Expression.SortField.builder()
+                        .expr(sb.fieldReference(input, 0))
```
nielspardon: Make the ordering column `R.I32` in the `namedScan` on line 44 so it matches the i32 offsets. As added, this fixture pairs an i64 ordering expression with `Preceding.of(sb.i32(5))`/`Following.of(sb.i32(7))`, and `add(i64, i32) -> i64` is not a declared arithmetic variant — so the fixture now encodes exactly the offset/ordering rule this PR leaves unchecked, and will need fixing again when that rule lands. Before this change these fixtures had no ordering expression at all, so no mismatch existed; same shape at `ExpressionCopyOnWriteVisitorTest.java:102`, where the sort expr is `sb.i64(1)`.

nielspardon (commented):
Add a sorts-carrying overload to `SubstraitBuilder.windowFn` — it never sets `sort()`, so after this change every RANGE frame with a `Preceding`/`Following` bound built through the DSL throws with no way for the caller to supply the ordering the check now demands (`sb.windowFn(FUNCTIONS_ARITHMETIC, "lead:any", R.I64, INITIAL_TO_RESULT, ALL, RANGE, WindowBound.Preceding.of(5), WindowBound.CURRENT_ROW)` → `...requires exactly one ordering expression, but found 0`); `ExpressionCreator.windowFunction` already takes a sort list. Also retitle to `fix(core)!:` — the squash-merge message comes from the PR title, so without the `!` the CHANGELOG entry carries no breaking marker.

Separately from the comparison-function clause already raised above: #1198's offset/ordering type-compatibility rule (`add(T, D) -> T`) is untouched, and that issue's own open question asked for a decision on it. Is `Closes #1198` intended here, or should that rule get a follow-up issue first?

anasik:
@nielspardon, I think it should get a follow-up issue.

alexandrefimov (commented):
Re-measured on `5127ae4`: with the `checkRangeOrdering` call dropped from `WindowFunctionInvocation.check()` and the relation's left in place, `:core:test` (743) and `:isthmus:test` (1239) are both green, so nothing pins the expression side. This mirror fails without that call and passes with it, and needs no import `ConsistentPartitionWindowRelRoundtripTest` doesn't already have:

```java
@Test
void rangePrecedingWithTwoOrderingExpressionsOnAnInvocationIsRejected() {
  SimpleExtension.WindowFunctionVariant declaration =
      extensions.getWindowFunction(
          SimpleExtension.FunctionAnchor.of(
              DefaultExtensionCatalog.FUNCTIONS_ARITHMETIC, "lead:any"));
  Expression.SortField sort =
      Expression.SortField.builder()
          .expr(sb.i64(1))
          .direction(Expression.SortDirection.ASC_NULLS_FIRST)
          .build();

  assertThrows(
      IllegalArgumentException.class,
      () ->
          Expression.WindowFunctionInvocation.builder()
              .declaration(declaration)
              .arguments(Collections.emptyList())
              .partitionBy(Collections.emptyList())
              .sort(Arrays.asList(sort, sort))
              .outputType(R.I64)
              .aggregationPhase(Expression.AggregationPhase.INITIAL_TO_RESULT)
              .invocation(Expression.AggregationInvocation.ALL)
              .lowerBound(WindowBound.Preceding.of(5))
              .upperBound(WindowBound.CURRENT_ROW)
              .boundsType(Expression.WindowBoundsType.RANGE)
              .build());
}
```

My other two points were about the description only — drop them if you'd rather not touch it.

core/src/main/java/io/substrait/expression/WindowBound.java
```diff
@@ -62,6 +63,46 @@ static void checkBoundsType(
...
+          "a RANGE bound with a Preceding or Following side requires exactly one ordering"
+              + " expression, but found "
+              + sorts.size());
+    }
+    if (sorts.get(0).direction() == Expression.SortDirection.CLUSTERED) {
+      throw new IllegalArgumentException(
+          "a RANGE bound with a Preceding or Following side cannot use"
+              + " SORT_DIRECTION_CLUSTERED for its ordering expression");
+    }
+  }
```
nielspardon, line 108: Include the offending function in both messages. `ConsistentPartitionWindow.check()` loops over every invocation, so with more than one window function neither message says which one failed, and on the proto read path the same bare string surfaces from `ProtoRelConverter.newConsistentPartitionWindow` with no indication of which relation in the plan was at fault — worth fixing here rather than deferring, since these two strings are the entire user-facing surface of a breaking change whose point is to start rejecting plans that previously parsed. The index is the load-bearing part rather than the key: this file's own `consistentPartitionWindowRoundtripMulti` fixture has two invocations that both declare `lead:any`. The `@throws` is retightened too — its trailing `or` currently parses as though CLUSTERED throws under ROWS as well.

```suggestion
  /**
   * Validates a RANGE window's ordering against its bounds, per the spec's rule that a RANGE frame
   * with a {@link Preceding} or {@link Following} bound must have exactly one ordering expression,
   * which must not use {@code SORT_DIRECTION_CLUSTERED}.
   *
   * @param boundsType the window's bounds type
   * @param lowerBound the window's lower bound
   * @param upperBound the window's upper bound
   * @param sorts the window's ordering expressions
   * @param function identifies the window function being validated, for the exception message
   * @throws IllegalArgumentException if {@code boundsType} is {@code RANGE} and either bound is
   *     {@link Preceding} or {@link Following}, and {@code sorts} does not hold exactly one
   *     ordering expression whose direction is not {@code SORT_DIRECTION_CLUSTERED}
   */
  static void checkRangeOrdering(
      Expression.WindowBoundsType boundsType,
      WindowBound lowerBound,
      WindowBound upperBound,
      List<Expression.SortField> sorts,
      String function) {
    boolean needsSingleOrdering =
        boundsType == Expression.WindowBoundsType.RANGE
            && (lowerBound instanceof Preceding
                || lowerBound instanceof Following
                || upperBound instanceof Preceding
                || upperBound instanceof Following);
    if (!needsSingleOrdering) {
      return;
    }
    if (sorts.size() != 1) {
      throw new IllegalArgumentException(
          function
              + ": a RANGE bound with a Preceding or Following side requires exactly one ordering"
              + " expression, but found "
              + sorts.size());
    }
    if (sorts.get(0).direction() == Expression.SortDirection.CLUSTERED) {
      throw new IllegalArgumentException(
          function
              + ": a RANGE bound with a Preceding or Following side cannot use"
              + " SORT_DIRECTION_CLUSTERED for its ordering expression");
    }
  }
```

Both call sites need the extra argument: `"window function " + i + " (" + windowFunction.declaration().key() + ")"` from an indexed loop in `ConsistentPartitionWindow.check()`, and `declaration().key()` from `Expression.WindowFunctionInvocation.check()`.

core/src/test/java/io/substrait/type/proto/ConsistentPartitionWindowRelRoundtripTest.java
```diff
@@ -254,6 +261,108 @@ void boundsTypeUnspecifiedWithRealBoundIsRejected() {
...
+            .sorts(
+                Arrays.asList(
+                    Expression.SortField.builder()
+                        .expr(sb.fieldReference(input, 0))
+                        .direction(Expression.SortDirection.ASC_NULLS_FIRST)
+                        .build(),
+                    Expression.SortField.builder()
+                        .expr(sb.fieldReference(input, 1))
+                        .direction(Expression.SortDirection.ASC_NULLS_FIRST)
+                        .build()));
```
nielspardon: Use the `sb.sortFields(input, 0, 1)` helper that already exists — it builds exactly this list, differing only in ASC_NULLS_LAST vs ASC_NULLS_FIRST, which is immaterial here since the check rejects only CLUSTERED. This PR adds eight hand-rolled `SortField.builder()` chains across its four test files; `sb.sortFields` covers the field-reference ones and `sb.sortField(expr, direction)` the literal and CLUSTERED ones, and `RelSamples` already uses both.

```suggestion
            .sorts(sb.sortFields(input, 0, 1));
```

nielspardon (commented):
Drop `Closes #1198` rather than filing a follow-up for the deferred rule: #1198's own "What core does today" item 3 is exactly `add(T, D) -> T` going unchecked, and its Open Questions section already hosts the "resolving a function from inside POJO validation is a different weight class" discussion — so leaving #1198 open tracks the remainder better than a new issue would, and #1230 is the isthmus half already pointing back to it.

Two doc spots sit outside the diff hunks, so I couldn't anchor them inline: `Preceding.of(long)` and `Following.of(long)` still read "For `BOUNDS_TYPE_ROWS` only", which this PR's own new tests contradict — they pair `Preceding.of(5)` with `RANGE` over an `R.I64` ordering column, and v0.102.0's `site/docs/expressions/window_functions.md:34` names `i64`/`i64` as an explicit valid pairing — and the no-sorts `windowFn` overload should say in its Javadoc that it supplies an empty ordering list and therefore rejects a RANGE `Preceding`/`Following` frame outright.

Nothing exercises the new sorts-carrying overload: all seven `sb.windowFn(` call sites still use the old one, and each is either ROWS or RANGE with `UNBOUNDED` bounds, so the single shape the overload was added to make expressible has no test behind it.

Unrelated to this PR but adjacent to the fixtures it touches: #1231 — `RelCopyOnWriteVisitor.visit(ConsistentPartitionWindow)` never visits its input, which the window CoW tests can't catch while their input is a bare `namedScan`.

anasik:
@nielspardon understood. Dropped the `Closes #1198`

core/src/test/java/io/substrait/relation/RelCopyOnWriteVisitorTest.java
```diff
@@ -61,12 +61,18 @@ private ConsistentPartitionWindow windowOver(Rel input, WindowBound lower, Windo
                     .upperBound(upper)
                     .boundsType(Expression.WindowBoundsType.RANGE)
                     .build()))
+        .sorts(
+            Arrays.asList(
+                Expression.SortField.builder()
+                    .expr(sb.fieldReference(input, 0))
+                    .direction(Expression.SortDirection.ASC_NULLS_FIRST)
+                    .build()))
```
nielspardon: `sb.sortFields(input, 0)` builds exactly this list, and this PR already switched the roundtrip fixture over to it. The only difference is `ASC_NULLS_LAST` vs `ASC_NULLS_FIRST`, immaterial here since the check rejects only `CLUSTERED` and none of the three callers assert on direction.

```suggestion
        .sorts(sb.sortFields(input, 0))
```

core/src/test/java/io/substrait/dsl/SubstraitBuilderTest.java
```diff
@@ -227,6 +228,28 @@ void testBooleanLogic() {
...
+              Expression.AggregationPhase.INITIAL_TO_RESULT,
+              Expression.AggregationInvocation.ALL,
+              builder.sortFields(scan, 0),
+              Expression.WindowBoundsType.RANGE,
+              WindowBound.Preceding.of(builder.i32(5)),
+              WindowBound.CURRENT_ROW,
+              builder.fieldReference(scan, 0));
+
+      assertNotNull(windowFn);
+      assertEquals(1, windowFn.sort().size());
```
nielspardon, line 252: Assert the thing this overload exists for. `assertNotNull` on the result of a builder that throws on failure is close to vacuous, whereas the sharper claim — and the reason the overload was added — is that the no-sorts overload cannot express this shape at all. Needs `import static org.junit.jupiter.api.Assertions.assertThrows;`.

```suggestion
      assertEquals(1, windowFn.sort().size());

      // The no-sorts overload cannot express this shape at all.
      assertThrows(
          IllegalArgumentException.class,
          () ->
              builder.windowFn(
                  DefaultExtensionCatalog.FUNCTIONS_ARITHMETIC,
                  "lead:any",
                  Type.I32.builder().nullable(false).build(),
                  Expression.AggregationPhase.INITIAL_TO_RESULT,
                  Expression.AggregationInvocation.ALL,
                  Expression.WindowBoundsType.RANGE,
                  WindowBound.Preceding.of(builder.i32(5)),
                  WindowBound.CURRENT_ROW,
                  builder.fieldReference(scan, 0)));
```

nielspardon (commented):
The rebase resolution is the right one — folding the ordering into `windowOver` covers its `Preceding`/`Following` caller and leaves the two `UNBOUNDED`/`CURRENT_ROW` callers unaffected. Two small things inline, neither blocking.

Worth putting the citation on the PR, since the body carries none: v0.102.0 states this rule verbatim at `site/docs/expressions/window_functions.md:34`, and again in `proto/substrait/algebra.proto` at :457 (`WindowRelFunction.bounds_type`) and :1401 (`Expression.WindowFunction.sorts`) — so the count and the CLUSTERED half are both spec-defined, and gating them on a `Preceding`/`Following` side matches the spec's own scope rather than over-applying to every RANGE frame the way substrait-go does.

The only clause the new Javadoc's paraphrase drops is "or a custom comparison function" from that same sentence, which `Expression.SortField` cannot express today — it has `expr()` and `direction()` and no `comparison_function_reference` — so that half stays with #1198 and needs nothing here.

alexandrefimov (approved):
Re-ran `:core:test` on `e132db5e`: 762 tests, all green.

The expression-side gap I measured on 2026-09-02 is pinned now. Neutering the `checkRangeOrdering` call in `WindowFunctionInvocation.check()` fails exactly two tests — `rangePrecedingWithTwoOrderingExpressionsOnAnInvocationIsRejected` and `testWindowFunctionWithOrdering` — and nothing else.

nielspardon (approved):
LGTM

PR #1206, opened 2026-08-31
fix(isthmus)!: reject RANGE offsets that cannot be retyped to the ordering column

### LINKED ISSUE #1199: isthmus: a window offset that cannot be retyped to the ordering column falls back to a type-mismatched offset
## Description

`WindowBoundConverter.integralLiteralOfType` (added in #1156) retypes an integral literal offset to match the ordering column's type, so that `add(T, D) -> T` stays defined for a RANGE frame. When it cannot do that it returns `Optional.empty()`, and the caller's `.orElse(offset)` keeps whatever type Calcite gave the literal — which is exactly the mismatch the method exists to prevent, now emitted silently.

## When the fallback fires

Every guard in the method has a path to it:

- **Out of range for a narrow ordering column.** `ORDER BY smallint_col RANGE BETWEEN 100000 PRECEDING` — Calcite accepts it, because `SqlWindow.validateFrameBoundary` only checks the bound's type *family* against `orderTypeFam.allowableDifferenceTypes()` for RANGE, with no range check and no coercion to the ordering type. The i16 guard rejects 100000, so the offset stays an i32 literal and the plan carries `add(i16, i32)`, which is not defined.
- **Decimal precision.** `digitCount(value) + scale > precision` falls through the same way.
- **FP round-trip.** A value that does not survive `(long)(double) value` (or the float equivalent) falls through.
- **Any ordering type the method does not enumerate** — temporal columns, for instance. `ORDER BY timestamp_col RANGE BETWEEN 5 PRECEDING` keeps an integral offset against a timestamp, where the spec wants interval_day.

All of these are degenerate frames in practice: a distance larger than the column's own domain, or a unit mismatch. That is why this is low severity rather than none — the plan is well-formed but not spec-conformant, and nothing says so.

## What to decide

Throwing is the only thing that makes it visible, and `UnsupportedOperationException` is the established isthmus idiom for "this SQL has no faithful Substrait form". The counter-argument is that isthmus would then reject SQL that Calcite accepts and that it has converted since long before #1156, so a lenient-but-silent conversion may be the deliberate trade. Either way it is worth stating in the code rather than leaving as a fallthrough.

## Test gap

The guards and the decimal/FP branches have no coverage. `WindowBoundConverterTest` (added in #1156) covers the happy paths — decimal keeps its fraction, ROWS widens to i64, RANGE takes the ordering type, zero becomes `CurrentRow`, a non-literal offset passes through — but the `47f52f9d` commit that added the range checks and the decimal/FP branches added no tests, so every branch above is untested. Worth adding regardless of which way the question above goes, since the expected behaviour is observable either way.

## Context

Surfaced reviewing #1156. Same shape as #1138 (spark: interval_year literal conversion silently wraps on Int overflow): a narrowing conversion that produces a plausible-looking value instead of failing. The related core-side gap — nothing validates offset/ordering type compatibility for a plan built directly or read from proto — is #1198.

🤖 Generated with AI

anasik:
- Throws instead of silently keeping a type-mismatched offset when it can't be retyped to the ordering column 
- Adds tests for the 4 previously-uncovered branches
- Chooses to reject rather than widen: when the offset can't be retyped to the ordering column's exact type, there's no larger type to widen to that stays correct, so failing loudly is safer than silently picking a different type and getting the value wrong
- Mirrors a negative RANGE/ROWS offset to the opposite bound with its magnitude, per spec (direction lives in Preceding/Following, not offset sign); a negation that overflows (e.g. `Long.MIN_VALUE`) throws `UnsupportedOperationException` rather than crashing 
- Zero offset now normalizes to `CurrentRow` before retyping, so it's never rejected for not fitting the ordering column's type

Follow-up (deferred, not in this PR): - Negative-offset mirroring only covers integral literals; a negative decimal or floating-point offset keeps its sign and label unchanged - RANGE offset retyping only handles integral offsets against the ordering column's type; decimal, floating-point, and custom/extension ordering types are unchecked or rejected outright even when the underlying `add`/`subtract` function would accept them — needs resolution against the function catalog rather than structural type matching

BREAKING CHANGE: `SqlToSubstrait` now throws `UnsupportedOperationException` when a `RANGE` window's integral offset cannot be retyped to the ordering column's type (out of range, past a decimal's precision, an FP round-trip that doesn't survive, or an ordering type with no integral form at all — the one a temporal ordering column hits). It previously converted successfully but produced a type-mismatched, already spec-invalid `offset_expr`.

Closes #1199

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved window-bound handling for zero and negative offsets.
  * Ensured RANGE offsets match the ordering expression’s type and supported limits.
  * Added clear rejection for unsupported temporal offsets and extreme minimum values.
  * Correctly converts valid integral, decimal, and floating-point offsets.
  * Improved reliability when processing boundary values and preventing invalid offset conversions.


alexandrefimov (commented):
A zero offset stops converting where the ordering type has no integral form. `toWindowBound` normalizes before it checks for zero, so the new refusal fires first. Measured on this branch against its parent (`8088c0d`): `RexWindowBounds.preceding(0)` with a `TIMESTAMP`, `DATE` or `VARCHAR` ordering type returned `CurrentRow` before and throws now, while `SMALLINT` gives `CurrentRow` either way. Zero needs no retyping — the spec makes it equivalent to CURRENT ROW, which `toWindowBound` does two lines below the change. `zeroOffsetBecomesCurrentRow` stays green because it converts a ROWS bound with no ordering type, so it never reaches the new branch.

Reading the zero check before normalizing keeps both:

```java
    RexNode node = rexWindowBound.getOffset();
    Expression converted = node.accept(rexExpressionConverter);

    // Per the spec, zero is not a valid offset; it is equivalent to CurrentRow, and producers
    // should emit CurrentRow rather than a zero offset_expr.
    if (integralValue(converted).filter(value -> value == 0).isPresent()) {
      return WindowBound.CURRENT_ROW;
    }

    Expression offset =
        normalizeIntegralOffset(
            converted, isRows, orderingType, rexExpressionConverter.getTypeConverter());
```

With that, `preceding(5)` over a `TIMESTAMP` ordering still throws, the four new tests pass, and so does the rest of `:isthmus:test`.

Two smaller things. An interval offset leaves `normalizeIntegralOffset` before the ordering type is consulted, so a temporal ordering column with one converts exactly as before — checked on `TIMESTAMP` and `DATE`, both still give back the `IntervalDayLiteral` unchanged, so the change does not reach `RANGE INTERVAL ... PRECEDING`. And the `BREAKING CHANGE` note names three causes — out of range, past a decimal's precision, an FP round trip that does not survive — where there is a fourth: an ordering type with no integral form at all, which is what `rangeOffsetAgainstUnsupportedOrderingTypeThrows` covers and what a temporal ordering column runs into. Maybe worth naming it there, since it is the one a reader is most likely to meet.

isthmus/src/main/java/io/substrait/isthmus/expression/WindowBoundConverter.java
```diff
@@ -82,10 +84,21 @@ private static Expression normalizeIntegralOffset(
       // The spec requires a BOUNDS_TYPE_ROWS offset_expr to be int64.
       return ExpressionCreator.i64(false, value.get());
     }
-    // BOUNDS_TYPE_RANGE: keep add(T, D) -> T defined for the ordering expression's type T.
+    // BOUNDS_TYPE_RANGE: keep add(T, D) -> T defined for the ordering expression's type T. If the
+    // ordering type is unknown, there is nothing to retype against, so the literal is left as-is.
     return orderingType
```
nielspardon, line 111: Check for a zero offset before retyping — a zero offset needs no representation in the ordering expression's type, so it should never reach this guard. On this branch `order by d55 range between 0 preceding` (`DECIMAL(5,5)`) throws where `main` returns `CurrentRow`, because `digitCount(0)` is 1 and `1 + 5 > 5`; the same move also turns the current `Preceding{0}` output for `DECIMAL(10,2)` and `DOUBLE` ordering columns into the `CurrentRow` the spec asks for.

In `toWindowBound`, replacing lines 49-61 with:

```java
    RexNode node = rexWindowBound.getOffset();
    Expression converted = node.accept(rexExpressionConverter);

    // Per the spec, zero is not a valid offset; it is equivalent to CurrentRow, and producers
    // should emit CurrentRow rather than a zero offset_expr. Checked before retyping: a zero
    // offset needs no representation in the ordering expression's type.
    if (integralValue(converted).filter(value -> value == 0).isPresent()) {
      return WindowBound.CURRENT_ROW;
    }

    Expression offset =
        normalizeIntegralOffset(
            converted, isRows, orderingType, rexExpressionConverter.getTypeConverter());
```

keeps the rejection this PR is adding (`100000` over a `SMALLINT` column still throws) and leaves `WindowBoundConverterTest` and `WindowFunctionTest` green.

isthmus/src/main/java/io/substrait/isthmus/expression/WindowBoundConverter.java
```diff
@@ -82,10 +84,21 @@ private static Expression normalizeIntegralOffset(
...
         .map(typeConverter::toSubstrait)
-        .flatMap(type -> integralLiteralOfType(type, value.get()))
+        .map(
+            type ->
+                integralLiteralOfType(type, value.get())
+                    .orElseThrow(
+                        () ->
+                            new UnsupportedOperationException(
```
nielspardon, line 118: Reword this to say what isthmus cannot build rather than what the spec forbids. `functions_arithmetic_decimal.yaml` derives `add(decimal<P1,S1>, decimal<P2,S2>)` with `init_prec = init_scale + max(P1-S1, P2-S2) + 1`, so `add(T, D) -> T` holds for no decimal pair at all — not even `decimal(5,2)` with itself, which this method accepts. The spec's own examples (`timestamp`/`interval_day`, `decimal`/`decimal`) show `D` need not equal `T`, and a wider decimal offset would satisfy that pairing here, so requiring `D == T` is an isthmus policy rather than a spec verdict.

The `@throws` at line 34 and the comment at line 87 carry the same claim, and both omit the `subtract(T, D) -> T` half of the requirement.

nielspardon (commented):
Move the zero-offset check above `normalizeIntegralOffset`: `RANGE BETWEEN 0 PRECEDING` over a `DECIMAL(5,5)` ordering column returns `CurrentRow` on `main` and throws on this branch, and the same reordering also stops the existing `Preceding{0.00}` emission for decimal and FP ordering columns that the spec tells producers not to write.

Separately, the `add(T, D) -> T` framing needs a second look. The spec asks only that `D` be *compatible* with `T` (and requires `subtract(T, D) -> T` as well), and for decimals `add` always widens the precision — so no decimal offset ever literally satisfies the rule the new message cites, including the ones the code accepts. Since #1199 left reject-vs-widen open and the spec says nothing about what a producer should do when it cannot represent the offset, it would help to state in the PR which option you chose and why.

One thing to fix before merge because it cannot be corrected after a release: the `BREAKING CHANGE:` footer is the only part of the body published verbatim, and it currently ends `` `offset_expr.` `` with the period inside the code span. It also omits the fourth rejection cause — any ordering type `integralLiteralOfType` has no case for — and the commit title is missing the `!` that the PR title carries.

alexandrefimov (approved):
Measured on the merge head (`00d510d`): a `DATE` or `VARCHAR` ordering — the two the new tests don't cover — is back to `CurrentRow` at zero, a non-zero offset over either still throws, and `SMALLINT` retypes as before.

The reorder can't drop a zero the old placement caught: `normalizeIntegralOffset` maps an integral literal to an integral literal of the same value, so `integralValue` sees the same number on either side of it — including on the ROWS path, where zero used to arrive as `i64(0)`.

nielspardon (commented):
`RANGE BETWEEN -5 PRECEDING` still converts to `Preceding{I32Literal{-5}}` from plain SQL here, and it's the same spec sentence this PR implements: `site/docs/expressions/window_functions.md:34` at the pinned `v0.102.0` requires "a non-negative distance whose type `D` is compatible with the ordering expression's type `T`", and `algebra.proto` gives the remedy on `Preceding.offset` — "Use CurrentRow for offset zero and Following for negative offsets." So it parallels the zero case you just hoisted: negate and flip to the mirror bound rather than throw, which keeps isthmus accepting what Calcite accepts (Calcite validates the sign for ROWS only) while emitting a valid plan. Whether that belongs here or in a follow-up is your call; the other three comments are smaller. Filed #1229 and #1230 for two adjacent gaps that predate this PR — nothing needed here for either.

isthmus/src/main/java/io/substrait/isthmus/expression/WindowBoundConverter.java
```diff
@@ -45,19 +47,19 @@ public static WindowBound toWindowBound(
...
     // Per the spec, zero is not a valid offset; it is equivalent to CurrentRow, and producers
-    // should emit CurrentRow rather than a zero offset_expr.
-    if (integralValue(offset).filter(value -> value == 0).isPresent()) {
+    // should emit CurrentRow rather than a zero offset_expr. Checked before retyping: a zero
+    // offset needs no representation in the ordering expression's type.
+    if (integralValue(converted).filter(value -> value == 0).isPresent()) {
       return WindowBound.CURRENT_ROW;
     }
 
+    Expression offset =
```
nielspardon, line 71: Negate and flip a negative offset to the mirror bound instead of emitting it — `RANGE BETWEEN -5 PRECEDING` reaches here as `Preceding{I32Literal{-5}}`, because Calcite's `SqlWindow#validateFrameBoundary` checks the sign only inside `if (isRows)`, and over a DECIMAL ordering column `digitCount` calls `.abs()` so it clears the new precision check too. Same shape as the zero case above: the spec carries the direction in the Preceding/Following choice, so the mirror bound is the equivalent rather than an error. Applied over your branch this yields `Following{5}` / `Preceding{5}` / decimal `+5`, leaves positive offsets untouched, and keeps the 46 window tests green:

```java
    // The spec carries a bound's direction in the Preceding/Following choice, not in the sign of
    // the offset: a negative offset is invalid, and the mirror bound with the magnitude is its
    // equivalent. Calcite only rejects a negative offset for ROWS, so RANGE reaches here.
    boolean preceding = rexWindowBound.isPreceding();
    Optional<Long> negative = integralValue(converted).filter(value -> value < 0);
    if (negative.isPresent()) {
      preceding = !preceding;
      converted = negate(converted, negative.get());
    }

    Expression offset =
        normalizeIntegralOffset(
            converted, isRows, orderingType, rexExpressionConverter.getTypeConverter());

    if (preceding) {
      return WindowBound.Preceding.of(offset);
    }
    if (rexWindowBound.isFollowing() || negative.isPresent()) {
      return WindowBound.Following.of(offset);
    }
```

with a helper alongside `integralValue`:

```java
  private static Expression negate(Expression offset, long value) {
    return integralLiteralOfType(offset.getType(), Math.negateExact(value))
        .orElseThrow(
            () ->
                new UnsupportedOperationException(
                    "window offset "
                        + value
                        + " cannot be negated within its own type "
                        + offset.getType().accept(new StringTypeVisitor())));
  }
```

isthmus/src/main/java/io/substrait/isthmus/expression/WindowBoundConverter.java
```diff
@@ -82,10 +84,19 @@ private static Expression normalizeIntegralOffset(
...
+        .map(
+            type ->
+                integralLiteralOfType(type, value.get())
+                    .orElseThrow(
+                        () ->
+                            new UnsupportedOperationException(
+                                "RANGE window offset "
+                                    + value.get()
+                                    + " does not fit the ordering expression's type "
+                                    + type)))
```
nielspardon: Render the ordering type with `StringTypeVisitor` so the message names it: `type` here is an `ImmutableType.Decimal`, whose `toString()` prints the generated class dump instead of `decimal<5,2>`. It's the established idiom in this package — `RexExpressionConverter#operandTypeName` does exactly this. Needs `import io.substrait.type.StringTypeVisitor;`.

```suggestion
                                "RANGE window offset "
                                    + value.get()
                                    + " does not fit the ordering expression's type "
                                    + type.accept(new StringTypeVisitor()))))
```

isthmus/src/main/java/io/substrait/isthmus/expression/WindowBoundConverter.java
```diff
@@ -31,6 +31,8 @@ public class WindowBoundConverter {
    * @return the corresponding Substrait {@link WindowBound}
    * @throws IllegalStateException if the bound is not one of CURRENT ROW, UNBOUNDED, PRECEDING, or
    *     FOLLOWING
+   * @throws UnsupportedOperationException if a RANGE offset's integral literal does not fit the
```
nielspardon, line 36: Say that the throw also covers an ordering type `integralLiteralOfType` has no case for at all — that's 20 of `TypeVisitor`'s 27 kinds, and "does not fit the ordering expression's exact type" reads as an out-of-range value rather than an unrepresentable one. It's worth naming because it's a behavior change: a `UserTypeMapper`-mapped ordering column converts to `Preceding{I32Literal{5}}` on `fb6a54ad` and throws here, and `add(u!myord, i32) -> u!myord` is declarable today (`functions_custom.yaml`, exercised by `CustomFunctionTest`). Separately, the class Javadoc at line 18 still promises "bounds with an arbitrary offset expression", which this PR narrows.

isthmus/src/test/java/io/substrait/isthmus/WindowBoundConverterTest.java
```diff
@@ -96,6 +97,64 @@ void rangeIntegralOffsetTakesTheOrderingExpressionType() {
...
+  @Test
+  void rangeOffsetOutOfRangeForOrderingTypeThrows() {
+    // Calcite's SqlWindow#validateFrameBoundary only checks the bound's type family against the
+    // ordering type for RANGE, not its range, so an offset that doesn't fit the ordering column's
+    // narrower type must be rejected here rather than silently kept as the literal's own type.
+    RexNode offset = c(100000, SqlTypeName.INTEGER);
+    RexWindowBound bound = RexWindowBounds.preceding(offset);
+    RelDataType orderingType = t(SqlTypeName.SMALLINT);
+
+    assertThrows(
```
nielspardon, line 136: Assert the messages on the four `assertThrows` tests — the bodies are near-identical and only the message distinguishes which guard fired, so any one of them would pass on a throw from the wrong branch. Worth adding the accept side of the boundary as well (`Short.MAX_VALUE` retypes, `+1` throws), and a `DATE` or `VARCHAR` case so the no-such-case path isn't covered by `TIMESTAMP` alone.

isthmus/src/main/java/io/substrait/isthmus/expression/WindowBoundConverter.java
anasik: @nielspardon agreed the custom-type case is real. Keeping the throw as-is here since it's still strictly safer than the old pass-through. Should we file a follow-up issue to handle the custom-type resolution case?

isthmus/src/main/java/io/substrait/isthmus/expression/WindowBoundConverter.java
```diff
@@ -45,30 +49,51 @@ public static WindowBound toWindowBound(
...
 
-    if (rexWindowBound.isPreceding()) {
+    // The spec carries a bound's direction in the Preceding/Following choice, not in the sign of
+    // the offset: a negative offset is invalid, and the mirror bound with the magnitude is its
+    // equivalent. Calcite only rejects a negative offset for ROWS, so RANGE reaches here.
+    boolean preceding = rexWindowBound.isPreceding();
+    Optional<Long> negative = integralValue(converted).filter(value -> value < 0);
+    if (negative.isPresent()) {
+      preceding = !preceding;
+      converted = negate(converted, negative.get());
```
alexandrefimov, line 68: The mirror only reaches integral offsets. `integralValue` recognises `i8` through `i64`, so a negative decimal offset keeps its `Preceding` label and its negative value: `order by O_TOTALPRICE range between -5.5 preceding` converts without error, and the same bound through `toWindowBound` gives `Preceding{decimal<19,1> -5.5}`. That is the shape this change removes for `i32`.

The integral path has a boundary of its own. `order by O_ORDERKEY range between -2147483648 preceding` over that `BIGINT` column throws `UnsupportedOperationException: window offset -2147483648 cannot be negated within its own type i32`, while `-2147483647` gives `Following{I64Literal{2147483647}}`: `negate` builds the flipped literal in the offset's own type, where 2147483648 has no form, though the ordering column is `i64`.

Mirroring the value instead of the literal covers the second one, but not by moving the flip behind `normalizeIntegralOffset` — with no `orderingType` that method returns the offset untouched, and the sign would survive into a `Following` bound. Would you rather take the decimal case here, or leave it for a follow-up and say so in the description?

isthmus/src/main/java/io/substrait/isthmus/expression/WindowBoundConverter.java
```diff
@@ -45,30 +49,51 @@ public static WindowBound toWindowBound(
...
+    if (rexWindowBound.isFollowing() || negative.isPresent()) {
       return WindowBound.Following.of(offset);
     }
 
     throw new IllegalStateException(
         "window bound was none of CURRENT ROW, UNBOUNDED, PRECEDING or FOLLOWING");
   }
 
+  private static Expression negate(Expression offset, long value) {
+    return integralLiteralOfType(offset.getType(), Math.negateExact(value))
```
alexandrefimov: `Math.negateExact` throws `ArithmeticException` for `Long.MIN_VALUE`, and it leaves `toWindowBound` uncaught — the method documents `IllegalStateException` and `UnsupportedOperationException`, not this one. It is reachable from SQL: `range between -9223372036854775808 preceding` over a `BIGINT` ordering column gives `java.lang.ArithmeticException: long overflow`. Wrapping it in the `UnsupportedOperationException` the javadoc already promises would keep the contract.

isthmus/src/main/java/io/substrait/isthmus/expression/WindowBoundConverter.java
anasik: Might as well take it here.

anasik: @alexandrefimov, on second thought, lets make a follow up issue and I will pick it up right after this is merged.

alexandrefimov: Agreed, let's keep the negative decimal case and negation before widening in a follow-up. Please mention those remaining cases in the PR description.

Rechecked `a8bf3d5a`: `Long.MIN_VALUE` now throws `UnsupportedOperationException`, including through SQL. Both `PRECEDING` and `FOLLOWING` are covered, and `Long.MIN_VALUE + 1` still mirrors correctly.

anasik: @alexandrefimov, I updated the PR description

alexandrefimov (approved):

isthmus/src/main/java/io/substrait/isthmus/expression/WindowBoundConverter.java
```diff
@@ -45,30 +49,58 @@ public static WindowBound toWindowBound(
...
+      throw new UnsupportedOperationException("window offset " + value + " cannot be negated");
+    }
+    return integralLiteralOfType(offset.getType(), negated)
+        .orElseThrow(
+            () ->
+                new UnsupportedOperationException(
+                    "window offset "
+                        + value
+                        + " cannot be negated within its own type "
+                        + offset.getType().accept(new StringTypeVisitor())));
```
nielspardon: Negate into `i64` and let `normalizeIntegralOffset` retype it, so the magnitude only has to fit the ordering column rather than the offset literal's own type. As written, `RANGE BETWEEN -2147483648 PRECEDING` over a `BIGINT` ordering column throws `cannot be negated within its own type i32`, while `2147483648 PRECEDING` over that same column converts to `I64Literal{2147483648}` — so the mirror rejects an input it can represent. With this the 66 isthmus window tests stay green, `Long.MIN_VALUE` still throws, and `offset` becomes unused so the signature can drop to `negate(long value)`.

```suggestion
    // Widen rather than negate in place: the magnitude need not fit the offset literal's own type,
    // only the type normalizeIntegralOffset then retypes it to.
    return ExpressionCreator.i64(false, negated);
```

nielspardon (commented):
The mirror negates inside the offset literal's own type, so it throws for a magnitude the ordering column can hold: on `0bda1ff7` `RANGE BETWEEN -2147483648 PRECEDING` over a `BIGINT` ordering column fails, where `0e0db6ed` converted it and this same branch accepts `2147483648 PRECEDING` — that's the "negation before widening" case from the thread above, but it turns out to be one line rather than follow-up work. The other two comments are Javadoc. Say the word and I'll open the deferred-work issue you and @alexandrefimov agreed on — nothing tracks it outside the PR body, where both bullets are on one line so they render as a single run-on paragraph.

isthmus/src/main/java/io/substrait/isthmus/expression/WindowBoundConverter.java
```diff
@@ -31,6 +33,8 @@ public class WindowBoundConverter {
    * @return the corresponding Substrait {@link WindowBound}
    * @throws IllegalStateException if the bound is not one of CURRENT ROW, UNBOUNDED, PRECEDING, or
    *     FOLLOWING
+   * @throws UnsupportedOperationException if a RANGE offset's integral literal does not fit the
+   *     ordering expression's exact type
```
nielspardon: Cover the negation throws here too — both fire outside what this clause describes, and one of them isn't RANGE-specific at all: a ROWS bound with `Long.MIN_VALUE` and an empty `orderingType` throws this same exception with no ordering type involved.

```suggestion
   * @throws UnsupportedOperationException if a RANGE offset's integral literal does not fit the
   *     ordering expression's exact type, or if a negative offset's magnitude has no positive
   *     representation
```

isthmus/src/main/java/io/substrait/isthmus/expression/WindowBoundConverter.java
```diff
@@ -16,7 +17,8 @@
  * Utility for converting Calcite {@link RexWindowBound} to Substrait {@link WindowBound}.
  *
  * <p>Supports {@code CURRENT ROW}, {@code UNBOUNDED}, and {@code PRECEDING}/{@code FOLLOWING}
- * bounds with an arbitrary offset expression.
+ * bounds with an arbitrary offset expression. A RANGE bound's integral literal offset must match
+ * the ordering expression's exact type.
```
nielspardon: Say that the converter now normalizes the bound, not just validates it: `-5 PRECEDING` comes back as `Following{5}`, which is a rewrite of the caller's bound and worth stating up front alongside the zero case.

```suggestion
 * <p>Supports {@code CURRENT ROW}, {@code UNBOUNDED}, and {@code PRECEDING}/{@code FOLLOWING}
 * bounds with an arbitrary offset expression. A RANGE bound's integral literal offset must match
 * the ordering expression's exact type. A negative integral offset is mirrored to the opposite
 * bound with its magnitude, and a zero offset becomes {@code CURRENT ROW}.
```

isthmus/src/test/java/io/substrait/isthmus/WindowBoundConverterTest.java
```diff
@@ -108,4 +218,78 @@ void zeroOffsetBecomesCurrentRow() {
...
+
+    WindowBound converted =
+        WindowBoundConverter.toWindowBound(
+            bound, false, Optional.of(orderingType), rexExpressionConverter);
+
+    assertEquals(WindowBound.Preceding.of(ExpressionCreator.i32(false, 5)), converted);
+  }
+
+  @Test
+  void negativeOffsetOverflowingLongIsRejectedRatherThanThrowingArithmeticException() {
```
nielspardon, line 297: Cover the widen: both mirror tests use `-5` over an `INTEGER` ordering column, where the magnitude fits the offset literal's own type either way, so reverting `f11cdeee` leaves all 18 green. This one fails on the pre-`f11cdeee` form with `window offset -2147483648 cannot be negated within its own type i32`.

```suggestion
  @Test
  void negativeOffsetIsMirroredWhenItsMagnitudeDoesNotFitItsOwnType() {
    // The magnitude only has to fit the ordering column, not the offset literal's own type:
    // Integer.MIN_VALUE negates to 2147483648, which is not an i32 but is an i64.
    RexNode offset = c(Integer.MIN_VALUE, SqlTypeName.INTEGER);
    RexWindowBound bound = RexWindowBounds.preceding(offset);
    RelDataType orderingType = t(SqlTypeName.BIGINT);

    WindowBound converted =
        WindowBoundConverter.toWindowBound(
            bound, false, Optional.of(orderingType), rexExpressionConverter);

    assertEquals(WindowBound.Following.of(ExpressionCreator.i64(false, 2147483648L)), converted);
  }

  @Test
  void negativeOffsetOverflowingLongIsRejectedRatherThanThrowingArithmeticException() {
```

nielspardon (commented):
Add a test for the widen — reverting `f11cdeee` leaves all 18 tests green, so nothing holds the fix in place. CodeRabbit's FP-saturation finding is #1229, already filed and deliberately deferred, so no need to take it here. #1291's second case *is* fixed here though: the issue's own `RANGE BETWEEN -2147483648 PRECEDING` over `O_ORDERKEY` now converts to `Following{i64 2147483648}`, so on merge it narrows to the negative-decimal case — linking #1291 from the follow-up section would also break those bullets onto separate lines, which currently render as one run-on paragraph.

nielspardon (approved):
LGTM

PR #1224, opened 2026-09-01
build: update vulnerability scan

bestbeforetoday:
- Use latest version of OSV-Scanner rather than a fixed version.
- Refactor Makefile to avoid recursive calls to make.
- Download OSV-Scanner binary to remove dependency on Docker.


nielspardon (approved):
LGTM

PR #1246, opened 2026-09-03
build: surface scan failures in build summary

bestbeforetoday:
Currently a scan failures requires inspection of the build log for the failing scan job. This change outputs the failure to the step summary so it is visible directly in the build summary page.

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -52,17 +52,11 @@ jobs:
           distribution: 'temurin'
       - name: Setup Gradle
         uses: gradle/actions/setup-gradle@v6.3.0
-      - name: Generate lockfile
-        run: make ${{ matrix.project }}/gradle.lockfile
       - name: Scan
         id: scan
-        run: |
-          make install-osv-scanner
-          ./osv-scanner scan source \
-            --config=osv-scanner.toml \
-            --lockfile=${{ matrix.project }}/gradle.lockfile \
-            --format=markdown \
-            --output-file=osv-scanner.md
+        run: make scan-${{ matrix.project }}
+        env:
+          OSV_SCANNER_ARGS: --format=markdown --output-file=osv-scanner.md
       - name: Report failure
         if: ${{ failure() && steps.scan.conclusion == 'failure' }}
         run: |
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -52,8 +52,23 @@ jobs:
           distribution: 'temurin'
       - name: Setup Gradle
         uses: gradle/actions/setup-gradle@v6.3.0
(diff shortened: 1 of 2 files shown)


bestbeforetoday:
You can see how how scan failures are surfaced by this updated workflow in this [build summary page](https://github.com/substrait-io/substrait-java/actions/runs/33747626072). Scroll down that page for the job summaries.

.github/workflows/pr.yml
```diff
@@ -52,8 +52,23 @@ jobs:
...
+      - name: Generate lockfile
+        run: make ${{ matrix.project }}/gradle.lockfile
       - name: Scan
-        run: make scan-${{ matrix.project }}
+        id: scan
+        run: |
+          docker run --rm \
+            --volume './:/src' \
+            ghcr.io/google/osv-scanner:latest \
+            scan source \
```
nielspardon: I'm a little confused about this using the docker image given that you just proposed migrating off the docker image here: https://github.com/substrait-io/substrait-java/pull/1224

What am I missing here?

bestbeforetoday: The Makefile can be used at the command-line by developers who might not have Docker installed. The GitHub Actions runners most certainly do have Docker installed, and the Docker image is hosted in GitHub Packages so is very easily accessible to them. Using the Docker image seemed an easier approach than downloading the binary. However... I can easily change this workflow to use the binary rather than a Docker image if you prefer that approach.

nielspardon: I mean the binary is also pulled from Github's releases so they are probably very similar to each other. I was just wondering whether we would want consistency here. Would also be an opportunity to have Makefile exercised regularly through CI.

bestbeforetoday: I agree the more the Makefile is exercised to confirm it works the better. I have pushed another commit to use the binary and some (not all) of the Makefile implementation. The workflow uses slightly different arguments so I have just added the osv-scanner command here. I could change the Makefile with a new target or to allow overrides to the osv-scanner arguments based on environment variables if you prefer.

nielspardon: Thanks, makes sense to me. I don't have a strong opinion on whether we should add new targets or just keep it the way you have it right now.

nielspardon (commented):
LGTM

.github/workflows/pr.yml
bestbeforetoday: Looking again I prefer using the Makefile entirely. It exercises the Makefile better and is more robust since the workflow does not need to know how the Makefile runs the scan or where it puts the binary. I have pushed another commit with this change. Unless you spot any mistakes, I think the change is good to go now.

nielspardon (approved):
LGTM

PR #1247, opened 2026-09-03
build: bump editorconfig-checker action to v2.2.0

bestbeforetoday:
The v2 tag for editorconfig-checker/action-editorconfig-checker is not maintained with the latest v2.x release state. The code in this version was last updated in 2023. This results in build warnings due to the back-level Node version used:

Node.js 20 is deprecated. The following actions target Node.js 20 but are being forced to run on Node.js 24:
editorconfig-checker/action-editorconfig-checker@v2.

This change updates to the latest release version, which uses Node 24.

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -15,7 +15,7 @@ jobs:
       - uses: actions/checkout@v7
         with:
           fetch-depth: 0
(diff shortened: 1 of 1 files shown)


nielspardon (approved):
LGTM

Issue #1185, opened 2026-08-28
core: RelCopyOnWriteVisitor throws for Expand, ExtensionWrite, NamedDdl and ExtensionDdl, so rewriting a plan holding one fails


nielspardon:
`RelCopyOnWriteVisitor` rebuilds a relation when something below it changed and returns `Optional.empty()` when nothing did, for every relation except four, which throw a bare `UnsupportedOperationException()` with no message: `Expand`, `ExtensionWrite`, `NamedDdl` and `ExtensionDdl`.

https://github.com/substrait-io/substrait-java/blob/main/core/src/main/java/io/substrait/relation/RelCopyOnWriteVisitor.java#L275-L306

The gap is already worked around once in main source, which is the clearest statement of what the base class should do. `OuterReferenceConverter.RelRewriter` overrides `visit(NamedDdl)` with the comment "The copy-on-write base throws for DDL; a view definition is an independent (top-level) query that may itself contain correlated subqueries, so traverse it rather than throw", and then does exactly that traversal.

Where it is not worked around it fails. `SchemaCollector.TableGatherer` overrides `visit(NamedScan)` and `visit(NamedWrite)` but not `visit(NamedDdl)`, so `SubstraitToCalcite.convert(Rel)` without a catalog reader — the path that resolves the schema by walking the plan — throws the bare exception on any plan holding a `CREATE VIEW`, naming neither the relation nor the reason.

The asymmetry inside the class is the part worth closing first: `visit(NamedWrite)` rebuilds normally while `visit(NamedDdl)` throws, even though a `NamedDdl`'s `view_definition` is a `Rel` in exactly the way a `NamedWrite`'s `input` is. That matters more now that #1181 gives the DDL relations a declared schema to carry, since a consumer using this visitor to rewrite a plan is the case the schema exists for.

Each of the four can rebuild the way its siblings do — `NamedDdl` by traversing `view_definition` as `RelRewriter` already shows, `ExtensionWrite` on the model of `NamedWrite`, `Expand` through its input and its fields, `ExtensionDdl` through its own definition. Where one genuinely cannot be rewritten, the throw should at least name the relation it refused.

Found while reviewing #1181, whose new view round-trip test carries a comment explaining that it cannot use `assertFullRoundTrip` for this reason.

PR #1248, opened 2026-09-03
fix(core)!: visit a window relation's input when rewriting it

### LINKED ISSUE #1231: core: RelCopyOnWriteVisitor.visit(ConsistentPartitionWindow) never visits its input, so a rewrite below a window relation is silently dropped
## Description

`RelCopyOnWriteVisitor.visit(ConsistentPartitionWindow)` rewrites the relation's window functions, partition expressions and sorts, but never calls `getInput().accept(this, context)`. A rewrite that should apply below the window relation is silently discarded, and because the method decides "nothing changed" from those three lists alone, an input-only rewrite collapses the whole traversal to `Optional.empty()` rather than returning a partially-updated tree.

`ConsistentPartitionWindow` is the only `SingleInputRel` in `:core` with this omission. `Aggregate`, `Fetch`, `Filter`, `Project`, `Sort`, `TopN`, `ExtensionSingle`, `NamedWrite` and all five `AbstractExchangeRel` subclasses call `getInput().accept(this, context)`; `Expand`, `ExtensionWrite`, `NamedDdl` and `ExtensionDdl` throw instead, which is #1185. So this is a distinct defect from #1185 — that one fails loudly, this one loses the rewrite with no signal.

## Reproduction

Using the `negateI32LiteralsVisitor` helper already in `RelCopyOnWriteVisitorTest`, over the same `Project` input in both cases:

```java
Rel scan = sb.namedScan(Arrays.asList("t"), Arrays.asList("a"), Arrays.asList(R.I32));
Rel project = sb.project(in -> List.of(sb.i32(5)), sb.remap(1), scan);

// Filter: the input's i32(5) becomes i32(-5)
sb.filter(in -> sb.bool(true), project)
    .accept(negateI32LiteralsVisitor(), EmptyVisitationContext.INSTANCE);
// => Optional[Filter{input=Project{..., expressions=[I32Literal{value=-5}]}, ...}]

// ConsistentPartitionWindow over the identical Project: rewrite lost entirely
ConsistentPartitionWindow.builder()
    .input(project)
    .windowFunctions(Arrays.asList(/* lead:any, UNBOUNDED / CURRENT_ROW, RANGE */))
    .partitionExpressions(Collections.emptyList())
    .sorts(Collections.emptyList())
    .build()
    .accept(negateI32LiteralsVisitor(), EmptyVisitationContext.INSTANCE);
// => Optional.empty
```

## Suggested fix

Visit the input alongside the other three lists, fold it into the `allEmpty(...)` check, and set it on the builder — the shape `visit(Aggregate)` already uses. A test that nests a rewritable relation under a `ConsistentPartitionWindow` would pin it; the existing window CoW tests all use a bare `namedScan` input, which has nothing to rewrite.

🤖 Generated with AI

alexandrefimov:
`RelCopyOnWriteVisitor.visit(ConsistentPartitionWindow)` rewrote the window functions, the partition expressions and the sorts, but never visited the relation's input. A rewrite that belonged below a window was therefore dropped, and one that applied nowhere else made the whole traversal come back as `Optional.empty()` — the visitor's way of saying nothing changed. It visits the input now and counts it in that decision. `visit(MultiBucketExchange)` computed a rewritten expression and then left it out of the same decision, so it counts that too.

`OuterReferenceConverter` is built on this visitor, which is where the defect was reachable from outside: a correlated subquery in a window's input never got to it, so neither direction of the offset-based/id-based conversion touched the reference. The new tests pin that, and drive the input check from the shared samples so a relation added without visiting its input fails too — on the base that check names `ConsistentPartitionWindow`.

Closes #1231

BREAKING CHANGE: a plan carrying an `Expand`, an `ExtensionWrite`, a `NamedDdl` or an `ExtensionDdl` below a `ConsistentPartitionWindow` now throws `UnsupportedOperationException` from `io.substrait.relation.RelCopyOnWriteVisitor`, which used to stop above those relations and return quietly. Those four visits refuse rather than descend, so a consumer needing such a plan rewritten has to override them; `OuterReferenceConverter` runs this visitor, which makes the throw reachable without naming the class. Rewrites that belong below a window relation, or that touch only a `MultiBucketExchange`'s expression, now take effect instead of being discarded.

### DIFF
--- a/core/src/main/java/io/substrait/relation/RelCopyOnWriteVisitor.java
+++ b/core/src/main/java/io/substrait/relation/RelCopyOnWriteVisitor.java
@@ -392,7 +392,7 @@ public class RelCopyOnWriteVisitor<E extends Exception>
     Optional<Expression> expression =
         exchange.getExpression().accept(getExpressionCopyOnWriteVisitor(), context);
 
-    if (allEmpty(input)) {
+    if (allEmpty(input, expression)) {
       return Optional.empty();
     }
 
--- a/core/src/main/java/io/substrait/relation/RelCopyOnWriteVisitor.java
+++ b/core/src/main/java/io/substrait/relation/RelCopyOnWriteVisitor.java
@@ -611,6 +611,7 @@ public class RelCopyOnWriteVisitor<E extends Exception>
   public Optional<Rel> visit(
       ConsistentPartitionWindow consistentPartitionWindow, EmptyVisitationContext context)
       throws E {
+    Optional<Rel> input = consistentPartitionWindow.getInput().accept(this, context);
     Optional<List<ConsistentPartitionWindow.WindowRelFunctionInvocation>> windowFunctions =
         transformList(
             consistentPartitionWindow.getWindowFunctions(), context, this::visitWindowRelFunction);
@@ -622,13 +623,14 @@ public class RelCopyOnWriteVisitor<E extends Exception>
     Optional<List<Expression.SortField>> sorts =
         transformList(consistentPartitionWindow.getSorts(), context, this::visitSortField);
 
-    if (allEmpty(windowFunctions, partitionExpressions, sorts)) {
+    if (allEmpty(input, windowFunctions, partitionExpressions, sorts)) {
       return Optional.empty();
     }
 
     return Optional.of(
         ConsistentPartitionWindow.builder()
             .from(consistentPartitionWindow)
+            .input(input.orElse(consistentPartitionWindow.getInput()))
             .partitionExpressions(
                 partitionExpressions.orElse(consistentPartitionWindow.getPartitionExpressions()))
             .sorts(sorts.orElse(consistentPartitionWindow.getSorts()))
--- a/core/src/test/java/io/substrait/relation/RelCopyOnWriteVisitorTest.java
+++ b/core/src/test/java/io/substrait/relation/RelCopyOnWriteVisitorTest.java
@@ -7,6 +7,7 @@ import io.substrait.expression.Expression;
 import io.substrait.expression.WindowBound;
 import io.substrait.extension.DefaultExtensionCatalog;
 import io.substrait.extension.SimpleExtension;
+import io.substrait.relation.physical.MultiBucketExchange;
 import io.substrait.util.EmptyVisitationContext;
 import java.util.Arrays;
 import java.util.Collections;
@@ -105,11 +106,7 @@ class RelCopyOnWriteVisitorTest extends TestBase {
         window.accept(negateI32LiteralsVisitor(), EmptyVisitationContext.INSTANCE));
   }
 
-  /**
-   * The window relation's own lists are not the only thing below it. A rewrite that applies to the
-   * input has nowhere else to be found: a window relation is the only single-input relation in this
-   * visitor that used to decide "unchanged" without asking its input.
-   */
+  /** A rewrite that applies only below a window relation comes back with the input replaced. */
   @Test
   void consistentPartitionWindowRewritesItsInput() {
     SimpleExtension.WindowFunctionVariant declaration =
@@ -144,5 +141,28 @@ class RelCopyOnWriteVisitorTest extends TestBase {
             .input(sb.project(in -> Arrays.asList(sb.i32(-5)), sb.remap(1), scan))
             .build();
     assertEquals(Optional.of(expected), rewritten);
+  }
+
+  /**
+   * The same guard, one relation over: a rewrite that touches only the exchange's own expression
+   * comes back, where the input alone used to decide whether anything changed.
+   */
+  @Test
+  void multiBucketExchangeRewritesItsExpression() {
+    Rel scan = sb.namedScan(Arrays.asList("test"), Arrays.asList("a"), Arrays.asList(R.I32));
+    MultiBucketExchange exchange =
+        MultiBucketExchange.builder()
+            .input(scan)
+            .expression(sb.i32(5))
+            .constrainedToCount(true)
+            .partitionCount(1)
+            .build();
+
+    Optional<Rel> rewritten =
+        exchange.accept(negateI32LiteralsVisitor(), EmptyVisitationContext.INSTANCE);
+
+    assertEquals(
+        Optional.of(MultiBucketExchange.builder().from(exchange).expression(sb.i32(-5)).build()),
+        rewritten);
   }
 }
--- a/core/src/test/java/io/substrait/relation/RelCopyOnWriteVisitorTest.java
+++ b/core/src/test/java/io/substrait/relation/RelCopyOnWriteVisitorTest.java
@@ -104,4 +104,45 @@ class RelCopyOnWriteVisitorTest extends TestBase {
         Optional.empty(),
         window.accept(negateI32LiteralsVisitor(), EmptyVisitationContext.INSTANCE));
   }
+
+  /**
+   * The window relation's own lists are not the only thing below it. A rewrite that applies to the
+   * input has nowhere else to be found: a window relation is the only single-input relation in this
+   * visitor that used to decide "unchanged" without asking its input.
+   */
+  @Test
+  void consistentPartitionWindowRewritesItsInput() {
+    SimpleExtension.WindowFunctionVariant declaration =
+        extensions.getWindowFunction(
+            SimpleExtension.FunctionAnchor.of(
+                DefaultExtensionCatalog.FUNCTIONS_ARITHMETIC, "lead:any"));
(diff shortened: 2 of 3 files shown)


core/src/main/java/io/substrait/relation/RelCopyOnWriteVisitor.java
```diff
@@ -622,13 +623,14 @@ public Optional<Rel> visit(
     Optional<List<Expression.SortField>> sorts =
         transformList(consistentPartitionWindow.getSorts(), context, this::visitSortField);
 
-    if (allEmpty(windowFunctions, partitionExpressions, sorts)) {
+    if (allEmpty(input, windowFunctions, partitionExpressions, sorts)) {
```
nielspardon, line 626: Fold `expression` into the guard in `visit(MultiBucketExchange)` too — line 395 computes it and then checks `allEmpty(input)` only, so an expression-only rewrite is discarded exactly the way the window input was. `visit(SingleBucketExchange)` at line 376 already has the correct form. No suggestion block because 395 is outside this diff.

core/src/test/java/io/substrait/relation/RelCopyOnWriteVisitorTest.java
```diff
@@ -104,4 +104,45 @@ void consistentPartitionWindowWithUnchangedBoundsIsNotCopied() {
         Optional.empty(),
         window.accept(negateI32LiteralsVisitor(), EmptyVisitationContext.INSTANCE));
   }
+
+  /**
+   * The window relation's own lists are not the only thing below it. A rewrite that applies to the
+   * input has nowhere else to be found: a window relation is the only single-input relation in this
+   * visitor that used to decide "unchanged" without asking its input.
+   */
```
nielspardon: Say what the test asserts rather than what the visitor used to do — the past-tense uniqueness claim is accurate today but nothing enforces it.

```suggestion
  /** A rewrite that applies only below a window relation comes back with the input replaced. */
```

nielspardon (commented):
#1061 already carries this exact `visit(ConsistentPartitionWindow)` hunk plus two adjacent fixes in the same file. Landing this one first is fine — I'll drop that hunk and its window test from #1061 on rebase. One sibling worth folding in here, below.

core/src/main/java/io/substrait/relation/RelCopyOnWriteVisitor.java
alexandrefimov: Folded, and the breaking note names both relations now.

core/src/test/java/io/substrait/relation/RelCopyOnWriteVisitorTest.java
alexandrefimov: Taken as written.

core/src/test/java/io/substrait/relation/RelCopyOnWriteVisitorTest.java
```diff
@@ -104,4 +105,64 @@ void consistentPartitionWindowWithUnchangedBoundsIsNotCopied() {
         Optional.empty(),
         window.accept(negateI32LiteralsVisitor(), EmptyVisitationContext.INSTANCE));
   }
+
+  /** A rewrite that applies only below a window relation comes back with the input replaced. */
+  @Test
+  void consistentPartitionWindowRewritesItsInput() {
```
nielspardon, line 103: Worth adding the `RelSamples`-driven version alongside this: loop `samples()`, skip `getInputs().isEmpty()`, and assert a rewrite below each relation propagates, exempting `Expand` and `ExtensionWrite` (#1185). It fails on the base, so it guards this class of bug for the next relation added too.

core/src/test/java/io/substrait/relation/RelCopyOnWriteVisitorTest.java
```diff
@@ -104,4 +105,64 @@ void consistentPartitionWindowWithUnchangedBoundsIsNotCopied() {
...
+            SimpleExtension.FunctionAnchor.of(
+                DefaultExtensionCatalog.FUNCTIONS_ARITHMETIC, "lead:any"));
+    Rel scan = sb.namedScan(Arrays.asList("test"), Arrays.asList("a"), Arrays.asList(R.I32));
+    Rel input = sb.project(in -> Arrays.asList(sb.i32(5)), sb.remap(1), scan);
+    ConsistentPartitionWindow window =
+        ConsistentPartitionWindow.builder()
+            .input(input)
+            .windowFunctions(
+                Arrays.asList(
+                    ConsistentPartitionWindow.WindowRelFunctionInvocation.builder()
```
nielspardon: Third copy of this fixture in the file — the three differ only in the input and the bounds, so a private `windowOver(input, lower, upper)` helper would collapse all of them and take the no-op `.arguments(Collections.emptyList())` with it.

nielspardon (commented):
Add a case to `OuterReferenceConverterTest` for a correlated subquery under a window relation — that is the public-API bug this actually fixes, and a round trip cannot pin it because both directions were no-ops before, so assert on the id-based plan directly. Separately, consider dropping the `!` and the footer: #1136 was authored that way and merged as plain `fix(core):`. If you keep it, the footer should mention that descending now surfaces the #1185 `UnsupportedOperationException` for an `Expand` or DDL relation under a window.

core/src/test/java/io/substrait/relation/RelCopyOnWriteVisitorTest.java
alexandrefimov: Added as `everyRelationWithInputsPropagatesARewriteBelowIt`; on `fb6a54a` it names `ConsistentPartitionWindow`.

core/src/test/java/io/substrait/relation/RelCopyOnWriteVisitorTest.java
alexandrefimov: Collapsed into `windowOver(input, lower, upper)`, and the empty argument list went with it.

alexandrefimov:
Added `outerReferenceUnderAWindowRelationIsConverted` to `OuterReferenceConverterTest`, asserting the anchor and the `rel_reference` on the id-based plan. It fails on the base, `fb6a54a`.

Kept the `!`. #1136 changed a derived type and no consumer started throwing; here a plan with an `Expand`, an `ExtensionWrite` or a DDL relation below a window goes from a silent `Optional.empty()` to the `UnsupportedOperationException` those relations' visits throw, and the footer leads with that now rather than with the rewrite. Say the word if you would still rather it went out as a plain `fix(core):`.

The editorconfig job is red on its own download — `Error: The binary 'ec-linux-amd64*' not found`, from the action's `latest` lookup — and re-running it needs rights I do not have. Everything else on this head is green.

core/src/test/java/io/substrait/relation/RelCopyOnWriteVisitorTest.java
```diff
@@ -78,30 +86,88 @@ void consistentPartitionWindowBoundOffsetsAreRewritten() {
...
+        rewritten);
+  }
+
+  /**
+   * Every relation that has inputs hands back a rewrite made below it, so a relation added without
+   * visiting its input fails here rather than dropping rewrites silently. Driven by the shared
+   * samples, whose own test keeps them exhaustive over the model.
+   */
+  @Test
+  void everyRelationWithInputsPropagatesARewriteBelowIt() {
```
nielspardon, line 149: Rename-a-scan only exercises the relation half of the guard, so this would not have caught the `MultiBucketExchange` bug in this same PR — with the old production code it reports only `ConsistentPartitionWindow`, and the exchange passes. A second pass over the samples with the `negateI32LiteralsVisitor` above would close the expression half.

nielspardon (commented):
Keep the `!` — you are right, and the footer leads with the right thing now. The red editorconfig check is #1247's action bump on main, not yours: every PR based on `944b9213` fails the same way while #1250 on the older base passes, so it needs its own fix. One thing left, below.

nielspardon (approved):

core/src/test/java/io/substrait/relation/RelCopyOnWriteVisitorTest.java
alexandrefimov: Right, and it does not catch it: on the base visitor the sweep reports `ConsistentPartitionWindow` and the exchange passes. `negateI32LiteralsVisitor` could not close it either -- no sample carries an i32 literal, and widening it to i64 reaches only `Fetch`, `Filter` and `TopN`, never the exchange, whose sample expression is a field reference. So the second sweep marks every `FieldReference` instead and pins the ten relations whose samples carry one; on the base `MultiBucketExchange` is the one missing from that list, which is what the failure says.

PR #1249, opened 2026-09-03
fix(isthmus): report an unsupported type whatever its fields are called

### LINKED ISSUE #1244: isthmus: TypeConverter's unsupported-type message is passed to String.format as its format string
`TypeConverter`'s unsupported-type arm builds its message by concatenation and then hands the result to `String.format` as the format string, with no arguments:

```java
default:
      String.format("Unable to convert the type " + type.toString()));
```

So the exception *class* depends on how the type happens to render. A `%` in the rendered type reaches `String.format` as a conversion specifier, and callers catching `UnsupportedOperationException` see an unrelated unchecked exception escape instead:

```java
RelDataType struct = TYPE_FACTORY.createStructType(
    List.of(TYPE_FACTORY.createSqlType(SqlTypeName.INTEGER)), List.of("%s"));
TypeConverter.DEFAULT.toSubstrait(TYPE_FACTORY.createMultisetType(struct, -1));

// toString() = RecordType(INTEGER %s) MULTISET
// -> java.util.MissingFormatArgumentException: Format specifier '%s'
// (with a normal field name: UnsupportedOperationException, as intended)
```

A field name is the easiest carrier since `RelRecordType.generateTypeString` interpolates it verbatim, and `MULTISET` is one of the names that falls to this `default`. `TypeConverter.DEFAULT.toSubstrait(RelDataType)` is public, so this is reachable without a custom type; I did not find a route that gets there through SQL alone.

The `String.format` wrapper buys nothing here — drop it, or write `String.format("Unable to convert the type %s", type)`.

Scanning every `String.format` call in `isthmus/src/main`, `core/src/main`, `isthmus-cli/src/main` and `examples` for a single concatenated argument, this is the only remaining site. PR #585 swept the same anti-pattern out of `ArithmeticFunctionTest`; no static analysis would catch it, since the PMD ruleset enables no format-string rule and there is no errorprone/SpotBugs/Checkstyle in the build.

Noticed while reviewing #1150, whose new test asserts `UnsupportedOperationException` from this arm. Not caused by that PR and out of scope for it.

Measured on `main` at fb6a54ad403cb09680bdd37b0e40f15a0bd57b6e.

alexandrefimov:
`TypeConverter`'s unsupported-type arm built its message by concatenation and handed the result to `String.format` as the format string, so the exception class depended on how the type rendered: a `%` in a field name arrived as a conversion specifier, and a caller catching `UnsupportedOperationException` saw a `MissingFormatArgumentException` escape instead. `toSubstrait(RelDataType)` is public, so a struct field named `%s` inside a multiset reaches it with stock Calcite types. The message is unchanged, and #1244 reports this as the last concatenated format string in the main sources.

The test asserts that message rather than the exception class alone, because two of the five field names it covers never changed the class: `%n` and `%%` rewrote the message and left an `UnsupportedOperationException` behind, while `100%` threw an `UnknownFormatConversionException`.

Closes #1244

### DIFF
--- a/isthmus/src/test/java/io/substrait/isthmus/CalciteTypeTest.java
+++ b/isthmus/src/test/java/io/substrait/isthmus/CalciteTypeTest.java
@@ -365,4 +365,22 @@ class CalciteTypeTest extends CalciteObjs {
     assertEquals(expression, converter.toSubstrait(calciteType));
     assertEquals(calciteType, converter.toCalcite(type, expression, dfsFieldNames));
   }
+
+  /**
+   * The unsupported-type arm reports the type it was handed, and a type renders whatever a field
+   * name says. Passing that message to {@code String.format} as its format string made the
+   * exception class depend on the name: a {@code %s} in it arrived as a conversion specifier.
+   */
+  @Test
+  void anUnsupportedTypeIsReportedWhateverItsFieldsAreCalled() {
+    RelDataType struct =
+        type.createStructType(
+            Arrays.asList(type.createSqlType(SqlTypeName.INTEGER)), Arrays.asList("%s"));
+
+    UnsupportedOperationException e =
+        assertThrows(
+            UnsupportedOperationException.class,
+            () -> TypeConverter.DEFAULT.toSubstrait(type.createMultisetType(struct, -1)));
+    assertTrue(e.getMessage().contains("%s"), e.getMessage());
+  }
 }
--- a/isthmus/src/test/java/io/substrait/isthmus/CalciteTypeTest.java
+++ b/isthmus/src/test/java/io/substrait/isthmus/CalciteTypeTest.java
@@ -368,19 +368,26 @@ class CalciteTypeTest extends CalciteObjs {
 
   /**
    * The unsupported-type arm reports the type it was handed, and a type renders whatever a field
-   * name says. Passing that message to {@code String.format} as its format string made the
-   * exception class depend on the name: a {@code %s} in it arrived as a conversion specifier.
+   * name says. Passing that message to {@code String.format} as its format string made the name
+   * decide the outcome: {@code %s} and {@code %d} arrived as conversion specifiers and changed the
+   * exception class, while {@code %n} and {@code %%} changed the message instead and threw nothing.
+   *
+   * <p>Both fixture choices are forced. {@code MULTISET} is the only {@link SqlTypeName} reaching
+   * that arm whose rendering carries a field name -- {@code ROW}, {@code ARRAY} and {@code MAP}
+   * recurse past it and every other name renders a fixed string -- and {@code -1} is the only
+   * maximum cardinality {@code createMultisetType} accepts.
    */
-  @Test
-  void anUnsupportedTypeIsReportedWhateverItsFieldsAreCalled() {
+  @ParameterizedTest
+  @ValueSource(strings = {"%s", "%d", "100%", "%n", "%%"})
+  void anUnsupportedTypeIsReportedWhateverItsFieldsAreCalled(String fieldName) {
     RelDataType struct =
         type.createStructType(
-            Arrays.asList(type.createSqlType(SqlTypeName.INTEGER)), Arrays.asList("%s"));
+            Arrays.asList(type.createSqlType(SqlTypeName.INTEGER)), Arrays.asList(fieldName));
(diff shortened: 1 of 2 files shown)


isthmus/src/test/java/io/substrait/isthmus/CalciteTypeTest.java
```diff
@@ -365,4 +365,22 @@ private void testType(
...
+  void anUnsupportedTypeIsReportedWhateverItsFieldsAreCalled() {
+    RelDataType struct =
+        type.createStructType(
+            Arrays.asList(type.createSqlType(SqlTypeName.INTEGER)), Arrays.asList("%s"));
+
+    UnsupportedOperationException e =
+        assertThrows(
+            UnsupportedOperationException.class,
+            () -> TypeConverter.DEFAULT.toSubstrait(type.createMultisetType(struct, -1)));
+    assertTrue(e.getMessage().contains("%s"), e.getMessage());
```
nielspardon: Parameterize this over the specifiers and assert the exact message — as written it passes on the unfixed code for two of them. A field name of `%n` or `%%` never threw: `%n` rendered a newline into the message and `%%` collapsed to a single `%`, both still `UnsupportedOperationException`, so `assertThrows` on the class plus `contains("%s")` stays green. `100%` is a likelier column name than `%s` and threw a third class, `UnknownFormatConversionException`.

```suggestion
  /**
   * The unsupported-type arm reports the type it was handed, and a type renders whatever a field
   * name says. Passing that message to {@code String.format} as its format string made the name
   * decide the outcome: {@code %s} and {@code %d} arrived as conversion specifiers and changed the
   * exception class, while {@code %n} and {@code %%} changed the message instead and threw nothing.
   *
   * <p>Both fixture choices are forced. {@code MULTISET} is the only {@link SqlTypeName} reaching
   * that arm whose rendering carries a field name — {@code ROW}, {@code ARRAY} and {@code MAP}
   * recurse past it and every other name renders a fixed string — and {@code -1} is the only
   * maximum cardinality {@code createMultisetType} accepts.
   */
  @ParameterizedTest
  @ValueSource(strings = {"%s", "%d", "100%", "%n", "%%"})
  void anUnsupportedTypeIsReportedWhateverItsFieldsAreCalled(String fieldName) {
    RelDataType struct =
        type.createStructType(
            Arrays.asList(type.createSqlType(SqlTypeName.INTEGER)), Arrays.asList(fieldName));
    RelDataType multiset = type.createMultisetType(struct, -1);

    UnsupportedOperationException e =
        assertThrows(
            UnsupportedOperationException.class, () -> TypeConverter.DEFAULT.toSubstrait(multiset));
    assertEquals("Unable to convert the type " + multiset, e.getMessage());
```

nielspardon (commented):
You renamed this away from the commit's own title — was that deliberate? `report an unsupported type whatever its fields are called` is the more accurate changelog line: the current one reads as a universal claim, and a `RelCrossType` from `RelDataTypeFactory.createJoinType` still NPEs at `switch (type.getSqlTypeName())` before it can reach this arm. Either way, please move commit `5779cb31`'s body into the description — squash-merge keeps the PR body, not the commit's, so the rationale you wrote there is what gets dropped.

isthmus/src/test/java/io/substrait/isthmus/CalciteTypeTest.java
alexandrefimov: Taken as written, with the dashes as `--` like the rest of the file.

isthmus/src/test/java/io/substrait/isthmus/CalciteTypeTest.java
```diff
@@ -365,4 +365,29 @@ private void testType(
     assertEquals(expression, converter.toSubstrait(calciteType));
     assertEquals(calciteType, converter.toCalcite(type, expression, dfsFieldNames));
   }
+
+  /**
```
nielspardon: Move this above the `testType` helpers at line 330 — it's now the only test below the helper block, so the next reader scanning for tests stops there.

nielspardon (commented):
One leftover from the first pass.

isthmus/src/test/java/io/substrait/isthmus/CalciteTypeTest.java
alexandrefimov: Moved above the block.

nielspardon (approved):

PR #1250, opened 2026-09-03
fix(examples): print a day-time interval under its own label

### LINKED ISSUE #1237: examples: ExpressionStringify prints day-time intervals under the IntervalYearLiteral label
`ExpressionStringify.visit(IntervalDayLiteral, ...)` (examples/substrait-spark/src/main/java/io/substrait/examples/util/ExpressionStringify.java:133) prints its output under the `IntervalYearLiteral` label — a copy-paste from the method directly above it:

```java
public String visit(IntervalYearLiteral expr, EmptyVisitationContext context) {
  return "<IntervalYearLiteral " + expr.months() + " " + expr.years() + ">";
}

public String visit(IntervalDayLiteral expr, EmptyVisitationContext context) {
  return "<IntervalYearLiteral " + expr.seconds() + " " + expr.days() + ">";
}
```

So a plan carrying both interval kinds dumps two entries that both say `IntervalYearLiteral`, and nothing distinguishes them.

Compounding it, both methods print their components in reverse declaration order with no field labels — months before years, seconds before days — as does `visit(IntervalCompoundLiteral, ...)` for all five of its fields. `<IntervalYearLiteral 1 12>` could be 1 year 12 months, 12 years 1 month, or a day-time interval of 1 second and 12 days.

`ExpressionStringify` is one of the direct `ExpressionVisitor` implementors, so it is the tool a maintainer reaches for when chasing exactly the kind of interval mis-conversion #1138 describes.

## Fix

Correct the label, and label the components while there:

```java
return "<IntervalDayLiteral days=" + expr.days() + " seconds=" + expr.seconds() + ">";
```

with the same treatment for `IntervalYearLiteral` and `IntervalCompoundLiteral`.

Found while reviewing #1140.

alexandrefimov:
All three interval cases now name their components, and the day-time one prints under its own label instead of `<IntervalYearLiteral`.

Past your snippet in one respect: `IntervalDayLiteral` and `IntervalCompoundLiteral` also print `precision`, since a `subseconds` count cannot be read without the scale it is counted in.

Closes #1237

### DIFF
--- a/examples/substrait-spark/src/main/java/io/substrait/examples/util/ExpressionStringify.java
+++ b/examples/substrait-spark/src/main/java/io/substrait/examples/util/ExpressionStringify.java
@@ -124,13 +124,21 @@ public class ExpressionStringify extends ParentStringify
   @Override
   public String visit(IntervalYearLiteral expr, EmptyVisitationContext context)
       throws RuntimeException {
-    return "<IntervalYearLiteral " + expr.months() + " " + expr.years() + ">";
+    return "<IntervalYearLiteral years=" + expr.years() + " months=" + expr.months() + ">";
   }
 
   @Override
   public String visit(IntervalDayLiteral expr, EmptyVisitationContext context)
       throws RuntimeException {
-    return "<IntervalYearLiteral " + expr.seconds() + " " + expr.days() + ">";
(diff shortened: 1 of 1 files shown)


nielspardon (approved):
LGTM

Issue #1173, opened 2026-08-26
isthmus: scalar function options are never written and are dropped when read, so an option-selected behaviour is lost


alexandrefimov:
A scalar function option present in a plan does not survive a round trip through Isthmus. Converting a `concat` invocation that carries `null_handling` to Calcite and back:

```
before: options=[FunctionOption{name=null_handling, values=[ACCEPT_NULLS]}]
after:  options=[]
```

Nothing sets them on the way out either — `ScalarFunctionConverter.generateBinding` builds the invocation from `outputType`, `declaration` and `addAllArguments` and stops there. So Isthmus writes scalar invocations without options and silently discards the ones it reads.

The aggregate side treats them as load-bearing, which is the contrast that makes this look unintended rather than decided: `AggregateFunctionConverter` sets `.options(binding.function().options())`, and on the way back `SubstraitRelNodeConverter.carriesOpaqueSemantics` refuses to reconstruct an invocation whose option set is non-empty, on the grounds that two invocations differing only in options would be indistinguishable once converted. That reasoning applies to scalar functions equally.

## Why an absent option is not a harmless omission

An option has no default, and cannot have one: in `simple_extensions_schema.yaml` an option is an object with `required: [values]` and properties `description` and `values` — there is no `default` key in the schema at all. So an unspecified option is undetermined rather than defaulted, and a consumer may pick either value.

For `concat` the two values differ on ordinary data. The declaration says it outright: `IGNORE_NULLS` skips null arguments, `ACCEPT_NULLS` makes the result null if any argument is null. Measured on a live Impala cluster by the consumer who ran into this — their numbers, not mine — `concat('a', cast(null as string))` returns `NULL`, i.e. `ACCEPT_NULLS`; `IGNORE_NULLS` would return `'a'` for the same call. Isthmus converting Calcite's concatenation, whose SQL semantics propagate null, has a specific one of the two in mind and emits neither.

## What this does not fix

[...]

Issue #1117, opened 2026-08-18
isthmus: Calcite→Substrait takes the function output type from the Rex node, so date ± interval_day is emitted with a non-conformant output type


alexandrefimov:
`FunctionFinder` takes the Substrait output type from the Calcite node it is converting (`FunctionConverter.java:744`):

```java
Type outputType = typeConverter.toSubstrait(call.getType());
```

The return type declared by the matched function variant is never consulted, so whatever Calcite inferred is what lands in the plan. For `date - interval_day` that means Isthmus emits an invocation the spec does not declare.

`SELECT event_date - INTERVAL '5' DAY FROM events` over `CREATE TABLE events (event_id INT, event_date DATE, ...)`:

```
key             = subtract:date_iday
declared output = Date{nullable=true}
args            = [FieldReference{type=Date{nullable=true}},
                   IntervalDayLiteral{days=5, precision=6}]
```

`functions_datetime.yaml` declares `subtract(date, interval_day<P>) -> precision_timestamp<P>`. Calcite's `MINUS_DATE` infers `DATE` for a DAY interval, and that is what is written out. `add:date_iday` behaves the same way through `DATETIME_PLUS`.

The plan is self-consistent, so it round-trips through Isthmus unnoticed; a consumer that validates the invocation against the YAML sees the wrong output type.

#1116 covers the Substrait -> Calcite half of the same disagreement, and #1099 fixes only that direction for subtraction — its `DATETIME_SUBTRACT` operator never sees the plans Isthmus itself produces from SQL, because those are built from a stock Calcite operator.

On the `P` in `precision_timestamp<P>`: on the SQL path it is 6, because that is the scale Calcite gives every parsed interval — `INTERVAL '5' DAY` is `INTERVAL DAY` with `scale=6`. Until #1120, `LiteralConverter` reported 6 regardless of what the literal's type carried; now it reads the type, which on this path still says 6. A narrower `P` reaches here only from a Substrait plan converted to Calcite and back, which is what #1118 was about.

Raised by @nielspardon while reviewing #1099.

nielspardon:
Two notes on the last paragraph, since they point at different fixes.

[...]

PR #1251, opened 2026-09-03
feat(isthmus): convert REVERSE and INITCAP from Calcite

### LINKED ISSUE #1222: isthmus: REVERSE and INITCAP have no `FunctionMappings` entry, so neither converts from Calcite
`REVERSE` and `INITCAP` exist on both sides of the Calcite ⇄ Substrait boundary, yet neither converts:

```
SELECT REVERSE(l_comment) FROM lineitem
  -> IllegalArgumentException: Unable to convert call REVERSE(varchar<44>?).
SELECT INITCAP(l_comment) FROM lineitem
  -> IllegalArgumentException: Unable to convert call INITCAP(varchar<44>?).
```

Both are declared in `functions_string.yaml` (spec v0.102.0) with character implementations, and both have a Calcite operator — `SqlLibraryOperators.REVERSE` and `SqlStdOperatorTable.INITCAP`. What is missing is the entry that joins them: `FunctionConverter` builds its `signatures` map exclusively from the `FunctionMappings` sig list, with no fall-back that matches an operator to a same-named Substrait function, so an operator absent from that table simply has no `FunctionFinder` and every call to it fails.

`AutomaticDynamicFunctionMappingConverterProvider` does not cover this. It synthesises Calcite operators for unmapped *Substrait* functions, which is the Substrait → Calcite direction; a real Calcite operator going the other way still needs a finder, and both queries above fail identically under it.

The fix is two entries:

```java
s(SqlLibraryOperators.REVERSE, "reverse"),
s(SqlStdOperatorTable.INITCAP, "initcap"),
```

Passing exactly those to `ScalarFunctionConverter` as `additionalSignatures` makes both queries convert, so nothing beyond the mapping is missing — the yaml signatures already match a `varchar` argument.

Two notes for whoever picks this up. 42 of the 126 scalar function names have no `FunctionFinder`, so this is a sample of a wider gap rather than the whole of it; `reverse` and `initcap` are called out because both were confirmed to have a Calcite counterpart whose signature already lines up, which is not true of every name on that list. And if you take an inventory yourself, `trim`/`ltrim`/`rtrim` will look unbound when they are not: all three map onto the single Calcite `TRIM` operator and `signatures` is keyed by operator, so two of the three are overwritten — inert here, because `TrimFunctionMapper` handles `TRIM` ahead of the finder path.

#1012 would restructure this mapping wholesale; these two entries are worth adding independently of it. #781 is the earlier instance of the same shape, for `ANY_VALUE`.

Found while reviewing #1169. Checked on `main` @ 85c03c76.

alexandrefimov:
`REVERSE` and `INITCAP` had no `FunctionMappings` entry, and `FunctionConverter` builds its signature map from that list alone, so neither had a `FunctionFinder` and no call to either converted. `functions_string.yaml` declares both, so the mapping was the whole gap; without it a call fails with `Unable to convert call REVERSE(char<16>?)`.

`INITCAP` binds Calcite's operator. `REVERSE` binds a local `SqlFunction` with `ARG0_NULLABLE`, registered in `SUBSTRAIT_SCALAR_OPERATOR_TABLE` where a lookup reaches it before the libraries, because Calcite's is wrong twice over. `SqlLibraryOperators.REVERSE` infers its return with `ARG0_NULLABLE_VARYING`, which widens a `CHAR`: `reverse(c16)` bound the `fixedchar` variant and then recorded `varchar<16>` as its output type, where the declared `fixedchar<L1>` makes it `fixedchar<16>`. And Calcite gives the Spark library's operator the same name, so with both libraries enabled the validator finds two candidates and discards both, which is why an `ORDER BY` over the call failed to validate. The same collision catches `least` and `greatest`, filed as #1274.

Neither fault is visible to a round trip, which carries the recorded `output_type` verbatim in both directions, so the two new tests assert the invocation's output type and the three `ORDER BY` forms directly.

Closes #1222

### DIFF
--- a/isthmus/src/main/java/io/substrait/isthmus/expression/FunctionMappings.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/expression/FunctionMappings.java
@@ -76,6 +76,10 @@ public class FunctionMappings {
    * Substrait declaration says {@code fixedchar<L1>}. Its name also collides with the Spark library
    * operator Calcite likewise calls {@code REVERSE}, which leaves neither candidate reachable once
    * both libraries are enabled. Isthmus defines its own to keep the declared return and the name.
+   *
+   * <p>Calcite's Enumerable convention cannot execute this operator because it has no {@code
+   * RexImpTable} implementor, unlike the library operator it replaces. It is used to preserve
+   * conversion semantics and operator identity.
    */
   public static final SqlFunction REVERSE =
       SqlBasicFunction.create(
--- a/isthmus/src/main/java/io/substrait/isthmus/expression/FunctionMappings.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/expression/FunctionMappings.java
@@ -71,6 +71,17 @@ public class FunctionMappings {
           OperandTypes.family(SqlTypeFamily.INTEGER, SqlTypeFamily.INTEGER));
 
   /**
+   * The {@code REVERSE(string)} function. Calcite's own is {@link SqlLibraryOperators#REVERSE},
+   * whose {@code ARG0_NULLABLE_VARYING} return widens a {@code CHAR} to {@code VARCHAR}, where the
+   * Substrait declaration says {@code fixedchar<L1>}. Its name also collides with the Spark library
+   * operator Calcite likewise calls {@code REVERSE}, which leaves neither candidate reachable once
+   * both libraries are enabled. Isthmus defines its own to keep the declared return and the name.
+   */
+  public static final SqlFunction REVERSE =
+      SqlBasicFunction.create(
+          "REVERSE", ReturnTypes.ARG0_NULLABLE, OperandTypes.CHARACTER, SqlFunctionCategory.STRING);
+
+  /**
    * The Substrait datetime subtraction function.
    *
    * <p>Calcite's datetime subtraction operators either infer their return type from a third
@@ -133,7 +144,7 @@ public class FunctionMappings {
               s(SqlStdOperatorTable.LOWER, "lower"),
               s(SqlStdOperatorTable.UPPER, "upper"),
               s(SqlStdOperatorTable.INITCAP, "initcap"),
-              s(SqlLibraryOperators.REVERSE, "reverse"),
+              s(REVERSE, "reverse"),
               s(SqlStdOperatorTable.BETWEEN),
               s(SqlStdOperatorTable.IS_NOT_DISTINCT_FROM, "is_not_distinct_from"),
               s(SqlStdOperatorTable.COALESCE, "coalesce"),
--- a/isthmus/src/main/java/io/substrait/isthmus/expression/FunctionMappings.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/expression/FunctionMappings.java
@@ -132,6 +132,8 @@ public class FunctionMappings {
               s(SqlStdOperatorTable.CHAR_LENGTH, "char_length"),
               s(SqlStdOperatorTable.LOWER, "lower"),
               s(SqlStdOperatorTable.UPPER, "upper"),
+              s(SqlStdOperatorTable.INITCAP, "initcap"),
+              s(SqlLibraryOperators.REVERSE, "reverse"),
               s(SqlStdOperatorTable.BETWEEN),
               s(SqlStdOperatorTable.IS_NOT_DISTINCT_FROM, "is_not_distinct_from"),
               s(SqlStdOperatorTable.COALESCE, "coalesce"),
--- a/isthmus/src/test/java/io/substrait/isthmus/StringFunctionTest.java
+++ b/isthmus/src/test/java/io/substrait/isthmus/StringFunctionTest.java
@@ -54,6 +54,20 @@ final class StringFunctionTest extends PlanTestBase {
 
   @ParameterizedTest
   @ValueSource(strings = {"c16", "vc32"})
+  void initcap(String column) throws Exception {
+    String query = String.format("SELECT initcap(%s) FROM strings", column);
+    assertFullRoundTrip(query, CREATES);
+  }
+
+  @ParameterizedTest
+  @ValueSource(strings = {"c16", "vc32"})
+  void reverse(String column) throws Exception {
+    String query = String.format("SELECT reverse(%s) FROM strings", column);
+    assertFullRoundTrip(query, CREATES);
+  }
+
+  @ParameterizedTest
+  @ValueSource(strings = {"c16", "vc32"})
   void replace(String column) throws Exception {
     String query =
         String.format("SELECT replace(%s, replace_from, replace_to) FROM replace_strings", column);
--- a/isthmus/src/test/java/io/substrait/isthmus/StringFunctionTest.java
+++ b/isthmus/src/test/java/io/substrait/isthmus/StringFunctionTest.java
@@ -8,6 +8,7 @@ import io.substrait.expression.Expression.FixedCharLiteral;
 import io.substrait.isthmus.sql.SubstraitCreateStatementParser;
 import io.substrait.plan.Plan;
 import io.substrait.relation.Project;
+import io.substrait.type.TypeCreator;
 import java.util.List;
 import org.apache.calcite.prepare.CalciteCatalogReader;
 import org.apache.calcite.sql.parser.SqlParseException;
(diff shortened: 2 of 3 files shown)


isthmus/src/main/java/io/substrait/isthmus/expression/FunctionMappings.java
```diff
@@ -132,6 +132,8 @@ public class FunctionMappings {
               s(SqlStdOperatorTable.CHAR_LENGTH, "char_length"),
               s(SqlStdOperatorTable.LOWER, "lower"),
               s(SqlStdOperatorTable.UPPER, "upper"),
+              s(SqlStdOperatorTable.INITCAP, "initcap"),
+              s(SqlLibraryOperators.REVERSE, "reverse"),
```
nielspardon: `SqlLibraryOperators.REVERSE` is the wrong operator to bind, for two independent reasons.

It is unreachable in an `ORDER BY`. `SubstraitOperatorTable` enables both `SPARK` and `BIG_QUERY`, and Calcite names `REVERSE_SPARK` "REVERSE" as well, so `lookupOperatorOverloads` returns two candidates, `filterOperatorRoutinesByKind` keeps neither, and validation fails:

```
SELECT c16 FROM strings ORDER BY reverse(c16)     -> No match found for function signature REVERSE(<CHARACTER>)
SELECT reverse(c16) FROM strings ORDER BY 1       -> same
SELECT reverse(c16) AS r FROM strings ORDER BY r  -> same
```

And its `ReturnTypes.ARG0_NULLABLE_VARYING` widens CHAR to VARCHAR, so `reverse(c16)` emits `reverse:fchar` with `output_type = varchar<16>` where the declaration says `fixedchar<L1>` — against `algebra.proto`'s "Must be set to the return type of the function, exactly as derived using the declaration in the extension". That is the #1117 mechanism. `INITCAP` uses `ARG0_NULLABLE` and is unaffected.

An Isthmus-local operator fixes both, in the shape `RIGHTSHIFT` already uses in this file:

```suggestion
              s(REVERSE, "reverse"),
```

with, beside `RIGHTSHIFT`:

```java
/** Isthmus-local REVERSE: non-varying return type, matching the Substrait declaration. */
public static final SqlFunction REVERSE =
    SqlBasicFunction.create(
        "REVERSE", ReturnTypes.ARG0_NULLABLE, OperandTypes.CHARACTER, SqlFunctionCategory.STRING);
```

and adding `FunctionMappings.REVERSE` to `SUBSTRAIT_SCALAR_OPERATOR_TABLE` in `SubstraitOperatorTable`, which is consulted first and early-returns. I verified that combination: overload count 2 -> 1, all three queries above pass, `reverse:fchar` emits `fixedchar<16>`, `:isthmus:check` and `:isthmus-cli:check` green.

Worth saying that `least` and `greatest` are already broken the same way, so the collision is a pre-existing trap rather than anything you introduced — I filed it as #1274. If you would rather keep this PR to the two entries, the local-operator half alone still fixes the output type; say so and the `ORDER BY` half can go with #1274 instead.

isthmus/src/main/java/io/substrait/isthmus/expression/FunctionMappings.java
```diff
@@ -132,6 +132,8 @@ public class FunctionMappings {
               s(SqlStdOperatorTable.CHAR_LENGTH, "char_length"),
               s(SqlStdOperatorTable.LOWER, "lower"),
               s(SqlStdOperatorTable.UPPER, "upper"),
+              s(SqlStdOperatorTable.INITCAP, "initcap"),
```
nielspardon, line 150: Not for this PR, just so it is on the record: every `initcap` implementation declares `options: char_set: [UTF8, ASCII_ONLY]` and Isthmus emits no `FunctionOption`, while Calcite's `SqlFunctions.initcap` treats only `[A-Za-z0-9]` as word characters. So `initcap('élan vital')` is `'éLan Vital'` here but `'Élan Vital'` for a consumer that picks UTF8, and the plan pins neither. Legal per `algebra.proto` ("If omitted, the consumer chooses its own behavior") and tracked by #1173 — noting it because `lower`/`upper` are Unicode-aware, so their precedent for dropping the option does not carry to `initcap`.

isthmus/src/test/java/io/substrait/isthmus/StringFunctionTest.java
```diff
@@ -52,6 +52,20 @@ void upper(String column) throws Exception {
     assertFullRoundTrip(query, CREATES);
   }
 
+  @ParameterizedTest
+  @ValueSource(strings = {"c16", "vc32"})
```
nielspardon: Add the fixture's third column: `CREATES` already declares `vc VARCHAR`, which exercises the `initcap:str` implementation — the one a producer emitting unbounded strings hits, and currently the only one of the three left untested.

```suggestion
  @ValueSource(strings = {"c16", "vc32", "vc"})
```

isthmus/src/test/java/io/substrait/isthmus/StringFunctionTest.java
```diff
@@ -52,6 +52,20 @@ void upper(String column) throws Exception {
...
+  void initcap(String column) throws Exception {
+    String query = String.format("SELECT initcap(%s) FROM strings", column);
+    assertFullRoundTrip(query, CREATES);
+  }
+
+  @ParameterizedTest
+  @ValueSource(strings = {"c16", "vc32"})
```
nielspardon: Same here for `reverse:str`:

```suggestion
  @ValueSource(strings = {"c16", "vc32", "vc"})
```

Separately, worth knowing what these two tests can and cannot catch: `assertFullRoundTrip` asserts only POJO1 == POJO2 == POJO3, and both directions carry the recorded `output_type` verbatim, so the `varchar<16>` mismatch above round-trips green. If you want them to pin the mapping, `testPositionParameterOrdering` further down this file asserts on the `ScalarFunctionInvocation` directly, and `assertSqlRoundTrip` (~line 161, used by the other `SqlLibraryOperators`-backed tests here) adds the Substrait -> SQL leg.

nielspardon (commented):
Map `reverse` to an Isthmus-local operator instead of `SqlLibraryOperators.REVERSE` — as written it is unreachable in an `ORDER BY` and emits a non-conformant output type, and one change fixes both. `initcap` is correct as it stands.

One note on the PR body, since title + body become the squash-merge commit that `semantic-release` turns into `CHANGELOG.md`: "exactly the ones you named" has no referent in `git log`, and the round-trip-test paragraph is what the diff already shows. The `FunctionConverter`-builds-its-map-from-`FunctionMappings`-alone sentence is the rationale worth keeping, and `Unable to convert call REVERSE(char<16>?)` is worth keeping as the symptom.

isthmus/src/main/java/io/substrait/isthmus/expression/FunctionMappings.java
alexandrefimov: Taken whole, both halves -- the local operator with `ARG0_NULLABLE` and the entry in `SUBSTRAIT_SCALAR_OPERATOR_TABLE`.

Two tests pin them, since as you say a round trip cannot: one asserts the invocation carries `fixedchar<16>` for `reverse(c16)`, the other that all three `ORDER BY` forms convert. On the library operator all four cases fail -- `expected: <FixedChar{nullable=true, length=16}> but was: <VarChar{nullable=true, length=16}>`, and your `No match found for function signature REVERSE(<CHARACTER>)` for the rest.

isthmus/src/main/java/io/substrait/isthmus/expression/FunctionMappings.java
alexandrefimov: Noted, and the distinction lands: `lower` and `upper` being Unicode-aware is exactly why their precedent does not carry here. Leaving it to #1173 -- writing an option would mean choosing a value on the consumer's behalf, which `algebra.proto` says is its choice when the field is omitted.

isthmus/src/test/java/io/substrait/isthmus/StringFunctionTest.java
alexandrefimov: Taken as written.

isthmus/src/test/java/io/substrait/isthmus/StringFunctionTest.java
alexandrefimov: Taken as written. The second half is answered by the two direct assertions above -- neither fault is one `assertFullRoundTrip` could have caught.

isthmus/src/main/java/io/substrait/isthmus/expression/FunctionMappings.java
```diff
@@ -70,6 +70,17 @@ public class FunctionMappings {
           ReturnTypes.ARG0_NULLABLE,
           OperandTypes.family(SqlTypeFamily.INTEGER, SqlTypeFamily.INTEGER));
 
+  /**
+   * The {@code REVERSE(string)} function. Calcite's own is {@link SqlLibraryOperators#REVERSE},
+   * whose {@code ARG0_NULLABLE_VARYING} return widens a {@code CHAR} to {@code VARCHAR}, where the
+   * Substrait declaration says {@code fixedchar<L1>}. Its name also collides with the Spark library
+   * operator Calcite likewise calls {@code REVERSE}, which leaves neither candidate reachable once
+   * both libraries are enabled. Isthmus defines its own to keep the declared return and the name.
+   */
```
nielspardon, line 83: Worth adding the `RexImpTable` caveat that `DATETIME_SUBTRACT` carries a few lines below, because this operator is the one case where it costs something. Confirmed by probing `RexImpTable.INSTANCE.get(op)`:

```
calcite REVERSE     kind=REVERSE          hasEnumerableImpl=true
isthmus REVERSE     kind=OTHER_FUNCTION   hasEnumerableImpl=false
isthmus RIGHTSHIFT  kind=OTHER_FUNCTION   hasEnumerableImpl=false
calcite INITCAP     kind=OTHER_FUNCTION   hasEnumerableImpl=true
```

`RIGHTSHIFT` and `DATETIME_SUBTRACT` replaced nothing — Calcite had no equivalent — so their lack of an implementor takes nothing away. This one displaces an operator that *did* have one, so a consumer executing Isthmus' Substrait→Calcite output on the Enumerable convention can no longer run `reverse`. That is clearly the right trade against an operator that is wrong twice over, but it is the kind of thing worth finding in the Javadoc rather than in a stack trace.

```suggestion
   *
   * <p>Calcite's Enumerable convention cannot execute this operator because it has no {@code
   * RexImpTable} implementor, unlike the library operator it replaces. It is used to preserve
   * conversion semantics and operator identity.
   */
```

nielspardon (commented):
Verified the fix: the local operator and the `SUBSTRAIT_SCALAR_OPERATOR_TABLE` entry both landed as discussed, and I reproduced your negative result — reverting just the two main-source files fails all four new cases, `expected: <FixedChar{nullable=true, length=16}> but was: <VarChar{nullable=true, length=16}>` plus the three `ORDER BY` forms. `:isthmus:check` and `:isthmus-cli:check` are green on the merge with current `main`, which has since moved to `substrait-packaging` 0.102.0 (`functions_string.yaml` is byte-identical to 0.101.0, so the declarations the change relies on are unchanged).

One small documentation point inline. The rewritten PR body reads well and `Unable to convert call REVERSE(char<16>?)` is the right thing to have kept.

isthmus/src/main/java/io/substrait/isthmus/expression/FunctionMappings.java
alexandrefimov: Taken as written, and I reproduced the probe before writing it into the source: `hasEnumerableImpl` is `true` for the library operator, `false` for this one and for `RIGHTSHIFT`. The distinction is the one that matters -- `RIGHTSHIFT` and `DATETIME_SUBTRACT` displaced nothing, so their own lack of an implementor took nothing away.

nielspardon (approved):
LGTM

Issue #1204, opened 2026-08-31
isthmus: the read relations drop filter, best_effort_filter and projection in both directions


nielspardon:
`ReadRel` carries `filter`, `best_effort_filter` and `projection` (a `MaskExpression`), and `AbstractReadRel` exposes all three. `SubstraitRelNodeConverter` reads none of them: `visit(NamedScan)` (205), `visit(LocalFiles)` (211) and `visit(VirtualTableScan)` (808) each build their Calcite relation from the schema alone, and `getFilter()` / `getBestEffortFilter()` / `getProjection()` appear nowhere in the file. `SubstraitRelVisitor` never sets them on the way back, so the loss is silent in both directions.

A read carrying a mandatory filter therefore converts to a Calcite plan that reads every row, with no error — wrong results rather than a refusal. The projection mask is worse than dropped: `AbstractReadRel.deriveRecordType` applies it (`MaskExpressionTypeProjector.project(projection, base)`), so the POJO's record type is the masked schema while the emitted Calcite row type is the unmasked one, and every field index a parent relation computed against the read is off.

```
NamedScan(names=[t], schema=[a i64, b string], filter=<pred>, projection=<mask>)
```
converts as if it were
```
NamedScan(names=[t], schema=[a i64, b string])
```

Refusing an unsupported field would be safer than dropping it, the way `visitOther` and `visit(Join)` refuse what they cannot express.

Reproduced on `main` at 7310fc8c. Distinct from #1160, which is about the emit mapping (`Rel.getRemap()`), not these three fields. Found while reviewing #1151.

PR #1260, opened 2026-09-03
fix: preserve mandatory read and post-join filters

bvolpato:
Mandatory ReadRel filters and JoinRel post-join filters are currently dropped when converting plans to Calcite or Spark. A read carrying FALSE therefore returns its input rows, and an outer join loses predicates that should inspect its null-extended output.

For example, this scan must produce no rows, but Isthmus currently returns an unrestricted TableScan:

```java
NamedScan scan = NamedScan.builder()
    .addNames("t")
    .initialSchema(NamedStruct.of(
        List.of("id"), TypeCreator.REQUIRED.struct(TypeCreator.REQUIRED.I32)))
    .filter(ExpressionCreator.bool(false, false))
    .build();
new SubstraitToCalcite(ConverterProvider.DEFAULT).convert(scan);
```

Apply mandatory scan filters before emit remapping and post-join filters after the join, using the actual output schema. Decode protobuf post-join references against that same schema so outer-join nullability is preserved and IS NULL cannot be incorrectly simplified away. These semantics follow spec v0.102.0.

Partially addresses #1204 for mandatory read filtering.

### DIFF
--- a/core/src/main/java/io/substrait/relation/ProtoRelConverter.java
+++ b/core/src/main/java/io/substrait/relation/ProtoRelConverter.java
@@ -1013,15 +1013,20 @@ public class ProtoRelConverter {
     Type.Struct unionedStruct = Type.Struct.builder().from(leftStruct).from(rightStruct).build();
     ProtoExpressionConverter converter =
         new ProtoExpressionConverter(lookup, extensions, unionedStruct, this);
+    Join.JoinType joinType = Join.JoinType.fromProto(rel.getType());
     ImmutableJoin.Builder builder =
         Join.builder()
             .left(left)
             .right(right)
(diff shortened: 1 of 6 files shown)


alexandrefimov:
Heads-up on an overlap: `ReadRel.projection`, one of the three fields #1204 names, is in #1280 -- it applies the mask in `visit(NamedScan)` and `visit(VirtualTableScan)`, where your `applyFilter` calls also sit.

The two compose: the filter stays on the scan, the mask goes above it, the emit mapping above that -- a filter is read against the direct schema, before the projection (spec v0.102.0, Read Filtering). Whichever of us lands second rebases, and I am glad to be that one.

Issue #1175, opened 2026-08-27
isthmus: NamedUpdate indexes the schema's depth-first name list with a top-level column ordinal, so it names the wrong update column


nielspardon:
`SubstraitRelNodeConverter.visit(NamedUpdate)` builds its update column list by indexing the schema's flattened name list with a *top-level* column ordinal:

https://github.com/substrait-io/substrait-java/blob/main/isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java#L721

`NamedStruct.names()` is one name per struct field at every level, in depth-first order, while `TransformExpression.getColumnTarget()` is an index among the top-level columns. The two agree only for a flat schema.

For a table whose schema is `struct(struct(i32, i32), i32)` the names are `["s", "x", "y", "n"]`, so an update targeting column 1 — the top-level column `n` — resolves to `x`, a nested field. The resulting `LogicalTableModify` names a column that does not exist at the top level, or silently the wrong one when a nested name collides with a real column.

This is the same flat-vs-depth-first confusion as #1152, which #1153 fixed for `VirtualTableScan`; the sibling method in the same file was left as it was. A struct-literal transformation targeting a struct column also carries Calcite's placeholder field names rather than the schema's, which is the other half of what #1153 fixed.

PR #1265, opened 2026-09-03
fix(isthmus): preserve update targets in nested schemas

bvolpato:
For a table src(s ROW(x INTEGER), x INTEGER, n INTEGER), UPDATE src SET n = 99 converts back to UPDATE src SET x = 99. The transform's top-level column ordinal is incorrectly used to index the flattened depth-first name list [S, X, X, N].

Reconstruct the declared row type before resolving target names so nested fields do not shift top-level update columns. This also preserves multiple-assignment order when nested and top-level names collide.

Partially addresses #1175; struct-literal assignment field names are separate from target-column resolution.

### DIFF
--- a/isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
@@ -771,8 +771,7 @@ public class SubstraitRelNodeConverter
     context.enterScope(AnchoredInput.of(update.getRelAnchor(), relBuilder.peek().getRowType()));
     RexNode condition = update.getCondition().accept(expressionRexConverter, context);
 
-    NamedStruct tableSchema = update.getTableSchema();
-    List<String> fieldNames = tableSchema.names();
+    List<String> fieldNames = toRowType(update.getTableSchema()).getFieldNames();
 
     List<String> updateColumnList = new ArrayList<>();
     List<RexNode> sourceExpressionList = new ArrayList<>();
@@ -1061,10 +1060,10 @@ public class SubstraitRelNodeConverter
   }
 
   /**
-   * Converts the schema a DDL relation declares into the row type that describes it.
+   * Converts a declared schema into the row type that describes it.
    *
    * @param schema the declared schema, whose names are one per field at every level of the struct
-   * @return the row type of the object the statement creates
+   * @return the row type with field names at their corresponding nesting levels
    */
   private RelDataType toRowType(NamedStruct schema) {
     return typeConverter.toCalcite(typeFactory, schema.struct(), schema.names());
--- a/isthmus/src/test/java/io/substrait/isthmus/NestedUpdateTargetTest.java
+++ b/isthmus/src/test/java/io/substrait/isthmus/NestedUpdateTargetTest.java
@@ -1,0 +1,53 @@
+package io.substrait.isthmus;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertInstanceOf;
+
+import io.substrait.isthmus.sql.SubstraitCreateStatementParser;
+import io.substrait.plan.Plan;
+import io.substrait.plan.PlanProtoConverter;
+import io.substrait.plan.ProtoPlanConverter;
+import io.substrait.relation.NamedUpdate;
+import java.util.List;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+import org.apache.calcite.prepare.Prepare;
+import org.apache.calcite.rel.core.TableModify;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.Arguments;
+import org.junit.jupiter.params.provider.MethodSource;
+
+class NestedUpdateTargetTest {
+
+  static Stream<Arguments> schemasAndTargets() {
+    return Stream.of(
+        Arguments.of("x INTEGER, n INTEGER", List.of("N")),
+        Arguments.of("s ROW(x INTEGER), x INTEGER, n INTEGER", List.of("N")),
+        Arguments.of("s ROW(x INTEGER), x INTEGER, n INTEGER", List.of("X")),
+        Arguments.of("s ROW(x INTEGER), x INTEGER, n INTEGER", List.of("N", "X")),
+        Arguments.of("x INTEGER, s ROW(n INTEGER), n INTEGER", List.of("X", "N")),
+        Arguments.of("s ROW(a ROW(x INTEGER), n INTEGER), x INTEGER, n INTEGER", List.of("N", "X")),
+        Arguments.of("s ROW(a INTEGER, b INTEGER), n INTEGER", List.of("N")));
+  }
+
+  @ParameterizedTest
+  @MethodSource("schemasAndTargets")
+  void preservesTopLevelUpdateTargets(String schema, List<String> targets) throws Exception {
+    Prepare.CatalogReader catalog =
+        SubstraitCreateStatementParser.processCreateStatementsToCatalog(
+            "CREATE TABLE src (" + schema + ")");
+    String assignments =
+        targets.stream().map(name -> name + " = 99").collect(Collectors.joining(", "));
+    Plan plan = new SqlToSubstrait().convert("UPDATE src SET " + assignments, catalog);
+    Plan decoded = new ProtoPlanConverter().from(new PlanProtoConverter().toProto(plan));
+    NamedUpdate update = assertInstanceOf(NamedUpdate.class, decoded.getRoots().get(0).getInput());
+
+    TableModify converted =
+        assertInstanceOf(
+            TableModify.class,
+            new SubstraitToCalcite(ConverterProvider.DEFAULT, catalog).convert(update));
+
+    assertEquals(targets, converted.getUpdateColumnList());
+    assertEquals(targets.size(), converted.getSourceExpressionList().size());
+  }
+}


nielspardon (commented):
Two changes to the conversion and one to the test. The fix for #1175 itself is right, and it matches the ordinal the outbound direction already emits, so this makes the two directions agree rather than picking a reading; the adjacent WHERE-clause gap is covered by your #1259.

One point sits outside the diff, so no suggestion is attached: `core/src/main/java/io/substrait/relation/AbstractUpdate.java:53`. `getColumnTarget()`'s Javadoc says only "the index of the target column to update", while `NamedStruct.names()` advertises the flattened depth-first list — the list this bug came from reaching for. The spec says only "index of the column to apply the transformation to" and never says which of the two it indexes, so it is worth stating on the accessor that it is a top-level ordinal into `getTableSchema().struct()`, and that this is the library's reading rather than something the spec settles.

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
```diff
@@ -771,8 +771,7 @@ public RelNode visit(NamedUpdate update, Context context) {
     context.enterScope(AnchoredInput.of(update.getRelAnchor(), relBuilder.peek().getRowType()));
     RexNode condition = update.getCondition().accept(expressionRexConverter, context);
 
-    NamedStruct tableSchema = update.getTableSchema();
-    List<String> fieldNames = tableSchema.names();
+    List<String> fieldNames = toRowType(update.getTableSchema()).getFieldNames();
```
nielspardon, line 774: Read the top-level names with the depth-first walk this class already uses in `applyOutputNames` instead of building the declared row type. `toRowType` runs the full Substrait→Calcite type mapping, so an UPDATE now fails because of a column it never touches: a flat `struct(uuid, i32)` with names `[U, N]` and `column_target = 1` returns `[N]` on `main` and throws `UnsupportedOperationException: Unknown expression type.` here — likewise for `interval_compound`, `precision_timestamp(9)` and any unmapped user-defined type. It only bites the supplied-catalog constructor, which is the foreign-plan path.

```suggestion
    NamedStruct tableSchema = update.getTableSchema();
    List<String> declaredNames = tableSchema.names();
    List<Type> schemaFields = tableSchema.struct().fields();
    List<String> fieldNames = new ArrayList<>(schemaFields.size());
    for (int field = 0, nameIndex = 0; field < schemaFields.size(); field++) {
      fieldNames.add(declaredNames.get(nameIndex));
      nameIndex += 1 + NamedFieldCountingTypeVisitor.countNames(schemaFields.get(field));
    }
```

This does add a fourth copy of that walk to this class, which #1178 tracks consolidating, so extracting the `applyOutputNames` one into a private helper would be better still — and either way `toRowType` goes back to serving only the DDL and write paths, so the Javadoc change below can be dropped.

isthmus/src/main/java/io/substrait/isthmus/SubstraitRelNodeConverter.java
```diff
@@ -771,8 +771,7 @@ public RelNode visit(NamedUpdate update, Context context) {
...
 
     List<String> updateColumnList = new ArrayList<>();
     List<RexNode> sourceExpressionList = new ArrayList<>();
```
nielspardon, line 777: Bound `column_target` before indexing it, at line 781 just below — that line is outside the diff, so I cannot attach this as a suggestion. Narrowing the names to the top-level fields also narrows the accepted range from the depth-first name count to the top-level field count, so a plan that previously resolved — wrongly, but without failing — now dies on a bare `ArrayIndexOutOfBoundsException` naming neither the update nor the table. Rejecting it is right, since Calcite's validator can only ever put a top-level name in `updateColumnList`, but it is worth saying so the way `Context.getInputFieldType` does in this class and the mirror guard does in `SubstraitRelVisitor`.

```java
      int columnTarget = transform.getColumnTarget();
      if (columnTarget < 0 || columnTarget >= fieldNames.size()) {
            String.format(
                "Update column target %d is outside the table schema's %d top-level columns",
                columnTarget, fieldNames.size()));
      }
      updateColumnList.add(fieldNames.get(columnTarget));
```

isthmus/src/test/java/io/substrait/isthmus/NestedUpdateTargetTest.java
```diff
@@ -0,0 +1,53 @@
...
+
+    TableModify converted =
+        assertInstanceOf(
+            TableModify.class,
+            new SubstraitToCalcite(ConverterProvider.DEFAULT, catalog).convert(update));
+
+    assertEquals(targets, converted.getUpdateColumnList());
+    assertEquals(targets.size(), converted.getSourceExpressionList().size());
+  }
+}
```
nielspardon, line 53: Assert the ordinal that crosses the wire, and make every case load-bearing. Three of these seven pass with the production fix reverted, because the nested field names collide with the top-level ones the buggy lookup lands on — and since both directions share one name list, emitting depth-first ordinals from `SubstraitRelVisitor` while reverting the line above keeps all seven green while changing every `column_target` on the wire. Renaming the nested fields, carrying the expected ordinals as test data and giving each assignment a distinct literal takes it to six of seven failing when the fix is reverted (the flat case cannot fail by construction) and pins the column-to-expression pairing, which `getSourceExpressionList().size()` cannot — Calcite's `TableModify` constructor already enforces that equality.

```suggestion
package io.substrait.isthmus;

import static org.junit.jupiter.api.Assertions.assertEquals;
import static org.junit.jupiter.api.Assertions.assertInstanceOf;

import io.substrait.isthmus.sql.SubstraitCreateStatementParser;
import io.substrait.plan.Plan;
import io.substrait.plan.PlanProtoConverter;
import io.substrait.plan.ProtoPlanConverter;
import io.substrait.relation.AbstractUpdate;
import io.substrait.relation.NamedUpdate;
import java.util.LinkedHashMap;
import java.util.List;
import java.util.Map;
import java.util.stream.Collectors;
import java.util.stream.IntStream;
import java.util.stream.Stream;
import org.apache.calcite.prepare.Prepare;
import org.apache.calcite.rel.core.TableModify;
import org.apache.calcite.rex.RexLiteral;
import org.junit.jupiter.params.ParameterizedTest;
import org.junit.jupiter.params.provider.Arguments;
import org.junit.jupiter.params.provider.MethodSource;

/**
 * A transformation's column target is an index among the schema's <em>top-level</em> columns, so
 * converting a {@link NamedUpdate} to Calcite must resolve it against the top-level names and not
 * against the depth-first name list a {@code NamedStruct} carries.
 */
class NestedUpdateTargetTest {

  /** The value assigned to the target at {@code index}; distinct per assignment. */
  private static int valueFor(int index) {
    return 11 * (index + 1);
  }

  static Stream<Arguments> schemasAndTargets() {
    return Stream.of(
        // A flat schema, where the depth-first and top-level names coincide: a control, and the
        // one shape whose target cannot shift.
        Arguments.of("x INTEGER, n INTEGER", List.of("N"), List.of(1)),
        // The nested field names below collide with nothing at the top level, so a top-level
        // ordinal read against the depth-first list cannot land on the expected name by accident.
        Arguments.of("s ROW(inner1 INTEGER), x INTEGER, n INTEGER", List.of("N"), List.of(2)),
        Arguments.of("s ROW(inner1 INTEGER), x INTEGER, n INTEGER", List.of("X"), List.of(1)),
        Arguments.of(
            "s ROW(inner1 INTEGER), x INTEGER, n INTEGER", List.of("N", "X"), List.of(2, 1)),
        Arguments.of(
            "x INTEGER, s ROW(inner1 INTEGER), n INTEGER", List.of("X", "N"), List.of(0, 2)),
        Arguments.of(
            "s ROW(inner1 ROW(deep1 INTEGER), inner2 INTEGER), x INTEGER, n INTEGER",
            List.of("N", "X"),
            List.of(2, 1)),
        Arguments.of("s ROW(inner1 INTEGER, inner2 INTEGER), n INTEGER", List.of("N"), List.of(1)));
  }

  @ParameterizedTest
  @MethodSource("schemasAndTargets")
  void preservesTopLevelUpdateTargets(
      String schema, List<String> targets, List<Integer> columnTargets) throws Exception {
    Prepare.CatalogReader catalog =
        SubstraitCreateStatementParser.processCreateStatementsToCatalog(
            "CREATE TABLE src (" + schema + ")");
    String assignments =
        IntStream.range(0, targets.size())
            .mapToObj(i -> targets.get(i) + " = " + valueFor(i))
            .collect(Collectors.joining(", "));
    Plan plan = new SqlToSubstrait().convert("UPDATE src SET " + assignments, catalog);
    Plan decoded = new ProtoPlanConverter().from(new PlanProtoConverter().toProto(plan));
    NamedUpdate update = assertInstanceOf(NamedUpdate.class, decoded.getRoots().get(0).getInput());

    // The ordinal that crosses the wire, asserted on its own: the two conversions agreeing with
    // each other does not establish which column a consumer will read.
    assertEquals(
        columnTargets,
        update.getTransformations().stream()
            .map(AbstractUpdate.TransformExpression::getColumnTarget)
            .collect(Collectors.toList()));

    TableModify converted =
        assertInstanceOf(
            TableModify.class,
            new SubstraitToCalcite(ConverterProvider.DEFAULT, catalog).convert(update));

    assertEquals(targets, converted.getUpdateColumnList());

    // Each target takes a distinct value, so this pins the column-to-expression pairing.
    // Calcite's TableModify constructor already requires the two lists to be the same length.
    Map<String, Integer> expected = new LinkedHashMap<>();
    IntStream.range(0, targets.size()).forEach(i -> expected.put(targets.get(i), valueFor(i)));
    Map<String, Integer> actual = new LinkedHashMap<>();
    for (int i = 0; i < converted.getUpdateColumnList().size(); i++) {
      RexLiteral value =
          assertInstanceOf(RexLiteral.class, converted.getSourceExpressionList().get(i));
      actual.put(converted.getUpdateColumnList().get(i), value.getValueAs(Integer.class));
    }
    assertEquals(expected, actual);
  }
}
```

Issue #1256, opened 2026-09-03
isthmus: SqlExpressionToSubstrait builds its base schema with the names-discarding toSubstrait, so nested field names are dropped


nielspardon:
`SqlExpressionToSubstrait` has its own private `toNamedStruct`, and it builds the struct with the single-argument `TypeConverter.toSubstrait(RelDataType)`:

```java
for (Map.Entry<String, RelDataType> entry : nameToTypeMap.entrySet()) {
  names.add(entry.getKey());
  types.add(converterProvider.getTypeConverter().toSubstrait(entry.getValue()));
}
return NamedStruct.of(names, Type.Struct.builder().fields(types).nullable(false).build());
```

That overload allocates a throwaway list for the depth-first field names and never hands it back, so only the top-level column names reach `names` while the struct keeps every nested field. `NamedStruct.of` does not check the two against each other, so the malformed schema is written out:

```
new SqlExpressionToSubstrait().convert("n + 1", List.of("CREATE TABLE t (s ROW(x INT, y VARCHAR), n BIGINT)"))

// baseSchema.names = [S, N]          -- 2 names
// baseSchema.struct = struct{ struct{ i32, string }, i64 }   -- 4 depth-first name slots
```

`TypeConverter.toNamedStruct` on the same row type produces `[s, x, y, n]`, which is what the plan path emits, so plans and extended expressions disagree on the same DDL. Reading the extended expression back and applying the names — `toCalcite(factory, struct, names)` — then indexes past the end of the list.

The fix is to call the existing `TypeConverter.toNamedStruct` rather than re-deriving the schema, which also removes one of the duplicate depth-first walkers noted in #1178.

Distinct from #1200 (the same class binds column indexes within their own table), #1201 and #1202.

Measured on `main` at 944b921303ddcfce5648fc6e5fd80540c651eaaf.

PR #1286, opened 2026-09-07
fix(isthmus): index extended expression fields against the combined schema

### LINKED ISSUE #1200: isthmus: SqlExpressionToSubstrait indexes a column within its own table, so every column outside the first converts to the wrong base-schema field
`SqlExpressionToSubstrait` binds each column name to a `RexInputRef` carrying that column's index **within its own table**, while the base schema it emits is the concatenation of **every** registered table. From the second table onward the two disagree, and the expression converts to a reference to whatever sits at that index in the concatenated schema.

`registerCreateTablesForExtendedExpression` builds both maps in one loop over each table's fields:

```java
nameToTypeMap.merge(field.getName(), field.getType(), ...);                        // base schema, in insertion order
nameToNodeMap.merge(field.getName(), new RexInputRef(field.getIndex(), ...), ...); // index within this table
```

`field.getIndex()` is the position in that table's row type. `nameToTypeMap` is a `LinkedHashMap` accumulated across all tables and `toNamedStruct` turns it into the base schema in insertion order, so the two agree only for the first table.

```
CREATE TABLE A (A1 BIGINT, A2 BIGINT, A3 BIGINT)
CREATE TABLE B (B1 BIGINT, B2 BIGINT)

                base schema              expression   emitted field
A3              [A1, A2, A3, B1, B2]     A3           2   correct
B2              [A1, A2, A3, B1, B2]     B2           1   wrong -- that is A2
```

With the 8-statement TPC-H DDL the schema has 61 fields and the skew is large: `L_ORDERKEY` sits at position 38 and emits `struct_field {}` (field 0, `P_PARTKEY`); `L_COMMENT` sits at 53 and emits `field: 15` (`S_COMMENT`). No exception, and the emitted field usually has a compatible type, so the wrong column is simply read.

A single registered table is unaffected, which is why the common case looks healthy — Arrow Java's documented use of this class (`ScanOptions.substraitFilter` / `substraitProjection`, built via `new SqlExpressionToSubstrait().convert(...)`) passes one table's DDL.

The existing tests cannot catch this: `SimpleExtendedExpressionsTest` asserts a proto → POJO → proto round trip, which is satisfied whatever the conversion produced, and every case references `L_ORDERKEY` under the multi-table TPC-H schema.

Two fixes are possible: make `nameToNodeMap` carry the running offset across tables, or bind names to the field of a single synthesized row type that `toNamedStruct` also reads. Either way a test should assert the emitted field index against the base schema position rather than round-tripping.

Reproduced on `main` @ 8088c0d2.

alexandrefimov:
Extended expressions share one `base_schema`, but references to fields from later tables used indices local to each table. After a three-column table, `B2` in a second table was emitted as field 1 instead of field 4.

Build references in the same registration order as the combined schema, following spec v0.102.0.

Closes #1200

### DIFF
--- a/isthmus/src/main/java/io/substrait/isthmus/SqlExpressionToSubstrait.java
+++ b/isthmus/src/main/java/io/substrait/isthmus/SqlExpressionToSubstrait.java
@@ -208,6 +208,8 @@ public class SqlExpressionToSubstrait extends SqlConverterBase {
         for (SubstraitTable t : tList) {
           rootSchema.add(t.getName(), t);
           for (RelDataTypeField field : t.getRowType(factory).getFieldList()) {
+            // Field references index the combined base schema in insertion order.
+            int fieldIndex = nameToTypeMap.size();
             nameToTypeMap.merge( // to validate the sql expression tree
                 field.getName(),
                 field.getType(),
@@ -217,7 +219,7 @@ public class SqlExpressionToSubstrait extends SqlConverterBase {
                 });
             nameToNodeMap.merge( // to convert sql expression into RexNode
                 field.getName(),
-                new RexInputRef(field.getIndex(), field.getType()),
+                new RexInputRef(fieldIndex, field.getType()),
                 (v1, v2) -> {
                   throw new IllegalArgumentException(
                       "There is no support for duplicate column names: " + field.getName());
--- a/isthmus/src/test/java/io/substrait/isthmus/SimpleExtendedExpressionsTest.java
+++ b/isthmus/src/test/java/io/substrait/isthmus/SimpleExtendedExpressionsTest.java
@@ -6,9 +6,11 @@ import static org.junit.jupiter.api.Assertions.assertThrows;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 
 import io.substrait.isthmus.expression.RexExpressionConverter;
+import io.substrait.proto.Expression;
 import io.substrait.proto.Expression.RexTypeCase;
 import io.substrait.proto.ExtendedExpression;
 import java.io.IOException;
+import java.util.List;
 import java.util.stream.Stream;
 import org.apache.calcite.sql.parser.SqlParseException;
 import org.junit.jupiter.api.Test;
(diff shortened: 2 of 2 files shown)


isthmus/src/main/java/io/substrait/isthmus/SqlExpressionToSubstrait.java
```diff
@@ -208,6 +208,8 @@ private Result registerCreateTablesForExtendedExpression(List<String> tables)
         for (SubstraitTable t : tList) {
           rootSchema.add(t.getName(), t);
           for (RelDataTypeField field : t.getRowType(factory).getFieldList()) {
+            // Field references index the combined base schema in insertion order.
+            int fieldIndex = nameToTypeMap.size();
```
nielspardon: `nameToTypeMap.size()` is the index into `base_schema.struct.types`, not into `names`, which the spec defines as the depth-first flattened list — the two coincide only while #1256 is open. Naming the right list keeps a future reader from retargeting the counter at `names`.

```suggestion
            // Index into base_schema.struct.types, whose order is this map's insertion order.
            // Read the size before merging, so it is the position this field will occupy.
            int fieldIndex = nameToTypeMap.size();
```

isthmus/src/main/java/io/substrait/isthmus/SqlExpressionToSubstrait.java
```diff
@@ -208,6 +208,8 @@ private Result registerCreateTablesForExtendedExpression(List<String> tables)
...
             nameToTypeMap.merge( // to validate the sql expression tree
```
nielspardon, line 226: Consider building one combined `RelDataType` with `factory.builder()` and letting `toNamedStruct` read that, so `RelDataTypeField.getIndex()` supplies the index and no counter is needed. That would also give `base_schema` the depth-first flattened names #1256 asks for, so it is scope expansion — your call whether it belongs here or stays a follow-up. As written the index space and the field space are two objects that agree only via the `LinkedHashMap` choice, the loop order, and reading `size()` before the merge.

isthmus/src/main/java/io/substrait/isthmus/SqlExpressionToSubstrait.java
```diff
@@ -208,6 +208,8 @@ private Result registerCreateTablesForExtendedExpression(List<String> tables)
...
                 field.getName(),
                 field.getType(),
```
nielspardon, line 228: Two items outside the diff, so I could not attach suggestions:

- `SqlExpressionToSubstrait.java:89` and `:102` — `convert`'s Javadoc still describes `createStatements` only as "table creation statements defining fields referenced by the expressions", but this PR makes three rules load-bearing for callers: `base_schema` concatenates tables in statement order then column order, bare column names must be unique across all tables, and references must be unqualified. It is the only place a consumer can learn what `base_schema`'s field order means.
- `SqlExpressionToSubstrait.java:213` — the duplicate-name `IllegalArgumentException` escapes a method declaring only `throws SqlParseException`, and is undocumented.

isthmus/src/test/java/io/substrait/isthmus/SimpleExtendedExpressionsTest.java
```diff
@@ -20,6 +22,78 @@ class SimpleExtendedExpressionsTest extends ExtendedExpressionTestBase {
...
+    // Reverse the expression order so a reference's index cannot accidentally be its position
+    // in the output expression list. All columns have the same type, so types cannot detect this.
+    String[] expressions = new String[columnNames.size()];
+    for (int index = 0; index < expressions.length; index++) {
+      expressions[index] = columnNames.get(columnNames.size() - index - 1);
+    }
+    ExtendedExpression converted = new SqlExpressionToSubstrait().convert(expressions, tables);
+
+    assertEquals(columnNames, converted.getBaseSchema().getNamesList());
+    assertEquals(columnNames.size(), converted.getBaseSchema().getStruct().getTypesCount());
```
nielspardon, line 57: Consider asserting through the reverse direction rather than the raw index: `new ProtoExtendedExpressionConverter().from(converted)` re-derives each reference's type from `base_schema`, so with `B (B1 BIGINT, B2 VARCHAR(5))` the `B2` reference comes back `VarChar(5)` after this fix and `I64` before it. That asserts the damage the issue describes instead of an index literal, and pins the types `getTypesCount()` only counts. Routing one multi-table case through a round trip would additionally make `FieldReference.StructField.constructOnRoot`'s bound check fire.

isthmus/src/test/java/io/substrait/isthmus/SimpleExtendedExpressionsTest.java
```diff
@@ -20,6 +22,78 @@ class SimpleExtendedExpressionsTest extends ExtendedExpressionTestBase {
...
+    assertEquals(expressions.length, converted.getReferredExprCount());
+    for (int index = 0; index < expressions.length; index++) {
+      assertEquals(
+          columnNames.size() - index - 1,
```
nielspardon: `columnNames.size() - index - 1` is both the expectation here and the input construction at line 49, so an edit to the reversal cancels out on both sides and the test stays green while asserting a different mapping. Asserting `columnNames.indexOf(expressions[index])` ties the expectation to the schema order instead.

isthmus/src/test/java/io/substrait/isthmus/SimpleExtendedExpressionsTest.java
```diff
@@ -20,6 +22,78 @@ class SimpleExtendedExpressionsTest extends ExtendedExpressionTestBase {
...
+    assertEquals(4, selectedField(multipleTables.getReferredExpr(0).getExpression()));
+    assertEquals(1, selectedField(singleTable.getReferredExpr(0).getExpression()));
+  }
+
+  private static int selectedField(Expression expression) {
+    assertEquals(RexTypeCase.SELECTION, expression.getRexTypeCase());
+    assertTrue(expression.getSelection().hasRootReference());
+    assertTrue(expression.getSelection().getDirectReference().hasStructField());
+    return expression.getSelection().getDirectReference().getStructField().getField();
+  }
```
nielspardon: Consider moving `selectedField` to `ExtendedExpressionTestBase` — `SubqueryPlanTest` open-codes this same walk eight times, and this is the only copy that also checks `RexTypeCase.SELECTION` and `hasRootReference()`. Note also that its three internal assertions run before the outer `assertEquals`, so the `expressions[index]` message at line 60 never applies when one of them fails.

nielspardon (commented):
Retitle this `fix(isthmus)!:` and add a `BREAKING CHANGE:` footer naming the changed indices — the emitted `field` moves for every caller that passes more than one CREATE statement, and `-c/--create` is repeatable, so the CLI reaches it too. Only the title reaches `CHANGELOG.md`, so as titled this is invisible to downstreams, and #1161, #1169, #1171, #1189, #1151 and #1248 all carry `!` plus a footer for exactly this shape. Also drop "following spec v0.102.0" from the body — v0.102.0 changed nothing about extended expressions, so the marker claims a spec change that did not happen.

The fix itself is right. The rest is inline: one design question about how the index is derived, plus test and doc suggestions.

isthmus/src/main/java/io/substrait/isthmus/SqlExpressionToSubstrait.java
alexandrefimov: Applied.

isthmus/src/main/java/io/substrait/isthmus/SqlExpressionToSubstrait.java
alexandrefimov: Leaving it to #1256: it also changes the names in `base_schema`.

isthmus/src/main/java/io/substrait/isthmus/SqlExpressionToSubstrait.java
alexandrefimov: Both added to `convert`'s Javadoc.

isthmus/src/test/java/io/substrait/isthmus/SimpleExtendedExpressionsTest.java
alexandrefimov: Done: `B2` is `VARCHAR(5)` now and comes back as `VarChar(5)`; before the fix, `I64`.

isthmus/src/test/java/io/substrait/isthmus/SimpleExtendedExpressionsTest.java
alexandrefimov: Done.

isthmus/src/test/java/io/substrait/isthmus/SimpleExtendedExpressionsTest.java
alexandrefimov: Kept local: `SubqueryPlanTest` extends `PlanTestBase` and reads outer references, so `hasRootReference()` does not hold there. The message now applies to the helper's assertions too.

Issue #1242, opened 2026-09-03
core: return-type derivation cannot evaluate a multi-line return program, so decimal arithmetic returns are not derivable


nielspardon:
Split out of #1139, which scoped the integer-parameter shapes (landed as #1141) and deferred this one: "a return program needs an assignment evaluator that does not exist at all".

A declaration may state its return as a multi-line program — assignments followed by a final type line — rather than a single type expression:

```yaml
        return: |-
          precision = integer_parameter(precision)
          precision_time<precision>
```

`TypeExpression.ReturnProgram` has no `visit` override in `ReturnTypeEvaluator`, so every such declaration reaches the throwing base. On the pinned catalog that is the fifteen variants `ParameterizedReturnTypeTest` pins as not-yet-derived:

```
add:dec_dec          divide:dec_dec       multiply:dec_dec     strptime_timestamp:str_str_i8
assume_timezone:date_str_i8              round:dec_i32        strptime_timestamp:str_str_str_i8
bitwise_and:dec_dec  floor:dec            strptime_time:str_str_i8
bitwise_or:dec_dec   modulus:dec_dec      subtract:dec_dec
bitwise_xor:dec_dec  ceil:dec
```

Worth singling out, as #1139 did: `add:dec_dec` is decimal addition — the case the `DECIMAL<P,S>` support exists for — and it is not derivable, because its return is a program rather than a type expression.

## What the work is

An assignment evaluator over the same `ParameterBindings` the single-line path already builds: evaluate each assignment's integer expression, write the result into the bindings, then evaluate the final type line — which is an ordinary type expression whose parameters are now bound. Note the final line is generally a shape #1141 already handles (`assume_timezone:date_str_i8` ends in `precision_timestamp_tz<precision>`, the decimal variants in `DECIMAL<prec, scale>`), so the return-program case reduces largely to evaluating the assignments.

The missing piece is integer-expression evaluation. Today `resolveInteger(String token)` takes a bare token and handles a literal or a bound name; a program's assignments need arithmetic, `min`/`max`, and the conditional form. Generalising it to evaluate a `TypeExpression` (integer literal, bound name, `BinaryOperation`, `IfOperation`) would cover the assignments and, at the same time, the single-line arithmetic forms below.

## Related gap worth folding in

[...]

PR #1288, opened 2026-09-07
feat(core): evaluate return-type programs and inline arithmetic

alexandrefimov:
Decimal arithmetic functions declare their return type as assignments followed by a final type expression. The evaluator currently rejects these programs, so their precision and scale cannot be derived from the function declaration.

Evaluate sequential assignments and arithmetic inside type parameters using the existing argument bindings. Support integer arithmetic, min/max, comparisons, boolean operations and conditionals from [spec v0.102.0](https://github.com/substrait-io/substrait/blob/v0.102.0/site/docs/expressions/scalar_functions.md#return-type-expressions). Use signed 64-bit intermediate values and reject arithmetic overflow or narrowing outside the type parameter's 32-bit range.

Integer division truncates toward zero. AND/OR evaluate both operands, while conditionals evaluate only the selected branch, matching [substrait-go's evaluator](https://github.com/substrait-io/substrait-go/blob/cb2d6e648bc0ee1549e354dbfe705d9f8a1ce2b1/types/type_derivation.go).

This enables 11 of the 15 catalog return programs: eight decimal arithmetic and three rounding variants. Four temporal variants still reference integer_parameter(precision) without binding precision from an argument type and remain unsupported.

Unary abs in return expressions also remains unsupported.

Refs #1242

## Summary by CodeRabbit

- **New Features**
  - Expanded parameterized return-type derivation with arithmetic, conditionals, local values, bound parameters, and parameterized type construction.
  - Added support for `>=`, `<=`, and `!=` comparisons.
  - Added validated decimal, numeric, nullable decimal, and varchar type derivation.

- **Bug Fixes**
  - Improved validation for narrowing, overflow, division by zero, invalid expressions, and unresolved parameters.
  - Decimal function results are checked against derived precision, scale, and nullability.
  - Nested return programs are rejected consistently.

- **Documentation**
  - Documented supported return-program semantics and limitations.

### DIFF
--- a/core/src/main/java/io/substrait/type/TypeExpressionEvaluator.java
+++ b/core/src/main/java/io/substrait/type/TypeExpressionEvaluator.java
@@ -27,16 +27,19 @@ import java.util.OptionalInt;
  * type classes whose parameter is an integer to substitute: {@code DECIMAL<P,S>}, {@code
  * varchar<L1>}, {@code fixedchar<L1>}, {@code fixedbinary<L1>}, {@code precision_time<P>}, {@code
  * precision_timestamp<P>}, {@code precision_timestamp_tz<P>}, {@code interval_day<P>} and {@code
- * interval_compound<P>}. No standard extension declares a parameterized {@code fixedbinary} or
- * {@code interval_compound} at all, as an argument or as a return -- those two are supported for
+ * interval_compound<P>}. Integer arithmetic, comparisons, boolean operations and conditionals can
+ * appear in those parameters or in the assignments of a multi-line return program. Arithmetic uses
+ * signed 64-bit values; overflow and narrowing to a type parameter's 32-bit representation are
+ * checked rather than wrapped. No standard extension declares a parameterized {@code fixedbinary}
+ * or {@code interval_compound} at all, as an argument or as a return -- those two are supported for
  * symmetry, and pinned against hand-written declarations rather than the catalog.
  *
  * <p>A {@code list} return still fails whatever its element, because the evaluator does not descend
  * into a container -- so an element parameter it would otherwise substitute, as in {@code
- * list<varchar<L1>>}, is out of reach just as an element type to evaluate is. A multi-line return
- * program still fails because evaluating one needs integer arithmetic over the bound parameters
- * rather than substitution. And a plain {@code any} cannot be derived at all: unlike {@code any1}
- * it names nothing, so there is no identity to bind.
+ * list<varchar<L1>>}, is out of reach just as an element type to evaluate is. A program referring
+ * to an argument's value rather than a parameter of its type still fails: this API receives only
+ * argument types. A plain {@code any} cannot be derived at all: unlike {@code any1} it names
+ * nothing, so there is no identity to bind. Type-covering expressions are not supported either.
  *
  * <p>Which shipped variants those cover is pinned by {@code ParameterizedReturnTypeTest} against
  * the declarations the catalog ships, and deliberately not repeated here -- the catalog is owned
@@ -85,7 +88,13 @@ public class TypeExpressionEvaluator {
       // The declared return type is already concrete; nothing to derive.
       return (Type) returnExpression;
     }
-    return returnExpression.accept(new ReturnTypeEvaluator(returnExpression, bindings));
+    try {
+      return new ReturnTypeEvaluator(returnExpression, bindings)
+          .evaluate(returnExpression, Type.class);
+    } catch (ArithmeticException e) {
+      throw new UnsupportedOperationException(
+          "Cannot evaluate return-type arithmetic: " + e.getMessage(), e);
+    }
   }
 
   /**
@@ -311,9 +320,12 @@ public class TypeExpressionEvaluator {
    * throwing base, keeping unsupported derivations fail-closed.
    */
   private static final class ReturnTypeEvaluator
-      extends TypeExpressionVisitor.TypeExpressionThrowsVisitor<Type, RuntimeException> {
+      extends TypeExpressionVisitor.TypeExpressionThrowsVisitor<Object, RuntimeException> {
 
     private final ParameterBindings bindings;
+    // The derivation language has three value kinds: integer, boolean and type. Local assignments
+    // can hold any of them; each operation checks the kind it consumes.
+    private final Map<String, Object> locals = new HashMap<>();
 
     private ReturnTypeEvaluator(TypeExpression returnExpression, ParameterBindings bindings) {
       super("Cannot evaluate return-type expression: " + returnExpression);
@@ -322,60 +334,73 @@ public class TypeExpressionEvaluator {
 
     @Override
     public Type visit(ParameterizedType.Decimal decimal) {
-      int precision = resolveInteger(decimal.precision().value());
-      int scale = resolveInteger(decimal.scale().value());
+      int precision = resolveInteger(decimal.precision());
+      int scale = resolveInteger(decimal.scale());
       return TypeCreator.of(decimal.nullable()).decimal(precision, scale);
     }
 
     @Override
     public Type visit(ParameterizedType.FixedChar fixedChar) {
-      return TypeCreator.of(fixedChar.nullable())
-          .fixedChar(resolveInteger(fixedChar.length().value()));
+      return TypeCreator.of(fixedChar.nullable()).fixedChar(resolveInteger(fixedChar.length()));
     }
 
     @Override
     public Type visit(ParameterizedType.VarChar varChar) {
-      return TypeCreator.of(varChar.nullable()).varChar(resolveInteger(varChar.length().value()));
+      return TypeCreator.of(varChar.nullable()).varChar(resolveInteger(varChar.length()));
     }
 
     @Override
     public Type visit(ParameterizedType.FixedBinary fixedBinary) {
       return TypeCreator.of(fixedBinary.nullable())
-          .fixedBinary(resolveInteger(fixedBinary.length().value()));
+          .fixedBinary(resolveInteger(fixedBinary.length()));
     }
 
     @Override
     public Type visit(ParameterizedType.PrecisionTime precisionTime) {
       return TypeCreator.of(precisionTime.nullable())
-          .precisionTime(resolveInteger(precisionTime.precision().value()));
+          .precisionTime(resolveInteger(precisionTime.precision()));
     }
 
     @Override
     public Type visit(ParameterizedType.PrecisionTimestamp precisionTimestamp) {
       return TypeCreator.of(precisionTimestamp.nullable())
-          .precisionTimestamp(resolveInteger(precisionTimestamp.precision().value()));
+          .precisionTimestamp(resolveInteger(precisionTimestamp.precision()));
     }
 
     @Override
     public Type visit(ParameterizedType.PrecisionTimestampTZ precisionTimestampTZ) {
       return TypeCreator.of(precisionTimestampTZ.nullable())
-          .precisionTimestampTZ(resolveInteger(precisionTimestampTZ.precision().value()));
+          .precisionTimestampTZ(resolveInteger(precisionTimestampTZ.precision()));
     }
 
     @Override
     public Type visit(ParameterizedType.IntervalDay intervalDay) {
       return TypeCreator.of(intervalDay.nullable())
-          .intervalDay(resolveInteger(intervalDay.precision().value()));
+          .intervalDay(resolveInteger(intervalDay.precision()));
     }
 
     @Override
     public Type visit(ParameterizedType.IntervalCompound intervalCompound) {
       return TypeCreator.of(intervalCompound.nullable())
-          .intervalCompound(resolveInteger(intervalCompound.precision().value()));
+          .intervalCompound(resolveInteger(intervalCompound.precision()));
     }
 
     @Override
-    public Type visit(ParameterizedType.StringLiteral stringLiteral) {
+    public Object visit(ParameterizedType.StringLiteral stringLiteral) {
+      Object local = locals.get(stringLiteral.value());
+      if (local != null) {
+        return local instanceof Type
+            ? ((Type) local).withNullable(stringLiteral.nullable())
+            : local;
+      }
+      Integer integer = bindings.boundInteger(stringLiteral.value());
+      if (integer != null) {
+        return integer.longValue();
+      }
+      OptionalInt literal = parseIntegerLiteral(stringLiteral.value());
+      if (literal.isPresent()) {
+        return (long) literal.getAsInt();
+      }
       // A wildcard return (e.g. min(any1) -> any1) resolves to the bound argument type, taking the
       // nullability declared on the return expression in both directions (a required return forces
       // the type non-null, a nullable one forces it nullable). MIRROR policy, if any, is applied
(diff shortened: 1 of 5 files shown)


core/src/test/java/io/substrait/type/ReturnProgramTypeTest.java
```diff
@@ -0,0 +1,156 @@
...
+    assertEquals(R.varChar(10), evaluate("varchar<L>"));
+    assertThrows(UnsupportedOperationException.class, () -> evaluate("varchar<a>"));
+  }
+
+  @Test
+  void conditionsSelectOnlyTheChosenBranch() {
+    assertEquals(R.varChar(10), evaluate("varchar<L > 5 ? L : missing>"));
+    assertEquals(R.varChar(10), evaluate("varchar<L < 5 ? 1 / 0 : L>"));
+    assertEquals(R.I64, evaluate("(L = 10) ? i64 : string"));
+    assertEquals(R.varChar(10), evaluate("varchar<!(L < 5) AND L > 0 ? L : 1>"));
```
nielspardon: Pin the grammar's actual associativity here — this case yields `varChar(10)` under both `(!(L < 5)) AND (L > 0)` and `!((L < 5) AND (L > 0))`, so it cannot catch a change in either direction. substrait-io/substrait#1228 moves `!` above the arithmetic and boolean tiers upstream, so asserting today's parse turns the eventual `substrait-packaging` bump into a failing test rather than a silently different derived type.

```suggestion
    // `!` binds looser than AND in the shipped grammar, so this is !((L < 5) AND (L < 0)).
    assertEquals(R.varChar(1), evaluate("varchar<!(L < 5) AND L < 0 ? 1 : 2>"));
```

core/src/main/java/io/substrait/type/TypeExpressionEvaluator.java
```diff
@@ -388,16 +413,149 @@ public Type visit(ParameterizedType.StringLiteral stringLiteral) {
       return bound.withNullable(stringLiteral.nullable());
```
nielspardon: Consult `boundType` before `parseIntegerLiteral`: every wildcard derivation — the whole `any1` family — now constructs and fills a `NumberFormatException` that is caught and discarded, because `parseIntegerLiteral` reports "not a number" by throwing. Before this PR the method went straight to `boundType`. A bound type name is never a numeral, so the swap is behaviour-preserving.

```suggestion
      // A wildcard return (e.g. min(any1) -> any1) resolves to the bound argument type, taking the
      // nullability declared on the return expression in both directions (a required return forces
      // the type non-null, a nullable one forces it nullable). MIRROR policy, if any, is applied
      // afterwards by the caller. This is checked before parsing the token as an integer literal
      // because parseIntegerLiteral reports "not a number" by throwing, and a bound type name is
      // never a numeral.
      Type bound = bindings.boundType(stringLiteral.value());
      if (bound != null) {
        return bound.withNullable(stringLiteral.nullable());
      }
      OptionalInt literal = parseIntegerLiteral(stringLiteral.value());
      if (literal.isPresent()) {
        return (long) literal.getAsInt();
      }
      throw new UnsupportedOperationException(
          "Unbound type parameter '" + stringLiteral.value() + "' in return-type expression");
```

core/src/main/java/io/substrait/type/TypeExpressionEvaluator.java
```diff
@@ -388,16 +413,149 @@ public Type visit(ParameterizedType.StringLiteral stringLiteral) {
...
+    public Object visit(TypeExpression.BinaryOperation operation) {
+      switch (operation.opType()) {
+        case AND:
+        case OR:
+          boolean left = evaluate(operation.left(), Boolean.class);
+          boolean right = evaluate(operation.right(), Boolean.class);
+          return operation.opType() == TypeExpression.BinaryOperation.OpType.AND
+              ? left && right
+              : left || right;
+        case COVERS:
```
nielspardon: Drop this arm and the matching Javadoc sentence at line 42: no parsed expression can carry `COVERS`, because `expr` in `SubstraitType.g4` has no production consuming `:=`, and the second switch's `default:` already reports an unevaluable operator. The spelling the spec does define, `covers(a, b)`, fails earlier in `ParseToPojo.getFunctionType` with an `IllegalStateException`, so the Javadoc's fail-closed promise is not kept either way.

core/src/main/java/io/substrait/type/TypeExpressionEvaluator.java
```diff
@@ -388,16 +413,149 @@ public Type visit(ParameterizedType.StringLiteral stringLiteral) {
       return bound.withNullable(stringLiteral.nullable());
     }
 
-    private int resolveInteger(String token) {
-      Integer bound = bindings.boundInteger(token);
-      if (bound != null) {
-        return bound;
+    @Override
+    public Type visit(TypeExpression.Decimal decimal) {
```
nielspardon, line 422: Consider collapsing these nine methods into the nine above them — `ParameterizedType.StringLiteral` is itself a `TypeExpression` and `resolveInteger` now takes a `TypeExpression`, so nine private helpers plus eighteen one-line delegations would remove the duplication. As written, every future integer-parameterized type class must be added in two places, and changing how nullability is applied is eighteen edits.

core/src/main/java/io/substrait/type/TypeExpressionEvaluator.java
```diff
@@ -27,16 +27,19 @@
  * type classes whose parameter is an integer to substitute: {@code DECIMAL<P,S>}, {@code
  * varchar<L1>}, {@code fixedchar<L1>}, {@code fixedbinary<L1>}, {@code precision_time<P>}, {@code
  * precision_timestamp<P>}, {@code precision_timestamp_tz<P>}, {@code interval_day<P>} and {@code
- * interval_compound<P>}. No standard extension declares a parameterized {@code fixedbinary} or
- * {@code interval_compound} at all, as an argument or as a return -- those two are supported for
+ * interval_compound<P>}. Integer arithmetic, comparisons, boolean operations and conditionals can
+ * appear in those parameters or in the assignments of a multi-line return program. Arithmetic uses
+ * signed 64-bit values; overflow and narrowing to a type parameter's 32-bit representation are
```
nielspardon: Note in this paragraph that the spec settles neither of two choices made here: integer division truncating toward zero, and `AND`/`OR` evaluating both operands. Both match substrait-go, which is the right call, but AGENTS.md asks for a spec gap to be stated rather than filled silently, and the PR body reaches neither `git blame` nor the changelog. Separately, "signed 64-bit values" holds only for intermediates — `IntegerLiteral.value()` is an `int`, so `varchar<4000000000 - 3999999990>` dies in the parser with `NumberFormatException`.

nielspardon (commented):
All five are mechanical; the derivation logic itself is sound. Start with the `!` test — it passes under either reading of the shipped grammar, so it pins nothing, and substrait-io/substrait#1228 is fixing that grammar upstream.

core/src/test/java/io/substrait/type/ReturnProgramTypeTest.java
alexandrefimov: Done. It gives `varChar(1)` today.

core/src/main/java/io/substrait/type/TypeExpressionEvaluator.java
alexandrefimov: Done.

core/src/main/java/io/substrait/type/TypeExpressionEvaluator.java
alexandrefimov: Done.

core/src/main/java/io/substrait/type/TypeExpressionEvaluator.java
alexandrefimov: Done.

core/src/main/java/io/substrait/type/TypeExpressionEvaluator.java
alexandrefimov: Done, plus a note that a literal must fit in 32 bits.

core/src/main/java/io/substrait/type/TypeExpressionEvaluator.java
```diff
@@ -322,82 +337,254 @@ private ReturnTypeEvaluator(TypeExpression returnExpression, ParameterBindings b
...
+
+    @Override
+    public Type visit(TypeExpression.IntervalCompound intervalCompound) {
+      return intervalCompound(intervalCompound.nullable(), intervalCompound.precision());
+    }
+
+    @Override
+    public Object visit(TypeExpression.ReturnProgram program) {
+      for (TypeExpression.ReturnProgram.Assignment assignment : program.assignments()) {
+        locals.put(assignment.name(), evaluate(assignment.expr(), Object.class));
```
nielspardon, line 478: Reject a nested `ReturnProgram` rather than evaluating it: `locals` is one flat map with no scope, so an inner program's assignments land in the enclosing scope and outlive it. `a = 1\nb = a = 99\ni64\nvarchar<a>` derives `varchar<99>` where scoping would give `varchar<1>`. The grammar reaches `#MultilineDefinition` from `expr`, so an assignment's right-hand side can be a program, and no sibling binding defines what that means.

core/src/main/java/io/substrait/type/TypeExpressionEvaluator.java
```diff
@@ -322,82 +337,254 @@ private ReturnTypeEvaluator(TypeExpression returnExpression, ParameterBindings b
...
+      return evaluate(program.finalExpression(), Type.class);
+    }
+
+    @Override
+    public Long visit(TypeExpression.IntegerLiteral literal) {
+      return (long) literal.value();
+    }
+
+    @Override
+    public Object visit(TypeExpression.IfOperation conditional) {
```
nielspardon, line 489: Check that both branches have the same value kind, rather than only the selected one: `L > 5 ? i64 : 3` derives `I64`, while the same declaration called with a shorter argument throws `Expected Type in return-type expression, got 3`. A declaration whose branches disagree in kind is wrong in the declaration, but it only surfaces for the half of the argument space that picks the bad branch, so a broken extension can ship and resolve wherever the good branch happens to win.

core/src/main/java/io/substrait/type/TypeExpressionEvaluator.java
```diff
@@ -311,9 +323,12 @@ private static OptionalInt parseIntegerLiteral(String token) {
...
-      extends TypeExpressionVisitor.TypeExpressionThrowsVisitor<Type, RuntimeException> {
+      extends TypeExpressionVisitor.TypeExpressionThrowsVisitor<Object, RuntimeException> {
 
     private final ParameterBindings bindings;
+    // The derivation language has three value kinds: integer, boolean and type. Local assignments
+    // can hold any of them; each operation checks the kind it consumes.
+    private final Map<String, Object> locals = new HashMap<>();
 
     private ReturnTypeEvaluator(TypeExpression returnExpression, ParameterBindings bindings) {
       super("Cannot evaluate return-type expression: " + returnExpression);
```
nielspardon: Build this message lazily — a `Supplier<String>` in the throwing base, or inside `t()`. It renders the whole expression tree on every construction, which before this PR was always the message of a thrown exception; now the catalog's decimal `add`/`subtract`/`multiply`/`divide`/`modulus` succeed, so each derivation stringifies a seven-assignment program and discards it. `resolveType` runs per function binding, so it scales with plan size.

nielspardon (commented):
Four smaller things, none blocking — the helper collapse reads well.

core/src/test/java/io/substrait/type/ReturnProgramTypeTest.java
```diff
@@ -0,0 +1,157 @@
...
+      value = {
+        "varchar<L + 1>; varchar<11>",
+        "fixedchar<L * 2>; fixedchar<20>",
+        "fixedbinary<L / 2>; fixedbinary<5>",
+        "decimal<L + 2, L - 8>; decimal<12,2>",
+        "precision_time<min(L, 6)>; precision_time<6>",
+        "precision_timestamp<max(3, L - 4)>; precision_timestamp<6>",
+        "precision_timestamp_tz<L - 4>; precision_timestamp_tz<6>",
+        "interval_day<L - 7>; interval_day<3>",
+        "interval_compound<L - 7>; interval_compound<3>"
```
nielspardon: Give one or two of these rows a `?` return: all nine declare a REQUIRED type, so the `TypeCreator.of(nullable)` call in each of the nine new helpers is uncovered. The collapse made those helpers the single place nullability is applied for both AST shapes, so hard-coding `REQUIRED` in one of them — or inverting the flag — leaves the whole suite green.

core/src/main/java/io/substrait/type/TypeExpressionEvaluator.java
alexandrefimov: Done, it throws now.

core/src/main/java/io/substrait/type/TypeExpressionEvaluator.java
alexandrefimov: Agreed. substrait-go doesn't check this either, so I'll send a syntactic kind check as a follow-up.

core/src/main/java/io/substrait/type/TypeExpressionEvaluator.java
alexandrefimov: Done, with a `Supplier<String>` constructor on the throwing visitors.

core/src/test/java/io/substrait/type/ReturnProgramTypeTest.java
alexandrefimov: Done, added `varchar?` and `decimal?` rows.

core/src/main/java/io/substrait/type/TypeExpressionEvaluator.java
alexandrefimov: It turned out small, so it's in this PR after all: bf3fc2d5.

core/src/main/java/io/substrait/type/TypeExpressionEvaluator.java
```diff
@@ -311,93 +327,345 @@ private static OptionalInt parseIntegerLiteral(String token) {
...
+    public Type visit(TypeExpression.IntervalCompound intervalCompound) {
+      return intervalCompound(intervalCompound.nullable(), intervalCompound.precision());
+    }
+
+    @Override
+    public Object visit(TypeExpression.ReturnProgram program) {
+      // The grammar lets an assignment's right-hand side be a program of its own, and nothing
+      // defines its scope. The locals here are one flat map, so an inner program's assignments
+      // would
+      // leak into the enclosing one; refuse it instead.
```
nielspardon: Rejoin this comment — `would` ended up alone on its own line. google-java-format leaves line comments as written, so the reflow sticks.

```suggestion
      // defines its scope. The locals here are one flat map, so an inner program's assignments
      // would leak into the enclosing one; refuse it instead.
```

nielspardon (commented):
Just a stray line wrap left. The kind check reads well — keeping it to the branches' shape means the untaken branch is still never evaluated.

core/src/main/java/io/substrait/type/TypeExpressionEvaluator.java
alexandrefimov: Done.

nielspardon (approved):
LGTM

Issue #1230, opened 2026-09-03
isthmus: a fractional RANGE window offset is emitted against an integral ordering column


nielspardon:
`normalizeIntegralOffset` in `WindowBoundConverter` returns early when the offset is not an integral literal, so the ordering expression's type is never consulted for a fractional offset. On `fb6a54ad`, `select sum(O_SHIPPRIORITY) over (order by O_SHIPPRIORITY range between 1.5 preceding and current row) from ORDERS` converts to `Preceding{DecimalLiteral{1.5}}` against an `i32` ordering column.

Per `site/docs/expressions/window_functions.md:34` at spec v0.102.0, a RANGE `offset_expr`'s type `D` must be compatible with the ordering expression's type `T` — `add(T, D) -> T` and `subtract(T, D) -> T` must be defined. `functions_arithmetic.yaml` defines `add`/`subtract` only for matched pairs, and `site/docs/types/type_system.md:17` rules out coercion ("Substrait employs a strict type system without any coercion rules"), so no `add(i32, decimal<P,S>) -> i32` exists and the emitted plan is invalid.

The integral half of that rule is now handled: #1206 makes an integral offset that cannot be retyped to the ordering column throw. A fractional offset takes the early return above that check, so the two halves behave differently — an oversized integral offset is rejected while a type-mismatched fractional one is emitted. Calcite does not catch it either, since `SqlWindow#validateFrameBoundary` checks only the type family for a RANGE bound.

The existing tests `precedingWithDecimalOffsetKeepsItsFraction` and `followingWithNonLiteralOffset` both pass `Optional.empty()` for the ordering type, so neither exercises the retype path.

#1198 tracks the same rule going unchecked on the core side, for plans built directly or read from proto; this is the isthmus producer half for non-integral offsets.

PR #1309, opened 2026-09-21
fix(isthmus)!: mirror negative and normalize zero decimal and floating-point RANGE offsets

### LINKED ISSUE #1291: isthmus: normalize negative RANGE offsets before retyping
Follow-up to #1206, as agreed in https://github.com/substrait-io/substrait-java/pull/1206#discussion_r3950929318.

Reproduced at `ee9f3d20` with the test TPC-H `ORDERS` schema. Two cases remain:

```sql
SELECT SUM(O_TOTALPRICE) OVER (
  ORDER BY O_TOTALPRICE RANGE BETWEEN -5.5 PRECEDING AND CURRENT ROW
) FROM ORDERS;
```

The converter emits a `Preceding` bound with a decimal offset of `-5.5`. The magnitude should be positive and the direction should become `Following`.

```sql
SELECT SUM(O_ORDERKEY) OVER (
  ORDER BY O_ORDERKEY RANGE BETWEEN -2147483648 PRECEDING AND CURRENT ROW
) FROM ORDERS;
```

The converter now emits a `Preceding` bound with an `i64` offset of `-2147483648`. It no longer throws the exception reported at `a8bf3d5a`; it produces an invalid negative distance instead. The positive magnitude `2147483648` fits the `i64` ordering type, but the sign and direction are not normalized.

[Spec v0.102.0](https://github.com/substrait-io/substrait/blob/v0.102.0/site/docs/expressions/window_functions.md) requires a non-negative RANGE distance. Normalize the sign and direction while preserving the exact value, then represent the magnitude in the ordering expression's compatible offset type.

anasik:
- #1206 mirrors a negative integral RANGE offset to the opposite bound with its magnitude, and normalizes a zero integral offset to CurrentRow, since the spec requires a non-negative RANGE distance. A decimal or floating-point offset skipped both checks entirely, so a negative one kept its sign and bound label, and a zero one stayed as Preceding/Following{0} instead of CurrentRow — the same spec-invalid offset_expr shapes the integral fix was meant to prevent.
- Extends both checks to Expression.DecimalLiteral, FP64Literal, and FP32Literal, reusing the same sign-check-and-flip logic as the integral case.
- Decimal negation is exact (BigDecimal), never needs widening, unlike integral where negating a type's minimum value can overflow.  
- renamed `negate` to `negateIntegral` because it sounded too general.

Closes #1291  

BREAKING CHANGE: a RANGE window's negative decimal or floating-point offset,  previously converted with its sign and bound label unchanged, is now mirrored  to the opposite bound with its positive magnitude, matching the existing  integral behavior. A zero-valued decimal or floating-point offset, previously left as Preceding/Following{0}, now converts to CurrentRow. Both match the existing integral behavior.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected conversion of negative decimal and floating-point window offsets to the appropriate preceding or following bounds.
  * Preserved decimal precision, scale, and numeric types when converting window offsets.
  * Zero-valued decimal and floating-point offsets now correctly produce current-row bounds.
  * Positive floating-point preceding offsets continue to retain their expected behavior.

* **Tests**
  * Added coverage for zero, positive, and negative decimal, double, and real window offsets.


anasik:
@nielspardon looks like the CodeRabbit is breaking the PR Title Check

nielspardon:
> @nielspardon looks like the CodeRabbit is breaking the PR Title Check

it looks like the reason is that the PR title is too long with 115 characters vs max 100 characters per conventional commits

anasik:
> > @nielspardon looks like the CodeRabbit is breaking the PR Title Check
> 
> it looks like the reason is that the PR title is too long with 115 characters vs max 100 characters per conventional commits

Fixed

isthmus/src/main/java/io/substrait/isthmus/expression/WindowBoundConverter.java
```diff
@@ -85,7 +86,56 @@ public static WindowBound toWindowBound(
...
+      Expression.DecimalLiteral decimal = (Expression.DecimalLiteral) expression;
+      return Optional.of(
+          DecimalUtil.getBigDecimalFromBytes(decimal.value().toByteArray(), decimal.scale(), 16));
+    }
+    return Optional.empty();
+  }
+
+  private static Expression decimalLiteralOfType(Type type, BigDecimal value) {
+    Type.Decimal decimal = (Type.Decimal) type;
+    return ExpressionCreator.decimal(type.nullable(), value, decimal.precision(), decimal.scale());
```
nielspardon: Pass `false` rather than `type.nullable()` here and at lines 162/164: `negateIntegral` (line 148) and every branch of `integralLiteralOfType` hardcode `false`, so as written a mirrored offset's nullability depends on which numeric family the literal came from. A literal distance can never be null, so `false` is the consistent answer — if you'd rather change the integral side instead that works too, but the three families should agree.

```suggestion
    return ExpressionCreator.decimal(false, value, decimal.precision(), decimal.scale());
```

isthmus/src/main/java/io/substrait/isthmus/expression/WindowBoundConverter.java
```diff
@@ -98,6 +148,24 @@ private static Expression negate(Expression offset, long value) {
...
+    return Optional.empty();
+  }
+
+  private static Expression floatingLiteralOfType(Type type, double value) {
+    if (type instanceof Type.FP64) {
+      return ExpressionCreator.fp64(type.nullable(), value);
+    } else if (type instanceof Type.FP32) {
+      return ExpressionCreator.fp32(type.nullable(), (float) value);
+    }
+    throw new IllegalStateException("expected a floating-point type, got " + type);
```
nielspardon: This throw is unreachable — `floatingValue` only matches `FP64Literal`/`FP32Literal`, whose `getType()` always returns `Type.FP64`/`Type.FP32` — while `decimalLiteralOfType` guards the same impossible case with a bare `(Type.Decimal)` cast at line 134, so two sibling helpers treat one non-case two different ways.

Both helpers exist only because the extractors discard the literal they matched and hand on `offset.getType()`; negating inside the matched branch (`ExpressionCreator.fp32(lit.nullable(), -lit.value())`, `ExpressionCreator.decimal(lit.nullable(), value.negate(), lit.precision(), lit.scale())`) removes both helpers, the cast, the dead branch, and the `float`→`double`→`float` round trip in `floatingValue`.

If the throw does stay, render the type with `type.accept(new StringTypeVisitor())` as line 194 does — a `Type`'s `toString` is the Immutables structural dump, not `fp64`.

isthmus/src/main/java/io/substrait/isthmus/expression/WindowBoundConverter.java
```diff
@@ -85,7 +86,56 @@ public static WindowBound toWindowBound(
...
-  private static Expression negate(Expression offset, long value) {
+  /**
+   * Reports whether {@code offset} is a decimal, floating-point, or integral literal with a value
+   * of zero.
+   *
+   * @param offset the offset expression to check
+   * @return {@code true} if {@code offset} is a zero-valued literal
+   */
+  private static boolean isZero(Expression offset) {
+    return decimalValue(offset).map(value -> value.signum() == 0).orElse(false)
```
nielspardon: `decimalValue` decodes the `ByteString` here and again as the eager head of `negateIfNegative`'s chain at line 109, and `integralValue` now runs three times per conversion. Decoding once into a local `Optional<BigDecimal>` in `toWindowBound` and branching on `signum()` answers both questions in one pass — worth more as a seam than as a saving, since the zero test and the sign test are currently two independently maintained `instanceof` chains that can drift apart.

Either way the call order deserves a comment: `-0.0 == 0` is true but `-0.0 < 0` is false, so a `-0.0` offset collapses to `CurrentRow` only because `isZero` runs before `negateIfNegative`.

isthmus/src/test/java/io/substrait/isthmus/WindowBoundConverterTest.java
```diff
@@ -307,4 +342,41 @@ void negativeOffsetOverflowingLongIsRejectedRatherThanThrowingArithmeticExceptio
...
+
+  @Test
+  void negativeDecimalPrecedingOffsetIsFlippedToFollowingWithItsMagnitude() {
+    // The integral mirror above has a decimal counterpart: RANGE BETWEEN -5.5 PRECEDING is
+    // equivalent to FOLLOWING 5.5.
+    RexNode offset = c(new BigDecimal("-5.5"), SqlTypeName.DECIMAL, 19, 1);
+    RexWindowBound bound = RexWindowBounds.preceding(offset);
+
+    WindowBound converted =
+        WindowBoundConverter.toWindowBound(bound, false, Optional.empty(), rexExpressionConverter);
```
nielspardon, line 354: All six new tests pass `Optional.empty()` for `orderingType`, cover PRECEDING only, and use `isRows=false`, where the integral negative tests at lines 288/303 pass `Optional.of(orderingType)` and cover both directions. A decimal case with `Optional.of(t(SqlTypeName.DECIMAL, 5, 2))` would pin down which behaviour is intended ahead of #1230 instead of leaving it unasserted, and a negative-FOLLOWING decimal case would cover the flip in the other direction.

A positive-FP case is the other gap: nothing fails today if `isZero` or `negateIfNegative` over-fires on an FP value, because `precedingWithDecimalOffsetKeepsItsFraction` only covers DECIMAL.

nielspardon (commented):
Narrow the title from "non-integral" to "decimal and floating-point" — a negative or zero `interval_day`/`interval_year` offset still keeps its sign and its bound label (`range between interval '-1' day preceding` converts to `Preceding{IntervalDayLiteral{days=-1}}`, and `interval '0' day preceding` stays `Preceding{...days=0}`), and the title is what lands in `CHANGELOG.md`. `fix(isthmus)!: mirror negative and normalize zero decimal and floating-point RANGE offsets` is 90 chars, inside the limit.

Same word in reverse at line 22: the class Javadoc's "A negative **integral** offset is mirrored to the opposite bound with its magnitude" now covers decimal and FP too.

The retype half of #1291 is already tracked as #1230, so leaving it out here is right — just confirm you still want `Closes #1291`. The interval and ROWS gaps I'll file separately; the inline comments below are all non-blocking cleanups.

nielspardon (commented):
Last thing from my first pass: the class Javadoc at lines 21-23 still reads "A negative **integral** offset is mirrored to the opposite bound with its magnitude" — that now covers decimal and FP too. The `@param orderingType` note just below it stays accurate, and your new `negativeDecimalPrecedingOffsetIsNotRetypedAgainstTheOrderingType` test is what makes the distinction legible.

isthmus/src/test/java/io/substrait/isthmus/WindowBoundConverterTest.java
nielspardon: The `notRetypedAgainstTheOrderingType` test is the right shape — worth noting for #1230 that its assertion is the one that has to flip when the retype lands.

anasik:
@nielspardon fixed

nielspardon (approved):
LGTM

